Issue #6, opened 2019-05-06
pg_hba.conf error when creating secondary if reverse DNS is unavailable 


marcocitus:
I tried setting up pg_auto_failover on a CentOS 7 Azure VM (locally), but when setting up the standby I ran into the following issue:
```
$ pg_autoctl create postgres --monitor postgres://autoctl_node@pgaf-test.dmqu0iik0f1uppllrfdl2ayvwe.bx.internal.cloudapp.net:6000/pg_auto_failover --pgport 6002 --nodename `hostname --fqdn`  --pgdata node2   
...
03:39:15 ERROR pg_basebackup: could not connect to server: FATAL:  no pg_hba.conf entry for replication connection from host "10.1.19.4", user "pgautofailover_replicator", SSL off

03:39:15 ERROR Failed to run pg_basebackup: exit code 1
03:39:15 ERROR Failed initialise standby server, see above for details
03:39:15 ERROR Failed to transition from state "wait_standby" to state "catchingup", see above.
```

In the pg_hba.conf, the following entries were added:
```
host all "pgautofailover_monitor" pgaf-test.dmqu0iik0f1uppllrfdl2ayvwe.bx.internal.cloudapp.net trust # Auto-generated by pg_auto_failover
host replication "pgautofailover_replicator" pgaf-test.dmqu0iik0f1uppllrfdl2ayvwe.bx.internal.cloudapp.net trust # Auto-generated by pg_auto_failover
host "postgres" "pgautofailover_replicator" pgaf-test.dmqu0iik0f1uppllrfdl2ayvwe.bx.internal.cloudapp.net trust # Auto-generated by pg_auto_failover
```

Using hostnames in pg_hba.conf relies on reverse DNS of the client IP. However, reverse DNS of internal IPs does not work on Azure.

Adding pg_hba.conf records for `10.1.19.4/32` on the primary and running `SELECT pg_reload_conf()` before creating the standby resolved the problem.

DimCitus:
See https://github.com/citusdata/pg_auto_failover/pull/9#issuecomment-490107561 for more context about a search of a good solution for this one. In particular:

> We might want to revisit the whole HBA management in pg_auto_failover and come up with a nicer overall implementation here, including facilities for users to provide their own HBA rules and then have pg_auto_failover only check that everything works as intended.

Discussion can continue from here.

DimCitus:
[...]

PR #9, opened 2019-05-06
Implement pg_autoctl show ipaddr|cidr.

DimCitus:
Rather than using `hostname -fqdn` we introduce new commands `show ipaddr`
which displays the discovered LAN ip address, and `show cidr` which allows
for debugging pg_autoctl HBA edits.

Those commands would be nice additions to the PG_AUTOCTL_DEBUG mode for the
command line, but it seems that in some environments with broken reverse DNS
we need an easy way to discover the LAN IP address from the command line. A
quick search shows it's pretty hard to do that with classic OS tooling in a
portable way, and we already have had to write code for that feature in
pg_auto_failover anyway. This patch exposes the feature to the end users.

See #6.

### DIFF
--- a/docs/ref/reference.rst
+++ b/docs/ref/reference.rst
@@ -56,6 +56,8 @@ keeper::
       uri     Show the postgres uri to use to connect to pg_auto_failover nodes
       events  Prints monitor's state of nodes in a given formation and group
       state   Prints monitor's state of nodes in a given formation and group
+      ipaddr  Print this node's IP address information
+      cidr    Print this node's CIDR information
 
     pg_autoctl enable
       secondary  Enable secondary nodes on a formation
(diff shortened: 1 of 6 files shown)


docs/ref/reference.rst
```diff
@@ -207,6 +209,21 @@ commands can be used, from any node in the setup.
     For details about the options to the command, see above in the ``pg_autoctl
     show events`` command.
 
+  - ``pg_autoctl show uri``
```
Bogdanp: Please excuse the drive-by comment, but it looks like this might have been mistakenly copied from above. This should say `pg_autoctl show ipaddr`, no?

DimCitus: Hi @Bogdanp, you're right, I've updated the PR (rebased on top of current master) and fixed the reference. I'm also going to add a message about the current state of this PR. Thanks!

DimCitus:
This PR leaves several problems open, so while it's interesting as a prototype and opens the discussion, it's not currently considered as a good solution for the problem at hand:

  - exposing a public API in the command line that might be best though as a debugging facility, namely the commands `pg_autoctl do show ipaddr|cidr` would make more sense, maybe, than having them exposed as in this PR.
  - it might be better to use `samenet` in pg_hba.conf, see https://www.postgresql.org/docs/current/auth-pg-hba-conf.html ; and then we don't need to put down the hostname of the other host in the HBA file, and we bypass reverse DNS lookups entirely.
  - discovering which LAN IP address to use on a host with several IP addresses is going to be guess work at best, in the simple case where there's a single internal (non-routable) IP address that might be ok though, but the method we implemented in `fetchLocalIPAddress` (https://github.com/citusdata/pg_auto_failover/blob/master/src/bin/pg_autoctl/ipaddr.c#L39) feels like it might be too magic for its own good.
  - about `samenet`, some hosting facilities only provide internet routable IP addresses, we might want to refrain from granting connection to them in that case (but we could use a /32 as we do now when installing the IP address from the secondary node).

We might want to revisit the whole HBA management in pg_auto_failover and come up with a nicer overall implementation here, including facilities for users to provide their own HBA rules and then have pg_auto_failover only check that everything works as intended.

PR #15, opened 2019-05-14
Code cleanup prepared transaction name

DimCitus:


src/bin/pg_autoctl/state.h
```diff
@@ -95,8 +95,6 @@ typedef struct
 	uint64_t last_secondary_contact;
 	int64_t xlog_lag;
 	int keeper_is_paused;
-
-	/* keeper's opened prepared transaction for master_update_node() */
```
thanodnl: seems strange to remove the comment while keeping the member

DimCitus: Well spotted, thanks! fixed now.

PR #16, opened 2019-05-15
Implement a default value for --nodename.

### LINKED ISSUE #6: pg_hba.conf error when creating secondary if reverse DNS is unavailable

DimCitus:
It is not always possible to obtain a default value for --nodename, because
of the way Postgres HBA rules matching is implemented. We need to have an
hostname for the local server that resoves in DNS to an IP that is on the
same server.

So in this patch we find which interface is used to connect to an external
service, either the pg_auto_failover monitor or the Google 8.8.8.8:53
service, and then do a reverse DNS lookup for our local routed IP. If we get
a candidate --nodename, we then do a DNS lookup of it and check that we get
back an IP address that we have on the local server. When all those
conditions are met then we take this hostname as a default --nodename.

When a --nodename is provided, we also check that we can resolve it with the
current DNS setup, and WARN the user when that's not the case. It might be
that we don't have enough information in our tests, so we continue with the
--nodename.

Fixes #6.


src/bin/pg_autoctl/cli_create_drop_node.c
```diff
@@ -506,3 +559,130 @@ cli_drop_node(int argc, char **argv)
...
+	if (!fetchLocalIPAddress(ipAddr, BUFSIZE, monitorHostname, monitorPort))
+	{
+		log_fatal("Failed to find a local IP address, "
+				  "please provide --nodename.");
+		return false;
+	}
+
+	/* from there on we can take the ipAddr as the default --nodename */
+	strlcpy(nodename, ipAddr, size);
+	log_debug("check_or_discover_nodename: local ip %s", ipAddr);
```
thanodnl: the first part of the logline seems to refer to a function name that is different from the one we are currently in. Is this a left over of a refactoring or by design?

src/bin/pg_autoctl/ipaddr.c
```diff
@@ -200,6 +232,32 @@ countSetBits(unsigned int n)
...
+	int i = 0;
+	unsigned int count = 0;
+
+	for(i=0; i<16; i++)
+	{
+		unsigned char n = addr[i];
+
+		while (n)
+		{
+			count += n & 1;
```
thanodnl: have you considered using `__builtin_popcount` for counting the bits set? Your code works, however this is such a common operation for the cpu there is a specialised implementation for it.

src/bin/pg_autoctl/cli_do_show.c
```diff
@@ -0,0 +1,235 @@
...
+	fprintf(stdout, "%s\n", cidr);
+}
+
+
+/*
+ * cli_check_nodename checks that the --nodename argument is either an IP
+ * address that exists on the local list of interfaces, or a hostname that a
+ * DNS lookup solves to an IP address we have on the local machine.
+ *
+ * TODO: then do a reverse lookup of that IP address!
```
thanodnl: is this still a todo? Seems like you already perform a lookup based on an ip address.

If it is still a TODO can we implement this already in the current PR instead of leaving the todo hanging here?

src/bin/pg_autoctl/ipaddr.c
```diff
@@ -244,3 +302,226 @@ isTestEnv(void)
...
+
+/*
+ * From /Users/dim/dev/PostgreSQL/postgresql/src/backend/libpq/hba.c
+ */
+static bool
+ipv6eq(struct sockaddr_in6 *a, struct sockaddr_in6 *b)
+{
+	int			i;
+
+	for (i = 0; i < 16; i++)
```
thanodnl: missing braces

src/bin/pg_autoctl/ipaddr.c
```diff
@@ -158,19 +160,49 @@ fetchLocalCIDR(const char *localIpAddress, char *localCIDR, int size)
...
 
-		if (strcmp(address, localIpAddress) != 0)
+		if (strcmp(address, localIpAddress) == 0)
 		{
-			continue;
+			found = true;
+			break;
 		}
-		break;
     }
```
thanodnl: nit; indenting seems messed up here

src/bin/pg_autoctl/ipaddr.c
```diff
@@ -244,3 +302,226 @@ isTestEnv(void)
...
+/*
+ * findHostnameLocalAddress does a reverse DNS lookup given an hostname
+ * (--nodename), and if the DNS lookup fails or doesn't return any local IP
+ * address, then returns false.
+ */
+bool
+findHostnameLocalAddress(const char *hostname, char *localIpAddress, int size)
+{
+	int error;
+    struct addrinfo *dns_lookup_addr;
```
thanodnl: indenting

src/bin/pg_autoctl/ipaddr.c
```diff
@@ -244,3 +302,226 @@ isTestEnv(void)
...
+	for (i = 0; i < 16; i++)
+		if (a->sin6_addr.s6_addr[i] != b->sin6_addr.s6_addr[i])
+			return false;
+
+	return true;
+}
+
+
+/*
+ * findHostnameLocalAddress does a reverse DNS lookup given an hostname
```
thanodnl: s/an hostname/a hostname/

src/bin/pg_autoctl/ipaddr.c
```diff
@@ -244,3 +302,226 @@ isTestEnv(void)
...
+ * (--nodename), and if the DNS lookup fails or doesn't return any local IP
+ * address, then returns false.
+ */
+bool
+findHostnameLocalAddress(const char *hostname, char *localIpAddress, int size)
+{
+	int error;
+    struct addrinfo *dns_lookup_addr;
+	struct addrinfo *dns_addr;
+	struct ifaddrs *ifaddr, *ifa;
```
thanodnl: I find the `ifa` name a bit non-telling, it seems to be iterating over `ifaddr`, so maybe the first should be `ifaddrs` (plural) and we can use `ifaddr` for iteration. Although the type is already called the same, don't exactly know if such a collision is allowed

src/bin/pg_autoctl/ipaddr.c
```diff
@@ -244,3 +302,226 @@ isTestEnv(void)
...
+				  strerror(errno));
+		return false;
+	}
+
+	/*
+	 * Compare both addresses list (dns lookup and list of interface
+	 * addresses) in a nested loop fashion: lists are not sorted, and we
+	 * expect something like a dozen entry per list anyway.
+	 */
+	for (dns_addr=dns_lookup_addr; dns_addr; dns_addr = dns_addr->ai_next)
```
thanodnl: spaces around `=`
and can we compare dns_addr to `NULL`, it is a bit hard to read like this.

src/bin/pg_autoctl/ipaddr.c
```diff
@@ -244,3 +302,226 @@ isTestEnv(void)
...
+	{
+		/*
+		 * Skip loopback addresses, we want to be able to connect to
+		 * hostname from other nodes.
+		 */
+		if (!(dns_addr->ai_flags & IFF_LOOPBACK))
+		{
+			for (ifa = ifaddr; ifa; ifa = ifa->ifa_next)
```
thanodnl: compare `ifa` to NULL please

src/bin/pg_autoctl/ipaddr.c
```diff
@@ -244,3 +302,226 @@ isTestEnv(void)
...
+			{
+				/*
+				 * TODO: implement ipv6 support
```
thanodnl: seems to be implemented from the code, if so please remove. Otherwise I would suggest implement it as we do support ipv6 on most locations.

src/bin/pg_autoctl/ipaddr.c
```diff
@@ -244,3 +302,226 @@ isTestEnv(void)
...
+	 * addresses) in a nested loop fashion: lists are not sorted, and we
+	 * expect something like a dozen entry per list anyway.
+	 */
+	for (dns_addr=dns_lookup_addr; dns_addr; dns_addr = dns_addr->ai_next)
+	{
+		/*
+		 * Skip loopback addresses, we want to be able to connect to
+		 * hostname from other nodes.
+		 */
+		if (!(dns_addr->ai_flags & IFF_LOOPBACK))
```
thanodnl: can we rewrite this in such a way that we can remove an indent from the block? It is hard to follow what is happening here due to the big body and many indents.

If I understand it correctly this can be replaced with a `continue;` when the lookback bit is set.

src/bin/pg_autoctl/ipaddr.c
```diff
@@ -244,3 +302,226 @@ isTestEnv(void)
...
+				/*
+				 * TODO: implement ipv6 support
+				 */
+				if (ifa->ifa_addr->sa_family == AF_INET
+					&& dns_addr->ai_family == AF_INET)
+				{
+					struct sockaddr_in *ip =
+						(struct sockaddr_in *) ifa->ifa_addr;
+
+					if (ipv4eq(ip, (struct sockaddr_in *) dns_addr->ai_addr))
```
thanodnl: wonder here if it would be more readable by either moving this check into the if-statement above, or replacing with a `continue;` on failure. Again to remove a bit of the indenting.

Same for the block below.

src/bin/pg_autoctl/ipaddr.c
```diff
@@ -244,3 +302,226 @@ isTestEnv(void)
...
+		log_trace("hostname \"%s\" is ipv6", hostname);
+		return IPTYPE_V6;
+	}
+	return IPTYPE_NONE;
+}
+
+
+/*
+ * findHostnameFromLocalIpAddress does a reverse DNS lookup from the local IP
+ * address that we found in fetchLocalIPAddress, given as a parameter.
```
thanodnl: the comment is a bit specific about how to get to the `localIpAddress`, wonder if we could word it a bit more generic as I think it is already used from multiple locations

src/bin/pg_autoctl/pgsql.c
```diff
@@ -1048,9 +1050,29 @@ hostname_from_uri(const char *pguri, char *hostname, int maxHostLength)
...
+			}
+		}
+
+		if (strcmp(option->keyword, "port") == 0)
+		{
+			if (option->val)
+			{
+				/* we expect a single port number in a monitor's URI */
+				*port = atoi(option->val);
+
```
thanodnl: nit; whiteline is not adding much value IMHO

tests/pgautofailover_utils.py
```diff
@@ -194,7 +194,7 @@ def create(self):
         create_command = [shutil.which('pg_autoctl'), 'create',
                           self.role.command(),
                         '--pgdata', self.datadir,
-                        '--nodename', str(self.vnode.address),
+                        #'--nodename', str(self.vnode.address),
```
thanodnl: dead code

src/bin/pg_autoctl/cli_create_drop_node.c
```diff
@@ -96,7 +102,8 @@ cli_create_config(Keeper *keeper, KeeperConfig *config)
 	 *   - configuration exists already, we need PGDATA
 	 *   - configuration doesn't exist already, we need PGDATA, and more
 	 */
-	if (!keeper_config_set_pathnames_from_pgdata(&config->pathnames, config->pgSetup.pgdata))
+	if (!keeper_config_set_pathnames_from_pgdata(&config->pathnames,
```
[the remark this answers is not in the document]
DimCitus: I was editing the general area and it just happened, I guess. I find the long lines very distracting in my setup so sometimes I would fix them just to help my reading of the code. In this instance it means there's some noise in the PR diff, sorry about that. Should I re-install the previous line length?

src/bin/pg_autoctl/cli_create_drop_node.c
```diff
@@ -219,10 +226,39 @@ cli_create_postgres(int argc, char **argv)
...
+	if (IS_EMPTY_STRING_BUFFER(config.nodename))
+	{
+		char monitorHostname[_POSIX_HOST_NAME_MAX];
+		int monitorPort = 0;
+
+		if (!hostname_from_uri(config.monitor_pguri,
+							   monitorHostname, _POSIX_HOST_NAME_MAX,
+							   &monitorPort))
+		{
+			log_fatal("Failed to parse --monitor Postgres URI: \"%s\"",
```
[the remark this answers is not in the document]
DimCitus: Fixed.

src/bin/pg_autoctl/cli_create_drop_node.c
```diff
@@ -219,10 +226,39 @@ cli_create_postgres(int argc, char **argv)
...
+					  config.monitor_pguri);
+			exit(EXIT_CODE_BAD_ARGS);
+		}
+
+		if (!discover_nodename((char *) &(config.nodename),
+							   _POSIX_HOST_NAME_MAX,
+							   monitorHostname,
+							   monitorPort))
+		{
+			log_fatal("Failed to setup --nodename, see above for details");
```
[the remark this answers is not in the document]
DimCitus: Thanks for the suggestion, fixed.

src/bin/pg_autoctl/cli_create_drop_node.c
```diff
@@ -219,10 +226,39 @@ cli_create_postgres(int argc, char **argv)
...
+			exit(EXIT_CODE_BAD_ARGS);
+		}
+	}
+	else
+	{
+		(void) check_nodename(config.nodename);
```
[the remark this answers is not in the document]
DimCitus: Added a comment about the explicit choice of WARN and continue.

src/bin/pg_autoctl/defaults.h
```diff
@@ -66,6 +66,10 @@
 /* Citus support */
 #define CITUS_EXTENSION_NAME "citus"
 
+/* Default external service provider to use to discover local IP address */
+#define PG_AUTOCTL_DEFAULT_SERVICE_NAME "8.8.8.8"
```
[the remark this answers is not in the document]
DimCitus: Renamed to `DEFAULT_INTERFACE_LOOKUP_SERVICE_NAME`.

src/bin/pg_autoctl/cli_create_drop_node.c
DimCitus: Left-over from refactoring. Good spot!

src/bin/pg_autoctl/ipaddr.c
DimCitus: Oh you mean like we had in the 80s in another part of the computing world, with the function `logcount` in Common Lisp? See http://clhs.lisp.se/Body/f_logcou.htm.

Well I didn't consider that because I didn't know about `__builtin_popcount`, thanks for mentioning. Looking into it though, it seems that we don't have to sweat much over it if our goal is performance, as explained in https://lemire.me/blog/2016/05/23/the-surprising-cleverness-of-modern-compilers/. Also, we don't run that loop nearly often enough that we would care, right?

src/bin/pg_autoctl/cli_do_show.c
DimCitus: I don't think there's a TODO left with the revision I ended up sending. Removed.

src/bin/pg_autoctl/ipaddr.c
DimCitus: The bracing policy in PostgreSQL is different from ours... added now!

src/bin/pg_autoctl/ipaddr.c
DimCitus: I don't see it here?

src/bin/pg_autoctl/ipaddr.c
DimCitus: Fixed.

src/bin/pg_autoctl/ipaddr.c
DimCitus: Fixed.

src/bin/pg_autoctl/ipaddr.c
DimCitus: Fixed.

src/bin/pg_autoctl/ipaddr.c
DimCitus: Fixed.

src/bin/pg_autoctl/ipaddr.c
DimCitus: Fixed.

src/bin/pg_autoctl/ipaddr.c
DimCitus: Removed TODO entry.

src/bin/pg_autoctl/ipaddr.c
DimCitus: Good thinking, done that way now!

src/bin/pg_autoctl/ipaddr.c
DimCitus: Done.

src/bin/pg_autoctl/ipaddr.c
DimCitus: Improved comment.

src/bin/pg_autoctl/pgsql.c
DimCitus: Removed white line,

tests/pgautofailover_utils.py
DimCitus: Removed, added a comment instead that we explicitly choose to not pass --nodename.

src/bin/pg_autoctl/cli_create_drop_node.c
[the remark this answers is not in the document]
thanodnl: no, not really.
I think we should add `citus_indent` soon so we have tooling correcting the long lines right when they are committed.

src/bin/pg_autoctl/ipaddr.c
thanodnl: the only reason to replace it would be readability. The algorithm is not trivial to read, however, popcount is also not widely known in the community. For performance in this case it is neglectable. Lets leave it like this for now.

src/bin/pg_autoctl/ipaddr.c
thanodnl: I added a commit to fix it, yet again, lets get `citus_indent` setup in the future.

marcocitus:
Tried this out on my laptop (CentOS). All seems fine, but if I use the URL returned by `SELECT pgautofailover.formation_uri()` I get:

```
$ psql postgres://172.17.74.231:6002,172.17.74.231:6001/postgres?target_session_attrs=read-write
psql: FATAL:  no pg_hba.conf entry for host "172.17.74.231", user "marco", database "postgres", SSL off
```

Not sure if that's a bad thing, putting it here for informational purposes.

DimCitus:
> When trying out the code I see this message;
> `14:38:46 WARN Failed to resolve hostname from address "10.46.110.35": nodename nor servname provided, or not known`
> what does `servname` stand for here? It is terminology that does not ring a bell to me.

This error message comes from `strerror(errno)` and is the libc wording in case of a DNS failure. I can see some documentation about it at https://sourceware.org/glibc/wiki/NameResolver ; and I think we should leave it as it is... I think `servname` is the name of the service, as in `/etc/service`, which is an english name for each well-known (reserved) port number.

src/bin/pg_autoctl/ipaddr.c
DimCitus: Yeah I we would have to figure out which version of which C compiler actually has popcount built-in, and if they are all using the same name for the feature, etc. Agree with leaving it as it is.

DimCitus:
> Tried this out on my laptop (CentOS). All seems fine, but if I use the URL returned by `SELECT pgautofailover.formation_uri()` I get:
> 
> ```
> $ psql postgres://172.17.74.231:6002,172.17.74.231:6001/postgres?target_session_attrs=read-write
> psql: FATAL:  no pg_hba.conf entry for host "172.17.74.231", user "marco", database "postgres", SSL off
> ```
> 
> Not sure if that's a bad thing, putting it here for informational purposes.

Thanks for testing and confirming that it worked for you. It's one of those patches where testing in many different OS / Network setup configuration makes sense!

About the error that's because we don't track the username in the monitor database, only the dbname. That's a design choice, it's a classic security advice to have the application code use a SQL role that is not the owner of the database, and the `--username` option is going to be that database owner.

In my testing scripts I use `psql -U approle <connection string>` and it works well. I guess we should improve our documentation about that, even when the lack of a user in the connection string is obvious: the reason why it's lacking is not.

marcocitus:
I don't think the username is the reason. I can connect just fine with `psql -p 6001 -h localhost`, but when I connect to `172.17.74.231`, my client IP also becomes `172.17.74.231` instead of `127.0.0.1` and there is no `pg_hba.conf` entry for `172.17.74.231`.

DimCitus:
> when I connect to `172.17.74.231`, my client IP also becomes `172.17.74.231` instead of `127.0.0.1` and there is no `pg_hba.conf` entry for `172.17.74.231`.

You're right, further testing showed that to me as well. The path that I tested worked because I would not copy/paste the `pg_autoctl show uri --formation default` but rather use what I know works :/ I have added an HBA rule that makes it work from the same host by default.

src/bin/pg_autoctl/ipaddr.c
```diff
@@ -274,20 +308,39 @@ fetchIPAddressFromInterfaceList(char *localIpAddress, int size)
...
-		if (ifa->ifa_addr->sa_family == AF_INET && !(ifa->ifa_flags & IFF_LOOPBACK))
+		if (ifaddr->ifa_flags & IFF_LOOPBACK)
+		{
+			continue;
+		}
+
+		/*
+		 * We only support IPv4 here, also this function is only called in test
+		 * environment where we run in a docker container with a network
+		 * namespace in which we use only IPv4, so that's ok.
```
thanodnl: maybe a good idea to have some tests run with ipv6, given the `SIGABORT` we ran into today.
Not something to do in this PR, but maybe open a ticket.

DimCitus:
> do we need the verbose output in our tests after you found the problem?

Given that the test suite only outputs the output of the command when they fail, I think it's better to have the `-vvv` (very very verbose) output there, yes.

src/bin/pg_autoctl/monitor_pg_init.c
```diff
@@ -223,9 +226,11 @@ monitor_install(PostgresSetup pgSetupOption, bool checkSettings)
 	 * Now allow nodes on the same network to connect to pg_auto_failover
 	 * database.
 	 */
-	if (!pghba_enable_lan_cidr(&postgres.sqlClient, HBA_DATABASE_DBNAME,
-							  PG_AUTOCTL_MONITOR_DBNAME,
-							  PG_AUTOCTL_MONITOR_USERNAME, "trust", NULL))
+	if (!pghba_enable_lan_cidr(&postgres.sqlClient,
+							   HBA_DATABASE_DBNAME,
+							   PG_AUTOCTL_MONITOR_DBNAME,
+							   (char *)nodename,
```
thanodnl: instead of casting here, can we make `pghba_enable_lan_cidr` a `const`? I don't see any reason to change the nodename by that function.
seeing the definition below, I think all `char *` should be `const`s

DimCitus: Good point, done that way now.

PR #17, opened 2019-05-16
Implement checking and upgrading of the pgautofailover extension.

DimCitus:
The pg_autoctl tool embeds the version of the extension that it is
compatible with on the monitor. The idea is that those versions are kept in
sync at the source code and packaging level, and a given pg_autoctl command
line is compatible with a single version of the monitor.

At startup on the monitor the command `pg_autoctl run` checks the monitor's
version and if necessary runs a command `ALTER EXTENSION ... UPDATE TO ...`
the expected version. It is expected that the extension packaging provides
the necessary upgrade paths.

On the keeper, we check that the version number of the extension on the
monitor is compatible with our expectations, and exit before entering the
main loop when that's not the case. We refrain from doing "upgrade at a
distance" and tell the admin (who's reading the logs, maybe?) to please take
care of the situation on the monitor node.

### DIFF
--- a/src/bin/pg_autoctl/cli_do_monitor.c
+++ b/src/bin/pg_autoctl/cli_do_monitor.c
@@ -116,7 +116,9 @@ static CommandLine monitor_resume_node_command =
 
 static CommandLine monitor_version_command =
 	make_command("version",
-				 "Check and display monitor's extension version",
+				 "Check that monitor version is "
+				 PG_AUTOCTL_EXTENSION_VERSION
+				 "; alter extension update if not",
 				 " [ --pgdata ]",
 				 KEEPER_CLI_PGDATA_OPTION,
 				 keeper_cli_getopt_pgdata,
@@ -574,7 +576,10 @@ cli_monitor_resume_node(int argc, char **argv)
 
 
 /*
- * cli_monitor_version checks and displays the monitor extension version.
+ * cli_monitor_version ensures that the version of the monitor is the one that
+ * is expected by pg_autoctl too. When that's not the case, the command issues
+ * an ALTER EXTENSION ... UPDATE TO ... to ensure that the monitor is now
+ * running the expected version number.
  */
 static void
 cli_monitor_version(int argc, char **argv)
--- a/src/bin/pg_autoctl/cli_do_monitor.c
+++ b/src/bin/pg_autoctl/cli_do_monitor.c
@@ -43,6 +43,7 @@ static void keeper_cli_monitor_register_node(int argc, char **argv);
 static void keeper_cli_monitor_node_active(int argc, char **argv);
 static void cli_monitor_pause_node(int argc, char **argv);
 static void cli_monitor_resume_node(int argc, char **argv);
+static void cli_monitor_version(int argc, char **argv);
 
 
 static CommandLine monitor_get_primary_command =
@@ -113,12 +114,21 @@ static CommandLine monitor_resume_node_command =
 				 keeper_cli_getopt_pgdata,
 				 cli_monitor_resume_node);
 
+static CommandLine monitor_version_command =
+	make_command("version",
+				 "Check and display monitor's extension version",
+				 " [ --pgdata ]",
+				 KEEPER_CLI_PGDATA_OPTION,
+				 keeper_cli_getopt_pgdata,
+				 cli_monitor_version);
+
 static CommandLine *monitor_subcommands[] = {
 	&monitor_get_command,
 	&monitor_register_command,
 	&monitor_node_active_command,
 	&monitor_pause_node_command,
 	&monitor_resume_node_command,
+	&monitor_version_command,
 	NULL
 };
 
@@ -560,4 +570,31 @@ cli_monitor_resume_node(int argc, char **argv)
 
 	log_info("Node %s:%d will exit from maintenance state soon",
 			 keeper.config.nodename, keeper.config.pgSetup.pgport);
+}
+
+
+/*
+ * cli_monitor_version checks and displays the monitor extension version.
+ */
+static void
+cli_monitor_version(int argc, char **argv)
+{
+	KeeperConfig config = keeperOptions;
+	Monitor monitor = { 0 };
+	MonitorExtensionVersion version = { 0 };
+
+	if (!monitor_init_from_pgsetup(&monitor, &config.pgSetup))
+	{
+		/* errors have already been logged */
+		exit(EXIT_CODE_BAD_ARGS);
+	}
+
+	/* Check version compatibility */
+	if (!monitor_ensure_extension_version(&monitor, &version))
+	{
+		/* errors have already been logged */
+		exit(EXIT_CODE_MONITOR);
+	}
+
+	fprintf(stdout, "%s\n", version.installedVersion);
 }
--- a/src/bin/pg_autoctl/monitor.c
+++ b/src/bin/pg_autoctl/monitor.c
@@ -1271,8 +1271,8 @@ parseExtensionVersion(void *ctx, PGresult *result)
 
 	if (PQgetisnull(result, 0, 0) || PQgetisnull(result, 0, 1))
 	{
-		log_error("pgautofailover default_version or installed_version "
-				  "returned by monitor is NULL");
+		log_error("default_version or installed_version for extension \"%s\" "
+				  "is NULL ", PG_AUTOCTL_MONITOR_EXTENSION_NAME);
 		context->parsedOK = false;
 		return;
 	}
@@ -1322,8 +1322,8 @@ monitor_extension_update(Monitor *monitor, char *targetVersion)
  * not the case, we blindly try to update the extension version on the monitor
  * to the target version we have in our default.h.
  *
- * NOTE: we don't check here if that's an upgrade or a downgrade, we rely on
- * the extension's upgrade path to be free of downgrade paths (such as
+ * NOTE: we don't check here if the update is an upgrade or a downgrade, we
+ * rely on the extension's update path to be free of downgrade paths (such as
  * pgautofailover--1.2--1.1.sql).
  */
 bool
@@ -1360,7 +1360,7 @@ monitor_ensure_extension_version(Monitor *monitor)
 		if (!prepare_connection_to_current_system_user(monitor,
 													   &dbOwnerMonitor))
 		{
-			log_error("Failed to upgrade extension \"%s\" to version \"%s\": "
+			log_error("Failed to update extension \"%s\" to version \"%s\": "
 					  "failed prepare a connection string to the "
 					  "monitor as the database owner",
 					  PG_AUTOCTL_MONITOR_EXTENSION_NAME,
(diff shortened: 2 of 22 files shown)


src/bin/pg_autoctl/monitor.c
```diff
@@ -1190,3 +1199,261 @@ monitor_get_notifications(Monitor *monitor)
...
+	if (PQnfields(result) != 2)
+	{
+		log_error("Query returned %d columns, expected 2", PQnfields(result));
+		context->parsedOK = false;
+		return;
+	}
+
+	if (PQgetisnull(result, 0, 0) || PQgetisnull(result, 0, 1))
+	{
+		log_error("pgautofailover default_version or installed_version "
```
thanodnl: can we use `PG_AUTOCTL_MONITOR_EXTENSION_NAME` instead of hardcoding the name of the extension?

src/bin/pg_autoctl/monitor.c
```diff
@@ -1190,3 +1199,261 @@ monitor_get_notifications(Monitor *monitor)
...
+					  "monitor as the database owner",
+					  PG_AUTOCTL_MONITOR_EXTENSION_NAME,
+					  PG_AUTOCTL_EXTENSION_VERSION);
+			return false;
+		}
+
+		if (!monitor_extension_update(&dbOwnerMonitor,
+									  PG_AUTOCTL_EXTENSION_VERSION))
+		{
+			log_fatal("Failed to update extension \"%s\" to version \"%s\" "
```
thanodnl: we are talking about `update` and `upgrade` in the same function. Lets choose one and stick with it for the warnings in this function.

src/bin/pg_autoctl/monitor.c
```diff
@@ -1190,3 +1199,261 @@ monitor_get_notifications(Monitor *monitor)
...
+	return true;
+}
+
+
+/*
+ * prepare_connection_to_current_system_user changes a given pguri to remove
+ * its "user" connection parameter, filling in the pre-allocated keywords and
+ * values string arrays.
+ *
+ * Postgres docs at the following address show 30 connection parameters, so the
```
thanodnl: Is this perhaps defined in a constant in libpq? would be good if we could use that, otherwise we need to keep track when libpq adds a new parameter.

src/bin/pg_autoctl/monitor.c
```diff
@@ -1190,3 +1199,261 @@ monitor_get_notifications(Monitor *monitor)
...
+		if (strcmp(option->keyword, "user") == 0)
+		{
+			/* skip the user, $USER is what we want to use here */
+			continue;
+		}
+		else if (option->val)
+		{
+			keywords[argCount] = option->keyword;
+			values[argCount] = option->val;
+			++argCount;
```
thanodnl: lets be defensive here and error when we go over the size of the array.

src/bin/pg_autoctl/pgsql.c
```diff
@@ -1344,3 +1344,74 @@ pgsql_listen(PGSQL *pgsql, char *channels[])
...
+	{
+		log_error("Failed to update extension \"%s\" to version \"%s\": %s",
+				  extname, version,
+				  PQerrorMessage(connection));
+		pgsql_finish(pgsql);
+		return false;
+	}
+
+	/* now build the SQL command */
+	snprintf(command, BUFSIZE, "ALTER EXTENSION %s UPDATE TO %s",
```
thanodnl: guess we want to error when `snprintf` needed to cutoff the string formatting due to reaching `BUFSIZE`. What woud the behaviour be in the current implementation? Guess it sends an incomplete query to postgres :(

src/bin/pg_autoctl/monitor.c
DimCitus: In `postgresql/src/interfaces/libpq/fe-connect.c` we see the following code, which means we're out of luck. I pushed it to 41 (10 extra) and added a check.

``` c
static const internalPQconninfoOption PQconninfoOptions[] = {
...
	/*
	 * Get enough memory for all options in PQconninfoOptions, even if some
	 * end up being filtered out.
	 */
	options = (PQconninfoOption *) malloc(sizeof(PQconninfoOption) * sizeof(PQconninfoOptions) / sizeof(PQconninfoOptions[0]));
```

src/bin/pg_autoctl/monitor.c
DimCitus: Fixed.

src/bin/pg_autoctl/monitor.c
DimCitus: Fixed.

src/bin/pg_autoctl/pgsql.c

DimCitus: Oh and 1024 bytes for `ALTER EXTENSION %s UPDATE TO %s`, and with the extension name of type `name` (limited to 64 bytes including the null byte at the end of the string), you need a pretty long version string before that can happen. I'd happily call it a bug. That said, we can implement some checking, that's easy and might save us in case of strange data corruption on file names or something.

src/bin/pg_autoctl/monitor.c
DimCitus: Fixed.

DimCitus:
> Is there an easy way to test this locally? And more importantly, do you think it is feasible to test this automatically? Otherwise I guess we should make it part of the packaging preparations to make sure it can upgrade cleanly.

For testing locally, I edit PG_AUTOCTL_EXTENSION_VERSION to "1.1" and then start `pg_autoctl run`, which I can see does try the `alter extension ... update to ...`, which then fails, on the monitor. On a postgres node, it just refuses to start.

I'm not sure how to unit-test that without providing an newer version of the extension that would not be installed by default and then activating the upgrade path facility. It also requires to change the code each time, because the version we checked is hard-coded (which is good, it means each `pg_autoctl` binary knows which version of the extension it requires).

Finally, yes I think providing and testing upgrade path is a packaging problem.

marcocitus:
> It is expected that the extension packaging provides the necessary upgrade paths.

The SQL steps for the ALTER EXTENSION should be in this repo. You can follow the example in Citus: https://github.com/citusdata/citus/tree/master/src/backend/distributed

> I'm not sure how to unit-test that without providing an newer version of the extension that would not be installed by default and then activating the upgrade path facility.

You can create an extension with a given version, see Citus:
https://github.com/citusdata/citus/blob/master/src/test/regress/sql/multi_extension.sql#L59

marcocitus:
For upgrading an extension with C implementations, it is essential to keep the schema and the binary in sync to prevent crashes. The ALTER EXTENSION should be done immediately after a restart that upgrades the binary.

In Citus, we also protect against any incompatibility between SQL and binary by starting every single Citus operation with `CheckCitusVersion`, which errors out immediately if the SQL is behind on the binary. For example in `create_distributed_table` https://github.com/citusdata/citus/blob/master/src/backend/distributed/commands/create_distributed_table.c#L133

DimCitus:
I implemented the version checks, using an SPI query because performance characteristics of pgautofailover are not comparable with the constraints of Citus itself. Let's keep things as simple as possible.

Then again, about the testing of the upgrade path... I agree with you that the Postgres level packaging of the extension (SQL upgrade paths) are going to be maintained in the main repository and taken care of by `make install`. At the moment though we have a single version of the extension, and also `pg_autoctl create monitor` directly installs the right version of it. So to build a test case with an upgrade path might necessitate us coming with new testing tools / framework / facilities. Working on that next.

src/bin/pg_autoctl/cli_do_monitor.c
```diff
@@ -561,3 +571,30 @@ cli_monitor_resume_node(int argc, char **argv)
 	log_info("Node %s:%d will exit from maintenance state soon",
 			 keeper.config.nodename, keeper.config.pgSetup.pgport);
 }
+
+
+/*
+ * cli_monitor_version checks and displays the monitor extension version.
```
thanodnl: This comment is not reflecting what we are doing.

the call to `monitor_ensure_extension_version` has side effects that it tries to upgrade as well. I think it is undesirable to have a command that should print the installed version to try and change the version installed.

Could we make this command adhere to the comment on this function? eg. show the version, without ensuring it is the version we are expecting given the version of `pg_autoctl`.

EDIT: I now see this is under the do commands, in that case we can decide that it will update to the right version, but that should be reflected both in this comment, and in the `--help` of the cli command

src/bin/pg_autoctl/monitor.c
```diff
@@ -1190,3 +1199,291 @@ monitor_get_notifications(Monitor *monitor)
...
+ * pg_available_extension_versions catalogs.
+ */
+static void
+parseExtensionVersion(void *ctx, PGresult *result)
+{
+	MonitorExtensionVersionParseContext *context =
+		(MonitorExtensionVersionParseContext *) ctx;
+
+	char *value = NULL;
+	int length;
```
thanodnl: nit; uninitialized value

src/monitor/pg_auto_failover.c
```diff
@@ -75,6 +76,11 @@ StartMonitorNode(void)
 {
 	BackgroundWorker worker;
 
+	DefineCustomBoolVariable("pgautofailover.enable_version_checks",
+							 "Enable extension version compatiblity checks",
+							 NULL, &EnableVersionChecks, true, PGC_SIGHUP,
```
thanodnl: searching through this PR I came to the conclusion this variable is not read ever. What is the use of this GUC? should we have a check for the value of `EnableVersionChecks ` in `checkPgAutoFailoverVersion` to blindly continue when this GUC is set to false?

src/bin/pg_autoctl/cli_do_monitor.c
DimCitus: Good point, fixed the comments and the help message that way:

```
$ PG_AUTOCTL_DEBUG=1 pg_autoctl do monitor 
Available commands:
  pg_autoctl do monitor
  + get       Get information from the monitor
    register  Register the current node with the monitor
    active    Call in the pg_auto_failover Node Active protocol
    pause     Pause pg_auto_failover on this node
    resume    Resume pg_auto_failover on this node
    version   Check that monitor version is 1.0; alter extension update if not
```

src/bin/pg_autoctl/monitor.c
DimCitus: Fixed.

src/monitor/pg_auto_failover.c
DimCitus: Good spot! now fixed.

src/bin/pg_autoctl/cli_do_monitor.c
```diff
@@ -116,7 +116,9 @@ static CommandLine monitor_resume_node_command =
 
 static CommandLine monitor_version_command =
 	make_command("version",
-				 "Check and display monitor's extension version",
+				 "Check that monitor version is "
```
thanodnl: I expected this change to reflect in the documentation as well, but the PR does not show updated documentation. You might want to update https://github.com/citusdata/pg_auto_failover/blob/d8ba26f47a60eaf1e1dc63ae67279553984d84f5/docs/ref/reference.rst#pg_autoctl-do with this PR as well

PR #24, opened 2019-06-14
Systemd integration

DimCitus:
Add the ability for `pg_autoctl` to produce the systemd unit configuration file for the pgautofailover service, using both the current PGDATA and absolute filename for the current program in use.


DimCitus:
In `config.c` a new function `normalize_filename()` is introduced. I though I needed it in the completion of this PR, but actually I don't. Still, the code is cleaner as a result, so I kept the refactoring around in the PR.

src/bin/pg_autoctl/cli_systemd.c
```diff
@@ -0,0 +1,127 @@
...
+cli_systemd_cat_service_file(int argc, char **argv)
+{
+	SystemdServiceConfig config = systemdOptions;
+	PostgresSetup pgSetup = { 0 };
+
+	(void) systemd_config_init(&config, pgSetup.pgdata);
+
+	log_info("HINT: to complete a systemd integration, "
+			 "run the following commands:");
+	log_info("pg_autoctl -q show systemd --pgdata %s | sudo tee %s",
```
thanodnl: `pgdata` could contain spaces, I think it is beneficial to output the directory with quotes around it.

src/bin/pg_autoctl/systemd_config.c
```diff
@@ -0,0 +1,119 @@
...
+
+	/* time to setup config->pathnames.systemd */
+	snprintf(config->pathnames.systemd, MAXPGPATH,
+			 "/etc/systemd/system/%s", KEEPER_SYSTEMD_FILENAME);
+
+	/* adjust defaults to known values from the config */
+	strlcpy(config->WorkingDirectory, config->pgSetup.pgdata, MAXPGPATH);
+
+	snprintf(config->EnvironmentPGDATA, BUFSIZE,
+			 "\"PGDATA=%s\"", config->pgSetup.pgdata);
```
thanodnl: I think we might want to test this with a space in the directory as well, will that work with systemd like this or do we need more quoting?

I was reading the spec and there seems to be different behaviour with quoting and `$X` vs `${X}` in systemd. https://www.freedesktop.org/software/systemd/man/systemd.service.html#Command%20lines

src/bin/pg_autoctl/cli_systemd.c
DimCitus: Good spot, fixed now!

src/bin/pg_autoctl/systemd_config.c
DimCitus: Using `${X}` is useful when using an environment variable in the command line, not when defining the environment variable value, if I read correctly. At definition time it seems best to use single quoting, as per the documentation you're sharing (thanks!). I've done that in the PR.

DimCitus: Will test tomorrow with a PGDATA path that contains spaces, then squash&merge!

PR #28, opened 2019-06-19
Add --auth option to configure authentication method

mtuncer:
we can now configure auth param in pg_hba for following users

Monitor:
autoctl_node : user to connect to monitor from pg_autoctl nodes

Nodes:
pgautofailover_monitor : user to connect to pg_autoctl nodes from the monitor

Replication user auth mode is also set to provided value if the replication user password is also set.

### DIFF
--- a/docs/operations.rst
+++ b/docs/operations.rst
@@ -19,6 +19,8 @@ environment variable, and checks whether PostgreSQL is already running. If
 Postgres is detected, the new node is registered in SINGLE mode, bypassing
 the monitor's role assignment policy.
 
+.. _pg_auto_failover_security:
+
 Security
 --------
 Connections between monitor and data nodes use *trust* authentication by
@@ -28,6 +30,15 @@ parameter when creating monitor or data Node. Any auth method supported by
 PostgreSQL could be used here. Please refer to
 https://www.postgresql.org/docs/current/auth-pg-hba-conf.html
 for available options.
+
+Security for following connections should be considered when setting up
+`.pgpass` file.
+  1. health check connection from monitor for `pgautofailover_monitor` user
+  2. connections for `pg_autoctl` command from data nodes to monitor for `autoctl_node` user
+  3. replication connections from secondary to primary data nodes for `replication` user.
+     Notice that primary and secondary nodes change during failover. Thus this setting
+     should be done on both primary and secondary nodes.
+  4. settings need to be updated after a new node is added.
 
 Operations
 ----------
--- a/docs/operations.rst
+++ b/docs/operations.rst
@@ -19,6 +19,16 @@ environment variable, and checks whether PostgreSQL is already running. If
 Postgres is detected, the new node is registered in SINGLE mode, bypassing
 the monitor's role assignment policy.
 
+Security
+--------
+Connections between monitor and data nodes use *trust* authentication by
+default. This lets accounts used by ``pg_auto_failover`` to connect to nodes
+without needing a password. Default behaviour could be changed using ``--auth``
+parameter when creating monitor or data Node. Any auth method supported by
+PostgreSQL could be used here. Please refer to
+https://www.postgresql.org/docs/current/auth-pg-hba-conf.html
+for available options.
+
 Operations
 ----------
 
--- a/docs/operations.rst
+++ b/docs/operations.rst
@@ -27,9 +27,10 @@ Connections between monitor and data nodes use *trust* authentication by
 default. This lets accounts used by ``pg_auto_failover`` to connect to nodes
 without needing a password. Default behaviour could be changed using ``--auth``
 parameter when creating monitor or data Node. Any auth method supported by
-PostgreSQL could be used here. Please refer to
-https://www.postgresql.org/docs/current/auth-pg-hba-conf.html
+PostgreSQL could be used here. Please refer to `PostgreSQL pg_hba documentation`__
 for available options.
+
+__ https://www.postgresql.org/docs/current/auth-pg-hba-conf.html
 
 Security for following connections should be considered when setting up
 `.pgpass` file.
@@ -39,6 +40,10 @@ Security for following connections should be considered when setting up
      Notice that primary and secondary nodes change during failover. Thus this setting
      should be done on both primary and secondary nodes.
   4. settings need to be updated after a new node is added.
+See `PostgreSQL documentation`__ on setting up `.pgpass` file.
+
+__ https://www.postgresql.org/docs/current/libpq-pgpass.html
+
 
 Operations
 ----------
--- a/docs/ref/reference.rst
+++ b/docs/ref/reference.rst
@@ -91,6 +91,7 @@ commands are dealing with the monitor:
         --pgdata      path to data directory
         --pgport      PostgreSQL's port number
         --nodename    hostname by which postgres is reachable
+        --auth        authentication method for connections from data nodes
 
      The ``--pgdata`` option is mandatory and default to the environment
      variable ``PGDATA``. The ``--pgport`` default value is 5432, and the
@@ -119,6 +120,11 @@ commands are dealing with the monitor:
      You may use the `--nodename` command line option to bypass the whole
      DNS lookup based process and force the local node name to a fixed
      value.
+     
+     The ``--auth`` option allows setting up authentication method to be used
+     for connections from data nodes with ``autoctl_node`` user.
+     If this option to be used, please make sure password is manually set on
+     the monitor, and appropriate setting is added to `.pgpass` file on data node.      
 
   - ``pg_autoctl run``
 
@@ -269,6 +275,7 @@ The other commands accept the same set of options.
     --nodename    pg_auto_failover node
     --formation   pg_auto_failover formation
     --monitor     pg_auto_failover Monitor Postgres URL
+    --auth        authentication method for connections from monitor
     --allow-removing-pgdata   Allow pg_autoctl to remove the database directory
 
 Three different modes of initialization are supported by this command,
@@ -320,6 +327,11 @@ prepare a new secondary system from scratch.
 When `--nodename` is omitted, it is computed as above (see
 :ref:`_pg_autoctl_create_monitor`), with the difference that step 1 uses the
 monitor IP and port rather than the public service 8.8.8.8:53.
+
+The ``--auth`` option allows setting up authentication method to be used when
+monitor node makes a connection to data node with `pgautofailover_monitor` user.
+If this option to be used, please make sure password is manually set on the data
+node, and appropriate setting is added to `.pgpass` file on monitor node.
 
 .. _pg_autoctl_configuration:
 
--- a/docs/ref/reference.rst
+++ b/docs/ref/reference.rst
@@ -123,8 +123,10 @@ commands are dealing with the monitor:
      
      The ``--auth`` option allows setting up authentication method to be used
      for connections from data nodes with ``autoctl_node`` user.
-     If this option to be used, please make sure password is manually set on
-     the monitor, and appropriate setting is added to `.pgpass` file on data node.      
+     If this option is used, please make sure password is manually set on
+     the monitor, and appropriate setting is added to `.pgpass` file on data node.
+     
+     See :ref:`pg_auto_failover_security` for notes on `.pgpass`
 
   - ``pg_autoctl run``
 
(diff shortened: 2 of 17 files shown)


src/bin/pg_autoctl/primary_standby.c
```diff
@@ -411,7 +411,17 @@ primary_add_standby_to_hba(LocalPostgresServer *postgres, char *standbyHostname,
 	PGSQL *pgsql = &(postgres->sqlClient);
 	PostgresSetup *postgresSetup = &(postgres->postgresSetup);
 	char hbaFilePath[MAXPGPATH];
-	char* authMethod = replicationPassword ? "md5" : "trust";
+	char* authMethod = NULL;
```
DimCitus: I am more used the the following spelling of the same logic, what do you think?

``` c
    char *authMethod = "trust";

    if (replicationPassword)
    {
```

Also please write `char *` with a space before the star, rather than how you did it... if only to look like the rest of the whole code base.

mtuncer: agreed, I had only changed the right part of the declaration. Now fixed * part.

tests/pgautofailover_utils.py
```diff
@@ -96,6 +98,10 @@ def connection_string(self):
         Returns a connection string which can be used to connect to this postgres
         node.
         """
+        if (self.password is not None):
```
DimCitus: Should be written:

```
    if self.password:
```

tests/pgautofailover_utils.py
```diff
@@ -358,10 +364,21 @@ def create(self):
                         '--pgdata', self.datadir,
                         '--pgport', str(self.port),
                         '--nodename', self.nodename]
+        
+        if self.auth_method is not None:
```
DimCitus: Should be written:

```
    if self.auth_method:
```

tests/pgautofailover_utils.py
```diff
@@ -358,10 +364,21 @@ def create(self):
...
+            init_command.extend(['--auth', self.auth_method])    
+        
         init_proc = self.vnode.run(init_command)
         wait_or_timeout_proc(init_proc,
                              name="create monitor",
                              timeout=COMMAND_TIMEOUT)
+        if self.password is not None:
+            alter_user_set_passwd_command = 'alter user ' + self.username + ' with password' + ' \'' + self.password + '\''
```
DimCitus: Please use Python string building facilities here:

``` python
    alter_user_set_passwd_command = "alter user %s with password \"%s\"" % (self.username, self.password)
```

tests/test_auth.py
```diff
@@ -0,0 +1,41 @@
...
+    monitor.wait_until_pg_is_running()
+
+def test_001_init_primary():
+    global node1
+    node1 = cluster.create_datanode("/tmp/node1", auth_method="md5", password="EasyPassword")
+    node1.create()
+    node1.run()
+    assert node1.wait_until_state(target_state="single")
+
+def test_001_stop_postgres():
```
DimCitus: This test is not useful here I think.

tests/test_auth.py
```diff
@@ -0,0 +1,41 @@
+import pgautofailover_utils as pgautofailover
+from nose.tools import *
+
+cluster = None
+node1 = None
+node2 = None
+node3 = None
```
DimCitus: Unused, please remove.

src/bin/pg_autoctl/primary_standby.c
```diff
@@ -411,7 +411,17 @@ primary_add_standby_to_hba(LocalPostgresServer *postgres, char *standbyHostname,
 	PGSQL *pgsql = &(postgres->sqlClient);
 	PostgresSetup *postgresSetup = &(postgres->postgresSetup);
 	char hbaFilePath[MAXPGPATH];
-	char* authMethod = replicationPassword ? "md5" : "trust";
+	char *authMethod = NULL;
```
DimCitus: I think it would be easier to read that way:

``` C
	char *authMethod = "trust";

	if (replicationPassword)
	{
		authMethod = pg_setup_get_auth_method(postgresSetup);
	}    
```

tests/pgautofailover_utils.py
```diff
@@ -41,28 +41,28 @@ def __init__(self, networkNamePrefix="pgauto", networkSubnet="172.27.1.0/24"):
         self.monitor = None
         self.datanodes = []
 
-    def create_monitor(self, datadir, port=5432, nodename=None):
+    def create_monitor(self, datadir, port=5432, nodename=None, auth_method=None, password=None):
```
DimCitus: What about using `authMethod`?

tests/pgautofailover_utils.py
DimCitus: I think we have 4 different Postgres users that we use at different points:

  1. the `--username`, `docker` in the test environment (OS logged in user)
  2. `autoctl`, the owner of the `pg_auto_failover` database
  3. `autoctl_node`, used by pg_autoctl in the API to the monitor, as per `pg_autoctl show uri`, or the one used in the `pg_autoctl create Postgres --monitor` option
  4. `pgautofailover_monitor` when the monitor connects to the Postgres nodes for health checks

Which user are we setting a password for in that code path? and which user are we providing a password to in the connection string?

docs/ref/reference.rst
```diff
@@ -119,6 +120,11 @@ commands are dealing with the monitor:
      You may use the `--nodename` command line option to bypass the whole
      DNS lookup based process and force the local node name to a fixed
      value.
+     
+     The ``--auth`` option allows setting up authentication method to be used
+     for connections from data nodes with ``autoctl_node`` user.
+     If this option to be used, please make sure password is manually set on
```
DimCitus: I think this should be written “If this option is used”, or maybe “is to be used”, though I prefer the first variant.

docs/ref/reference.rst
```diff
@@ -321,6 +328,11 @@ When `--nodename` is omitted, it is computed as above (see
 :ref:`_pg_autoctl_create_monitor`), with the difference that step 1 uses the
 monitor IP and port rather than the public service 8.8.8.8:53.
 
+The ``--auth`` option allows setting up authentication method to be used when
+monitor node makes a connection to data node with `pgautofailover_monitor` user.
+If this option to be used, please make sure password is manually set on the data
```
DimCitus: Same as before, “if this option is used”. Could also be written “When using this option”.

docs/ref/reference.rst
```diff
@@ -119,6 +120,11 @@ commands are dealing with the monitor:
      You may use the `--nodename` command line option to bypass the whole
      DNS lookup based process and force the local node name to a fixed
      value.
+     
+     The ``--auth`` option allows setting up authentication method to be used
+     for connections from data nodes with ``autoctl_node`` user.
+     If this option to be used, please make sure password is manually set on
+     the monitor, and appropriate setting is added to `.pgpass` file on data node.
```
DimCitus: I think we should expand on the `.pgpass` idea with a list of HBA rules needed for pg_auto_failover:

  - the monitor health check, from the monitor to the Postgres nodes,
  - the pg_autoctl connections from the Postgres nodes to the monitor, using the `autoctl_node` user,
  - the replication connection from the secondary to the primary,
  - the replication connection after a failover, so from the current primary to the current secondary,
  - what about replacing a node with a new one?

tests/pgautofailover_utils.py
mtuncer: fixed

tests/pgautofailover_utils.py
mtuncer: monitor and data nodes now provide which user needs to authenticated. user names and passwords are hard coded at this time.

connections string checks the node requires authentication, and if user password is supplied during provisioning, then creates the connection string accordingly

tests/test_auth.py
mtuncer: ok removed

tests/test_auth.py
mtuncer: done

src/bin/pg_autoctl/primary_standby.c
mtuncer: ok

tests/pgautofailover_utils.py
mtuncer: done

docs/ref/reference.rst
mtuncer: ok

docs/ref/reference.rst
mtuncer: fixed

docs/ref/reference.rst
mtuncer: added a paragraph on .pgpass on Security section

tests/pgautofailover_utils.py
```diff
@@ -121,6 +129,18 @@ def run_sql_query(self, query, *args):
             except psycopg2.ProgrammingError:
                 return None
 
+    def create_authenticated_users(self):
+        if self.authMethod is None:
```
DimCitus: I would do it the other way around, it'd be less surprising:

``` python
    if self.authMethod:
        for authenticatedUser in ...:
```

tests/pgautofailover_utils.py
```diff
@@ -284,6 +304,10 @@ def create(self):
                              name="keeper init",
                              timeout=COMMAND_TIMEOUT)
 
+        if (self.get_state() == 'primary'):
```
DimCitus: Now that is strange. Certainly we want to be able to authenticate the `pgautofailover_monitor` user before we get to be primary, like in WAIT_PRIMARY stage already? Also, what about monitoring the secondary? Feels strange.

tests/pgautofailover_utils.py
```diff
@@ -284,6 +304,10 @@ def create(self):
...
+            super().create_authenticated_users()
```
DimCitus: Reading the test cases more closely, I'm now thinking that:

  1. the method name sounds wrong, because it does not create any user, only adds passwords to them
  2. calling that automatically from the test _infrastructure_ looks like the wrong place to me now, I think we should expose the API and then call it in the test file where we pick the authMethod to be md5, that's where we want to install and control the passwords after all

tests/test_auth.py
```diff
@@ -0,0 +1,37 @@
...
+def teardown_module():
+    cluster.destroy()
+
+def test_000_create_monitor():
+    monitor = cluster.create_monitor("/tmp/monitor", authMethod="md5")
+    monitor.wait_until_pg_is_running()
+
+def test_001_init_primary():
+    global node1
+    node1 = cluster.create_datanode("/tmp/node1", authMethod="md5")
```
DimCitus: I think we should call the method that sets the passwords at this point in the test file, after having created a cluster with MD5 authentication policy.

tests/test_auth.py
```diff
@@ -0,0 +1,37 @@
...
+    node1.run()
+    assert node1.wait_until_state(target_state="single")
+
+def test_002_create_t1():
+    node1.run_sql_query("CREATE TABLE t1(a int)")
+    node1.run_sql_query("INSERT INTO t1 VALUES (1), (2)")
+
+def test_003_init_secondary():
+    global node2
+    node2 = cluster.create_datanode("/tmp/auth/node2", authMethod="md5")
```
DimCitus: And so we would call the password setting method at this point too.

mtuncer: secondary gets the password setting from primary

tests/test_auth.py
mtuncer: added set_user_password method to set individual user passwords

tests/pgautofailover_utils.py
mtuncer: removed

mtuncer: because it was not needed anymore after moving password setting responsibility to test

tests/pgautofailover_utils.py
mtuncer: function is gone altogether

docs/operations.rst
```diff
@@ -19,6 +19,27 @@ environment variable, and checks whether PostgreSQL is already running. If
...
+Connections between monitor and data nodes use *trust* authentication by
+default. This lets accounts used by ``pg_auto_failover`` to connect to nodes
+without needing a password. Default behaviour could be changed using ``--auth``
+parameter when creating monitor or data Node. Any auth method supported by
+PostgreSQL could be used here. Please refer to
+https://www.postgresql.org/docs/current/auth-pg-hba-conf.html
+for available options.
+
+Security for following connections should be considered when setting up
+`.pgpass` file.
```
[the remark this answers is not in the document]
mtuncer: added

tests/pgautofailover_utils.py
mtuncer: addressed both concerns
1 - removed the function, replaced with set_user_password function to set individual user password
2 - called the new function from test file

tests/test_auth.py
DimCitus: ah yes, indeed!

tests/pgautofailover_utils.py
DimCitus: I'm not sure I understand how that works wrt to setting the password both on the connection string and for the user on the target database...

DimCitus:
Consider squash-and-merge and editing a proper commit message! :shipit:

mtuncer:
we are using target node's connection_string() method when connecting.

PR #30, opened 2019-06-25
Fix install recipe

### LINKED ISSUE #13: README Please provide alternate recipe for adding repos
curl https://install.citusdata.com/community/deb.sh | sudo bash

Encouraging people to blindly execute content from the internet as root is as dangerous as it is convenient - on both security grounds and failure modes. Please provide alternate default instructions to add your repos.

(I hasten to add that this project looks awesome and I’ll be looking to see if we can drop some homegrown equivalents - thank you for opening this!)

DimCitus:
Add the manual steps to install our repositories and packages, in addition to the quick install method requiring the scary `curl | sudo bash` stanza.

Fixes #13.

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -12,6 +12,15 @@ PostgreSQL 11.
 Ubuntu or Debian
 ----------------
 
+Quick install
+~~~~~~~~~~~~~
+
+The following installation method downloads a bash script that automates
+several steps. The full script is available for review at our `package cloud
+installation instructions`__ page.
+
+__ https://packagecloud.io/citusdata/community/install#bash
+  
 .. code-block:: bash
 
   # add the required packages to your system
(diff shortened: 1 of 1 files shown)


docs/install.rst
```diff
@@ -12,6 +12,15 @@ PostgreSQL 11.
 Ubuntu or Debian
 ----------------
 
+Quick install
+~~~~~~~~~~~~~
+
+The following installation method downloads a bash script that automates
+several steps. The full script is available for review at our `package cloud
+installation instructions page`__ url.
```
mtuncer: may drop `url` word

mtuncer: more like
package cloud installation instructions __ page.

docs/install.rst
```diff
@@ -50,15 +116,15 @@ Here's a sample output from the command:
 
    $ pg_autoctl show systemd
```
mtuncer: `pg_autoctl show systemd` errors out when `--pgdata` is not provided. We are not setting `--pgdata` in the tutorial. Perhaps we should either set environment variable or provide the value

docs/install.rst
DimCitus: Fixed that way, thanks!

docs/install.rst
DimCitus: Fixed!

PR #35, opened 2019-07-01
Implement pg_autoctl enable|disable maintenance CLI.

### LINKED ISSUE #33: pg_autoctl enable|disable maintenance
It would be very convenient for me to be able to put a node into maintenance mode from that node's machine. Main use case is preparation for a minor PG upgrade.

DimCitus:
The feature was already there and the CLI hidden in our DEBUG tooling,
because we were not sure at the time of the initial implementation how to
expose the feature to our users.

Now that we have the enable|disable subcommands already, it looks like it is
a good fit for the maintenance feature.

Fixes #33.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -33,7 +33,7 @@ RUN PIPENV_PIPFILE=tests/Pipfile pipenv install --system --deploy
 
 COPY Makefile ./
 COPY ./src/ ./src
-RUN make clean && make install -j8
+RUN make -s clean && make -s install -j8
 
 COPY ./tests/ ./tests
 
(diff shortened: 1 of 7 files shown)


Dockerfile
```diff
@@ -33,7 +33,7 @@ RUN PIPENV_PIPFILE=tests/Pipfile pipenv install --system --deploy
 
 COPY Makefile ./
 COPY ./src/ ./src
-RUN make clean && make install -j8
+RUN make -s clean && make -s install -j8
```
mtuncer: not really relevant but ok

DimCitus: Yeah, I grew tired of seeing all the output when we just want to know that we had zero warnings, which `make -s` implements. I was not sure of adding that to the current PR but though same as you did, that it should be “ok”.

PR #34, opened 2019-07-01
Fix missing include header.

DimCitus:
Some of our C files are missing headers and compile then fails on FreeBSD.
I'm not sure how compile passes in other systems / compilers, and the fix
seems relevant and general enough for all platforms.

Fixes #32.

PR #37, opened 2019-07-02
Fix dockerfile build-dependencies.

DimCitus:
Now that we're using $(shell $(PG_CONFIG) --libs) to have the list of
librairies we depend on, we must align the build dependencies of
pg_auto_failover with the Postgres build dependencies on the platform where
we build.

In debian, that means having build dependencies towards readline
support (that's libedit), pam, zlib, xml, xstl, and selinux.

See #34.

### DIFF
--- a/src/bin/pg_autoctl/Makefile
+++ b/src/bin/pg_autoctl/Makefile
@@ -31,8 +31,8 @@ CFLAGS += -I $(shell $(PG_CONFIG) --cflags)
 CFLAGS += -Wformat
 CFLAGS += $(COMMON_LIBS)
 
-LIBS  = -L $(shell $(PG_CONFIG) --libdir)
-LIBS += -L $(shell $(PG_CONFIG) --pkglibdir)
+LIBS  = -L $(shell $(PG_CONFIG) --pkglibdir)
+LIBS += $(shell $(PG_CONFIG) --ldflags)
 LIBS += $(shell $(PG_CONFIG) --libs)
 LIBS += -lpq
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,6 +11,12 @@ RUN apt-get update \
     libicu-dev \
     libkrb5-dev \
     libssl-dev \
+    libedit-dev \
+    libpam-dev \
+    zlib1g-dev \
+    libxml2-dev \
+    libxslt1-dev \
+    libselinux1-dev \
     make \
     openssl \
     pipenv \


mtuncer:
it did not build until I added
```
LIBS += $(shell $(PG_CONFIG) --ldflags)
```
to `src/bin/pg_autoctl/Makefile`

DimCitus:
Then I guess it should replace the existing lines, right? Basically we would do:

``` diff
-LIBS  = -L $(shell $(PG_CONFIG) --libdir)
-LIBS += -L $(shell $(PG_CONFIG) --pkglibdir)
+LIBS  = $(shell $(PG_CONFIG) --ldflags)
```

mtuncer:
they are different in my system
```
pg_config --libdir ---> /opt/pgsql/11/lib
pg_config --pkglibdir --> /opt/pgsql/11/lib
pg_config --ldflags  --> -L/usr/local/opt/openssl/lib -Wl,-dead_strip_dylibs
```

mtuncer:
pg_config documentation describes
```
--libdir
Print the location of object code libraries.

--pkglibdir
Print the location of dynamically loadable modules, or where the server would search for them. (Other architecture-dependent data files might also be installed in this directory.)
```

I am thinking about if we should just remove `pkglibdir` line since it looked like something to do with loading libraries, not linking.

DimCitus:
Here's what I have:
``` bash
$ pg_config --ldflags
-L/Applications/Postgres.app/Contents/Versions/12/lib -L/Applications/Postgres.app/Contents/Versions/12/lib -Wl,-dead_strip_dylibs

$ pg_config --libdir
/Applications/Postgres.app/Contents/Versions/12/lib

$ pg_config --pkglibdir
/Applications/Postgres.app/Contents/Versions/12/lib/postgresql
```

DimCitus:
After playing with many different settings, it seems to me that the way to go is the following:

``` make
LIBS  = -L $(shell $(PG_CONFIG) --libdir)
LIBS  = -L $(shell $(PG_CONFIG) --pkglibdir)
LIBS += $(shell $(PG_CONFIG) --ldflags)
LIBS += $(shell $(PG_CONFIG) --libs)
LIBS += -lpq
```

src/bin/pg_autoctl/Makefile
```diff
@@ -32,7 +32,8 @@ CFLAGS += -Wformat
 CFLAGS += $(COMMON_LIBS)
 
 LIBS  = -L $(shell $(PG_CONFIG) --libdir)
-LIBS += -L $(shell $(PG_CONFIG) --pkglibdir)
+LIBS  = -L $(shell $(PG_CONFIG) --pkglibdir)
```
mtuncer: we should += to LIBS

DimCitus: Good spot! It's actually the other way round, I left the --libdir line even though it's not needed anymore, as I just re-tested. Re-uploaded new code (as a rebase).

PR #41, opened 2019-07-09
Fix IP address inquiries when ifa_addr is NULL.

### LINKED ISSUE #40: about the func findHostnameLocalAddress addr protocol judgment
ifa_addr is empty in the pptp protocol,"ifaddr->ifa_addr->sa_family == AF_INET" causes a segmentation fault.
Screenshot from 2019-07-09 16-26-56
Screenshot from 2019-07-09 16-10-05

DimCitus:
When using PPTP protocol some interfaces are exposed without an if_addr
field at all, which would trip our IP address discovery code and cause a
segfault.

Fixes #40.

### DIFF
--- a/src/bin/pg_autoctl/ipaddr.c
+++ b/src/bin/pg_autoctl/ipaddr.c
@@ -133,6 +133,19 @@ fetchLocalCIDR(const char *localIpAddress, char *localCIDR, int size)
 		char netmask[INET6_ADDRSTRLEN] = { 0 };
 		char address[INET6_ADDRSTRLEN] = { 0 };
 
+		/*
+		 * Some interfaces might have an empty ifa_addr, such as when using the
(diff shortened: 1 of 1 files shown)


src/bin/pg_autoctl/ipaddr.c
```diff
@@ -133,6 +133,19 @@ fetchLocalCIDR(const char *localIpAddress, char *localCIDR, int size)
 		char netmask[INET6_ADDRSTRLEN] = { 0 };
 		char address[INET6_ADDRSTRLEN] = { 0 };
 
+		/*
+		 * Some interfaces might have an empty ifa_addr, such as when using the
```
mtuncer: is there a way to test this ?

DimCitus: It sounds complex to build a test case for this and easy enough a code path to accept it without tests? better continue the look than crashing (segfault) on a NULL value anyway...

PR #43, opened 2019-07-09
Drop replication slot after failover being demoted to secondary

### LINKED ISSUE #42: multiple failovers may leave replication slot open in the secondary
after performing manual failover multiple times observed that `pg_replication_slots` table has an entry with `active = f`

ideally we should not have any replication slots entry in the secondary node.

mtuncer:
Replication slot record remains in pg_replication_slots after
primary gets demoted to be secondary. This commit explicitly
removes the record.

Fixes #42

### DIFF
--- a/src/bin/pg_autoctl/fsm_transition.c
+++ b/src/bin/pg_autoctl/fsm_transition.c
@@ -473,8 +473,9 @@ fsm_rewind_or_init(Keeper *keeper)
 
 	if (!primary_drop_replication_slot(postgres, config->replication_slot_name))
 	{
-		log_warn("Failed to drop replication slot \"%s\" used by the standby failed",
+		log_error("Failed to drop replication slot \"%s\" used by the standby failed",
 				  config->replication_slot_name);
+		return false;
 	}
 
 	return true;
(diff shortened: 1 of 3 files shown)


src/bin/pg_autoctl/fsm_transition.c
```diff
@@ -471,6 +471,12 @@ fsm_rewind_or_init(Keeper *keeper)
 		}
 	}
 
+	if (!primary_drop_replication_slot(postgres, config->replication_slot_name))
+	{
+		log_warn("Failed to drop replication slot \"%s\" used by the standby failed",
+				  config->replication_slot_name);
```
DimCitus: I think we should return false here, making sure that the transition fails and is retried. I can't see a good reason to continue as if everything was good when we fail to drop the replication slot. Also, the WARN should probably be raised to an ERROR.

DimCitus:
It is possible to add test coverage for the drop of the replication slot in an existing failover scenario in our test suite? Just a quick SQL query somewhere should do...

src/bin/pg_autoctl/fsm_transition.c
mtuncer: changed accordingly

mtuncer:
added manual failover test and verified replication slot removal on the secondary

PR #52, opened 2019-08-15
Fix destroy command to be resilient to missing monitor.

DimCitus:
In case when we can't contact the monitor at `pg_autoctl do destroy` time,
it's better to just continue with the destroy process rather than fail here.
After all, we might have destroyed the monitor already, and we want to get
rid of the local node anyway.

### DIFF
--- a/src/bin/pg_autoctl/cli_do_misc.c
+++ b/src/bin/pg_autoctl/cli_do_misc.c
@@ -507,9 +507,27 @@ keeper_cli_destroy_keeper_node(Keeper *keeper, KeeperConfig *config)
 		/* keeper_remove uses log_info() to explain what's happening */
 		if (!keeper_remove(keeper, config))
 		{
-			log_fatal("Failed to remove local node from the pg_auto_failover "
+			log_error("Failed to remove local node from the pg_auto_failover "
 					  "monitor, see above for details");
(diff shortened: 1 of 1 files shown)


src/bin/pg_autoctl/cli_do_misc.c
```diff
@@ -507,9 +507,27 @@ keeper_cli_destroy_keeper_node(Keeper *keeper, KeeperConfig *config)
...
-			exit(EXIT_CODE_INTERNAL_ERROR);
+
+			/* we still have to remove the state files now */
+			log_info("Removing local node state file: \"%s\"",
+					 config->pathnames.state);
+
+			if (!unlink_file(config->pathnames.state))
+			{
+				/* we already logged about errors */
+				exit(EXIT_CODE_BAD_STATE);
```
mtuncer: shall we try to unlink the other file too before retuning bad state. we can still return bad state, but we could remove the other file if possible

DimCitus: Good point. I think it also applies to other call sites of keeper_remove, so I made the change there and changed the function signature to have an extra parameter `bool ignore_monitor_errors`. It should be better now. 

In passing I removed what looked like dead code to me.

PR #53, opened 2019-08-22
Report current LSN of Postgres nodes rather than WAL lag.

mtuncer:
there is an issue with secondary getting stuck in catching up state

### DIFF
--- a/src/bin/pg_autoctl/fsm.c
+++ b/src/bin/pg_autoctl/fsm.c
@@ -280,13 +280,13 @@ keeper_fsm_step(Keeper *keeper)
 	log_info("Calling node_active for node %s/%d/%d with current state: "
 			 "PostgreSQL is running is %s, "
 			 "sync_state is \"%s\", "
-			 "WAL delta is %" PRId64 ".",
+			 "latest WAL LSN is %s.",
 			 config->formation,
 			 keeperState->current_node_id,
 			 keeperState->current_group,
 			 postgres->pgIsRunning ? "true" : "false",
 			 postgres->pgsrSyncState,
-			 postgres->walLag);
+			 postgres->receivedLsn);
 
 	if (!monitor_node_active(monitor,
 							 config->formation,
@@ -296,7 +296,7 @@ keeper_fsm_step(Keeper *keeper)
 							 keeperState->current_group,
 							 keeperState->current_role,
 							 postgres->pgIsRunning,
-							 postgres->walLag,
+							 postgres->receivedLsn,
 							 postgres->pgsrSyncState,
 							 &assignedState))
 	{
--- a/src/bin/pg_autoctl/fsm.c
+++ b/src/bin/pg_autoctl/fsm.c
@@ -286,7 +286,7 @@ keeper_fsm_step(Keeper *keeper)
 			 keeperState->current_group,
 			 postgres->pgIsRunning ? "true" : "false",
 			 postgres->pgsrSyncState,
-			 postgres->receivedLsn);
+			 postgres->currentLSN);
 
 	if (!monitor_node_active(monitor,
 							 config->formation,
@@ -296,7 +296,7 @@ keeper_fsm_step(Keeper *keeper)
 							 keeperState->current_group,
 							 keeperState->current_role,
 							 postgres->pgIsRunning,
-							 postgres->receivedLsn,
+							 postgres->currentLSN,
 							 postgres->pgsrSyncState,
 							 &assignedState))
 	{
--- a/src/bin/pg_autoctl/keeper.c
+++ b/src/bin/pg_autoctl/keeper.c
@@ -6,6 +6,9 @@
  * Licensed under the PostgreSQL License.
  *
  */
+
+#include <postgres.h>
+
 #include <fcntl.h>
 #include <inttypes.h>
 #include <stdlib.h>
@@ -19,7 +22,6 @@
 #include "keeper_config.h"
 #include "pgsetup.h"
 #include "state.h"
-
 
 static bool keeper_get_replication_state(Keeper *keeper);
 
@@ -329,14 +331,14 @@ keeper_update_pg_state(Keeper *keeper)
 
 	bool pg_is_not_running_is_ok = true;
 
-	log_debug("Update local PostgreSQL state");
+	log_warn("Update local PostgreSQL state");
 
 	memcpy(pgSetup, &(config->pgSetup), sizeof(PostgresSetup));
 
 	/* reinitialize the replication state values each time we update */
 	postgres->pgIsRunning = false;
 	memset(postgres->pgsrSyncState, 0, PGSR_SYNC_STATE_MAXLENGTH);
-	postgres->walLag = -1;
+	strcpy(postgres->receivedLsn, "0/0");
 
 	/*
 	 * In some states, it's ok to not have a PostgreSQL data directory at all.
@@ -479,12 +481,15 @@ keeper_update_pg_state(Keeper *keeper)
 	 * In some states, PostgreSQL isn't expected to be running, or not expected
 	 * to have a streaming replication to monitor at all.
 	 */
+	log_warn("keeperState->current_role = %s", NodeStateToString(keeperState->current_role));
 	switch (keeperState->current_role)
 	{
 		case PRIMARY_STATE:
 		case WAIT_PRIMARY_STATE:
 		case SECONDARY_STATE:
+		case CATCHINGUP_STATE:
 		{
+			log_warn("calling keeper_get_replication_state");
 			return postgres->pgIsRunning
 				&& keeper_get_replication_state(keeper);
 		}
@@ -587,13 +592,14 @@ keeper_get_replication_state(Keeper *keeper)
 				pgsql,
 				config->replication_slot_name,
 				postgres->pgsrSyncState,
-				&(postgres->walLag),
+				postgres->receivedLsn,
 				missing_wallag_ok);
+		log_warn("latest received lsn = %s", postgres->receivedLsn);
 	}
 	else
 	{
-		success =
-			pgsql_get_wal_lag_from_standby(pgsql, &(postgres->walLag));
+		success = pgsql_get_received_lsn_from_standby(pgsql, postgres->receivedLsn);
+		log_warn("latest received lsn = %s", postgres->receivedLsn);
 	}
 	pgsql_finish(pgsql);
 
--- a/src/bin/pg_autoctl/keeper.c
+++ b/src/bin/pg_autoctl/keeper.c
@@ -6,9 +6,6 @@
  * Licensed under the PostgreSQL License.
  *
  */
-
-#include <postgres.h>
-
 #include <fcntl.h>
 #include <inttypes.h>
 #include <stdlib.h>
@@ -22,6 +19,7 @@
 #include "keeper_config.h"
 #include "pgsetup.h"
 #include "state.h"
+
 
 static bool keeper_get_replication_state(Keeper *keeper);
 
@@ -331,14 +329,14 @@ keeper_update_pg_state(Keeper *keeper)
 
 	bool pg_is_not_running_is_ok = true;
 
-	log_warn("Update local PostgreSQL state");
+	log_debug("Update local PostgreSQL state");
 
 	memcpy(pgSetup, &(config->pgSetup), sizeof(PostgresSetup));
 
 	/* reinitialize the replication state values each time we update */
 	postgres->pgIsRunning = false;
 	memset(postgres->pgsrSyncState, 0, PGSR_SYNC_STATE_MAXLENGTH);
-	strcpy(postgres->receivedLsn, "0/0");
+	strcpy(postgres->currentLSN, "0/0");
 
 	/*
 	 * In some states, it's ok to not have a PostgreSQL data directory at all.
@@ -481,7 +479,7 @@ keeper_update_pg_state(Keeper *keeper)
 	 * In some states, PostgreSQL isn't expected to be running, or not expected
 	 * to have a streaming replication to monitor at all.
 	 */
-	log_warn("keeperState->current_role = %s", NodeStateToString(keeperState->current_role));
+	log_debug("keeperState->current_role = %s", NodeStateToString(keeperState->current_role));
 	switch (keeperState->current_role)
 	{
 		case PRIMARY_STATE:
@@ -489,7 +487,6 @@ keeper_update_pg_state(Keeper *keeper)
 		case SECONDARY_STATE:
 		case CATCHINGUP_STATE:
 		{
-			log_warn("calling keeper_get_replication_state");
 			return postgres->pgIsRunning
 				&& keeper_get_replication_state(keeper);
 		}
@@ -592,14 +589,14 @@ keeper_get_replication_state(Keeper *keeper)
 				pgsql,
 				config->replication_slot_name,
 				postgres->pgsrSyncState,
-				postgres->receivedLsn,
+				postgres->currentLSN,
+				NAMEDATALEN,
 				missing_wallag_ok);
-		log_warn("latest received lsn = %s", postgres->receivedLsn);
+		log_warn("latest received lsn = %s", postgres->currentLSN);
 	}
 	else
 	{
-		success = pgsql_get_received_lsn_from_standby(pgsql, postgres->receivedLsn);
-		log_warn("latest received lsn = %s", postgres->receivedLsn);
+		success = pgsql_get_received_lsn_from_standby(pgsql, postgres->currentLSN, NAMEDATALEN);
 	}
 	pgsql_finish(pgsql);
 
--- a/src/bin/pg_autoctl/keeper.c
+++ b/src/bin/pg_autoctl/keeper.c
@@ -479,7 +479,6 @@ keeper_update_pg_state(Keeper *keeper)
 	 * In some states, PostgreSQL isn't expected to be running, or not expected
 	 * to have a streaming replication to monitor at all.
 	 */
-	log_debug("keeperState->current_role = %s", NodeStateToString(keeperState->current_role));
 	switch (keeperState->current_role)
 	{
 		case PRIMARY_STATE:
@@ -567,8 +566,7 @@ keeper_get_replication_state(Keeper *keeper)
 	LocalPostgresServer *postgres = &(keeper->postgres);
 
 	PGSQL *pgsql = &(postgres->sqlClient);
-	bool missing_wallag_ok =
-		keeperState->current_role == WAIT_PRIMARY_STATE;
+	bool missingStateOk = keeperState->current_role == WAIT_PRIMARY_STATE;
 
 	bool success = false;
 
@@ -585,18 +583,19 @@ keeper_get_replication_state(Keeper *keeper)
 	if (pg_setup_is_primary(pgSetup))
 	{
 		success =
-			pgsql_get_sync_state_and_wal_lag(
+			pgsql_get_sync_state_and_current_lsn(
 				pgsql,
 				config->replication_slot_name,
 				postgres->pgsrSyncState,
 				postgres->currentLSN,
-				NAMEDATALEN,
-				missing_wallag_ok);
+				PG_LSN_MAXLENGTH,
+				missingStateOk);
 		log_warn("latest received lsn = %s", postgres->currentLSN);
 	}
 	else
 	{
-		success = pgsql_get_received_lsn_from_standby(pgsql, postgres->currentLSN, NAMEDATALEN);
+		success = pgsql_get_received_lsn_from_standby(pgsql, postgres->currentLSN,
+													  PG_LSN_MAXLENGTH);
 	}
 	pgsql_finish(pgsql);
 
--- a/src/bin/pg_autoctl/monitor.c
+++ b/src/bin/pg_autoctl/monitor.c
@@ -289,7 +289,7 @@ monitor_node_active(Monitor *monitor,
 					char *formation, char *host, int port, int nodeId,
 					int groupId, NodeState currentState,
 					bool pgIsRunning,
-					uint64_t replicationLag, char *pgsrSyncState,
+					char *receivedLSN, char *pgsrSyncState,
 					MonitorAssignedState *assignedState)
 {
 	PGSQL *pgsql = &monitor->pgsql;
@@ -298,7 +298,7 @@ monitor_node_active(Monitor *monitor,
 		"$6::pgautofailover.replication_state, $7, $8, $9)";
 	int paramCount = 9;
 	Oid paramTypes[9] = { TEXTOID, TEXTOID, INT4OID, INT4OID,
-						  INT4OID, TEXTOID, BOOLOID, INT8OID, TEXTOID };
+						  INT4OID, TEXTOID, BOOLOID, LSNOID, TEXTOID };
 	const char *paramValues[9];
 	MonitorAssignedStateParseContext parseContext = { assignedState, false };
 	const char *nodeStateString = NodeStateToString(currentState);
@@ -310,8 +310,10 @@ monitor_node_active(Monitor *monitor,
 	paramValues[4] = intToString(groupId).strValue;
 	paramValues[5] = nodeStateString;
 	paramValues[6] = pgIsRunning ? "true" : "false";
-	paramValues[7] = intToString(replicationLag).strValue;
+	paramValues[7] = receivedLSN;
 	paramValues[8] = pgsrSyncState;
+
+	log_info("calling monitor node active with lsn %s", receivedLSN);
 
 	if (!pgsql_execute_with_params(pgsql, sql,
 								   paramCount, paramTypes, paramValues,
@@ -320,10 +322,10 @@ monitor_node_active(Monitor *monitor,
 		log_error("Failed to get node state for node %d (%s:%d) "
 				  "in group %d of formation \"%s\" with initial state "
 				  "\"%s\", replication state \"%s\", "
-				  "and replication lag %" PRId64 ", "
+				  "and latest received lsn \"%s\", "
 				  "see previous lines for details",
 				  nodeId, host, port, groupId, formation, nodeStateString,
-				  pgsrSyncState, replicationLag);
+				  pgsrSyncState, receivedLSN);
 		return false;
 	}
 
@@ -334,11 +336,11 @@ monitor_node_active(Monitor *monitor,
 	{
 		log_error("Failed to get node state for node %d (%s:%d) in group %d of formation "
 				  "\"%s\" with initial state \"%s\", replication state \"%s\","
-				  " and replication lag %" PRId64
+				  " and latest received lsn \"%s\""
 				  " because the monitor returned an unexpected result, "
 				  "see previous lines for details",
 				  nodeId, host, port, groupId, formation, nodeStateString,
-				  pgsrSyncState, replicationLag);
+				  pgsrSyncState, receivedLSN);
 		return false;
 	}
 
--- a/src/bin/pg_autoctl/monitor.c
+++ b/src/bin/pg_autoctl/monitor.c
@@ -289,7 +289,7 @@ monitor_node_active(Monitor *monitor,
 					char *formation, char *host, int port, int nodeId,
 					int groupId, NodeState currentState,
 					bool pgIsRunning,
-					char *receivedLSN, char *pgsrSyncState,
+					char *currentLSN, char *pgsrSyncState,
 					MonitorAssignedState *assignedState)
 {
 	PGSQL *pgsql = &monitor->pgsql;
@@ -310,10 +310,8 @@ monitor_node_active(Monitor *monitor,
 	paramValues[4] = intToString(groupId).strValue;
 	paramValues[5] = nodeStateString;
 	paramValues[6] = pgIsRunning ? "true" : "false";
-	paramValues[7] = receivedLSN;
+	paramValues[7] = currentLSN;
 	paramValues[8] = pgsrSyncState;
-
-	log_info("calling monitor node active with lsn %s", receivedLSN);
 
 	if (!pgsql_execute_with_params(pgsql, sql,
 								   paramCount, paramTypes, paramValues,
@@ -322,10 +320,10 @@ monitor_node_active(Monitor *monitor,
 		log_error("Failed to get node state for node %d (%s:%d) "
 				  "in group %d of formation \"%s\" with initial state "
 				  "\"%s\", replication state \"%s\", "
-				  "and latest received lsn \"%s\", "
+				  "and current lsn \"%s\", "
 				  "see previous lines for details",
 				  nodeId, host, port, groupId, formation, nodeStateString,
-				  pgsrSyncState, receivedLSN);
+				  pgsrSyncState, currentLSN);
 		return false;
 	}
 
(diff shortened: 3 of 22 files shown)


src/bin/pg_autoctl/fsm.c
```diff
@@ -280,13 +280,13 @@ keeper_fsm_step(Keeper *keeper)
 	log_info("Calling node_active for node %s/%d/%d with current state: "
 			 "PostgreSQL is running is %s, "
 			 "sync_state is \"%s\", "
-			 "WAL delta is %" PRId64 ".",
+			 "latest WAL LSN %s.",
```
DimCitus: Missing verb?

src/bin/pg_autoctl/keeper.c
```diff
@@ -587,13 +589,13 @@ keeper_get_replication_state(Keeper *keeper)
...
-				&(postgres->walLag),
+				postgres->receivedLsn,
 				missing_wallag_ok);
 	}
 	else
 	{
-		success =
-			pgsql_get_wal_lag_from_standby(pgsql, &(postgres->walLag));
+		success = pgsql_get_received_lsn_from_standby(pgsql, postgres->receivedLsn);
+		log_info("latest end lsn = %s", postgres->receivedLsn);
```
DimCitus: Looks like a log_debug or even a log_trace to me...

src/bin/pg_autoctl/monitor.c
```diff
@@ -298,7 +298,7 @@ monitor_node_active(Monitor *monitor,
 		"$6::pgautofailover.replication_state, $7, $8, $9)";
 	int paramCount = 9;
 	Oid paramTypes[9] = { TEXTOID, TEXTOID, INT4OID, INT4OID,
-						  INT4OID, TEXTOID, BOOLOID, INT8OID, TEXTOID };
+						  INT4OID, TEXTOID, BOOLOID, TEXTOID, TEXTOID };
```
DimCitus: Should we use `LSNOID` here? Given that we're using the text protocol then the server would cast the literal values in the `pg_lsn` data type for us automatically, after having found the server-side function with the matching signature, IIRC.

src/bin/pg_autoctl/monitor.c
```diff
@@ -289,7 +289,7 @@ monitor_node_active(Monitor *monitor,
 					char *formation, char *host, int port, int nodeId,
 					int groupId, NodeState currentState,
 					bool pgIsRunning,
-					uint64_t replicationLag, char *pgsrSyncState,
+					char *latestLSN, char *pgsrSyncState,
```
DimCitus: Is it `receivedLSN` or `latestLSN`?

src/bin/pg_autoctl/pgsql.c
```diff
@@ -1187,18 +1187,18 @@ typedef struct PgsrSyncAndWALContext
 {
 	bool		parsedOk;
 	char		syncState[PGSR_SYNC_STATE_MAXLENGTH];
-	uint64_t	walLag;
+	char        latestLSN[NAMEDATALEN];
```
DimCitus: A comment explaining the choice of the length here would be necessary.

src/bin/pg_autoctl/pgsql.c
```diff
@@ -1187,18 +1187,18 @@ typedef struct PgsrSyncAndWALContext
...
 pgsql_get_sync_state_and_wal_lag(PGSQL *pgsql, const char *slotName,
-								 char *pgsrSyncState, int64_t *walLag,
+								 char *pgsrSyncState, char *latestLSN,
 								 bool missing_ok)
 {
 	PgsrSyncAndWALContext context = { 0 };
 	char *sql =
 		"select sync_state, "
-		"pg_current_wal_lsn() - flush_lsn as wal_lag "
+		"pg_current_wal_lsn() "
```
DimCitus: The same thing was named `receivedLSN` and then `latestLSN` but actually is the `current LSN` then?

src/bin/pg_autoctl/pgsql.c
```diff
@@ -1253,21 +1253,31 @@ parsePgsrSyncStateAndWAL(void *ctx, PGresult *result)
 
 		case 1:
 		{
-			char *strWalLag = PQgetvalue(result, 0, 1);
```
DimCitus: We did 1 then 0 because we have to declare variables before any piece of other code in C. Now that both parts are done the same way, I think it's fair to re-order them in their natural order, 0 then 1.

src/bin/pg_autoctl/pgsql.c
```diff
@@ -1253,21 +1253,31 @@ parsePgsrSyncStateAndWAL(void *ctx, PGresult *result)
...
 
 			/* we trust our length and PostgreSQL results */
 			strlcpy(context->syncState,
 					PQgetvalue(result, 0, 0),
 					PGSR_SYNC_STATE_MAXLENGTH);
 
-			context->walLag = strtoull(strWalLag, NULL, 10);
 
-			if (context->walLag == 0 && errno != 0)
+#if 0
```
DimCitus: This looks like an attempt to run PostgreSQL backend code in the client application...

src/bin/pg_autoctl/pgsql.c
```diff
@@ -1284,7 +1294,7 @@ parsePgsrSyncStateAndWAL(void *ctx, PGresult *result)
  * replication lag as seen from the pg_stat_wal_receiver system view.
  */
 bool
-pgsql_get_wal_lag_from_standby(PGSQL *pgsql, int64_t *walLag)
+pgsql_get_wal_lag_from_standbyi(PGSQL *pgsql, int64_t *walLag)
```
DimCitus: Typo?

src/bin/pg_autoctl/pgsql.c
```diff
@@ -1310,6 +1320,41 @@ pgsql_get_wal_lag_from_standby(PGSQL *pgsql, int64_t *walLag)
...
+ *
+ * received_lsn is the latest lsn known to be received and flushed to the disk. It does
+ * not specify if it is applied or not. Caller should have allocated necessary memory
+ * for result value.
+ */
+bool
+pgsql_get_received_lsn_from_standby(PGSQL *pgsql, char *received_lsn)
+{
+	SingleValueResultContext context;
+	char *sql =
```
DimCitus: Would fit in a single line this time?

src/bin/pg_autoctl/pgsql.c
```diff
@@ -1310,6 +1320,41 @@ pgsql_get_wal_lag_from_standby(PGSQL *pgsql, int64_t *walLag)
 }
 
 
+/*
+ * pgsql_get_received_lsn_from_standby queries a standby PostgreSQL server to get the
+ * received_lsn value from the pg_stat_wal_receiver system view.
```
DimCitus: Why choose the `received_lsn` value from the view? I think that's the main thing this comment should explain here. To be faire the next sentence does half of the job, explaining what the value is. It still does not explain why this is the one that we want to read.

src/bin/pg_autoctl/pgsql.c
```diff
@@ -1310,6 +1320,41 @@ pgsql_get_wal_lag_from_standby(PGSQL *pgsql, int64_t *walLag)
...
+ *
+ * received_lsn is the latest lsn known to be received and flushed to the disk. It does
+ * not specify if it is applied or not. Caller should have allocated necessary memory
+ * for result value.
+ */
+bool
+pgsql_get_received_lsn_from_standby(PGSQL *pgsql, char *received_lsn)
```
DimCitus: When passing a `char *` to be filled-in, all the other call sites would add the length of the passed-in buffer as a separate argument, usually named `size`. Please do the same thing here, otherwise the caller must know that this argument is hard-coded to take NAMEDATALEN bytes.

src/bin/pg_autoctl/pgsql.c
```diff
@@ -1310,6 +1320,41 @@ pgsql_get_wal_lag_from_standby(PGSQL *pgsql, int64_t *walLag)
...
+{
+	SingleValueResultContext context;
+	char *sql =
+		"SELECT received_lsn"
+		"  FROM pg_stat_wal_receiver";
+
+	context.resultType = PGSQL_RESULT_STRING;
+	context.parsedOk = false;
+
+	log_info("pgsql_get_received_lsn_from_standby : running %s", sql);
```
DimCitus: Looks like a log_trace to me?

src/bin/pg_autoctl/pgsql.c
```diff
@@ -1310,6 +1320,41 @@ pgsql_get_wal_lag_from_standby(PGSQL *pgsql, int64_t *walLag)
...
+	pgsql_execute_with_params(pgsql, sql, 0, NULL, NULL,
+							  &context, &parseSingleValueResult);
+
+	if (!context.parsedOk)
+	{
+		log_error("PostgreSQL cannot reach the primary server: "
```
DimCitus: Error message does not look like the other messages we usually display to the user. When using the ERROR level, we try to phrase the message so that the user knows what to do next, if an action is required. Otherwise, we could use a WARNING. In any case, most messages in `pg_autoctl` follow the form 

    Failed to <action>: <reason>

src/bin/pg_autoctl/pgsql.c
```diff
@@ -1310,6 +1320,41 @@ pgsql_get_wal_lag_from_standby(PGSQL *pgsql, int64_t *walLag)
...
+				  "the system view pg_stat_wal_receiver has no rows.");
+		return false;
+	}
+
+	strlcpy(received_lsn, context.strVal, NAMEDATALEN);
```
DimCitus: NAMEDATALEN here should be `size`, as given in argument.

src/bin/pg_autoctl/primary_standby.h
```diff
@@ -27,7 +27,7 @@ typedef struct LocalPostgresServer
 	PostgresSetup	postgresSetup;
 	bool			pgIsRunning;
 	char			pgsrSyncState[PGSR_SYNC_STATE_MAXLENGTH];
-	int64_t			walLag;
+	char            receivedLsn[NAMEDATALEN];
```
DimCitus: Here too we will need a comment about why NAMEDATALEN is a good fit for the LSN value. Oh, and maybe some comments about why keeping the value as a string internally makes sense (I think it does, by the way).
So that would be a _section_ comment rather than in-line, I suppose.

src/monitor/group_state_machine.c
```diff
@@ -329,7 +329,7 @@ AssignGoalState(AutoFailoverNode *pgAutoFailoverNode,
 						  pgAutoFailoverNode->nodeName,
 						  pgAutoFailoverNode->nodePort,
 						  pgAutoFailoverNode->pgsrSyncState,
-						  pgAutoFailoverNode->walDelta,
+						  pgAutoFailoverNode->currentLSN,
```
DimCitus: On the monitor side, we get updated every 5 seconds. I reckon we should use `reportedLSN` here rather than _current_. Like the other `reported...` fields we already have.

src/monitor/group_state_machine.c
```diff
@@ -345,27 +345,31 @@ WalDifferenceWithin(AutoFailoverNode *secondaryNode,
...
+
+	if (secondaryLsn == 0 || otherNodeLsn == 0)
 	{
-		walDifference = otherNode->walDelta;
+		/* we don't have any data yet */
+		return false;
 	}
-	else
+
+	if (secondaryLsn > otherNodeLsn)
```
DimCitus: I think this would be easier to read as

``` C
    walDifference = abs(otherNodeLsn - secondaryLsn);
```

But then I wonder why we accept that the LSN might be in any order in between the secondary and the primary. Certainly the primary LSN is supposed to always be greater than or equal to the secondary's LSN? If we keep using the absolute difference, then I think we need to explain why in a comment...

src/monitor/node_active_protocol.c
```diff
@@ -244,7 +246,8 @@ node_active(PG_FUNCTION_ARGS)
 	Oid currentReplicationStateOid = PG_GETARG_OID(5);
 
 	bool currentPgIsRunning = PG_GETARG_BOOL(6);
-	int64 currentWalDelta = PG_GETARG_INT64(7);
+	text *currentLSNText = PG_GETARG_TEXT_P(7);
```
DimCitus: I believe we could be `PG_GETARG_LSN` here, given a SQL declaration of the function that takes a `pg_lsn` input. Given the following code in `src/include/utils/pg_lsn.h` in PostgreSQL, we don't seem to have an `LSNOID`, so maybe sending the text representation and using the `INT8OID` to target a function that accept a `pg_lsn` data type would work?

``` C
#define DatumGetLSN(X) ((XLogRecPtr) DatumGetInt64(X))
#define LSNGetDatum(X) (Int64GetDatum((int64) (X)))

#define PG_GETARG_LSN(n)	 DatumGetLSN(PG_GETARG_DATUM(n))
#define PG_RETURN_LSN(x)	 return LSNGetDatum(x)
```

src/monitor/node_active_protocol.c
```diff
@@ -665,6 +675,24 @@ IsStateIn(ReplicationState state, List *allowedStates)
 }
 
 
+/* converts text representation of lsn to XlogRecPtr */
+static XLogRecPtr TextToLsn(char *lsnString)
```
DimCitus: It might be that we can make it work without this function, which I think would be best.

src/monitor/node_metadata.c
```diff
@@ -158,7 +158,7 @@ TupleToAutoFailoverNode(TupleDesc tupleDescriptor, HeapTuple heapTuple)
 	pgAutoFailoverNode->pgsrSyncState =
 		SyncStateFromString(TextDatumGetCString(pgsrSyncState));
 	pgAutoFailoverNode->reportTime = DatumGetTimestampTz(reportTime);
-	pgAutoFailoverNode->walDelta = DatumGetInt64(walDelta);
+	pgAutoFailoverNode->currentLSN = DatumGetInt64(latestLSN);
```
DimCitus: This is confusing, isn't it?

We should stick to the current norm for naming things. On the client side, `pg_autoctl` sees the CURRENT version of things, then sends it to the monitor, for storage and processing. The monitor might use the value immediately, or later from another's node request, at which point we know that the value is the latest reported value rather than the current one. So in the monitor we accept the CURRENT value and then register it as the REPORTED value. 

I don't think we use LATEST at the moment in our variable names, to avoid the confusion.

src/monitor/pgautofailover.sql
```diff
@@ -160,7 +160,7 @@ AS 'MODULE_PATHNAME', $$node_active$$;
 
 grant execute on function
       pgautofailover.node_active(text,text,int,int,int,
-                          pgautofailover.replication_state,bool,bigint,text)
+                          pgautofailover.replication_state,bool,text,text)
```
DimCitus: So this should be `pg_lsn` so that Postgres does the text to LSN transformation for us... can we make that work?

mtuncer:
There is an issue with multiple failovers in regression tests. multi failovers work just fine when triggered manually but in the regression test secondary fails to be promoted to primary. 

I disabled a test to see if there is anything else. Now it looks okay.

src/bin/pg_autoctl/pgsql.c
mtuncer: added

src/bin/pg_autoctl/pgsql.c
mtuncer: I use received_lsn at secondary node, and latest_lsn at monitor.

mtuncer: having second thoughts on this.

src/bin/pg_autoctl/pgsql.c
mtuncer: yes indeed, a failed attempt.

src/bin/pg_autoctl/pgsql.c
mtuncer: done

src/bin/pg_autoctl/pgsql.c
mtuncer: added some more explanation

src/bin/pg_autoctl/pgsql.c
mtuncer: ok

src/bin/pg_autoctl/pgsql.c
mtuncer: it was a debug helper while developing. made it trace

src/bin/pg_autoctl/pgsql.c
mtuncer: done

src/bin/pg_autoctl/monitor_pg_init.c
```diff
@@ -32,6 +32,13 @@
  */
 GUC monitor_default_settings[] = {
 	{ "shared_preload_libraries", "'pgautofailover'" },
+	{ "log_destination", "stderr"},
```
mtuncer: @DimCitus  we should re-evaluate if we want to have these changes.

src/bin/pg_autoctl/pgsql.c
mtuncer: per our discussion, it is `current lsn` at data nodes, and `reported lsn` at monitor.

src/bin/pg_autoctl/keeper.c
```diff
@@ -479,11 +479,13 @@ keeper_update_pg_state(Keeper *keeper)
 	 * In some states, PostgreSQL isn't expected to be running, or not expected
 	 * to have a streaming replication to monitor at all.
 	 */
+	log_debug("keeperState->current_role = %s", NodeStateToString(keeperState->current_role));
```
DimCitus: do we want to keep this debug output once the PR is merged? it seems helpful for cooking the PR, maybe not so much once we get it to being mergeable?

src/bin/pg_autoctl/keeper.c
```diff
@@ -587,13 +589,14 @@ keeper_get_replication_state(Keeper *keeper)
 				pgsql,
 				config->replication_slot_name,
 				postgres->pgsrSyncState,
-				&(postgres->walLag),
+				postgres->currentLSN,
+				NAMEDATALEN,
 				missing_wallag_ok);
+		log_warn("latest received lsn = %s", postgres->currentLSN);
```
DimCitus: I think we should rename the function `pgsql_get_sync_state_and_wal_lag` because it's now getting the CurrentLSN rather than the WAL lag, and also rename our boolean `missing_wallag_ok` to be closer to what happens here (`missing_currentLSN_ok` I guess?).

src/bin/pg_autoctl/monitor_pg_init.c
DimCitus: I think it would be best if we can isolate them in another PR...

src/bin/pg_autoctl/pgsql.c
```diff
@@ -1182,23 +1182,27 @@ validate_connection_string(const char *connectionString)
...
 	char		syncState[PGSR_SYNC_STATE_MAXLENGTH];
-	uint64_t	walLag;
+	char        currentLSN[NAMEDATALEN];
 } PgsrSyncAndWALContext;
 
 bool
 pgsql_get_sync_state_and_wal_lag(PGSQL *pgsql, const char *slotName,
-								 char *pgsrSyncState, int64_t *walLag,
+								 char *pgsrSyncState, char *currentLSN,
+								 int bufferLength,
```
DimCitus: Let's have a better name than `buferLength` which is not only very generic but also confusing: it's not the length of the string as per `strlen` but the size of the memory allocated. What about `currentLSNsize` instead?

src/bin/pg_autoctl/pgsql.c
```diff
@@ -1223,15 +1227,15 @@ pgsql_get_sync_state_and_wal_lag(PGSQL *pgsql, const char *slotName,
 	}
 
 	strlcpy(pgsrSyncState, context.syncState, PGSR_SYNC_STATE_MAXLENGTH);
-	*walLag = context.walLag;
+	strlcpy(currentLSN, context.currentLSN, bufferLength-1);
```
DimCitus: Do we actually need the `-1` here? We certainly don't do that elsewhere, so that reads strange. Are you telling me we need to fix all the other use cases for `strlcpy` in the code?

src/bin/pg_autoctl/pgsql.c
```diff
@@ -1279,21 +1277,32 @@ parsePgsrSyncStateAndWAL(void *ctx, PGresult *result)
...
- * replication lag as seen from the pg_stat_wal_receiver system view.
+ * pgsql_get_received_lsn_from_standby queries a standby PostgreSQL server to get the
+ * received_lsn value from the pg_stat_wal_receiver system view.
+ *
+ * received_lsn is the latest lsn known to be received and flushed to the disk. It does
+ * not specify if it is applied or not. Caller should have allocated necessary memory
+ * for result value.
+ *
+ * We are collecting the latest WAL entry that is received successfully. It will be
+ * eventually applied to the receiving database.  This information will later be
```
DimCitus: Add a note that explains that the data is known to be safe on the standby, and can be replayed later at end-of-recovery before promotion of the standby happens, please. That's the reasoning here, and it might be that we need to ensure that the WAL known to be received is actually replayed at promotion time, so please look into that (in Postgres documentation + source code if needed) and add your findings and references to the comment!

src/bin/pg_autoctl/pgsql.c
DimCitus: Nitpick: make the SQL query use a single line, it fits ;-)

src/bin/pg_autoctl/pgsql.c
```diff
@@ -1304,12 +1313,11 @@ pgsql_get_wal_lag_from_standby(PGSQL *pgsql, int64_t *walLag)
 		return false;
 	}
 
-	*walLag = context.bigint;
+	strlcpy(receivedLSN, context.strVal, bufferLength-1);
```
DimCitus: This usage of `bufferLength-1` is worrying me a lot, can you double-check?

src/bin/pg_autoctl/pgsql.c
```diff
@@ -1279,21 +1277,32 @@ parsePgsrSyncStateAndWAL(void *ctx, PGresult *result)
...
 	char *sql =
-		"SELECT coalesce(latest_end_lsn - received_lsn, -1) "
+		"SELECT received_lsn"
 		"  FROM pg_stat_wal_receiver";
 
-	context.resultType = PGSQL_RESULT_BIGINT;
+	context.resultType = PGSQL_RESULT_STRING;
 	context.parsedOk = false;
 
+	log_trace("pgsql_get_received_lsn_from_standby : running %s", sql);
```
DimCitus: I think the SQL statement is going to be logged by the `sql__execute_with_params` function at the DEBUG level anyway. Do we need this `log_trace` call here?

src/bin/pg_autoctl/pgsql.c
```diff
@@ -1279,21 +1277,32 @@ parsePgsrSyncStateAndWAL(void *ctx, PGresult *result)
...
+ * not specify if it is applied or not. Caller should have allocated necessary memory
+ * for result value.
+ *
+ * We are collecting the latest WAL entry that is received successfully. It will be
+ * eventually applied to the receiving database.  This information will later be
+ * used by monitor to decide which secondary has the latest data.
  */
 bool
-pgsql_get_wal_lag_from_standby(PGSQL *pgsql, int64_t *walLag)
+pgsql_get_received_lsn_from_standby(PGSQL *pgsql, char *receivedLSN, int bufferLength)
```
DimCitus: I think `bufferLength` is misnamed, and propose `receivedLSNsize` instead.

src/bin/pg_autoctl/primary_standby.h
```diff
@@ -20,14 +20,18 @@
  * LocalPostgresServer represents a local postgres database cluster that
  * we can manage via a SQL connection and operations on the database
  * directory contained in the PostgresSetup.
+ *
+ * currentLSN value is kept as text for better portability. We do not
+ * perform any operation on the value after it was read from database.
+ * NAMEDATALEN is long enough for text representation of the LSN value.
```
DimCitus: Maybe add some examples of normal and maximum size LSN text representation, to make it obvious that NAMEDATALEN is the proper choice here. If it is not, then maybe we can define a new size constant such as `LSN_MAXLENGTH` or something.

src/monitor/node_metadata.c
```diff
@@ -453,10 +453,12 @@ ReportAutoFailoverNodeState(char *nodeName, int nodePort,
 		"UPDATE " AUTO_FAILOVER_NODE_TABLE
 		" SET reportedstate = $1, reporttime = now(), "
 		"reportedpgisrunning = $2, reportedrepstate = $3, "
-		"waldelta = CASE $4 WHEN -1 THEN waldelta ELSE $4 END, "
-		"walreporttime = CASE $4 WHEN -1 THEN walreporttime ELSE now() END, "
+		"reportedlsn = CASE $4 WHEN '0/0'::pg_lsn THEN reportedlsn ELSE $4 END, "
+		"walreporttime = CASE $4 WHEN '0/0'::pg_lsn THEN walreporttime ELSE now() END, "
 		"statechangetime = now() WHERE nodename = $5 AND nodeport = $6";
 
+	elog(INFO, "update query %s", updateQuery);
```
DimCitus: Do we keep that in the merged PR? I don't think so?

tests/test_basic_operation.py
```diff
@@ -89,33 +89,40 @@ def test_013_drop_secondary():
...
+   monitor.failover()
+   assert node3.wait_until_state(target_state="primary")
+   assert node2.wait_until_state(target_state="secondary")
+   node2_replication_slots = node2.run_sql_query("select count(*) from pg_replication_slots")
+   assert node2_replication_slots == [(0,)]
+   node3_replication_slots = node3.run_sql_query("select count(*) from pg_replication_slots")
+   assert node3_replication_slots == [(1,)]
+   
+   node2.fail()
+   node2.run()
```
DimCitus: That's the test workaround we have been talking about. It would be best that we don't need it of course.

src/bin/pg_autoctl/keeper.c
mtuncer: removed

src/bin/pg_autoctl/keeper.c
mtuncer: made it `missingStateOk`

src/bin/pg_autoctl/pgsql.c
mtuncer: ok, made id `maxLSNSize`

mtuncer: and ditched -1

src/bin/pg_autoctl/pgsql.c
mtuncer: removed, it won't actually matter though.

src/bin/pg_autoctl/pgsql.c
mtuncer: added some more explanation

src/bin/pg_autoctl/pgsql.c
mtuncer: maxLSNSize

src/bin/pg_autoctl/pgsql.c
mtuncer: it fits

src/bin/pg_autoctl/pgsql.c
mtuncer: removed

src/bin/pg_autoctl/pgsql.c
mtuncer: removed - 1

src/monitor/node_metadata.c
mtuncer: right, removed

src/bin/pg_autoctl/primary_standby.h
mtuncer: pg_lsn.c defines MAXPG_LSNLEN as 17 and uses MAXPG_LSNLEN+1 length buffers to do lsn operations.

I defined PG_LSN_MAXLENGTH = 18 and used it in the code

src/bin/pg_autoctl/pgsql.c
DimCitus: I would feel better with some references (links) to Postgres documentation about the behaviour at promotion time where we can read that it will actually replay any WAL that has been received before promoting. Otherwise we have just implemented a data loss hasard. Also this is related to our choice of `synchronous_commit` default value, which is currently `on`, and might have to be `remote_apply` depending on your findings.

src/monitor/node_active_protocol.c
```diff
@@ -261,14 +262,21 @@ node_active(PG_FUNCTION_ARGS)
...
 	currentNodeState.replicationState =
 		EnumGetReplicationState(currentReplicationStateOid);
-	currentNodeState.walDelta = currentWalDelta;
+	currentNodeState.reportedLSN = currentLSN;
 	currentNodeState.pgsrSyncState = SyncStateFromString(currentPgsrSyncState);
 	currentNodeState.pgIsRunning = currentPgIsRunning;
 
+	ereport(INFO, (errmsg("before calling internal NodeActive with  %s, %d, " INT64_FORMAT ,
+						   nodeName, nodePort, currentNodeState.reportedLSN )));
+
```
DimCitus: Looks like debug logs, do we need them once the PR is merged?

src/monitor/node_active_protocol.c
```diff
@@ -261,14 +262,21 @@ node_active(PG_FUNCTION_ARGS)
 
 	checkPgAutoFailoverVersion();
 
+	ereport(INFO, (errmsg("node_active is called with %s, %d, " UINT64_FORMAT ,
+						   nodeName, nodePort,  currentLSN)));
+
+
```
DimCitus: This looks like debug level logs, do we need them once the PR is merged?

mtuncer: it could be removed

src/monitor/node_active_protocol.c
mtuncer: removed

src/bin/pg_autoctl/pgsql.c
mtuncer: https://www.postgresql.org/docs/12/warm-standby.html#STANDBY-SERVER-OPERATION states that

```
Standby mode is exited and the server switches to normal operation when pg_ctl promote is run or a trigger file is found (trigger_file). Before failover, any WAL immediately available in the archive or in pg_wal will be restored, but no attempt is made to connect to the master.
```

DimCitus: The documentation that we are interested in seems to be at https://www.postgresql.org/docs/current/warm-standby.html#STANDBY-SERVER-OPERATION where the documentation sayth:

> Standby mode is exited and the server switches to normal operation when pg_ctl promote is run or a trigger file is found (trigger_file). Before failover, any WAL immediately available in the archive or in pg_wal will be restored, but no attempt is made to connect to the master.

mtuncer: done

src/bin/pg_autoctl/monitor_pg_init.c
mtuncer: removed from this pr

PR #56, opened 2019-09-02
Use PQExpBuffer facility in more places.

### LINKED ISSUE #54: Fix length string buffers need to be converted to dynamic  buffers in client code
We have many uses of `char buffer[BUFFER_LENGTH]` and `strcpy`, `sprintf` to that buffer.  
Although we haven't seen a problem yet. Using fixed size buffer to host dynamic length content and having unchecked access to these buffers is not safe.

We should use  `PQExpBuffer` and `appendPQExpBufferStr()` functions to modify that buffer.

DimCitus:
Rather than playing fast and loose with sprintf and static buffers that
might or might not be big enough chunks of memory, use the PostgreSQL client
side library for dynamic string buffers where it makes sense.

Fixes #54

### DIFF
--- a/src/bin/pg_autoctl/pgctl.c
+++ b/src/bin/pg_autoctl/pgctl.c
@@ -1024,9 +1024,6 @@ pg_write_recovery_conf(const char *pgdata,
  *
  * The result is written to destination and the length of the result.
  */
-#define LOCAL_BOUND_CHECK \
-	{ if (destinationSize < escapedStringLength) {goto error;} }
-
 static bool
 escape_recovery_conf_string(char *destination, int destinationSize,
 							const char *recoveryConfString)
@@ -1038,7 +1035,10 @@ escape_recovery_conf_string(char *destination, int destinationSize,
 	/* we are going to add at least 3 chars: two quotes and a NUL character */
 	if (destinationSize < (length+3))
 	{
-		goto error;
+		log_error("BUG: failed to escape recovery parameter value \"%s\" "
+				  "in a buffer of %d bytes",
+				  recoveryConfString, destinationSize);
+		return false;
 	}
 
 	destination[escapedStringLength++] = '\'';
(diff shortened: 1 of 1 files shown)


src/bin/pg_autoctl/pgctl.c
```diff
@@ -1037,34 +1066,72 @@ escape_recovery_conf_string(char *destination, const char *recoveryConfString)
...
+		/* errors have already been logged. */
+		return false;
+	}
+	appendPQExpBuffer(buffer, "host = %s", escaped);
+	appendPQExpBuffer(buffer, "port = %d", primaryPort);
+
+	if (!escape_recovery_conf_string(escaped, BUFSIZE, replicationUsername))
+	{
+		/* errors have already been logged. */
+		return false;
```
marcocitus: Some missing destroys here.

DimCitus: Thanks! Fixed now.

src/bin/pg_autoctl/pgctl.c
```diff
@@ -975,57 +981,80 @@ pg_write_recovery_conf(const char *pgdata,
...
+	}
+
 	destination[escapedStringLength++] = '\'';
 
 	for (charIndex = 0; charIndex < length; charIndex++)
 	{
 		char currentChar = recoveryConfString[charIndex];
+
+		if (destinationSize < (charIndex+1))
```
mtuncer: we should compare against escapedStringLength

mtuncer: this was we could think about removing the previous `(destinationSize < (length + 3))` check

DimCitus: I improved the error checking here, thanks to your review. Please have another look?

src/bin/pg_autoctl/pgctl.c
```diff
@@ -975,96 +981,160 @@ pg_write_recovery_conf(const char *pgdata,
...
+							const char *recoveryConfString)
 {
 	int charIndex = 0;
 	int length = strlen(recoveryConfString);
 	int escapedStringLength = 0;
 
+	/* we are going to add at least 3 chars: two quotes and a NUL character */
+	if (destinationSize < (length+3))
+	{
+		goto error;
```
mtuncer: please no :(

src/bin/pg_autoctl/pgctl.c
```diff
@@ -975,96 +981,160 @@ pg_write_recovery_conf(const char *pgdata,
...
 	destination[escapedStringLength++] = '\'';
 
 	for (charIndex = 0; charIndex < length; charIndex++)
 	{
 		char currentChar = recoveryConfString[charIndex];
+
 		if (currentChar == '\'')
 		{
 			destination[escapedStringLength++] = '\'';
+			LOCAL_BOUND_CHECK;
```
mtuncer: I liked repeated error message better

src/bin/pg_autoctl/pgctl.c
```diff
@@ -975,96 +981,160 @@ pg_write_recovery_conf(const char *pgdata,
...
+		LOCAL_BOUND_CHECK;
 	}
 
 	destination[escapedStringLength++] = '\'';
 	destination[escapedStringLength] = '\0';
 
-	return escapedStringLength;
+	return true;
+
+error:
```
mtuncer: I don't consider goto as mother of all evil but not fond of it either.

src/bin/pg_autoctl/pgctl.c
DimCitus: Ok, I did it without goto then. I don't like goto either, though for error cases with the same error message I though it was a good way to share the code.

src/bin/pg_autoctl/pgctl.c
DimCitus: Ok, re-done that way now.

src/bin/pg_autoctl/pgctl.c
DimCitus: Got rid of it.

PR #59, opened 2019-09-10
Implement an http api and --disable-monitor

DimCitus:
The idea is to embed an HTTP server within pg_autoctl and expose an API that, when all is done, will allow:
  1. fetch the current status of things
  2. run pg_autoctl without a monitor node and instead control the FSM from the HTTP API

This PR is still missing tests, and the details of the ways the API are exposed might be improved a bit. For instance instead of `POST /api/1.0/fsm/assign/state` it should probably be `POST /api/1.0/fsm/assign?goalstate=state`.


marcocitus:
One high-level question I have is whether we are happy to expose the state machine directly or we should have some layer of abstraction on top. The API will effectively set the state machine in stone, because we cannot easily change it without breaking clients. 

An alternative approach could be to model the API after the actual transitions that we are asking pg_autoctl to make (e.g. "enable synchronous replication"").

DimCitus:
I think the FSM has to be part of the API we expose anyway, because that's what makes sense and how you can achieve robust HA implementations. The way I see our implementation, the FSM is central to the value added in pg_auto_failover.

That said I could see value in exposing the internal facilities in HTTP too, much like we do for the `pg_autoctl do ...` sub-commands. It would allow a more *build it yourself* integration and might be necessary in some cases, though I would still recommend using the FSM as much as possible.

Even if that means that we need to consider the FSM as being an exposed API for the development...

DimCitus:
Thinking some more about it, it appears to me that if we open our lower level API directly, then we won't be in a position to maintain any state, or at least not in a position to implement transitions. So then maybe it would make sense to have

  - `--disable-monitor`
  - `--disable-fsm`

It would be that disabling FSM would automatically disable the monitor too, obviously. The API would then expose all the entry points that we have with PG_AUTOCTL_DEBUG set, allowing to use our code to execute about any Postgres related operation from the outside of the container.

With such an approach I think it would be fair to say that the `--disable-fsm` work should be done in a separate (dedicated) PR.

src/bin/lib/README.md
```diff
@@ -20,3 +20,15 @@ The single-header library is used to implement parsing "modern" command lines.
 ## Configuration file parsing
 
 We utilize the "ini.h" ini-file reader from https://github.com/mattiasgustavsson/libs
+
+## HTTP and JSON
+
+We use https://github.com/vurtun/mmx implementation of an HTTP server in C,
```
fdr: that is some compact code.

src/bin/pg_autoctl/cli_do_fsm.c
```diff
@@ -176,12 +185,7 @@ keeper_cli_fsm_state(int argc, char **argv)
 	}
 
 	/* check that the state matches with running PostgreSQL instance */
-	if (!keeper_update_pg_state(&keeper))
-	{
-		log_fatal("Failed to update the keeper's state from the local "
-				  "PostgreSQL instance, see above.");
-		exit(EXIT_CODE_BAD_STATE);
-	}
+	(void) keeper_update_pg_state(&keeper);
```
DimCitus: That looks like a rebase “mistake” (either the tooling or myself). Will fix.

src/bin/pg_autoctl/fsm_transition.c
```diff
@@ -81,6 +81,41 @@ fsm_init_primary(Keeper *keeper)
...
+	}
+	else if (initState.pgInitState >= PRE_INIT_STATE_RUNNING)
+	{
+		log_error("PostgreSQL is already running at \"%s\", refusing to "
+				  "initialize a new cluster on-top of the current one.",
+				  pgSetup.pgdata);
+
+		return false;
+	}
+
```
DimCitus: Looks like another rebase mistake to deal with.

src/bin/pg_autoctl/keeper_listener.c
```diff
@@ -0,0 +1,398 @@
...
+ *      - pg_autoctl do fsm assign single
+ *      - pg_autoctl do fsm assign wait_primary
+ *      - pg_autoctl enable maintenance
+ *      - pg_autoctl disable maintenance
+ *    - postgres -p 5432 -h localhost -k /tmp
+ *
+ *   We still want the `postgres` process to run as a chile of the main
+ *   pg_autoctl service. When PostgreSQL is started by a listener command, this
+ *   means we also need a communication/execution channel with the pg_autoctl
+ *   parent process.
```
DimCitus: This part has not been implemented, it's unclear whether it's necessary to add that level of complexity. It's also something that could be done separately later in another PR.

DimCitus:
Since the creation of this PR, the following things did change:
  1. pg_autoctl is now the parent process of Postgres, and the keeper is the “node-active” sibling process
  2. most pg_autoctl commands know how to return a JSON output
  3. it is now possible to enable/disable a monitor online

Embedding a web server might not be better than having a cgi-bin based frontend, or hacking together a special web app in Go or Clojure or something self-contained that would either query the monitor or run `pg_autoctl` commands and parse their `--json` output. The advantage of embedding an http server written in C is unclear at this point.

PR #65, opened 2019-09-16
Remove get_current_GUC_value function

### LINKED ISSUE #60: pg_autoctl failed to create monitor on mac
I was unable to create monitor using 1.0.4 release with the following command

```
pg_autoctl create monitor --pgdata ${WORK_DIR}/monitor --pgport 6000 --nodename localhost
```
It failed with
```
 LOG:  syntax error in file "/Users/mtuncer/dev/ha/test/monitor/postgresql-auto-failover.conf" line 4, near end of line
 FATAL:  configuration file "/Users/mtuncer/dev/ha/test/monitor/postgresql.conf" contains errors
```

when I checked `postgresql-auto-failover.conf` failure was port setting file with no value set.

```
port = 
```

I checked `monitor/pg_autoctl.cfg` that is stored in temporary cache. Port setting was correctly applied. We just failed to move that setting into `postgresql-auto-failover.conf`

I tested this with ubuntu, it was fine.
I also testhed v1.0.3 with mac which is also fine. It must be something we introduced along 1.0.4

mtuncer:
`get_current_GUC_value` had an invalid memory usage. Removing the function was easier than fixing. 

Fixes #60

### DIFF
--- a/src/bin/pg_autoctl/pgctl.c
+++ b/src/bin/pg_autoctl/pgctl.c
@@ -40,8 +40,9 @@
 #define PROGRAM_NOT_RUNNING 3
 
 
-static const char *get_current_GUC_value(const char *name, const char *value,
-										 PostgresSetup *pgSetup);
+static bool get_current_GUC_value(const char *name, const char *value,
+								  PostgresSetup *pgSetup,
+								  PQExpBuffer valueBuffer);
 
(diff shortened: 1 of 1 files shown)


mtuncer:
decided to remove the function altogether since it does not do much anyway.

mtuncer:
done

PR #76, opened 2019-09-25
Fix sprintf errors for size_t type on Windows platform

pashagolub:
Use `%zu` format placeholder for `size_t`, and use `%lld` format placeholer for int64.

Error:
```
notifications.c:87:4: error: format '%lu' expects argument of type 'long unsigned int', but argument 5 has type 'size_t' {aka 'long long unsigned int'} [-Werror=format=]
   87 |    "S:%s:%s:%lu.%s:%d:%zd:%zu.%s:%d",
      |    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

### DIFF
--- a/src/monitor/notifications.c
+++ b/src/monitor/notifications.c
@@ -84,7 +84,7 @@ NotifyStateChange(ReplicationState reportedState,
 	 * string in the message. Parsing is then easier on the receiving side too.
 	 */
 	sprintf(payload,
-			"S:%s:%s:%lu.%s:%d:%ld:%lu.%s:%d",
+			"S:%s:%s:%zu.%s:%d:%lld:%zu.%s:%d",
 			ReplicationStateGetName(reportedState),
 			ReplicationStateGetName(goalState),
 			strlen(formationId),


DimCitus:
The travis build is now failing because on their platform the `%lld` is not the right way to do things. Maybe we need to play with `#include <inttypes.h>` here again and use something like `PRId64`?

```
2536 notifications.c:87:26: error: format ‘%lld’ expects argument of type ‘long long int’, but argument 8 has type ‘int64 {aka long int}’ [-Werror=format=]
2537    "S:%s:%s:%zu.%s:%d:%lld:%zu.%s:%d",
2538                       ~~~^
2539                       %ld
2540 cc1: all warnings being treated as errors
2541
```

pashagolub:
Oh, god... Dunno. I was just trying to compile it 💃

DimCitus:
Meanwhile the notification format has been switched to using JSON, and we don't use those non portable printf formatters, I believe. Closing this issue now, even though it would be best that we try to compile on windows again... I just don't know how to do that...

DimCitus:
See also #729 where we clean format strings to be compatible with i386. I believe that's improving/solving our case here.

PR #89, opened 2019-10-22
 Introduce monitor and cli variables for multi standby configuration

mtuncer:
This work introduces multi-standby configuration variables
- formation.number_sync_standbys : integer, number of nodes in write quorum
- node.candidate_priority : integer(1-100), priority of the candidate to be promoted, 0 if the node is never meant to be primary
- node.replication_quorum : boolean, true if the node joins in write quorum

`register_node` and `create_formation` methods have those parameters to set them during creation time. I have also added `set_formation_number_sync_standbys`, `set_node_candidate_priority,`, and set_node_replication_quorum` UDFs with their corresponding get functions

`node_active` now also return candidate_priority and replication_quorum parameters.


src/bin/pg_autoctl/monitor.c
```diff
@@ -496,6 +548,25 @@ parseNodeState(void *ctx, PGresult *result)
 		++errors;
 	}
 
+	value = PQgetvalue(result, 0, 3);
+	if (sscanf(value, "%d", &context->assignedState->candidatePriority) != 1)
```
DimCitus: We don't use `sscanf`, please see about using `strtol` and friends.

src/bin/pg_autoctl/pgsetup.c
```diff
@@ -327,6 +327,16 @@ pg_setup_init(PostgresSetup *pgSetup,
 				options->authMethod, NAMEDATALEN);
 	}
 
+	if (options->quorum >= 0)
```
DimCitus: Do we need this test? if yes, why? I would think we want to duplicate the default value when that's the one that is setup at this point...

src/monitor/node_metadata.c
```diff
@@ -521,6 +532,54 @@ ReportAutoFailoverNodeHealth(char *nodeName, int nodePort,
...
+		CStringGetTextDatum(nodeName),        /* nodename */
+		Int32GetDatum(nodePort)               /* nodeport */
+	};
+	const int argCount = sizeof(argValues) / sizeof(argValues[0]);
+	int spiStatus = 0;
+
+	const char *updateQuery =
+		"UPDATE " AUTO_FAILOVER_NODE_TABLE " SET "
+		"candidate_priority = $1, replication_quorum = $2, "
+		"statechangetime = now() WHERE nodename = $3 AND nodeport = $4";
```
DimCitus: Why do we update `statechangetime` here, when neither `goal state` nor `reported state` has been changed?

src/monitor/pgautofailover.sql
```diff
@@ -362,13 +378,16 @@ CREATE FUNCTION pgautofailover.current_state
...
+   OUT candidate_priority	int,
+   OUT replication_quorum	bool
  )
 RETURNS SETOF record LANGUAGE SQL STRICT
 AS $$
-   select nodename, nodeport, groupid, nodeid, reportedstate, goalstate
-     from pgautofailover.node
-    where formationid = formation_id
+   select nodename, nodeport, groupid, nodeid, reportedstate, goalstate,
+   		  candidate_priority, replication_quorum
```
DimCitus: Those new output columns are NOT printed in `monitor_print_state` as far as I can see. This is strange, we should either not report them or report them and then print them too. Why report them and not print them? (or did I miss the change in the `monitor_print_state` function maybe?)

src/bin/pg_autoctl/keeper.c
```diff
@@ -590,7 +590,6 @@ keeper_get_replication_state(Keeper *keeper)
 				postgres->currentLSN,
 				PG_LSN_MAXLENGTH,
 				missingStateOk);
-		log_warn("latest received lsn = %s", postgres->currentLSN);
```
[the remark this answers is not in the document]
mtuncer: aggreed, https://github.com/citusdata/pg_auto_failover/pull/93

src/bin/pg_autoctl/cli_common.c
```diff
@@ -216,7 +223,45 @@ cli_create_node_getopts(int argc, char **argv,
...
+				/* { "enable-write-quorum", no_argument, NULL, 'Q'} */
+				log_trace("--enable-write-quorum");
+				LocalOptionConfig.pgSetup.quorum = 1;
+				break;
+			}
+			case 'P':
+			{
+				/* { "failover-priority", required_argument, NULL, 'P'} */
+				int candidatePriority = -1;
+				int scanResult = sscanf(optarg, "%d", &candidatePriority);
```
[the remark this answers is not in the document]
mtuncer: done

src/bin/pg_autoctl/cli_common.c
```diff
@@ -216,7 +223,45 @@ cli_create_node_getopts(int argc, char **argv,
...
 
+				LocalOptionConfig.pgSetup.candidate_priority = candidatePriority;
+				log_trace("--failover-priority %d", candidatePriority);
+				break;
+			}
+			case 'q':
+			{
+				/* { "replication-quorum", required_argument, NULL, 'q'} */
+				int replicationQuorum = -1;
+				int scanResult = sscanf(optarg, "%d", &replicationQuorum);
```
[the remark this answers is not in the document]
mtuncer: ok

src/bin/pg_autoctl/cli_create_drop_node.c
```diff
@@ -74,6 +74,9 @@ CommandLine create_postgres_command =
 				 "  --formation   pg_auto_failover formation\n"
 				 "  --monitor     pg_auto_failover Monitor Postgres URL\n"
 				 "  --auth        authentication method for connections from monitor\n"
+				 "  --replication-quorum minimum number of standbys to confirm write\n"
+				 "  --failover-priority priority of the node to be promoted to become primary\n"
+				 "  --enable-write-quorum node participates in write quorum\n"
```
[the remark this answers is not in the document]
mtuncer: ok

src/bin/pg_autoctl/defaults.h
```diff
@@ -48,6 +48,10 @@
 #define POSTGRESQL_FAILS_TO_START_TIMEOUT 20
 #define POSTGRESQL_FAILS_TO_START_RETRIES 3
 
+#define FAILOVER_FORMATION_REPLICATION_QUORUM 1
+#define FAILOVER_NODE_CANDIDATE_PRIORITY 100
+#define FAILOVER_NODE_QUORUM 1
```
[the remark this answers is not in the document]
mtuncer: used the defaults in cli_common

src/bin/pg_autoctl/ini_file.c
```diff
@@ -39,6 +39,8 @@ read_ini_file(const char *filename, IniOption *optionList)
 	ini = ini_load(fileContents, NULL);
 	free(fileContents);
 
+	log_error("Reading configuration file at %s", filename);
```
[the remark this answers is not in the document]
mtuncer: it needs to be removed

src/bin/pg_autoctl/ini_file.c
```diff
@@ -552,7 +554,7 @@ ini_set_setting(const char *filename, IniOption *optionList,
 {
 	IniOption *option = NULL;
 
-	log_debug("Reading configuration from %s", filename);
+	log_error("Reading configuration from %s", filename);
```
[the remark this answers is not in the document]
mtuncer: leftover from debugging, removed

src/bin/pg_autoctl/keeper_config.c
```diff
@@ -127,6 +127,28 @@
 							&(config->postgresql_restart_failure_max_retries), \
 							POSTGRESQL_FAILS_TO_START_RETRIES)
 
+#define OPTION_FAILOVER_FORMATION_REPLICATION_QUORUM(config) \
+	make_int_option_default("failover", "replication_quorum", \
+							NULL, \
```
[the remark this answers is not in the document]
mtuncer: ok

src/bin/pg_autoctl/keeper_config.c
```diff
@@ -127,6 +127,28 @@
...
+							false, \
+							&(config->replication_quorum), \
+							FAILOVER_FORMATION_REPLICATION_QUORUM)
+
+#define OPTION_FAILOVER_NODE_CANDIDATE_PRIORITY(config) \
+	make_int_option_default("failover", "candidate_priority", \
+							NULL, \
```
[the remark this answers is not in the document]
mtuncer: ok

src/bin/pg_autoctl/keeper_config.c
```diff
@@ -127,6 +127,28 @@
...
+							false, \
+							&(config->pgSetup.candidate_priority), \
+							FAILOVER_NODE_CANDIDATE_PRIORITY)
+
+#define OPTION_FAILOVER_NODE_QUORUM(config) \
+	make_int_option_default("failover", "quorum", \
+							NULL, \
```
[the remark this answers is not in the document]
mtuncer: ok

src/bin/pg_autoctl/monitor.c
mtuncer: We used sscanf in the same function, please see how we parse nodeId and groupId fields.
We should do the same for all if we want to switch to strtol family

src/bin/pg_autoctl/pgsetup.c
mtuncer: they were there for additional checks, guess it is too much

src/monitor/node_metadata.c
mtuncer: I treated candidate_priority and replication_quorum changes as state changes. Fine with not setting it.

src/monitor/pgautofailover.sql
mtuncer: modified monitor_print_state to display new columns

src/bin/pg_autoctl/cli_common.c
```diff
@@ -68,6 +71,10 @@ cli_create_node_getopts(int argc, char **argv,
 	LocalOptionConfig.prepare_promotion_walreceiver = -1;
 	LocalOptionConfig.postgresql_restart_failure_timeout = -1;
 	LocalOptionConfig.postgresql_restart_failure_max_retries = -1;
+	LocalOptionConfig.replication_quorum = 1;
+	LocalOptionConfig.pgSetup.candidate_priority = 100;
+	LocalOptionConfig.pgSetup.quorum = 1;
```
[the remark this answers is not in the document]
mtuncer: Code is changed to use default constants.  There are help/error messages to direct user to correct usage of the functions. However, discussion is left in the documentation.

src/bin/pg_autoctl/ini_file.c
```diff
@@ -552,8 +552,6 @@ ini_set_setting(const char *filename, IniOption *optionList,
 {
 	IniOption *option = NULL;
 
-	log_debug("Reading configuration from %s", filename);
-
```
DimCitus: Why this PR is removing this debug entry?

src/bin/pg_autoctl/keeper_config.c
```diff
@@ -131,6 +131,7 @@
 							&(config->postgresql_restart_failure_max_retries), \
 							POSTGRESQL_FAILS_TO_START_RETRIES)
 
+
```
DimCitus: Nitpicking: extra whitespace added.

DimCitus: I realise the proper answer to those is to setup some automatic indenting rules on the repository, but we failed to do that up to now, so I guess we still have to be careful about it :(

src/bin/pg_autoctl/keeper_config.h
```diff
@@ -45,6 +45,7 @@ typedef struct KeeperConfig
 	int prepare_promotion_walreceiver;
 	int postgresql_restart_failure_timeout;
 	int postgresql_restart_failure_max_retries;
+
```
DimCitus: Extra white space.

src/bin/pg_autoctl/monitor.c
```diff
@@ -346,6 +358,217 @@ monitor_node_active(Monitor *monitor,
 }
 
 
+/*
+ * monitor_set_node_replication_settings updates the monitor on the changes
+ * in the node replication settings. Caller should first retrieve
+ * existing values if setting only one parameter.
```
DimCitus: Why build the API that way? It's fragile. We could maybe use the classic Common Lisp trick for function keyword parameters where for each parameter there's an extra bit of information: parameter-name-s-p, where s-p stands for supplied-predicate. It is set to true only when a specific value has been passed through for this parameter. That allows to pass NULL and say “yeah I intentionally passed NULL, it's not the default”.

So we could have two extra bool parameters here, set_candidate_priority and set_replication_quorum, that would only be true when that's what we want to do.

Another way to do it would be to write two functions, one per parameter. Do we have any call-sites that are setting both the values in a single call?

src/bin/pg_autoctl/cli_get_set_properties.c
```diff
@@ -0,0 +1,368 @@
...
+		/* errors have already been logged */
+		exit(EXIT_CODE_BAD_ARGS);
+	}
+
+	if (!monitor_get_node_replication_settings(&(keeper.monitor),
+			keeper.state.current_node_id,
+			&assignedCandidatePriority, &assignedReplicationQuorum))
+	{
+		exit(EXIT_CODE_BAD_ARGS);
+	}
```
DimCitus: This technique only works in the happy case. If two commands are run in parallel (by mistake, or because of a bug somewhere in the scripts calling `pg_autoctl`, etc), then we have race conditions all the way. On way to avoid them is to use `UPDATE table SET param = new value WHERE param = old value` but the function call API you've set up on the monitor does not allow that.

I think it would be easier and safer to just have two functions here, one per parameter. Do we have any call-site that wants to set both parameters at the same time?

src/bin/pg_autoctl/monitor.c
```diff
@@ -346,6 +358,217 @@ monitor_node_active(Monitor *monitor,
...
+	return true;
+}
+
+/*
+ * monitor_get_node_replication_settings retrieves replication settings
+ * from the monitor.
+ */
+bool
+monitor_get_node_replication_settings(Monitor *monitor, int nodeid,
+									  int *candidate_priority, bool *replicationQuorum)
```
DimCitus: We only have two parameters at the moment, though we might add more in the future. I think we should make a `typedef struct` for those parameters and use the struct here rather than every single parameter separately.

src/bin/pg_autoctl/monitor.c
```diff
@@ -346,6 +358,217 @@ monitor_node_active(Monitor *monitor,
...
+	}
+
+	*candidate_priority = parseContext.candidatePriority;
+	*replicationQuorum = parseContext.replicationQuorum;
+
+	return true;
+}
+
+
+static void parseNodeReplicationSettings(void *ctx, PGresult *result)
```
DimCitus: Missing function-level comments. Formatting is off, too.

src/bin/pg_autoctl/monitor.c
```diff
@@ -346,6 +358,217 @@ monitor_node_active(Monitor *monitor,
...
+		context->parsedOK = false;
+		return;
+	}
+
+	/* if we reach this line, then we're good. */
+	context->parsedOK = true;
+}
+
+
+bool monitor_get_formation_number_sync_standbys(Monitor *monitor, char *formation,
```
DimCitus: Missing function-level comments. Formatting.

src/bin/pg_autoctl/monitor.c
```diff
@@ -346,6 +358,217 @@ monitor_node_active(Monitor *monitor,
...
+
+	*numberSyncStandbys = parseContext.intVal;
+
+	return true;
+}
+
+
+bool
+monitor_set_formation_number_sync_standbys(Monitor *monitor, char *formation,
+										   int numberSyncStandbys)
```
DimCitus: Missing function-level comments.

src/bin/pg_autoctl/monitor.c
```diff
@@ -623,13 +865,15 @@ printCurrentState(void *ctx, PGresult *result)
 		}
 	}
 
-	fprintf(stdout, "%*s | %6s | %5s | %5s | %17s | %17s\n",
+	fprintf(stdout, "%*s | %6s | %5s | %5s | %17s | %17s | %18s | %18s\n",
 			maxNodeNameSize, "Name", "Port",
-			"Group", "Node", "Current State", "Assigned State");
+			"Group", "Node", "Current State", "Assigned State",
+			"Candidate Priority", "Replication Quorum");
```
DimCitus: Wow, +18 columns is a lot. Let's make it "prio" and "quorum" maybe? we're not going to fit in most screen sizes. Maybe it's time to have a verbose and a non-verbose output. In another branch I have implement `--json` for the output here, too.

Let's have a non-verbose default output with only name, port, current state, assigned state, priority, quorum; and a verbose one where we add in the group and node IDs. I don't think those are needed for normal users, now that I think about it.

src/bin/pg_autoctl/pgsetup.h
```diff
@@ -113,6 +113,8 @@ typedef struct pg_setup
 	PostgresControlData control;            /* pg_controldata pgdata */
 	PostgresPIDFile pidFile;                /* postmaster.pid information */
 	PgInstanceKind pgKind;					/* standalone/coordinator/worker */
+	int candidate_priority;					/* promotion candidate priority 0-100, 0 -> not a promotion candidate */
+	bool replication_quorum;				/* true if participates in write quorum, false otherwise */
```
DimCitus: We should have those in a separate structure for clarity. Also, as mentioned in another comment in this review, we might add more settings to the nodes and formations later.

src/monitor/node_active_protocol.c
```diff
@@ -965,7 +988,89 @@ stop_maintenance(PG_FUNCTION_ARGS)
...
+		char message[BUFSIZE];
+
+		ReportAutoFailoverNodeReplicationState(currentNode->nodeName,
+				currentNode->nodePort,
+				currentNode->candidatePriority,
+				currentNode->replicationQuorum);
+
+		LogAndNotifyMessage(
+			message, BUFSIZE,
+			"Updating candidatePriority and replicationQuorum.");
```
DimCitus: We should log the new values here?

src/monitor/node_active_protocol.c
```diff
@@ -965,7 +988,89 @@ stop_maintenance(PG_FUNCTION_ARGS)
...
+{
+	int32 nodeid = PG_GETARG_INT32(0);
+	int candidatePriority = PG_GETARG_INT32(1);
+	bool replicationQuorum = PG_GETARG_BOOL(2);
+	AutoFailoverNode *currentNode = NULL;
+	bool nodeUpdated = false;
+
+
+	ereport(LOG, (errmsg("updating node replication properties for node %d, candidate_priority=%d, replication_quorum=%s",
+						  nodeid, candidatePriority, replicationQuorum?"true":"false")));
```
DimCitus: This looks more like a debug message, but I'm not too sure. Do we want to keep it?

src/monitor/node_active_protocol.c
```diff
@@ -965,7 +988,89 @@ stop_maintenance(PG_FUNCTION_ARGS)
...
+
+	currentNode = GetAutoFailoverNodeWithId(nodeid);
+
+	if (currentNode == NULL)
+	{
+		ereport(ERROR, (errmsg("node %d is not registered", nodeid)));
+	}
+
+	LockFormation(currentNode->formationId, ShareLock);
+	LockNodeGroup(currentNode->formationId, currentNode->groupId, ExclusiveLock);
```
DimCitus: We lock the node after having read the current values from the table. In between those two operations, a concurrent function call could have changed the values already. Is there a way that we don't depend on the current values that way in the implementation of this function?

Maybe on the monitor side too we should have a specific function per parameter. Or a way to not have to guess if there's a value that's been given but that we don't actually want to update...

src/monitor/node_metadata.c
```diff
@@ -276,6 +286,56 @@ GetAutoFailoverNode(char *nodeName, int nodePort)
...
+
+	Datum argValues[] = {
+		Int32GetDatum(nodeid)        /* nodeid */
+	};
+	const int argCount = sizeof(argValues) / sizeof(argValues[0]);
+	int spiStatus = 0;
+
+	const char *selectQuery =
+		SELECT_ALL_FROM_AUTO_FAILOVER_NODE_TABLE
+		" WHERE nodeid = $1";
```
DimCitus: Do we need to add nodename and nodeport to the WHERE clause here? I think I remember that's what we do elsewhere, to prevent things from having changed concurrently.

src/monitor/node_metadata.c
```diff
@@ -457,6 +523,7 @@ ReportAutoFailoverNodeState(char *nodeName, int nodePort,
 		"walreporttime = CASE $4 WHEN '0/0'::pg_lsn THEN walreporttime ELSE now() END, "
 		"statechangetime = now() WHERE nodename = $5 AND nodeport = $6";
 
+
```
DimCitus: Extra while space?

src/monitor/node_metadata.c
```diff
@@ -521,6 +588,51 @@ ReportAutoFailoverNodeHealth(char *nodeName, int nodePort,
...
+/*
+ * ReportAutoFailoverNodeReplicationState persists the replication properties of
+ * a node.
+ *
+ * We use SPI to automatically handle triggers, function calls, etc.
+ */
+void
+ReportAutoFailoverNodeReplicationState(char *nodeName, int nodePort,
+							int candidatePriority,
+							bool replicationQuorum)
```
DimCitus: We are not reporting the **state** of the node replication here, but the **settings** of the node replication, right? I think we should change the name of that function to reflect that.

src/monitor/pgautofailover.sql
```diff
@@ -107,13 +123,15 @@ CREATE TABLE pgautofailover.event
...
+    nodeport         int not null,
     reportedstate    pgautofailover.replication_state not null,
     goalstate        pgautofailover.replication_state not null,
     reportedrepstate text,
     reportedlsn      pg_lsn not null default '0/0',
+    candidatepriority int,
+    replicationquorum bool,
     description      text,
-
+
```
DimCitus: Extra white space.

src/bin/pg_autoctl/cli_formation.c
```diff
@@ -199,7 +201,21 @@ keeper_cli_formation_create_getopts(int argc, char **argv)
...
-
+			case 'r':
+			{
+				/* { "number-sync-stanbys", required_argument, NULL, 'S'} */
+				int numberSyncStanbys = strtol(optarg, NULL, 10);
+
+				if (errno == EINVAL|| numberSyncStanbys < 0)
+				{
+					log_fatal("--number-sync-stanbys argument is not valid."
+							  " Use a non-negative integer value.");
```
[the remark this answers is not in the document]
mtuncer: I picked non-negative because we also support 0 there. Shall we drop supporting 0 ?

src/monitor/node_metadata.c
mtuncer: do you suggest filtering based on (nodeId, nodename, nodeport) ?
I originally had the latter two here, then you asked me to do checks based on nodeid.  Can  you clarify ?

src/monitor/node_metadata.c
mtuncer: `ReportAutoFailoverNodeReplicationSetting`

src/bin/pg_autoctl/cli_common.c
```diff
@@ -48,6 +48,10 @@ bool allowRemovingPgdata = false;
  *		{ "monitor", required_argument, NULL, 'm' },
  *		{ "allow-removing-pgdata", no_argument, NULL, 'R' },
  *		{ "help", no_argument, NULL, 0 },
+ *		{ "enable-replication-quorum", no_argument, NULL, 'Q'},
+ *		{ "disable-replication-quorum", no_argument, NULL, 'q'},
+ *		{ "candidate-priority", required_argument, NULL, 'P'},
+ *		{ "replication-quorum", required_argument, NULL, 'r'},
```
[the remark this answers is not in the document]
mtuncer: It was in the original design. I would love to remove enable/disable part.

src/bin/pg_autoctl/cli_create_drop_node.c
```diff
@@ -63,17 +63,21 @@ CommandLine create_postgres_command =
...
+				 "  --username                    PostgreSQL's username\n"
+				 "  --dbname                      PostgreSQL's database name\n"
+				 "  --nodename                    pg_auto_failover node\n"
+				 "  --formation                   pg_auto_failover formation\n"
+				 "  --monitor                     pg_auto_failover Monitor Postgres URL\n"
+				 "  --auth                        authentication method for connections from monitor\n"
+				 "  --candidate-priority          priority of the node to be promoted to become primary\n"
+				 "  --enable-replication-quorum   node participates in write quorum\n"
+				 "  --disable-replication-quorum  node does not participate in write quorum\n"
+				 "  --replication-quorum          1 if node participates in write quorum 0 otherwise\n"
```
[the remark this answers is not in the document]
mtuncer: done

src/bin/pg_autoctl/cli_formation.c
```diff
@@ -145,6 +146,7 @@ keeper_cli_formation_create_getopts(int argc, char **argv)
 		{ "enable-secondary", no_argument, NULL, 's' },
 		{ "disable-secondary", no_argument, NULL, 'S' },
 		{ "help", no_argument, NULL, 'h' },
+		{ "number-sync-stanbys", required_argument, NULL, 'r' },
```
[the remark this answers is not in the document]
mtuncer: `r` was a left over from previous name, changed

src/bin/pg_autoctl/cli_get_set_properties.c
```diff
@@ -0,0 +1,368 @@
...
+		/* errors have already been logged */
+		exit(EXIT_CODE_BAD_ARGS);
+	}
+
+	if (!monitor_get_node_replication_settings(&(keeper.monitor),
+			keeper.state.current_node_id,
+			candidatePriority, replicationQuorum))
+	{
+		exit(EXIT_CODE_MONITOR);
+	}
```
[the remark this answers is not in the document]
mtuncer: done

src/bin/pg_autoctl/cli_get_set_properties.c
```diff
@@ -0,0 +1,368 @@
...
+}
+
+
+static void
+cli_get_node_replication_quorum(int argc, char **argv)
+{
+	int candidatePriority = 0;
+	bool replicationQuorum = false;
+
+	get_node_replication_settings(&candidatePriority, &replicationQuorum);
```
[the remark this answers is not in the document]
mtuncer: opted for boolean as you suggested in an earlier comment

src/bin/pg_autoctl/cli_get_set_properties.c
```diff
@@ -0,0 +1,368 @@
...
+			keeper.state.current_node_id,
+			candidatePriority, replicationQuorum))
+	{
+		exit(EXIT_CODE_MONITOR);
+	}
+}
+
+
+static void
+cli_get_node_replication_quorum(int argc, char **argv)
```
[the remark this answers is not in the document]
mtuncer: added

src/bin/pg_autoctl/cli_get_set_properties.c
```diff
@@ -0,0 +1,368 @@
...
+	bool replicationQuorum = false;
+
+	get_node_replication_settings(&candidatePriority, &replicationQuorum);
+
+	fprintf(stdout, "%c\n", replicationQuorum?'t':'f');
+}
+
+
+static void
+cli_get_node_candidate_priority(int argc, char **argv)
```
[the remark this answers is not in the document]
mtuncer: done

src/bin/pg_autoctl/cli_get_set_properties.c
```diff
@@ -0,0 +1,368 @@
...
+{
+	int candidatePriority = 0;
+	bool replicationQuorum = false;
+
+	get_node_replication_settings(&candidatePriority, &replicationQuorum);
```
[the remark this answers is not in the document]
mtuncer: done

src/bin/pg_autoctl/cli_get_set_properties.c
```diff
@@ -0,0 +1,368 @@
...
+
+	fprintf(stdout, "%d\n", candidatePriority);
+}
+
+
+static void
+cli_get_formation_number_sync_standbys(int argc, char **argv)
```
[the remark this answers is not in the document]
mtuncer: added

src/bin/pg_autoctl/cli_get_set_properties.c
```diff
@@ -0,0 +1,368 @@
...
+	int *candidatePriority = NULL;
+	bool *replicationQuorum = NULL;
+	int assignedCandidatePriority = -1;
+	bool assignedReplicationQuorum = false;
+
+	keeper.config = keeperOptions;
+
+	if (argc != 2)
+	{
+		log_error("Expected node property name and a value, instead found %d arguments", argc);
```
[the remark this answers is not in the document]
mtuncer: ok

src/bin/pg_autoctl/cli_get_set_properties.c
```diff
@@ -0,0 +1,368 @@
...
+	bool missing_pgdata_is_ok = false;
+	bool pg_is_not_running_is_ok = true;
+	char *name = NULL;
+	char *value = NULL;
+	int numberSyncStandbys = -1;
+
+
+	if (argc != 2)
+	{
+		log_error("Expected formation property name and a value, instead found %d arguments", argc);
```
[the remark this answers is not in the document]
mtuncer: ok

src/bin/pg_autoctl/cli_get_set_properties.c
```diff
@@ -0,0 +1,368 @@
...
+	}
+
+	if (!monitor_init_from_pgsetup(&monitor, &config.pgSetup))
+	{
+		/* errors have already been logged */
+		exit(EXIT_CODE_BAD_ARGS);
+	}
+
+
+	monitor_set_formation_number_sync_standbys(&monitor, config.formation, numberSyncStandbys);
```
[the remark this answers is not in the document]
mtuncer: displayed an error message upon failure

src/bin/pg_autoctl/ini_file.c
mtuncer: looked like added by me in this PR. restored

src/bin/pg_autoctl/keeper_config.c
mtuncer: citus_indent would have been good

src/bin/pg_autoctl/keeper_config.h
mtuncer: done

src/bin/pg_autoctl/pgsetup.h
mtuncer: separated

src/monitor/node_active_protocol.c
mtuncer: separated functions in both cli and monitor. My earlier intention was to set them in a single call, specify NULL for not setting the value. It did not work as I expected.  Now resort to separate calls for everything.

src/monitor/node_metadata.c
mtuncer: removed

src/monitor/pgautofailover.sql
mtuncer: removed

src/bin/pg_autoctl/cli_get_set_properties.c
```diff
@@ -0,0 +1,368 @@
...
+
+CommandLine set_command =
+	make_command_set("set",
+					 "Set a pg_auto_failover node, or formation setting", NULL, NULL,
+					 NULL, set_subcommands);
+
+
+
+static void
+get_node_replication_settings(int *candidatePriority, bool *replicationQuorum)
```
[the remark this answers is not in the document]
mtuncer: done

src/bin/pg_autoctl/cli_formation.c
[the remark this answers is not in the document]
mtuncer: I checked citus and PG code for similar usage. Found that
PG has following snippet in  `src/backend/partitioning/partbounds.c` function `satisfies_hash_partition()`
```c
	if (remainder < 0)
		ereport(ERROR,
				(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
				 errmsg("remainder for hash partition must be a non-negative integer")));
```

So I am inclined to keep the message this way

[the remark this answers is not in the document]
mtuncer: I completely overlooked that you can't provide a negative number to cli command easily
when you wrote `pg_autoctl set node candidate-priority -1 --pgdata primary `  parameter parser treats -1 as a parameter and can't find it in valid options and we fail way before in the cli.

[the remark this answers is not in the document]
DimCitus: What about `-- candidate-priority -1` where the `--` usage signals to getup that there's no options to parse after that? Otherwise, agreed, let's use “non-negative” as in Citus, it might be the best way to have an easy to understand error message.

src/bin/pg_autoctl/cli_get_set_properties.c
```diff
@@ -0,0 +1,357 @@
...
+			log_error("candidate-priority value %s is not valid."
+					  " Valid values are integers from 0 to 100. ", value);
+			exit(EXIT_CODE_BAD_ARGS);
+		}
+
+		if (!monitor_set_node_candidate_priority(&(keeper.monitor),
+					keeper.state.current_node_id, keeper.config.nodename,
+					keeper.config.pgSetup.pgport, candidatePriority))
+		{
+			log_error("unable to set candidate-priority.");
```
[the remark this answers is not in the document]
DimCitus: Also printing the new value when the command is successful seems like the right thing to do. And that's what the `pg_autoctl config set` command does too.

src/bin/pg_autoctl/cli_get_set_properties.c
```diff
@@ -0,0 +1,357 @@
...
+
+	if (!monitor_init_from_pgsetup(&monitor, &config.pgSetup))
+	{
+		/* errors have already been logged */
+		exit(EXIT_CODE_BAD_ARGS);
+	}
+
+	if (!monitor_set_formation_number_sync_standbys(&monitor, config.formation, numberSyncStandbys))
+	{
+		log_error("Unable to set number-sync-standbys.");
```
DimCitus: Review error message, include more information, ensure non-zero EXIT_CODE_... in that case, and print out the new value when we could set it.

src/bin/pg_autoctl/monitor.c
```diff
@@ -346,6 +358,266 @@ monitor_node_active(Monitor *monitor,
...
+	paramValues[1] = nodeName,
+	paramValues[2] = intToString(nodePort).strValue;
+	paramValues[3] = candidatePriorityText;
+
+	if (!pgsql_execute_with_params(pgsql, sql,
+								   paramCount, paramTypes, paramValues,
+								   NULL, NULL))
+	{
+		log_error("Failed to update node candidate priority on node %d"
+				  " for candidate_priority : %s",
```
DimCitus: I think we should put the value in double-quotes to make it easier to read the error message here.

src/bin/pg_autoctl/monitor.c
```diff
@@ -346,6 +358,266 @@ monitor_node_active(Monitor *monitor,
...
+	paramValues[0] = intToString(nodeid).strValue;
+	paramValues[1] = nodeName;
+	paramValues[2] = intToString(nodePort).strValue;
+	paramValues[3] = replicationQuorumText;
+
+	if (!pgsql_execute_with_params(pgsql, sql, paramCount, paramTypes,
+								   paramValues, NULL, NULL))
+	{
+		log_error("Failed to update node replication quorum on node %d"
+				  " and replication_quorum  : %s",
```
DimCitus: Again, user-provided values in double-quotes.

src/bin/pg_autoctl/monitor.c
```diff
@@ -346,6 +358,266 @@ monitor_node_active(Monitor *monitor,
...
+	Oid paramTypes[2] = { TEXTOID, INT4OID};
+	const char *paramValues[2];
+	SingleValueResultContext parseContext = {PGSQL_RESULT_BOOL, false};
+	paramValues[0] = formation;
+	paramValues[1] = intToString(numberSyncStandbys).strValue;
+
+	if (!pgsql_execute_with_params(pgsql, sql, paramCount, paramTypes, paramValues,
+								   &parseContext, parseSingleValueResult))
+	{
+		log_error("Failed retrieve settings of formation \"%s\".", formation);
```
DimCitus: I think this is a copy-paste error, we are setting the parameter, not retrieving it.

src/bin/pg_autoctl/pgsetup.h
DimCitus: Well the values are still inline at the top-level of the pg_setup structure. What I meant is that we would have a  member of the struct using the new typedef `NodeReplicationSettings`. Then we would use `pgSetup.settings.candidate_priority`.

``` C
    NodeReplicationSettings settings;
```

src/monitor/node_active_protocol.c
```diff
@@ -965,7 +989,153 @@ stop_maintenance(PG_FUNCTION_ARGS)
...
+	int32 nodeId = PG_GETARG_INT32(0);
+	text *nodeNameText = PG_GETARG_TEXT_P(1);
+	char *nodeName = text_to_cstring(nodeNameText);
+	int32 nodePort = PG_GETARG_INT32(2);
+	int candidatePriority = PG_GETARG_INT32(3);
+	AutoFailoverNode *currentNode = NULL;
+	bool nodeUpdated = false;
+
+	ereport(LOG, (errmsg("updating candidate_priority for node %d, candidate_priority=%d",
+						  nodeId, candidatePriority)));
```
DimCitus: Looks like a debug message to me, do we want to keep it in the merge?

src/monitor/node_active_protocol.c
```diff
@@ -965,7 +989,153 @@ stop_maintenance(PG_FUNCTION_ARGS)
...
+						errmsg("invalid value for candidate_priority \"%d\" "
+							   "expected an integer value between 0 and 100", candidatePriority)));
+	}
+
+
+	if (candidatePriority != currentNode->candidatePriority)
+	{
+		currentNode->candidatePriority = candidatePriority;
+		nodeUpdated = true;
+	}
```
DimCitus: Why do we check if the new value is different from the old value here? I'm not sure there's any point in doing that. If there's a point though, as we said before, the fact that we're using the value from before we take a lock means there is a race condition. We should either stop doing that test (my preference), or make the test reliable.

src/monitor/node_active_protocol.c
```diff
@@ -965,7 +989,153 @@ stop_maintenance(PG_FUNCTION_ARGS)
...
+						  currentNode->nodePort,
+						  currentNode->pgsrSyncState,
+						  currentNode->reportedLSN,
+						  currentNode->candidatePriority,
+						  currentNode->replicationQuorum,
+						  message);
+
+	}
+
+	PG_RETURN_BOOL(nodeUpdated);
```
DimCitus: And so if the user sets the same value again, then we report a failure to update?

src/monitor/node_active_protocol.c
```diff
@@ -965,7 +989,153 @@ stop_maintenance(PG_FUNCTION_ARGS)
...
+	int32 nodeid = PG_GETARG_INT32(0);
+	text *nodeNameText = PG_GETARG_TEXT_P(1);
+	char *nodeName = text_to_cstring(nodeNameText);
+	int32 nodePort = PG_GETARG_INT32(2);
+	bool replicationQuorum = PG_GETARG_BOOL(3);
+	AutoFailoverNode *currentNode = NULL;
+	bool nodeUpdated = false;
+
+	ereport(LOG, (errmsg("updating node replication_quorum for node %d, replication_quorum=%s",
+						  nodeid, replicationQuorum?"true":"false")));
```
DimCitus: This looks like a debug message. Should we keep it in the merge?

src/monitor/node_active_protocol.c
```diff
@@ -965,7 +989,153 @@ stop_maintenance(PG_FUNCTION_ARGS)
...
+	}
+
+	LockFormation(currentNode->formationId, ShareLock);
+	LockNodeGroup(currentNode->formationId, currentNode->groupId, ExclusiveLock);
+
+	if (replicationQuorum != currentNode->replicationQuorum)
+	{
+		currentNode->replicationQuorum = replicationQuorum;
+		nodeUpdated = true;
+	}
```
DimCitus: Same comment as before about the semantics (returning false when we are setting the same value again) and race-conditions (get before lock, then use a value that might have been updated already). Again my preference is to not do that test at all and let Postgres handle the concurrency here as far as new values are concerned (let's still lock the formation).

src/monitor/node_metadata.c
```diff
@@ -521,6 +591,53 @@ ReportAutoFailoverNodeHealth(char *nodeName, int nodePort,
 	SPI_finish();
 }
 
+/*
+ * ReportAutoFailoverNodeReplicationState persists the replication properties of
```
DimCitus: Comment missed the function name update.

src/bin/pg_autoctl/monitor.c
```diff
@@ -346,6 +358,266 @@ monitor_node_active(Monitor *monitor,
...
 
+/*
+ * monitor_set_node_candidate_priority updates the monitor on the changes
+ * in the node candidate priority.
+ */
+bool
+monitor_set_node_candidate_priority(Monitor *monitor, int nodeid, char* nodeName, int nodePort, int candidate_priority)
+{
+	PGSQL *pgsql = &monitor->pgsql;
+	const char *sql = "SELECT * FROM pgautofailover.set_node_candidate_priority($1, $2, $3, $4)";
```
DimCitus: The server-side function `RETURNS bool`, which is a scalar value, not a composite, not a table. We should directly `SELECT pgautofailover.set_node_candidate_priority($1, $2, $3, $4)` I think.

src/bin/pg_autoctl/monitor.c
```diff
@@ -346,6 +358,266 @@ monitor_node_active(Monitor *monitor,
...
+ * monitor_set_node_replication_quorum updates the monitor on the changes
+ * in the node replication quorum.
+ */
+bool
+monitor_set_node_replication_quorum(Monitor *monitor, int nodeid,
+									char* nodeName, int nodePort,
+									bool replicationQuorum)
+{
+	PGSQL *pgsql = &monitor->pgsql;
+	const char *sql = "SELECT * FROM pgautofailover.set_node_replication_quorum($1, $2, $3, $4)";
```
DimCitus: Again, server-side function returns bool, a scalar.

src/bin/pg_autoctl/monitor.c
```diff
@@ -346,6 +358,265 @@ monitor_node_active(Monitor *monitor,
...
+	paramValues[0] = intToString(nodeid).strValue;
+	paramValues[1] = nodeName;
+	paramValues[2] = intToString(nodePort).strValue;
+	paramValues[3] = replicationQuorumText;
+
+	if (!pgsql_execute_with_params(pgsql, sql, paramCount, paramTypes,
+								   paramValues, NULL, NULL))
+	{
+		log_error("Failed to update node replication quorum on node %d"
+				  " and replication_quorum  : \"%s\"",
```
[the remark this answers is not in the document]
mtuncer: done

src/bin/pg_autoctl/monitor.c
```diff
@@ -346,6 +358,265 @@ monitor_node_active(Monitor *monitor,
...
+	paramValues[1] = nodeName,
+	paramValues[2] = intToString(nodePort).strValue;
+	paramValues[3] = candidatePriorityText;
+
+	if (!pgsql_execute_with_params(pgsql, sql,
+								   paramCount, paramTypes, paramValues,
+								   NULL, NULL))
+	{
+		log_error("Failed to update node candidate priority on node %d"
+				  " for candidate_priority : \"%s\"",
```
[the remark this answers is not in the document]
mtuncer: removed

src/bin/pg_autoctl/defaults.h
```diff
@@ -51,6 +51,10 @@
 #define POSTGRESQL_FAILS_TO_START_TIMEOUT 20
 #define POSTGRESQL_FAILS_TO_START_RETRIES 3
 
+#define FAILOVER_FORMATION_NUMBER_SYNC_STANBYS 1
+#define FAILOVER_NODE_CANDIDATE_PRIORITY 100
+#define FAILOVER_NODE_REPLICATION_QUORUM 1
```
DimCitus: In the PostgresSetup structure the slot NodeReplicationSettings is a boolean. We use this default value for filling-in this slot, if I'm following the PR correctly. The default reads as an integer rather than a boolean, is it on purpose?

src/bin/pg_autoctl/parsing.c
```diff
@@ -316,3 +316,101 @@ read_length_delimited_string_at(const char *ptr, char *buffer, int size)
...
+			break;
+	}
+
+	if (result)
+		*result = false;		/* suppress compiler warning */
+	return false;
+}
+
+bool
+parse_bool(const char *value, bool *result)
```
DimCitus: Missing function-level comments.

src/monitor/formation_metadata.c
```diff
@@ -520,3 +511,116 @@ IsCitusFormation(AutoFailoverFormation *formation)
...
+}
+
+
+/*
+ * AutoFailoverFormationGetDatum prepares a Datum from given formation.
+ * Caller is expected to provide fcinfo structure that contains compatible
+ * call result type.
+ */
+Datum
+AutoFailoverFormationGetDatum(FunctionCallInfo fcinfo, AutoFailoverFormation *formation)
```
DimCitus: I didn't say it before, I like the way you restructured that part of the code. Makes it clearer, thanks!

src/monitor/pgautofailover--1.0--1.1.sql
```diff
@@ -0,0 +1,265 @@
...
+ (
+    IN formation_id         text,
+    IN node_name            text,
+    IN node_port            int,
+    IN dbname               name,
+    IN desired_group_id     int default -1,
+    IN initial_group_role   pgautofailover.replication_state default 'init',
+    IN node_kind            text default 'standalone',
+    IN candidate_priority 	int default 100,
+    IN replication_quorum	bool default true,
```
DimCitus: data type alignment seems off?

mtuncer: done

src/bin/pg_autoctl/parsing.c
mtuncer: added

src/bin/pg_autoctl/defaults.h
mtuncer: it needs to be bool, fixed

Issue #62, opened 2019-09-13
Feature Request: skip management of pg_hba.conf


gmr:
I'm using Ansible to manage my PostgreSQL nodes and have to account for the line added by pg_auto_failover to pg_hba.conf in my Ansible config.

The line is also much more permissive than I'd prefer to allow.

Having an option that is along the lines of "--skip-pg_hba" that lets me manage that when using "create monitor" and "create postgres" would be ideal.

DimCitus:
Hi @gmr, this is something we have had in our internal list of wish list item and that was not given priority yet. I think it's fair to say we should have a look. Also it's a rather small contribution I believe, so if you would want to have a look I could provide help and guidance... otherwise I will just assign it to myself!

Bouke:
I'm looking for something similar as well. We use Puppet to manage `pg_hba.conf`. Any modifications outside of puppet are seen as deviations from the desired config state and are removed. However this also triggers a restart of the postgres service to make the new config apply. When the pgautofailover service starts, it'll make the same changes to `pg_hba.conf`, and on the next run of Puppet it'll detect those changes, causing another restart. As a result, the primary role continuously moves back and forth both nodes, causing disruption. For now, I think I'll disable the automatically triggered restart, but that would require manual intervention when I want to change those rules.

Markus-:
Anything new here?

gmr:
Yeah, I've slowly been making requested changes, just have been very limited on my time to do so. Sorry!

DimCitus:
Released now in version 1.2, `--skip-pg-hba`. Thanks for your contributions, enjoy the new release!

PR #91, opened 2019-10-28
Allow for conditional writing of pg_hba.conf (#62)

gmr:
What?
Add a cli option (`--no-pg-hba`) for create monitor and create postgres

Why?
In some environments it is preferable for configuration management create
the `pg_hba.conf` file and having `pg_auto_failover` create it will cause
configuration management and `pg_auto_failover` to overwrite each other.

In addition, the current implementation is too opinionated about the
CIDR block used. In my environment, I need something much more restrictive.

### DIFF
--- a/docs/ref/reference.rst
+++ b/docs/ref/reference.rst
@@ -92,6 +92,7 @@ commands are dealing with the monitor:
         --pgport      PostgreSQL's port number
         --nodename    hostname by which postgres is reachable
         --auth        authentication method for connections from data nodes
+        --no-pg-hba   Do not allow pg_autoctl to manage pg_hba.conf
 
      The ``--pgdata`` option is mandatory and default to the environment
      variable ``PGDATA``. The ``--pgport`` default value is 5432, and the
@@ -120,12 +121,12 @@ commands are dealing with the monitor:
      You may use the `--nodename` command line option to bypass the whole
      DNS lookup based process and force the local node name to a fixed
      value.
-     
+
      The ``--auth`` option allows setting up authentication method to be used
      for connections from data nodes with ``autoctl_node`` user.
      If this option is used, please make sure password is manually set on
      the monitor, and appropriate setting is added to `.pgpass` file on data node.
-     
+
      See :ref:`pg_auto_failover_security` for notes on `.pgpass`
 
   - ``pg_autoctl run``
@@ -279,6 +280,7 @@ The other commands accept the same set of options.
     --monitor     pg_auto_failover Monitor Postgres URL
     --auth        authentication method for connections from monitor
     --allow-removing-pgdata   Allow pg_autoctl to remove the database directory
+    --no-pg-hba   Do not allow pg_autoctl to manage pg_hba.conf
 
 Three different modes of initialization are supported by this command,
 corresponding to as many implementation strategies.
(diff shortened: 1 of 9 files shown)


docs/ref/reference.rst
```diff
@@ -92,6 +92,7 @@ commands are dealing with the monitor:
         --pgport      PostgreSQL's port number
         --nodename    hostname by which postgres is reachable
         --auth        authentication method for connections from data nodes
+        --no-pg-hba   Do not allow pg_autoctl to manage pg_hba.conf
```
DimCitus: I think I would like `--skip-hba` or `--skip-pg-hba` better. The rationale is to avoid negative commands and then the infamous *if not notThis then that* that makes every debugging session harder than necessary.

src/bin/pg_autoctl/cli_do_misc.c
```diff
@@ -255,12 +255,15 @@ keeper_cli_add_standby_to_hba(int argc, char **argv)
 		exit(EXIT_CODE_BAD_ARGS);
 	}
 
-	if (!primary_add_standby_to_hba(&postgres, standbyHostname, config.replication_password))
+	if (!noPgHba)
```
DimCitus: Here's the double negation I would like to avoid.

``` C
    if (!skipHBA)
    {
```

src/bin/pg_autoctl/fsm_transition.c
```diff
@@ -102,14 +103,26 @@ fsm_init_primary(Keeper *keeper)
...
 
-	if (!primary_create_user_with_hba(postgres,
-									  PG_AUTOCTL_HEALTH_USERNAME, password,
-									  monitorHostname, authMethod))
-	{
-		log_error("Failed to initialise postgres as primary because creating the "
-				  "database user that the pg_auto_failover monitor "
-				  "uses for health checks failed, see above for details");
-		return false;
+	if (!noPgHba) {
```
DimCitus: Style issue: braces on their own line.

src/bin/pg_autoctl/fsm_transition.c
```diff
@@ -102,14 +103,26 @@ fsm_init_primary(Keeper *keeper)
...
+		if (!primary_create_user(postgres,
+								 PG_AUTOCTL_HEALTH_USERNAME, password,
+								 monitorHostname, authMethod))
+		{
+			log_error("Failed to initialise postgres as primary because creating the "
+					  "database user that the pg_auto_failover monitor "
+					  "uses for health checks failed, see above for details");
+			return false;
+		}
+	} else {
```
DimCitus: Style: braces on their own lines.

src/bin/pg_autoctl/primary_standby.c
```diff
@@ -327,6 +327,33 @@ postgres_add_default_settings(LocalPostgresServer *postgres)
 }
 
 
+/*
+ * primary_create_user creates a user.
+ */
+bool
+primary_create_user(LocalPostgresServer *postgres, char *userName,
+					char *password, char *hostname, char *authMethod)
```
DimCitus: Is it really necessary to add an intermediate function here rather than calling `pgsql_create_user` directly. On a quick reading I don't see much logic added? Maybe it's worth it for the `pgsql_finish` call only, but we could certainly handle that at the caller level?

gmr: Sounds reasonable.

DimCitus:
Done in #169 and released in 1.2, the option is `--skip-pg-hba`.

PR #95, opened 2019-11-04
compile on alpine linux

haavardw:


DimCitus:
Hi @haavardw, thanks for your contribution! Do you have any pointer about why Alpine linux is doing such a change, and ideally a couple pointers to other projects and how they solved their compatibility with this OS? I'm quite surprised here, to be honest...

haavardw:
@DimCitus I guess saying Alpine Linux is to blame is just indirectly true. Alpine linux (https://alpinelinux.org) is all about size and is built upon musl libc. It's musl libc which defines stdin/-err/-out here in /usr/include/stdio.h:
https://git.musl-libc.org/cgit/musl/tree/include/stdio.h#n65
Looking through the commit log of this file, it seems like this has been the definition since the initial implementation (https://git.musl-libc.org/cgit/musl/commit/include/stdio.h?id=0b44a0315b47dd8eced9f3b7f31580cf14bbfc01)

musl libc is also all about size and optimization and reducing "waste". I'm guessing this is an optimization they've done, but that would just be me guessing.

Alpine linux is common when building minimal docker images. My concrete use case is building upon a docker image of timescaledb which again is built upon postgres on alpine linux.

haavardw:
@DimCitus The citus extension builds on Alpine Linux: https://github.com/citusdata/docker/blob/master/Dockerfile-alpine

I cannot see that they've done any particular adjustments to be compatible (jsut by grepping through the git log and skimming the config file in the main citus repository). I'm guessing they've just avoided variable names with this kind of conflicts..

DimCitus:
See #161.

Issue #74, opened 2019-09-24
Unable to create postgres on standby machine


Raves1209:
getting below error when trying to setup secondary instance :

```
[postgres@lxlab1569252301 postgres_log]$ pg_autoctl create postgres --pgdata /postgres_data --monitor postgres://autoctl_node@172.29.142.26:50001/pg_auto_failover
14:10:38 INFO  Using --nodename "lxlab1569252301.devlab.lowes.com", which resolves to IP address "172.29.142.23"
14:10:38 INFO  Continuing from a previous `pg_autoctl create` failed attempt
14:10:38 INFO  PostgreSQL state at registration time was: PGDATA does not exists
14:10:38 INFO  FSM transition from "wait_standby" to "catchingup": The primary is now ready to accept a standby
14:10:38 INFO  The primary node returned by the monitor is lxlab1569252313.devlab.lowes.com:50001
14:10:38 INFO  Initialising PostgreSQL as a hot standby
14:10:38 INFO  Target directory exists: "/postgres_data", stopping PostgreSQL
14:10:38 INFO  pg_ctl: no server running

14:10:38 INFO  pg_ctl stop failed, but PostgreSQL is not running anyway
could not remove file or directory "//backup": Permission denied
14:10:38 ERROR Failed to remove directory "//backup": Permission denied
14:10:38 ERROR Failed initialise standby server, see above for details
14:10:38 ERROR Failed to transition from state "wait_standby" to state "catchingup", see above.
```

i dont know from its checking /backup folder .

successfully created monitor and primary instance.
Version is 11.2 and Os types is linux x86_64.
Followed below link for setup 
https://github.com/citusdata/pg_auto_failover

adiep-cdpq:
@Raves1209 When initializing your slave node for the first time, your pgdata folder should be non-existing. In your case, the slave node shouldn't have /postgres_data when you execute your command.

adiep-cdpq:
However, on the docs, it says the following : 

> 3. Initialize a secondary node from scratch
> 
> This happens when --pgdata (or the environment variable PGDATA) points to a non-existing or empty directory, and when the monitor registration call assigns the state WAIT_STANDBY in step 1.
> 
> In that case, the pg_autoctl create command steps through the initial states of registering a secondary server, which includes preparing the primary server PostgreSQL HBA rules and creating a replication slot.
> 
[...]

Issue #64, opened 2019-09-14
Support mounted pgdata folder


ElfoLiNk:
When executing create postgres on secondary node it fails due to mounted PGDATA folder.

`ERROR Failed to remove directory "/opt/postgressql/pgdbdata": Device or resource busy`

Full log:
```
ubuntu@ip-10-61-5-136:~$ sudo -u postgres -i pg_autoctl create postgres  --pgdata /opt/postgressql/pgdbdata --monitor postgres://autoctl_node@MONITOR_NODE:5432/pg_auto_failover --nodename SECONDARY_NODE
15:13:01 INFO  Registered node SECONDARY_NODE with id 16 in formation "default", group 0.
15:13:01 INFO  Writing keeper init state file at "/var/lib/postgresql/.local/share/pg_autoctl/opt/postgressql/pgdbdata/pg_autoctl.init"
15:13:01 INFO  Successfully registered as "wait_standby" to the monitor.
15:13:01 INFO  FSM transition from "init" to "wait_standby": Start following a primary
15:13:01 INFO  Transition complete: current state is now "wait_standby"
15:13:06 INFO  FSM transition from "wait_standby" to "catchingup": The primary is now ready to accept a standby
15:13:06 INFO  The primary node returned by the monitor is PRIMARY_NODE:5432
15:13:06 INFO  Initialising PostgreSQL as a hot standby
15:13:06 INFO  Target directory exists: "/opt/postgressql/pgdbdata", stopping PostgreSQL
15:13:06 INFO  pg_ctl: no server running

15:13:06 INFO  pg_ctl stop failed, but PostgreSQL is not running anyway
15:13:09 INFO  Running /usr/lib/postgresql/11/bin/pg_basebackup -w -h PRIMARY_NODE -p 5432 --pgdata /opt/postgressql/backup -U pgautofailover_replicator --write-recovery-conf --max-rate 100M --wal-method=stream --slot pgautofailover_standby ...
15:18:27 INFO  pg_basebackup: initiating base backup, waiting for checkpoint to complete
pg_basebackup: checkpoint completed
pg_basebackup: write-ahead log start point: CB/6D000028 on timeline 3
pg_basebackup: starting background WAL receiver
       0/30758969 kB (0%), 0/1 tablespace (.../postgressql/backup/backup_label)
....
30759029/30759029 kB (100%), 0/1 tablespace (...gressql/backup/global/pg_control)
30759029/30759029 kB (100%), 1/1 tablespace
[...]

PR #96, opened 2019-11-05
Fix creating a secondary server in an already existing directory.

DimCitus:
This could have been a one-liner, but raises the question of the hard-coded
choice for the backup directory. The major reason of the target PGDATA
already existing is that it's targeting a mount-point. Then our default
choice of ${PGDATA}/../backup for the pg_basebackup target is not a good
one, because the renaming of the directory will cross file system boundaries
and not be atomic anymore.

This patch then allows to configure the backup directory. In passing, the
default value now includes the nodename, so that it's easier to run tests
environments with multiple instances on the same machine.

See #64 and #74.

### DIFF
--- a/docs/ref/configuration.rst
+++ b/docs/ref/configuration.rst
@@ -116,7 +116,7 @@ An example configuration file looks like the following::
   monitor = postgres://autoctl_node@192.168.1.34:6000/pg_auto_failover
   formation = default
   group = 0
-  nodename = node1.db.local.tld
+  nodename = node1.db
   nodekind = standalone
   
   [postgresql]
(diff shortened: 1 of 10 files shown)


docs/ref/configuration.rst
```diff
@@ -129,6 +129,7 @@ An example configuration file looks like the following::
   [replication]
   slot = pgautofailover_standby
   maximum_backup_rate = 100M
+  backup_directory = /data/backup/node1.db.local.tld
```
mtuncer: can use a more meaningful default here

DimCitus: The default is `${pgdata}/../backup/${nodename}` in the code, and so in the sample configuration file too. Do you mean that the code should use another default value? If yes, which one?

mtuncer: `node1.db.local.tld` looked non-trivial. 

I meant something like `/data/pg_autoctl/backup/nodedata.db` or something like that.

DimCitus: Good catch, changed to `node1.db`.

PR #100, opened 2019-11-07
Improve options and arguments parsing.

### LINKED ISSUE #99: Unable to get --pgdata value in pg_autoctl get config command
I hit this while implementing `pg_autoctl set node candidate-priority xxx` command. When investigated found existing `pg_autoctl config get ...` command also exhibits the same behavior.

I have a monitor and a primary node set up in the same machine.

When I run the command  `pg_autoctl config get timeout.network_partition_timeout --pgdata /Users/mtuncer/dev/ha/test/primary/`
 I got

```
11:01:32 INFO  Managing PostgreSQL installation at "/Users/mtuncer/dev/ha/test/primary/"
pg_autoctl config get: Get the value of a given pg_autoctl configuration variable
usage: pg_autoctl config get [ section.option ]

  --pgdata      path to data directory
```

if I set `PGDATA= /Users/mtuncer/dev/ha/test/primary/ ` and run  the same command without --pgdata I got successful. 
```
11:05:08 INFO  Managing PostgreSQL installation at "/Users/mtuncer/dev/ha/test/primary"
20
```

However, if I run the earlier command while PGDATA is set I still get the same error

```
pg_autoctl config get timeout.network_partition_timeout --pgdata /Users/mtuncer/dev/ha/test/primary/
11:06:13 INFO  Managing PostgreSQL installation at "/Users/mtuncer/dev/ha/test/primary"
pg_autoctl config get: Get the value of a given pg_autoctl configuration variable
usage: pg_autoctl config get [ section.option ]

  --pgdata      path to data directory
```

This is happening on Mac OS 10.14 with Apple Clang 11.0.0

DimCitus:
Our nested command and sub-commands parsing is not compatible with the
default GNU getopt behavior where they will re-order the command line so
that users are free to mix options and arguments as they want.

When we are on the final sub-command though, we could re-enable the default
behavior of GNU getopt_long for more flexibility to the user. That's what we
do with this patch, making it possible to use either form for the same
command as in the following example:

    $ pg_autoctl -q config get --pgdata /tmp/sm/a postgresql.port
    3001
    
    $ pg_autoctl config get postgresql.port --pgdata /tmp/sm/a -q
    3001

Without this patch, only the first case works.

As most of our commands only process options, only two places need to be
adjusted to make that behavior available. The other commands don't need this
at the moment.

Fixed #99.

### DIFF
--- a/src/bin/pg_autoctl/cli_common.c
+++ b/src/bin/pg_autoctl/cli_common.c
@@ -275,6 +275,16 @@ keeper_cli_getopt_pgdata(int argc, char **argv)
 	};
 	optind = 0;
 
+	/*
+	 * The only command lines that are using keeper_cli_getopt_pgdata are
+	 * terminal ones: they don't accept subcommands. In that case our option
+	 * parsing can happen in any order and we don't need getopt_long to behave
+	 * in a POSIXLY_CORRECT way.
+	 *
+	 * The unsetenv() call allows getopt_long() to reorder arguments for us.
+	 */
+	unsetenv("POSIXLY_CORRECT");
+
 	while ((c = getopt_long(argc, argv, "D:",
 							long_options, &option_index)) != -1)
 	{
--- a/src/bin/pg_autoctl/cli_common.c
+++ b/src/bin/pg_autoctl/cli_common.c
@@ -12,6 +12,7 @@
 #include <getopt.h>
 
 #include "cli_common.h"
+#include "cli_root.h"
 #include "commandline.h"
 #include "ipaddr.h"
 #include "keeper.h"
@@ -60,6 +61,7 @@ cli_create_node_getopts(int argc, char **argv,
 {
 	KeeperConfig LocalOptionConfig = { 0 };
 	int c, option_index, errors = 0;
+	int verboseCount = 0;
 
 	/* force some non-zero default values */
 	LocalOptionConfig.groupId = -1;
@@ -217,6 +219,39 @@ cli_create_node_getopts(int argc, char **argv,
 				break;
 			}
 
+			case 'V':
+			{
+				/* keeper_cli_print_version prints version and exits. */
+				keeper_cli_print_version(argc, argv);
+				break;
+			}
+
+			case 'v':
+			{
+				++verboseCount;
+				switch (verboseCount)
+				{
+					case 1:
+						log_set_level(LOG_INFO);
+						break;
+
+					case 2:
+						log_set_level(LOG_DEBUG);
+						break;
+
+					default:
+						log_set_level(LOG_TRACE);
+						break;
+				}
+				break;
+			}
+
+			case 'q':
+			{
+				log_set_level(LOG_ERROR);
+				break;
+			}
+
 			default:
 			{
 				/* getopt_long already wrote an error message */
@@ -266,11 +301,14 @@ int
 keeper_cli_getopt_pgdata(int argc, char **argv)
 {
 	KeeperConfig options = { 0 };
-	int c, option_index = 0;
+	int c, option_index = 0, errors = 0;
+	int verboseCount = 0;
 
 	static struct option long_options[] = {
 		{ "pgdata", required_argument, NULL, 'D' },
-		{ "help", no_argument, NULL, 'h' },
+		{ "version", no_argument, NULL, 'V' },
+		{ "verbose", no_argument, NULL, 'v' },
+		{ "quiet", no_argument, NULL, 'q' },
 		{ NULL, 0, NULL, 0 }
 	};
 	optind = 0;
@@ -285,7 +323,7 @@ keeper_cli_getopt_pgdata(int argc, char **argv)
 	 */
 	unsetenv("POSIXLY_CORRECT");
 
-	while ((c = getopt_long(argc, argv, "D:",
+	while ((c = getopt_long(argc, argv, "D:Vvq",
 							long_options, &option_index)) != -1)
 	{
 		switch (c)
@@ -296,7 +334,53 @@ keeper_cli_getopt_pgdata(int argc, char **argv)
 				log_trace("--pgdata %s", options.pgSetup.pgdata);
 				break;
 			}
+
+			case 'V':
+			{
+				/* keeper_cli_print_version prints version and exits. */
+				keeper_cli_print_version(argc, argv);
+				break;
+			}
+
+			case 'v':
+			{
+				++verboseCount;
+				switch (verboseCount)
+				{
+					case 1:
+						log_set_level(LOG_INFO);
+						break;
+
+					case 2:
+						log_set_level(LOG_DEBUG);
+						break;
+
+					default:
+						log_set_level(LOG_TRACE);
+						break;
+				}
+				break;
+			}
+
+			case 'q':
+			{
+				log_set_level(LOG_ERROR);
+				break;
+			}
+
+			default:
+			{
+				/* getopt_long already wrote an error message */
+				errors++;
+				break;
+			}
 		}
+	}
+
+	if (errors > 0)
+	{
+		commandline_help(stderr);
+		exit(EXIT_CODE_BAD_ARGS);
 	}
 
 	if (IS_EMPTY_STRING_BUFFER(options.pgSetup.pgdata))
@@ -316,7 +400,8 @@ keeper_cli_getopt_pgdata(int argc, char **argv)
 	log_info("Managing PostgreSQL installation at \"%s\"",
 			 options.pgSetup.pgdata);
 
-	if (!keeper_config_set_pathnames_from_pgdata(&options.pathnames, options.pgSetup.pgdata))
+	if (!keeper_config_set_pathnames_from_pgdata(&options.pathnames,
+												 options.pgSetup.pgdata))
 	{
 		/* errors have already been logged */
 		exit(EXIT_CODE_BAD_ARGS);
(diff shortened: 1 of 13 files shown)


src/bin/pg_autoctl/cli_create_drop_node.c
```diff
@@ -266,7 +269,9 @@ cli_create_monitor_getopts(int argc, char **argv)
 		{ "nodename", required_argument, NULL, 'n' },
 		{ "listen", required_argument, NULL, 'l' },
 		{ "auth", required_argument, NULL, 'A' },
-		{ "help", no_argument, NULL, 0 },
```
mtuncer: help was not explicitly handled before. now we are removing support for --help option. Is this intentional ?

src/bin/pg_autoctl/cli_common.c
```diff
@@ -266,16 +301,29 @@ int
 keeper_cli_getopt_pgdata(int argc, char **argv)
 {
 	KeeperConfig options = { 0 };
-	int c, option_index = 0;
+	int c, option_index = 0, errors = 0;
+	int verboseCount = 0;
 
 	static struct option long_options[] = {
 		{ "pgdata", required_argument, NULL, 'D' },
-		{ "help", no_argument, NULL, 'h' },
```
mtuncer: is removing --help intentional ?

mtuncer: `pg_autoctl config get --help` works but `pg_autoctl config get replication.slot --pgdata primary --help` does not

mtuncer: similary `pg_autoctl show state --help` but `pg_autoctl show state --formation default --help` fails with unrecognized option error.

src/bin/pg_autoctl/cli_common.c
```diff
@@ -266,16 +301,29 @@ int
 keeper_cli_getopt_pgdata(int argc, char **argv)
 {
 	KeeperConfig options = { 0 };
-	int c, option_index = 0;
+	int c, option_index = 0, errors = 0;
+	int verboseCount = 0;
 
 	static struct option long_options[] = {
```
mtuncer: root_option() function and this function is now very similar.

src/bin/pg_autoctl/cli_common.c
```diff
@@ -266,16 +301,29 @@ int
 keeper_cli_getopt_pgdata(int argc, char **argv)
 {
 	KeeperConfig options = { 0 };
-	int c, option_index = 0;
+	int c, option_index = 0, errors = 0;
```
mtuncer: not fun of having multiple declarations in the same line. guess it is okay.

src/bin/pg_autoctl/cli_common.c
```diff
@@ -217,6 +219,39 @@ cli_create_node_getopts(int argc, char **argv,
 				break;
 			}
 
+			case 'V':
```
mtuncer: commented list of options in the function header should be updated accordingly for this function and the other modified functions.

src/bin/pg_autoctl/cli_formation.c
```diff
@@ -61,11 +61,14 @@ keeper_cli_formation_getopts(int argc, char **argv)
 {
 	FormationConfig options = { 0 };
 	int c = 0, option_index = 0, errors = 0;
+	int verboseCount = 0;
 
 	static struct option long_options[] = {
 		{ "pgdata", required_argument, NULL, 'D' },
 		{ "formation", required_argument, NULL, 'f' },
-		{ "help", no_argument, NULL, 'h' },
+		{ "version", no_argument, NULL, 'V' },
```
mtuncer: we are having very similar looking functions.  We would need to consolidate them at some time

src/bin/pg_autoctl/cli_enable_disable.c
```diff
@@ -126,6 +129,39 @@ cli_secondary_getopts(int argc, char **argv)
 				break;
 			}
 
+			case 'V':
```
mtuncer: shall we consider `keeper_cli_formation_create_getopts` too

src/bin/pg_autoctl/cli_enable_disable.c
```diff
@@ -87,27 +87,30 @@ CommandLine disable_commands =
...
 cli_secondary_getopts(int argc, char **argv)
 {
 	KeeperConfig options = { 0 };
 	int c, option_index, errors = 0;
+	int verboseCount = 0;
 
 	static struct option long_options[] = {
-			{ "pgdata", required_argument, NULL, 'D' },
-			{ "formation", required_argument, NULL, 'f' },
-			{ "help", no_argument, NULL, 0 },
```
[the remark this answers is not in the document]
DimCitus: There was a problem in `src/bin/lib/commandline.c` that prevented `--help` to work everywhere, and so I removed support for it in a first pass. Thinking about it more, I could find the bug and fix it, and I have now re-introduced `--help` everywhere. Thanks!

src/bin/pg_autoctl/cli_common.c
DimCitus: The difference is `--pgdata`. Sharing code in C for handling `switch` cases is... well... I could do some advanced macrology but I'd rather not... what do you think?

src/bin/pg_autoctl/cli_common.c
DimCitus: Good point, done now!

src/bin/pg_autoctl/cli_enable_disable.c
DimCitus: Done now!

DimCitus:
I think it's all good now, please have another pass. Thanks!

src/bin/pg_autoctl/cli_common.c
mtuncer: This PR is not a good place for it but I have following thought.
1 - we provide long and short options for getopt_long() call. Both parameters need to be compatible and need to be changed together. Perhaps we could have a way to derive short options from long ones since we already specify corresponding short name when we are declaring long ones.

2 - Considering adding a group mask  to accompany `struct option` such that when we are declaring an option line we can also specify the group it belongs to. This way when parsing we also check the group mask. That could help handling similar parser functions. But this idea is pretty immature at this time. More thinking is needed.

mtuncer: when I say group mask I think of a bitmask, so that the same option could be part of multiple groups.
```c
#define GETOPT_PGDATA 0x00001
#define GETOPT_CREATE_NODE 0x0002
```
so that when we are declaring an option we can say
```c
GETOPT_PGDATA | GETOPT_CREATE_NODE
```
Prior to running `getopt_long()` we can run a function to determine the options we are looking for 
with a group number.

mtuncer: prefer functions over macros. Like the name `macrology`

src/bin/pg_autoctl/cli_create_drop_node.c
```diff
@@ -203,13 +203,16 @@ cli_create_postgres_getopts(int argc, char **argv)
...
+		{ "verbose", no_argument, NULL, 'v' },
+		{ "quiet", no_argument, NULL, 'q' },
+ 		{ "help", no_argument, NULL, 'h' },
 		{ NULL, 0, NULL, 0 }
 	};
 
 	int optind =
 		cli_create_node_getopts(argc, argv,
-								long_options, "C:D:h:p:l:U:A:d:n:f:m:R",
+								long_options, "C:D:h:p:l:U:A:d:n:f:m:RVvqh",
```
mtuncer: missing 'H'

src/bin/pg_autoctl/cli_do_root.c
```diff
@@ -244,13 +244,26 @@ keeper_cli_keeper_setup_getopts(int argc, char **argv)
...
+	 * in a POSIXLY_CORRECT way.
+	 *
+	 * The unsetenv() call allows getopt_long() to reorder arguments for us.
+	 */
+	unsetenv("POSIXLY_CORRECT");
+
 	int optind =
 		cli_create_node_getopts(argc, argv,
-								long_options, "C:D:h:p:l:y:U:A:d:n:f:g:m:R",
+								long_options, "C:D:h:p:l:y:U:A:d:n:f:g:m:RVvqh",
```
mtuncer: `H`

mtuncer: we will run out of letters soon as we add new parameters. Perhaps we should consider dropping short names for uncommon uses, or better all short options.

mtuncer: I have already lost short names and need to come up with new ones  for replication-quorum and candidate priority  options in my PR that is going to be merged after this.

src/bin/pg_autoctl/cli_common.c
DimCitus: It might be preferable to switch to using https://www.gnu.org/software/libc/manual/html_node/Argp.html#Argp rather than continue using the getopt API in that case, but I didn't study that.

src/bin/pg_autoctl/cli_create_drop_node.c
DimCitus: Fixed, good catch, thanks!

src/bin/pg_autoctl/cli_do_root.c
DimCitus: Fixed.

PR #103, opened 2019-11-12
Retry connecting to the monitor if it's not ready yet.

DimCitus:
In some initialisation scenarios, the monitor is started at the same time as
any other node, so that a keeper might be trying to connect to the monitor
too early. Implement a retry loop based on PQping in that case.


src/bin/pg_autoctl/pgsql.c
```diff
@@ -165,6 +205,116 @@ pgsql_open_connection(PGSQL *pgsql)
...
+	uint64_t startTime = time(NULL);
+
+	log_warn("Failed to connect to \"%s\", retrying until "
+			 "the server is ready", pgsql->connectionString);
+
+	while (retry)
+	{
+		uint64_t now = time(NULL);
+
+		if ((now - startTime) >= POSTGRES_PING_RETRY_TIMEOUT)
```
mtuncer: does this needs to be configurable. 15 min looks little long to me

src/bin/pg_autoctl/pgsql.c
```diff
@@ -151,9 +156,44 @@ pgsql_open_connection(PGSQL *pgsql)
...
+			case PGSQL_CONN_LOCAL:
+			{
+				log_error("Connection to database failed: %s",
+						  PQerrorMessage(connection));
+				pgsql_finish(pgsql);
+				return NULL;
+			}
+
+			case PGSQL_CONN_MONITOR:
+			case PGSQL_CONN_COORDINATOR:
```
mtuncer: Is this intentional (waiting until coordinator is available) ?

src/bin/pg_autoctl/pgsql.c
```diff
@@ -165,6 +205,116 @@ pgsql_open_connection(PGSQL *pgsql)
...
+ * when it could connect, false otherwise.
+ */
+static PGconn *
+pgsql_retry_open_connection(PGSQL *pgsql)
+{
+	PGconn *connection = NULL;
+	int attempts = 0;
+	bool retry = true;
+	bool connectionOk = false;
+	PGPing rv;
```
mtuncer: unused parameter

src/bin/pg_autoctl/pgsql.c
DimCitus: The expectation is that we face a transient error. We don't want to fail too soon, and I agree that 15 mins is a long time. For interactive use-cases, it's still possible to interrupt (^C) the loop easily, though. Also, implementing configuration parameters down in `pgsql.c` would require a lot of changes, and I don't think it's important enough to be able to change the default here.

Another (cheap) way to do it would be via an environment variable, maybe. I'm just not sure it's worth it.

src/bin/pg_autoctl/pgsql.c
DimCitus: Yes it is. The coordinator could be the last node provisioned and we want to handle that situation just fine.

src/bin/pg_autoctl/pgsql.c
DimCitus: Fixed, thanks!

DimCitus:
> there are some build failures on travis, also saw an unused variable warning

All solved now, thanks!

src/bin/pg_autoctl/pgsql.c
mtuncer: I would rather have this way than having a magic env variable.

src/bin/pg_autoctl/pgsql.c
```diff
@@ -165,6 +205,115 @@ pgsql_open_connection(PGSQL *pgsql)
...
+						  attempts);
+				retry = true;
+				break;
+			}
+
+			case PQPING_NO_ATTEMPT:
+			{
+				log_error("Failed to ping server \"%s\" because of "
+						  "client-side problems (no attempt were made)",
+						  pgsql->connectionString);
```
admilazz: Would pgsql->connectionstring contain passwords? It might be better to strip passwords out of the logs.

src/bin/pg_autoctl/pgsql.c
```diff
@@ -165,6 +205,115 @@ pgsql_open_connection(PGSQL *pgsql)
...
+		++attempts;
+		switch (PQping(pgsql->connectionString))
+		{
+			case PQPING_OK:
+			{
+				log_debug("PQping OK after %d attempts", attempts);
+				retry = false;
+
+				/*
+				 * Ping is now ok, and connection is still NULL because the
```
admilazz: I'm not familiar with pgping, but I suppose it's possible that after connecting the database is still not ready yet. Sometimes Postgres will let you connect but respond to queries with an error like "the database engine is starting up" or "the database engine is shutting down" or (if the query makes a change) "the database is in a read-only mode" (if it is performing recovery as part of startup).

In our own code we wrap not only connecting but also running a query in the retry loop and consider those transient errors from Postgres to be reasons to retry.

PR #104, opened 2019-11-16
Implement pg_autoctl show files.

DimCitus:
Our usage of the XDG standard for configuration and data files is complex to
understand and debug for most users. Having a command to help navigate and
diagnose the current situation is then useful.

TODO: add documentation about the new command.


src/bin/pg_autoctl/cli_show.c
```diff
@@ -528,3 +563,354 @@ cli_show_monitor_uri(int argc, char **argv)
...
+			{
+				if (!fprint_file_contents(config.pathnames.config))
+				{
+					/* errors have already been logged */
+					exit(EXIT_CODE_BAD_CONFIG);
+				}
+			}
+			else
+			{
+				fprintf(stdout, "\"%s\"\n", config.pathnames.config);
```
marcocitus: the extra quotes seem tedious for parsing the output.

DimCitus: I think it's needed when file names contain spaces and sometimes other characters too, so I'd rather have extra quoting always, just in case? Alternative is to actually quote and escape only when needed... is there a good way to do the escaping in a portable way? I figured always double-quoting is cheap and works almost always...

DimCitus:
The strange error is now fixed.

src/bin/pg_autoctl/cli_show.c
marcocitus: Why is it needed? If I wish to parse the output, can't I just read everything up to the end of line?

DimCitus: Good point, done that way now, with some more changes. Thinking more about it, now that we have JSON support internally and that we emit more of it, I think it would be a good format for the `--all` output. And `jq` makes it very easy to then use that format if you want to script around it.

```
$ pg_autoctl show file --pgdata /tmp/sm/x
{
    "config": "\/Users\/dim\/.config\/pg_autoctl\/tmp\/sm\/x\/pg_autoctl.cfg",
    "state": "\/Users\/dim\/.local\/share\/pg_autoctl\/tmp\/sm\/x\/pg_autoctl.state",
    "init": "\/Users\/dim\/.local\/share\/pg_autoctl\/tmp\/sm\/x\/pg_autoctl.init",
    "pid": "\/tmp\/pg_autoctl\/tmp\/sm\/x\/pg_autoctl.pid"
}

$ pg_autoctl show file --pgdata /tmp/sm/m
{
    "config": "\/Users\/dim\/.config\/pg_autoctl\/tmp\/sm\/m\/pg_autoctl.cfg"
}

$ pg_autoctl show file --pgdata /tmp/sm/m | jq .
{
  "config": "/Users/dim/.config/pg_autoctl/tmp/sm/m/pg_autoctl.cfg"
}

$ pg_autoctl show file --pgdata /tmp/sm/m | jq .config
"/Users/dim/.config/pg_autoctl/tmp/sm/m/pg_autoctl.cfg"

$ pg_autoctl show file --pgdata /tmp/sm/x | jq .config
"/Users/dim/.config/pg_autoctl/tmp/sm/x/pg_autoctl.cfg"

$ pg_autoctl show file --pgdata /tmp/sm/x --config
/Users/dim/.config/pg_autoctl/tmp/sm/x/pg_autoctl.cfg
```

PR #107, opened 2019-11-19
Move all the INIT to SINGLE initialisation code in the FSM transition.

DimCitus:
That way we will be able to re-use that code more easily from the HTTP API
when running without monitor.


DimCitus:
Good comment, I created a new issue to track that: https://github.com/citusdata/pg_auto_failover/issues/111. I'm not sure what the behaviour should be exactly, but it seems strange to accept the non-running Postgres instance in one case and refuse it in another case indeed.

PR #117, opened 2019-11-29
Work on improving tests reports in case of failures.

DimCitus:
Also implement --stop for the nose test driver, because the way our tests
suites are written it does not make sense to continue after a failure.

This PR introduces some failures on its own, and that still needs to be
figured out, unfortunately.

### DIFF
--- a/tests/pgautofailover_utils.py
+++ b/tests/pgautofailover_utils.py
@@ -55,14 +55,16 @@ class Cluster:
     # TODO group should auto sense for normal operations and passed to the
     # create cli as an argument when explicitly set by the test
     def create_datanode(self, datadir, port=5432, group=0,
-                        listen_flag=False, role=Role.Postgres, formation=None, authMethod=None):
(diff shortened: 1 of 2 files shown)


tests/pgautofailover_utils.py
```diff
@@ -92,28 +94,51 @@ def __init__(self, datadir, vnode, port, username, authMethod, database, role):
...
-                          '--pgdata', self.datadir]
+        run_command = [shutil.which('pg_autoctl'), 'run', '-vvv',
+                       '--pgdata', self.datadir]
         self.pg_autoctl_run_proc = self.vnode.run(run_command)
 
+        print("%s [%d]" % (" ".join(run_command),
+                           self.pg_autoctl_run_proc.pid))
+
+        # check that the process is still running after 1s
+        time.sleep(1)
```
mtuncer: that means we spend at least 1 sec at each run call

DimCitus:
This has been fixed in a better way in #139

PR #118, opened 2019-11-29
Avoid using stdin/stdout/stderr as variable names

eradman:
These are macros in C89/C99

On OpenBSD 6.6 this causes the following build error

```
In file included from pgctl.c:28:
/home/eradman/git/pg_auto_failover/src/bin/pg_autoctl/../lib/subcommands.c/runprogram.h:44:8: error: expected member name or
      ';' after declaration specifiers
        char *stdout;
        ~~~~  ^
```

```
$ cc --version
OpenBSD clang version 8.0.1 (tags/RELEASE_801/final) (based on LLVM 8.0.1)
Target: amd64-unknown-openbsd6.6
Thread model: posix
InstalledDir: /usr/bin
```

### DIFF
--- a/src/bin/lib/subcommands.c/runprogram.h
+++ b/src/bin/lib/subcommands.c/runprogram.h
@@ -41,8 +41,8 @@ typedef struct
 	int error;					/* save errno when something's gone wrong */
 	int returnCode;
 
-	char *stdout;
-	char *stderr;
+	char *stdOut;
+	char *stdErr;
 } Program;
 
 Program run_program(const char *program, ...);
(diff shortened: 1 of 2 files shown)


DimCitus:
See also #148

DimCitus:
Hi @eradman ; I have updated your PR to fix merges with the current master branch. See #161. Can you share a Dockerfile to make it easy for us to check that we indeed have fixed the compiling on OpenBSD (or Alpine Linux)?

eradman:
Tested this PR again on OpenBSD and it seems good.

I'm not as familiar with Docker or Alpine Linux but I'll see if a Dockerfile is easy to assemble

DimCitus:
See #161 which takes care of current master's branch.

Issue #86, opened 2019-10-10
monitor does not restart when running with pg_autoctl run


thanodnl:
If the postgres of the monitor gets killed or crashes it is not correctly restarted by the 'keeper'.
This can be reproduced in 2 sessions:

Session 1
```bash
pg_autoctl run --pgdata /path/of/monitor
```

Session 2
```bash
kill <pid of monitor postgres>
```
(or a killall postgres for that matter)

`pg_autoctl` will be in an error loop but not restarting the postgres process. This is different behaviour than we have when running a postgres node.

Bouke:
The loop prints the following:
```
14:51:25 WARN  Lost connection.
14:51:25 WARN  Re-establishing connection. We might miss notifications.
14:51:25 ERROR Connection to database failed: could not connect to server: No such file or directory
	Is the server running locally and accepting
	connections on Unix domain socket "/var/run/postgresql/.s.PGSQL.5432"?
```

PR #121, opened 2019-12-03
Ensure the monitor's Postgres is running as part of `pg_autoctl run`

mtuncer:
Keeper now checks the state of monitor and restart it if it is found to be not active.

Fixes :  #86

### DIFF
--- a/src/bin/pg_autoctl/cli_service.c
+++ b/src/bin/pg_autoctl/cli_service.c
@@ -191,14 +191,12 @@ cli_monitor_run(int argc, char **argv)
 	MonitorConfig mconfig = { 0 };
 	Monitor monitor = { 0 };
 	MonitorExtensionVersion version = { 0 };
-	char postgresUri[MAXCONNINFO];
 	bool missingPgdataIsOk = false;
 	bool pgIsNotRunningIsOk = true;
 	char connInfo[MAXCONNINFO];
 	char *channels[] = { "log", "state", NULL };
 
-	if (!monitor_config_init_from_pgsetup(&monitor,
-										  &mconfig,
+	if (!monitor_config_init_from_pgsetup(&mconfig,
 										  &kconfig.pgSetup,
 										  missingPgdataIsOk,
 										  pgIsNotRunningIsOk))
@@ -210,37 +208,7 @@ cli_monitor_run(int argc, char **argv)
 	pg_setup_get_local_connection_string(&(mconfig.pgSetup), connInfo);
 	monitor_init(&monitor, connInfo);
 
-	if (!pg_is_running(mconfig.pgSetup.pg_ctl, mconfig.pgSetup.pgdata))
-	{
-		log_info("Postgres is not running, starting postgres");
-
-		if (!pg_ctl_start(mconfig.pgSetup.pg_ctl,
-						  mconfig.pgSetup.pgdata,
-						  mconfig.pgSetup.pgport,
-						  mconfig.pgSetup.listen_addresses))
-		{
-			log_error("Failed to start PostgreSQL, see above for details");
-			exit(EXIT_CODE_PGCTL);
-		}
-	}
-
-	log_info("PostgreSQL is running in \"%s\" on port %d",
-			 mconfig.pgSetup.pgdata, mconfig.pgSetup.pgport);
-
-	/* Check version compatibility */
-	if (!monitor_ensure_extension_version(&monitor, &version))
-	{
-		/* errors have already been logged */
-		exit(EXIT_CODE_MONITOR);
-	}
-
-	/* Now get the the Monitor URI to display it to the user, and move along */
-	if (monitor_config_get_postgres_uri(&mconfig, postgresUri, MAXCONNINFO))
-	{
-		log_info("pg_auto_failover monitor is ready at %s", postgresUri);
-	}
-
-	(void) monitor_listen_loop(&monitor);
+	(void) monitor_service_run(&monitor, &mconfig);
 }
 
 
--- a/src/bin/pg_autoctl/monitor.c
+++ b/src/bin/pg_autoctl/monitor.c
@@ -19,6 +19,7 @@
 #include "defaults.h"
 #include "log.h"
 #include "monitor.h"
+#include "monitor_config.h"
 #include "parsing.h"
 #include "pgsql.h"
 #include "catalog/pg_type.h"
@@ -1865,13 +1866,73 @@ prepare_connection_to_current_system_user(Monitor *source, Monitor *target)
 
 
 /*
- * monitor_listen_loop loops over a LISTEN command that is notified at every
+ * ensure_monitor_pg_running checks if monitor is running, attempts to restart if it is not.
+ * The function verifies if the extension version is the same as expected.  It returns true
+ * if the monitor is up and running.
+ */
+bool
+ensure_monitor_pg_running(Monitor *monitor, struct MonitorConfig *mconfig)
+{
+	MonitorExtensionVersion version = { 0 };
+	char postgresUri[MAXCONNINFO];
+
+	if (!pg_is_running(mconfig->pgSetup.pg_ctl, mconfig->pgSetup.pgdata))
+	{
+		log_info("Postgres is not running, starting postgres");
+		pgsql_finish(&(monitor->pgsql));
+
+		if (!pg_ctl_start(mconfig->pgSetup.pg_ctl,
+						  mconfig->pgSetup.pgdata,
+						  mconfig->pgSetup.pgport,
+						  mconfig->pgSetup.listen_addresses))
+		{
+			log_error("Failed to start PostgreSQL, see above for details");
+			return false;
+		}
+
+		/*
+		 * Check version compatibility.
+		 *
+		 * The function terminates any existing connection during clenaup
+		 * therefore it is not called when PG is found to be running to not
+		 * to intervene pgsql_listen call.
+		 */
+		if (!monitor_ensure_extension_version(monitor, &version))
+		{
+			/* errors have already been logged */
+			return false;
+		}
+
+	}
+
+	return true;
+}
+
+
+/*
+ * monitor_service_run watches over monitor process, restarts if it is necessary,
+ * also loops over a LISTEN command that is notified at every
  * change of state on the monitor, and prints the change on stdout.
  */
 bool
-monitor_listen_loop(Monitor *monitor)
+monitor_service_run(Monitor *monitor,  struct MonitorConfig *mconfig)
 {
 	char *channels[] = { "log", "state", NULL };
+	char postgresUri[MAXCONNINFO];
+
+	/* We exit the loop if we can't get monitor to be running during the start */
+	if (!ensure_monitor_pg_running(monitor, mconfig))
+	{
+		/* errors were already logged */
+		log_warn("Failed to ensure PostgreSQL is running, exiting the service");
+		return false;
+	}
+
+	/* Now get the the Monitor URI to display it to the user, and move along */
+	if (monitor_config_get_postgres_uri(mconfig, postgresUri, MAXCONNINFO))
+	{
+		log_info("pg_auto_failover monitor is ready at %s", postgresUri);
+	}
 
 	log_info("Contacting the monitor to LISTEN to its events.");
 	pgsql_listen(&(monitor->pgsql), channels);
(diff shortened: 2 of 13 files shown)


src/bin/pg_autoctl/cli_service.c
```diff
@@ -30,6 +30,7 @@ static int stop_signal = SIGTERM;
 static void cli_service_run(int argc, char **argv);
 static void cli_keeper_run(int argc, char **argv);
 static void cli_monitor_run(int argc, char **argv);
+static void ensure_monitor_running(Monitor *monitor, MonitorConfig *mconfig);
```
DimCitus: I think `ensure_monitor_pg_is_running` would be a better name (use of verb, specialise to the Postgres instance of the monitor). And then, why are we not using `ensure_local_postgres_is_running`?

src/bin/pg_autoctl/cli_service.c
```diff
@@ -224,6 +195,8 @@ cli_monitor_run(int argc, char **argv)
 			log_warn("Re-establishing connection. We might miss notifications.");
 			pgsql_finish(&(monitor.pgsql));
 
+			ensure_monitor_running(&monitor, &mconfig);
```
DimCitus: Maybe the function should return a boolean and we should implement error handling when we fail to start/restart Postgres. Either quit with a non-zero code so that systemd or another supervisor can handle the situation, or implement a restart strategy, or maybe both.

src/bin/pg_autoctl/cli_service.c
```diff
@@ -238,6 +211,50 @@ cli_monitor_run(int argc, char **argv)
...
+	{
+		log_info("Postgres is not running, starting postgres");
+
+		if (!pg_ctl_start(mconfig->pgSetup.pg_ctl,
+						  mconfig->pgSetup.pgdata,
+						  mconfig->pgSetup.pgport,
+						  mconfig->pgSetup.listen_addresses))
+		{
+			log_error("Failed to start PostgreSQL, see above for details");
+			exit(EXIT_CODE_PGCTL);
```
DimCitus: I think a lower-level function should (almost) never exit itself, but return false so that the caller can decide the best way forward, which might still be an exit.

src/bin/pg_autoctl/cli_service.c
```diff
@@ -238,6 +211,50 @@ cli_monitor_run(int argc, char **argv)
...
+
+	log_info("PostgreSQL is running in \"%s\" on port %d",
+			 mconfig->pgSetup.pgdata, mconfig->pgSetup.pgport);
+
+	/* Check version compatibility */
+	if (!monitor_ensure_extension_version(monitor, &version))
+	{
+		/* errors have already been logged */
+		exit(EXIT_CODE_MONITOR);
+	}
```
DimCitus: I think we should only do this when we restart Postgres, no need to do that check at every loop, right? Also, again, I don't think we should exit in a lower level function.

src/bin/pg_autoctl/cli_service.c
```diff
@@ -238,6 +211,50 @@ cli_monitor_run(int argc, char **argv)
...
+						  mconfig->pgSetup.pgport,
+						  mconfig->pgSetup.listen_addresses))
+		{
+			log_error("Failed to start PostgreSQL, see above for details");
+			exit(EXIT_CODE_PGCTL);
+		}
+	}
+
+	log_info("PostgreSQL is running in \"%s\" on port %d",
+			 mconfig->pgSetup.pgdata, mconfig->pgSetup.pgport);
```
DimCitus: So this INFO line is getting printed at every loop iteration on the monitor output then? That would be too much chatter, let's only print that when we restart Postgres. It might be that reusing `ensure_local_postgres_is_running` would give that for us for free, though, please have a look if we can re-use that on the monitor side of things.

src/bin/pg_autoctl/cli_service.c
```diff
@@ -238,6 +211,50 @@ cli_monitor_run(int argc, char **argv)
...
+	{
+		/* errors have already been logged */
+		exit(EXIT_CODE_MONITOR);
+	}
+
+	/* Now get the the Monitor URI to display it to the user, and move along */
+	if (monitor_config_get_postgres_uri(mconfig, postgresUri, MAXCONNINFO))
+	{
+		log_info("pg_auto_failover monitor is ready at %s", postgresUri);
+	}
```
DimCitus: Do we want to print that at every loop?

src/bin/pg_autoctl/cli_service.c
```diff
@@ -210,27 +210,10 @@ cli_monitor_run(int argc, char **argv)
...
-			exit(EXIT_CODE_PGCTL);
-		}
-	}
-
-	log_info("PostgreSQL is running in \"%s\" on port %d",
-			 mconfig.pgSetup.pgdata, mconfig.pgSetup.pgport);
-
-	/* Check version compatibility */
-	if (!monitor_ensure_extension_version(&monitor, &version))
+	if (!ensure_monitor_pg_running(&monitor, &mconfig))
```
DimCitus: I think this call should go in `monitor_service_run` itself. It would be easier to follow the logic, because that's where we connect to the local Postgres too.

src/bin/pg_autoctl/monitor.c
```diff
@@ -1886,6 +1928,13 @@ monitor_listen_loop(Monitor *monitor)
 			log_warn("Re-establishing connection. We might miss notifications.");
 			pgsql_finish(&(monitor->pgsql));
 
+			if (!ensure_monitor_pg_running(monitor, mconfig))
+			{
+				log_warn("unable to start monitor, will re-attempt in %d seconds", PG_AUTOCTL_MONITOR_SLEEP_TIME);
```
DimCitus: This warning message is confusing, we're not starting the monitor but Postgres. Our users are already confused about what is `pg_autoctl` and where Postgres is running and what's the role of that thing we call a “monitor”, so I think we should be careful here.

What about the following message: "Failed to ensure PostgreSQL is running, retrying in %d seconds".

src/bin/pg_autoctl/monitor.c
```diff
@@ -1886,6 +1928,13 @@ monitor_listen_loop(Monitor *monitor)
 			log_warn("Re-establishing connection. We might miss notifications.");
 			pgsql_finish(&(monitor->pgsql));
 
+			if (!ensure_monitor_pg_running(monitor, mconfig))
```
DimCitus: This is very weird. Ensure Postgres is running should be done before we try to get notifications, and each time we want to connect to LISTEN to them, just before we connect. 

Maybe having a new function `monitor_listen` that ensure Postgres is running and then calls `pgsql_listen` would be a better design. We would need to make it so that we can call that function first in each loop iteration, so it needs to check if we are already connected and listening ; maybe with a new boolean to track that in the Monitor structure?

src/bin/pg_autoctl/monitor.h
```diff
@@ -14,13 +14,28 @@
...
+	char nodename[_POSIX_HOST_NAME_MAX];
+
+	/* PostgreSQL setup */
+	char role[NAMEDATALEN];
+
+	/* PostgreSQL setup */
+	PostgresSetup pgSetup;
+} MonitorConfig;
+
+
```
DimCitus: Why moving the structure definition to `monitor.h` rather than keeping it in `monitor_config.h` and including `monitor_config.h` in `monitor.c` as you did already? That's what we do for the `keeper...` files and I think we should have the same pattern in both the keeper and the monitor code.

src/bin/pg_autoctl/monitor_config.h
```diff
@@ -17,21 +17,6 @@
...
-	/* pg_autoctl setup */
-	char nodename[_POSIX_HOST_NAME_MAX];
-
-	/* PostgreSQL setup */
-	char role[NAMEDATALEN];
-
-	/* PostgreSQL setup */
-	PostgresSetup pgSetup;
-} MonitorConfig;
-
```
DimCitus: I don't think we want to do that?

tests/test_basic_operation.py
```diff
@@ -102,25 +103,30 @@ def test_014_add_new_secondary():
...
-   assert node3_replication_slots == [(0,)]
-                
+    monitor.failover()
+    assert node2.wait_until_state(target_state="primary")
+    assert node3.wait_until_state(target_state="secondary")
+    node2_replication_slots = node2.run_sql_query("select count(*) from pg_replication_slots");
+    assert node2_replication_slots == [(1,)]
+    node3_replication_slots = node3.run_sql_query("select count(*) from pg_replication_slots");
+    assert node3_replication_slots == [(0,)]
+
```
DimCitus: That's a lot of review noise. I agree that we need to take care of this indentation issue at some point, but I'm leaning towards making it its own PR if that's possible. What do you think?

tests/test_basic_operation.py
```diff
@@ -102,25 +103,30 @@ def test_014_add_new_secondary():
...
 def test_016_drop_primary():
-   node2.drop()
-   assert not node2.pg_is_running()
-   assert node3.wait_until_state(target_state="single")
+    node2.drop()
+    assert not node2.pg_is_running()
+    assert node3.wait_until_state(target_state="single")
+
+def test_017_stop_monitor():
```
DimCitus: Mmm. There's a confusion around what process implements the monitor in our code and maybe even docs, and then users are confused to. I propose we name this test `test_017_stop_postgres_monitor` to help us get to a better consistent terminology here.

src/bin/pg_autoctl/cli_service.c
mtuncer: `ensure_local_postgres_is_running` works on a much different data structure and performs additional checks. I was not sure about merging two functions

src/bin/pg_autoctl/monitor.c
mtuncer: I made `ensure_monitor_pg_running` to be called before pgsql_listen calls. Also,  terminated active connection monitor->pgsql->connection if the monitor is found to be not running.

tests/test_basic_operation.py
mtuncer: ok

src/bin/pg_autoctl/monitor.c
```diff
@@ -1865,13 +1866,76 @@ prepare_connection_to_current_system_user(Monitor *source, Monitor *target)
...
+		if (!monitor_ensure_extension_version(monitor, &version))
+		{
+			/* errors have already been logged */
+			return false;
+		}
+
+	}
+
+	log_debug("PostgreSQL is running in \"%s\" on port %d",
+			  mconfig->pgSetup.pgdata, mconfig->pgSetup.pgport);
```
DimCitus: I think we're going to display that at every iteration of the LISTEN loop, is that right? we should move that to just before the loop in the `monitor_service_run` function, right?

src/bin/pg_autoctl/monitor.c
```diff
@@ -1881,6 +1945,13 @@ monitor_listen_loop(Monitor *monitor)
 	 */
 	for (;;)
 	{
+		if (!ensure_monitor_pg_running(monitor, mconfig))
+		{
+			log_warn("Failed to ensure PostgreSQL is running, retrying in %d seconds", PG_AUTOCTL_MONITOR_SLEEP_TIME);
+			sleep(PG_AUTOCTL_MONITOR_SLEEP_TIME);
+			continue;
+		}
```
DimCitus: We still have the problem that when we lose the connection, we do `pgsql_listen` before we ensure that Postgres is running. Maybe your initial placement of the function call was right, or maybe we need to re-think that loop and make it a `do { ... } while (true)` one instead, where it might be easier to reason and include the first step in the main loop.

mtuncer: ensure_monitor_pg_running issues `pgsql_finish` if the db is known to be down. This way the first call to  `monitor_get_notifications` will fail, and issue `pgsql_listen` again. I don't see any benefit to move `monitor_get_notifications` before `ensure_monitor_pg_running`

src/bin/pg_autoctl/monitor.c
mtuncer: it used to be log_info. I was trying to reduce the frequency. Okay with moving this line outside.

mtuncer: maybe just remove it altogether

PR #122, opened 2019-12-10
WIP: Feature/multiple standbys

DimCitus:
Allow more than one standby per primary.


src/monitor/group_state_machine.c
```diff
@@ -295,14 +341,533 @@ ProceedGroupState(AutoFailoverNode *activeNode)
...
+	 * We reach this code when we don't have an healthy primary anymore, it's
+	 * been demoted or is draining now. Most probably it's dead. We have
+	 * collected the last received LSN from all the standby nodes in SECONDARY
+	 * state, and now we need to select one of the failover candidates.
+	 *
+	 * The selection is based on candidatePriority. If the candidate with the
+	 * higher priority doesn't have the most recent LSN, we have it fetch the
+	 * missing WAL bits from one of the standby which did receive them.
+	 */
+	if (reportedLSNCount == candidateCount)
```
marcocitus: If I understand correctly, we require that all standbys be up to continue with a failover. What would be a way to improve that?

src/monitor/group_state_machine.c
```diff
@@ -295,14 +341,533 @@ ProceedGroupState(AutoFailoverNode *activeNode)
...
+				"after %s:%d became unhealthy.",
+				selectedNode->nodeName, selectedNode->nodePort,
+				mostAdvancedNode->nodeName, mostAdvancedNode->nodePort,
+				primaryNode->nodeName, primaryNode->nodePort);
+
+			AssignGoalState(selectedNode,
+							REPLICATION_STATE_WAIT_FORWARD, message);
+
+			AssignGoalState(mostAdvancedNode,
+							REPLICATION_STATE_WAIT_CASCADE, message);
```
marcocitus: What's the reason we don't put the primary node into draining here?

src/monitor/group_state_machine.c
```diff
@@ -295,14 +341,533 @@ ProceedGroupState(AutoFailoverNode *activeNode)
...
+								message);
+
+				return true;
+			}
+		}
+	}
+
+	/*
+	 * when another node wants to become standby:
+	 *  primary -> join_primary
```
marcocitus: Is failover disabled while the new standby is joining? If the primary fails during this time window, is that something we can recover from?

src/bin/pg_autoctl/fsm_transition.c
```diff
@@ -412,40 +413,84 @@ prepare_replication(Keeper *keeper, bool other_node_missing_is_ok)
...
+	/*
+	 * Should we fail somewhere in this loop, we return false and fail the
+	 * whole transition. The transition is going to be tried again, and we are
+	 * going to try and add HBA entries and create replication slots again.
+	 * Both operations succeed when their target entry already exists.
+	 *
+	 * Note that primary_create_replication_slot() is idempotent thanks to
+	 * first dropping the target replication slot, then creating it again. We
+	 * might want to avoid drop/create noise on those servers that we managed
+	 * to process in the previous loop.
```
marcocitus: Yes, dropping the slot means standby that are a little behind might not be able to catch up anymore.

src/bin/pg_autoctl/pgsql.c
```diff
@@ -637,6 +648,350 @@ pgsql_drop_replication_slot(PGSQL *pgsql, const char *slotName, bool verbose)
...
+/*
+ * pgsql_drop_replication_slots drops all the replication slots. If the verbose
+ * flag is false, then no info message will be logged.
+ */
+bool
+pgsql_drop_replication_slots(PGSQL *pgsql, bool verbose)
+{
+	char *sql =
+		"SELECT pg_drop_replication_slot(slot_name) "
+		"  FROM pg_replication_slots ";
```
marcocitus: Might be useful to only drop 'streaming' slots and not the logical ones.

src/monitor/group_state_machine.c
```diff
@@ -295,14 +341,533 @@ ProceedGroupState(AutoFailoverNode *activeNode)
...
+
+			}
+			else if (otherNode->candidatePriority == 0)
+			{
+				/* also not a candidate */
+				--failoverCandidateCount;
+			}
+
+			/* disable synchronous replication to maintain availability */
+			if (failoverCandidateCount == 0)
```
marcocitus: Do we want to keep a minimum here?

src/monitor/group_state_machine.c
```diff
@@ -295,14 +341,533 @@ ProceedGroupState(AutoFailoverNode *activeNode)
...
+	 * If a failover is in progress, continue driving it.
+	 *
+	 * Here we have choosen a failover candidate, which is either being
+	 * promoted to being the new primary (when it already had all the most
+	 * recent WAL, or is done fetching them), or is fetching the most recent
+	 * WAL it's still missing from another standby node.
+	 */
+	if (candidateNode != NULL)
+	{
+		/*
```
marcocitus: This code block always returns. Might help readability to move it into a separate function. I keep going back and forth between the top and bottom of this function.

src/monitor/group_state_machine.c
```diff
@@ -295,14 +341,533 @@ ProceedGroupState(AutoFailoverNode *activeNode)
...
+			return true;
+		}
+
+		/*
+		 * The candidate had to fast forward with the activeNode, grabbing WAL
+		 * bits that only this node had. Now that's done. The activeNode has to
+		 * follow the new primary that's being promoted.
+		 */
+		if (IsCurrentState(activeNode, REPLICATION_STATE_WAIT_CASCADE) &&
+			IsCurrentState(candidateNode, REPLICATION_STATE_STOP_REPLICATION))
```
marcocitus: Does only the candidate go through stop replication? I assume that doesn't cause the primary to detect a network partition. If it doesn't and wére ok with that, is the state still useful?

src/monitor/group_state_machine.c
DimCitus: In that branch the primary has been put into draining already...

src/monitor/group_state_machine.c
DimCitus: We allow failover when in the new state JOIN_PRIMARY, that's why we have a new state here and don't re-use WAIT_PRIMARY.

src/monitor/group_state_machine.c
DimCitus: Done in a new function!

src/bin/pg_autoctl/fsm_transition.c
DimCitus: Good point. I implemented “if not exists” behaviour instead here.

src/bin/pg_autoctl/pgsql.c
DimCitus: And we have a pattern for the replication slot names that pg_auto_failover should care about, too. So now we're using it to drop only “our” slots. Thanks.

src/monitor/group_state_machine.c
DimCitus: All standby nodes that are currently considered as a potential candidate need to have reported their LSN. From one call to `node_active` to the next, that candidate list might change. I have improved how we build the list itself, and then the comments around this test. Oh and fixed it so that we prevent progress when candidateCount is zero...

src/monitor/group_state_machine.c
DimCitus: Having a single node go through stop replication will not make the primary a network partition. Adding to that, we also have all the other standby nodes re-connect to the new primary when they do REPORT_LSN to SECONDARY. That happens as soon as the candidate enters STOP REPLICATION. All combined, we now expect the primary to detect the network partition (because it no longer has any steaming replica connected).

DimCitus: Getting back to this point, I have now introduced a (yet another) state in between REPORT_LSN and SECONDARY. Nodes go through JOIN_SECONDARY now, and in that state we CHECKPOINT; then stop PostgreSQL. When going from JOIN_SECONDARY to SECONDARY we then edit the recovery configuration (`primary_conninfo`) and restart Postgres.

So while the selected candidate for failover goes through PREPARE_REPLICATION then STOP_REPLICATION, the other nodes go through REPORT_LSN then JOIN_SECONDARY. This gives the effect of shutting down replication streams with the primary, which will then realise it if it's alone in the loosing part of a network split.

src/bin/pg_autoctl/file_utils.c
```diff
@@ -519,3 +521,65 @@ set_program_absolute_path(char *program, int size)
...
+ * then write the same content to the filename again, but skipping lines that
+ * match with the given regex.
+ */
+bool
+rewrite_file_skipping_lines_matching(const char *filename, const char *regex)
+{
+	char *fileContents = NULL;
+	long fileSize = 0L;
+
+	char *fileLines[BUFSIZE];
```
mtuncer: This limits the files to 1024 lines

src/bin/pg_autoctl/file_utils.c
```diff
@@ -519,3 +521,65 @@ set_program_absolute_path(char *program, int size)
...
+
+	/* split the file contents in lines and filter them */
+	lineCount = splitLines(fileContents, fileLines, BUFSIZE);
+
+	for (lineNumber = 0; lineNumber < lineCount; lineNumber++)
+	{
+		char *currentLine = fileLines[lineNumber];
+
+		/* copy lines that are NOT matching the regex */
+		if (regexp_first_match(currentLine, regex) == NULL)
```
mtuncer: `regexp_first_match` returns `malloc` ed buffer. It should be freed

src/bin/pg_autoctl/file_utils.c
DimCitus: Added an array bound check, thanks!

src/bin/lib/log/src/log.h
```diff
@@ -23,6 +23,7 @@ enum { LOG_TRACE, LOG_DEBUG, LOG_INFO, LOG_WARN, LOG_ERROR, LOG_FATAL };
 #define log_warn(...)  log_log(LOG_WARN,  __FILE__, __LINE__, __VA_ARGS__)
 #define log_error(...) log_log(LOG_ERROR, __FILE__, __LINE__, __VA_ARGS__)
 #define log_fatal(...) log_log(LOG_FATAL, __FILE__, __LINE__, __VA_ARGS__)
+#define log_at_level(level, ...) log_log(level, __FILE__, __LINE__, __VA_ARGS__)
```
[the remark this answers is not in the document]
DimCitus: Will remove it from this PR when merging with the current main branch, where we have this:

``` c
#define log_level(level, ...) log_log(level, __FILE__, __LINE__, __VA_ARGS__)
```

src/bin/pg_autoctl/pgsql.c
```diff
@@ -1823,6 +1818,91 @@ parsePgMetadata(void *ctx, PGresult *result)
...
+	}
+
+	if (PQntuples(result) != 1)
+	{
+		log_error("Query returned %d rows, expected 1", PQntuples(result));
+		context->parsedOk = false;
+		return;
+	}
+
+	context->reachedLSN = strcmp(PQgetvalue(result, 0, 0), "t") == 0;
```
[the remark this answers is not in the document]
DimCitus: We use `parse_bool` for parsing user provided input. We trust Postgres boolean output to always be either NUILL, `t`, or `f`, and so we simplify the parsing down.

src/bin/pg_autoctl/cli_create_drop_node.c
```diff
@@ -189,11 +189,11 @@ cli_create_config(Keeper *keeper, KeeperConfig *config)
  * handling.
  */
 void
-cli_create_pg(Keeper *keeper)
+cli_create_pg(Keeper *keeper, NodeState initNodeState)
```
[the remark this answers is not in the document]
DimCitus: In this PR I toyed with the idea of implementing `pg_autoctl create standby`, but I think we should remove that. It was my first step towards deciding how to add a node to an existing group compared to creating a new group with a first node, but we have better ways to solve that. In particular, in the Postgres standalone case, a formation is only allowed to have a single group, so we don't need that anymore.

Will see about removing that part of the PR in the next iteration.

src/bin/pg_autoctl/fsm_transition.c
```diff
@@ -912,3 +978,288 @@ fsm_promote_standby(Keeper *keeper)
...
+	}
+
+	/*
+	 * We expect a single WAIT_CASCADE node in the system at any point.
+	 */
+	upstreamNode = &(keeper->otherNodes.nodes[0]);
+
+	if (!standby_init_replication_source(postgres,
+										 upstreamNode,
+										 PG_AUTOCTL_REPLICA_USERNAME,
```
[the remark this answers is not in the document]
DimCitus: In the current master's branch we now have:

``` c
bool
standby_init_database(LocalPostgresServer *postgres,
					  const char *nodename,
					  bool skipBaseBackup)
{
	PostgresSetup *pgSetup = &(postgres->postgresSetup);
	ReplicationSource *upstream = &(postgres->replicationSource);

...
```

and also

``` c
/*
 * standby_init_replication_source initializes a replication source structure
 * with given arguments. If the primaryNode is NULL, then the
 * replicationSource.primary structure slot is not updated.
 *
 * Note that we just store the pointers to all those const char *arguments
 * here, expect for the primaryNode there's no copying involved.
 */
bool
standby_init_replication_source(LocalPostgresServer *postgres,
								NodeAddress *primaryNode,
								const char *username,
								const char *password,
								const char *slotName,
								const char *maximumBackupRate,
								const char *backupDirectory,
								SSLOptions sslOptions,
								int currentNodeId)
{
	ReplicationSource *upstream = &(postgres->replicationSource);
...
```

So I think that is a good answer to your comment, and we will need to also change that when doing the merge for this PR.

src/bin/pg_autoctl/file_utils.c
```diff
@@ -742,6 +744,68 @@ set_program_absolute_path(char *program, int size)
 }
 
 
+/*
+ * rewrite_file_skipping_lines_matching reads filename contents in memory and
```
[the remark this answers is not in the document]
DimCitus: Yes, will see about cleaning it out.

src/bin/pg_autoctl/fsm_transition.c
```diff
@@ -654,6 +654,44 @@ fsm_stop_postgres(Keeper *keeper)
...
+	if (pgSetup->control.pg_control_version >= 1200)
+	{
+		if (!pgsql_reset_primary_conninfo(pgsql))
+		{
+			log_error("Failed to RESET primary_conninfo");
+			return false;
+		}
+	}
+
+	log_info("Issue a CHECKPOINT; command before stopping Postgres");
```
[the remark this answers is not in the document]
DimCitus: Good point, thanks.

src/bin/pg_autoctl/primary_standby.c
```diff
@@ -839,3 +847,258 @@ check_postgresql_settings(LocalPostgresServer *postgres, bool *settings_are_ok)
...
+								 pgSetup->pg_ctl,
+								 pgSetup->pgdata,
+								 pgsql))
+	{
+		log_error("Failed to clean-up Postgres replication settings, "
+				  "see above for details");
+		return false;
+	}
+
+	/* we might be back from maintenance and find Postgres is not running */
```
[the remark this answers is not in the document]
DimCitus: Yes that needs to happen when I later merge to the current main branch.

src/bin/pg_autoctl/fsm_transition.c
```diff
@@ -846,7 +898,8 @@ fsm_start_maintenance_on_standby(Keeper *keeper)
 bool
 fsm_restart_standby(Keeper *keeper)
 {
-	return fsm_start_postgres(keeper);
+	/* the primary might have change during our maintenance period */
+	return fsm_follow_new_primary(keeper);
```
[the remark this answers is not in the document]
DimCitus: Meanwhile in the main branch the code for restarting a standby after maintenance is now:

``` c
bool
fsm_restart_standby(Keeper *keeper)
{
	return fsm_rewind_or_init(keeper);
}
```

And this will fetch the current primary from the monitor. I think this also is going to be changed in the big merge to the current main branch.

src/bin/pg_autoctl/keeper.c
```diff
@@ -700,7 +701,9 @@ keeper_ensure_postgres_is_running(Keeper *keeper, bool updateRetries)
 		}
 		return true;
 	}
-	else if (ensure_local_postgres_is_running(postgres))
+	/* race condition? sometimes the funcall is happy and the pid 0 */
```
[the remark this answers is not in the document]
DimCitus: I think this got fixed in the following work: https://github.com/citusdata/pg_auto_failover/pull/244 ; so again something that we might fix just by having a nice merge to the current main branch.

src/bin/pg_autoctl/pgctl.c
```diff
@@ -920,6 +937,156 @@ pg_ctl_start(const char *pg_ctl,
...
+	char *fileContents;
+	long fileSize;
+
+	/* logLevel to use when introducing which file path logs come from */
+	int pathLogLevel = logLevel <= LOG_DEBUG ? LOG_DEBUG : LOG_WARN;
+
+	/* prepare startup.log file in PGDATA */
+	join_path_components(pgStartupPath, pgdata, "startup.log");
+
+	if (read_file(pgStartupPath, &fileContents, &fileSize) && fileSize > 0)
```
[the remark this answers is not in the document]
DimCitus: That's how Postgres work, as far as I understand. We certainly want to avoid losing those startup logs, because in some cases those are very useful.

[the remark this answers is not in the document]
onderkalaci: I agree that it is very useful. 

My point was, even if we're mostly interested in the latest (or maybe few of the latest) logs, we might be printing thousands of restart logs that are in `startup.log`. We can definitely live with that, though users might be confused seeing old logs -- which may even contain unrelated errors to our current error situation.

So, if there is an easy way to distinguish the latest startup logs, say the last 5 ones, it might be more interesting to see I guess.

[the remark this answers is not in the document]
DimCitus: We actually replace the file content at each Postgres start, because of the following code in `pg_ctl_postgres` where we open the file in write mode rather than in append mode:

``` c
	/* we want to redirect the output to logfile */
	logFileDescriptor = open(logfile, FOPEN_FLAGS_W, 0644);
```

Did you actually see `startup.log` content from multiple Postgres starts in your testing?

DimCitus:
Closed thanks to #326

PR #124, opened 2019-12-17
{WIP} Stop pg_autoctl service if postgres instance can not be started

mtuncer:
This is a prototype implementation of pg_autoctl service to exit if postgresql instance can not be started.

Now we have a new option `--exit-if-postgres-is-not-running` in create postgres command. If this is specified,   we check number of retry attempts if we detect that the database instance is down.  If timeout or retry limit is reached we stop restart attempts and terminate the service.

Our retry logic was only considering restart if the node state is in primary. This pr extends that to init and single states too. This is open to discussion.

### DIFF
--- a/src/bin/pg_autoctl/cli_create_drop_node.c
+++ b/src/bin/pg_autoctl/cli_create_drop_node.c
@@ -51,10 +51,10 @@ CommandLine create_monitor_command =
 	make_command("monitor",
 				 "Initialize a pg_auto_failover monitor node",
 				 " [ --pgdata --pgport --pgctl --nodename ] ",
-				 "  --pgctl       path to pg_ctl\n" \
-				 "  --pgdata      path to data directory\n" \
-				 "  --pgport      PostgreSQL's port number\n" \
-				 "  --nodename    hostname by which postgres is reachable\n" \
+				 "  --pgctl       path to pg_ctl\n"
+				 "  --pgdata      path to data directory\n"
+				 "  --pgport      PostgreSQL's port number\n"
+				 "  --nodename    hostname by which postgres is reachable\n"
 				 "  --auth        authentication method for connections from data nodes\n"
 				 "  --run         create node then run pg_autoctl service\n",
 				 cli_create_monitor_getopts,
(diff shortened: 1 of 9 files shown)


src/bin/pg_autoctl/cli_create_drop_node.c
```diff
@@ -64,19 +64,21 @@ CommandLine create_postgres_command =
...
+				 "  --username              			PostgreSQL's username\n"
+				 "  --dbname                			PostgreSQL's database name\n"
+				 "  --nodename              			pg_auto_failover node\n"
+				 "  --formation             			pg_auto_failover formation\n"
+				 "  --monitor               			pg_auto_failover Monitor Postgres URL\n"
+				 "  --auth                  			authentication method for connections from monitor\n"
+				 "  --candidate-priority    			priority of the node to be promoted to become primary\n"
+				 "  --replication-quorum    			true if node participates in write quorum\n"
+				 "  --run         						create node then run pg_autoctl service\n"
+				 "  --exit-if-postgres-is-not-running   exit pg_autoctl if postgres node can not be started\n"
```
DimCitus: Maybe we could keep the previous alignment and have the `--exit-...` option that is very long be off-alignment?

I'm not sure what the standard terminal width is nowadays, but I now I like it better when it's easy to copy/paste text from the terminal to an email or a document and have it fit on the page, which to this days still mean 80 columns, or in the email case if you want to make sure it fits, 72 columns...

src/bin/pg_autoctl/keeper.c
```diff
@@ -245,7 +245,7 @@ ReportPgIsRunning(Keeper *keeper)
 	int timeout = config->postgresql_restart_failure_timeout;
 	uint64_t now = time(NULL);
 
-	if (keeperState->current_role != PRIMARY_STATE)
+	if (keeperState->current_role != PRIMARY_STATE && keeperState->current_role != SINGLE_STATE)
```
DimCitus: We should add a comment here to explain why we are interested in only those two states.

The monitor only cares about the PgIsRunning information when it deals with a PRIMARY, CATCHINGUP, and SECONDARY. Then we add in SINGLE state because we want to update our local values for that state and shut down the keeper when PostgreSQL is not running (for the 3rd time) when `--exit-...` has been used.

Maybe it would be even better to have this state depend on `keeper->config.exit_keeper_if_postgres_not_running` to make it obvious, and to avoid extra computations when they are not needed?

src/bin/pg_autoctl/loop.c
```diff
@@ -216,6 +217,19 @@ keeper_service_run(Keeper *keeper, pid_t *start_pid)
 			}
 		}
 
+
+		if (!reportPgIsRunning && config->exit_keeper_if_postgres_not_running)
```
DimCitus: In a non-intuitive way I think I would prefer this test to be split in two as in the following example. It allows to have comments at the right places, make it obvious that we do something specific when `config->exit_keeper_if_postgres_not_running` is in use, and allows us to also add things to that case pretty easily, in the future, independently of `reportPgIsRunning`.

``` C
		if (config->exit_keeper_if_postgres_not_running)
		{
			/*
			 * 
			 */
			if (!reportPgIsRunning)
			{
				...
			}
		}
```

PR #133, opened 2019-12-26
Fix/error handling improvements

DimCitus:


src/bin/pg_autoctl/parsing.c
```diff
@@ -423,3 +430,38 @@ parse_bool(const char *value, bool *result)
 {
 	return parse_bool_with_len(value, strlen(value), result);
 }
+
+
+/*
+ * getNextErrorLine prepares a multi-line error message in a way that calling
```
mtuncer: comment and the function implementation is not in sync

DimCitus: Fixed, thanks!

PR #134, opened 2019-12-26
Feature/multiple other node

DimCitus:
Allow more than two nodes per group, by allowing another standby to join. Then change all the internal APIs to allow for _other nodes_ rather than a single one, both server-side and client-side.

In passing, implement new commands:

    $ pg_autoctl show nodes
    $ pg_autoctl do monitor get others

The first one can be useful as a complement to the existing `pg_autoctl show state` when debugging some situations, and the second one is the new version of the `pg_autoctl do monitor get other` that we used to have. They both can output the nodes in a tabular form or using the JSON format.


src/monitor/group_state_machine.c
```diff
@@ -82,27 +87,42 @@ ProceedGroupState(AutoFailoverNode *activeNode)
...
+	/*
+	 * We separate out the FSM for the primary server, because that one needs
+	 * to loop over every other node to take decisions. That induces some
+	 * complexity that is best managed in a specialized function.
+	 */
+	if (IsInPrimaryState(activeNode))
 	{
-		char message[BUFSIZE];
+		return ProceedGroupStateForPrimaryNode(activeNode);
+	}
```
marcocitus: Would it make sense to put the remainder of the function in a ProceedGroupStateForSecondaryNode?

DimCitus: I though about it, and decided to minimise the changes as much as I could, so that we can see the same code as before in the same function, but with more cases added to it.

In the final PR we have the following code for multiple standby failover orchestration for instance:

``` c
	/* Multiple Standby failover is handled in its own function */
	if (IsUnhealthy(primaryNode) && nodesCount > 2)
	{
		/*
		 * ProceedGroupStateForMSFailover chooses the failover candidate when
		 * there's more than one standby node around, by applying the
		 * candidatePriority and comparing the reportedLSN. The function also
		 * orchestrate fetching the missing WAL from the failover candidate if
		 * that's needed.
		 *
		 * When ProceedGroupStateForMSFailover returns true, it means it was
		 * successfull in driving the failover to the next step, and we should
		 * stop here. When it return false, it did nothing, and so we want to
		 * apply the common orchestration code for a failover.
		 */
		if (ProceedGroupStateForMSFailover(activeNode, primaryNode))
		{
			return true;
		}
	}
```

So it could make sense to have an even more explicit split, maybe, yes.

src/bin/pg_autoctl/cli_do_fsm.c
```diff
@@ -274,15 +274,20 @@ keeper_cli_fsm_assign(int argc, char **argv)
...
+			/* we just accept a single standby at a time here */
+			NodeAddress *otherNode = &(keeper.otherNodes.nodes[0]);
+
+			keeper.otherNodes.count = 1;
+
 			goalState = NodeStateFromString(argv[0]);
 
-			/* now prepare host and port in keeper.otherNode */
-			strlcpy(keeper.otherNode.host, argv[1], _POSIX_HOST_NAME_MAX);
+			/* now prepare host and port in keeper.otherNodes */
```
mtuncer: id parameter is expected to be before host and port. We are still getting those from their original offsets

mtuncer: and doing nothing regarding the id parameter

src/bin/pg_autoctl/cli_do_monitor.c
```diff
@@ -51,13 +55,13 @@ static CommandLine monitor_get_primary_command =
...
-	make_command("other",
-				 "Get the other node from the pg_auto_failover group of nodename/port",
+static CommandLine monitor_get_other_nodes_command =
+	make_command("others",
+				 "Get the other nodes from the pg_auto_failover group of nodename/port",
 				 " [ --pgdata ]",
 				 KEEPER_CLI_PGDATA_OPTION,
-				 keeper_cli_getopt_pgdata,
-				 keeper_cli_monitor_get_other_node);
+				 cli_do_monitor_get_other_nodes_getopts,
```
mtuncer: only difference between two functions is addition --json parameter. Would it be possible add this in the existing getopts function instead of introducing regular one

src/bin/pg_autoctl/cli_do_monitor.c
```diff
@@ -161,26 +165,173 @@ keeper_cli_monitor_get_primary_node(int argc, char **argv)
...
-			config.formation, config.groupId,
-			primaryNode.host, primaryNode.port);
+	(void) printNodeHeader(strlen(primaryNode.host));
+	(void) printNodeEntry(&primaryNode);
+	fprintf(stdout, "\n");
+}
+
+
+/*
+ * keeper_cli_monitor_state_getopts parses the command line options for the
```
mtuncer: name mismatch

src/bin/pg_autoctl/cli_do_monitor.c
```diff
@@ -161,26 +165,173 @@ keeper_cli_monitor_get_primary_node(int argc, char **argv)
...
+keeper_cli_monitor_get_other_nodes(int argc, char **argv)
 {
 	KeeperConfig config = keeperOptions;
 	bool missing_pgdata_is_ok = true;
 	bool pg_is_not_running_is_ok = true;
 
 	Monitor monitor = { 0 };
-	NodeAddress otherNode = { 0 };
+
+	/* arbitrary limit to 12 other nodes */
```
mtuncer: hardcoded limit gives me itches. 12 is pretty generous upper limit, but still can we do without it ?

src/bin/pg_autoctl/monitor.c
```diff
@@ -104,14 +248,77 @@ monitor_get_other_node(Monitor *monitor, char *myHost, int myPort, NodeAddress *
...
+	if (!context.parsedOk)
+	{
+		log_error("Failed to get the other nodes from the monitor while running "
+				  "\"%s\" with host %s and port %d because it returned an "
+				  "unexpected result. See previous line for details.",
+				  sql, myHost, myPort);
+		return false;
+	}
+
+	strlcpy(json, context.strVal, size);
```
mtuncer: if context.strVal is longer than size, output string would not have terminating  NULL at the end

mtuncer: my comment is not correct, however, if for some reason the data is truncated, it won't be well formed json

mtuncer: context.strVal is allocated inside  `parseSingleValueResult` via `strdup` but never deallocated

src/bin/pg_autoctl/cli_do_fsm.c
DimCitus: Good spot, fixed.

src/bin/pg_autoctl/cli_do_monitor.c
DimCitus: What about command that would then accept a `--json` option but fail to deliver JSON output? I guess we should have JSON output support everywhere now, rather than half baked easy-enough-to-parse-maybe output like we did at first. That sounds like another PR though.  Will work on that other PR separately.

src/bin/pg_autoctl/cli_do_monitor.c
DimCitus: Fixed, thanks.

src/bin/pg_autoctl/cli_do_monitor.c
DimCitus: I pondered the alternative, but that would require to introduce careful memory management, which we avoided up to now. I would much rather than an artificial limit or 12 nodes (for now) rather than deal with dynamic memory in many places...

src/bin/pg_autoctl/monitor.c
DimCitus: Error handling added, thanks.

DimCitus: I added a call to `free` the memory there, and I noticed other places that share the same problem. Another PR to fix those places would be good.

src/bin/pg_autoctl/cli_do_monitor.c
DimCitus: See https://github.com/citusdata/pg_auto_failover/pull/140.

DimCitus:
> We have changes in sql api. We would need to provide a new upgrade file or update existing one (1.0 -> 1.1)

I have edited the upgrade script, and tried to add it to the SQL regression tests (`make installcheck`). I have added the SQL callable function `get_other_node` in the .so (it errors out with ERRCODE_FEATURE_NOT_SUPPORTED).

PR #136, opened 2019-12-26
Review how to get the local Postgres metadata.

DimCitus:
Rather than doing multiple and different queries for the cases when the
local Postgres is a primary or a standby, do a single query that knows how
to fetch the metadata we are interested in.

### DIFF
--- a/src/bin/pg_autoctl/keeper.c
+++ b/src/bin/pg_autoctl/keeper.c
@@ -517,7 +517,16 @@ keeper_update_pg_state(Keeper *keeper)
 
 		case PRIMARY_STATE:
 		{
-			/* we expect replication to be in place */
+			/*
+			 * We expect to be able to read the current LSN, as always when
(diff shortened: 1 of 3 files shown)


src/bin/pg_autoctl/keeper.c
```diff
@@ -488,19 +509,33 @@ keeper_update_pg_state(Keeper *keeper)
...
+		{
+			/* we don't expect to have a streaming replica */
+			return postgres->pgIsRunning;
+		}
+
+		case PRIMARY_STATE:
+		{
+			/* we expect replication to be in place */
+			return postgres->pgIsRunning
+				&& !IS_EMPTY_STRING_BUFFER(postgres->currentLSN)
```
mtuncer: why we make emptiness check for lsn and syncstate is not explicit. perhaps single line explaining why

DimCitus: Good point, done now!

PR #138, opened 2020-01-02
Make destroy command public

### LINKED ISSUE #126: Move destroy to the main commands
When users discover and play around with pg_auto_failover for the first time there comes a point to wiping down the install and starting again. Or maybe because they were not using the right version of Postgres, or some other issue. It would be best to have a `pg_autoctl destroy` command for that, rather than the developers only `pg_autoctl do destroy` command as we have now.

Other elements to have in mind:

  - the `pg_autoctl drop` command should list remaining files
  - the `pg_autoctl drop` command should inform the user about the `destroy` command being available as a log_info line
  - the `pg_autoctl destroy` command should work even when a `pg_autoctl drop` command was used just before, and should then finish the cleaning.

mtuncer:
fixes #126


src/bin/pg_autoctl/cli_create_drop_node.c
```diff
@@ -632,6 +639,8 @@ cli_drop_node(int argc, char **argv)
 	log_info("Removed the keeper from the monitor and removed the state file "
 			 "from disk.");
 
+	log_info("Node data is not deleted. You can also use \"destroy\" "
+			 "command to delete all resources belonging to the node");
```
DimCitus: Negative log entries sound like a problem, or read like something should be done. It's not the case here. What about the following messages:

``` c
    log_warn("configuration file \"%s\" has been preserved", config.pathnames.config);\
    if (directory_exists(config.pgSetup.pgdata))
    {
        log_warn("Postgres Data Directory \"%s\" has been preserved", config.pgSetup.pgdata);
    }
    log_info("drop node keeps your data and setup safe, you can still run Postgres or re-join the pg_auto_failover cluster later");
    log_info("HINT: to completely remove your local Postgres instance and setup, consider the command pg_autoctl destroy");
```

We should refrain from the PGDATA related WARN message when the directory doesn't exist anyway, even though the `drop node` command itself is not deleting it. It could have been deleted by the user.

src/bin/pg_autoctl/cli_create_drop_node.c
```diff
@@ -855,3 +864,171 @@ check_nodename(const char *nodename)
...
+		{
+			log_error("Failed to remove directory \"%s\": %s",
+					  pgSetup->pgdata, strerror(errno));
+			exit(EXIT_CODE_INTERNAL_ERROR);
+		}
+	}
+	else
+	{
+		log_info("Skipping removal of \"%s\": directory does not exists",
+				 pgSetup->pgdata);
```
DimCitus: Now wondering if that should be a WARN message. It sounds like it should, what do you think?

src/bin/pg_autoctl/cli_create_drop_node.c
mtuncer: ok

src/bin/pg_autoctl/cli_create_drop_node.c
mtuncer: it feels like an unexpected situation, warn might be better

PR #139, opened 2020-01-09
Fix/test suite

DimCitus:


Makefile
```diff
@@ -6,6 +6,15 @@ TEST_CONTAINER_NAME = pg_auto_failover-test
 DOCKER_RUN_OPTS = --cap-add=SYS_ADMIN --cap-add=NET_ADMIN -ti --rm
 PDF = ./docs/_build/latex/pg_auto_failover.pdf
 
+# TEST indicates the testfile to run
```
mtuncer: Like the feature. wonder if using lowercase name would be nicer

DimCitus: The way make works, the variable can be set in the environment or just for the make command. That's why I believe it's usually uppercase.

```
$ TEST=test_auth make run-test
```

tests/pgautofailover_utils.py
```diff
@@ -473,7 +482,8 @@ def set_replication_quorum(self, replicationQuorum):
             Sets replication quorum via pg_autoctl
         """
 
-        set_command = [shutil.which('pg_autoctl'), 'set', 'node', '--pgdata', self.datadir,
+        set_command = [shutil.which('pg_autoctl'),
```
mtuncer: can we hava PGAutoCtl class handle this too ?

tests/pgautofailover_utils.py
```diff
@@ -657,14 +672,104 @@ def failover(self, formation='default', group=0):
...
+
+    def communicate(self):
+        """
+        Read all data from the Unix PIPE
+        """
+        self.out, self.err = self.run_proc.communicate()
+
+        return self.out, self.err
+
+    def readmore(self, secs=1):
```
mtuncer: can you re-think the name

tests/test_create_run.py
```diff
@@ -75,19 +75,3 @@ def test_010_drop_secondary():
...
     assert node2.wait_until_state(target_state="single")
-    
-def test_011_add_new_secondary():
-    global node3
-    node3 = cluster.create_datanode("/tmp/basic/node3")
-    node3.create(run = True)
-    assert node3.wait_until_state(target_state="secondary")
-    assert node2.wait_until_state(target_state="primary")
-
-def test_012_multiple_manual_failover():
```
mtuncer: any reason why we don't need this anymore ?

tests/pgautofailover_utils.py
DimCitus: done, thanks!

tests/pgautofailover_utils.py
DimCitus: Good point, I did that in the follow-up commit, now pushed.

tests/test_create_run.py
DimCitus: I don't believe those tests are expecting any specific behaviour in `create --run` and we test them already in the normal `create` then `run` cases.

PR #140, opened 2020-01-10
Implement --json for (almost) all commands.

DimCitus:
Have the option parsing for --json implemented in the common function
cli_getopt_pgdata and implement JSON formatted output in (almost) all
commands.

### DIFF
--- a/src/bin/pg_autoctl/cli_config.c
+++ b/src/bin/pg_autoctl/cli_config.c
@@ -38,25 +38,25 @@ static void cli_monitor_config_set(int argc, char **argv);
 static CommandLine config_check =
 	make_command("check",
 				 "Check pg_autoctl configuration",
-				 " [ --pgdata ]",
-				 KEEPER_CLI_PGDATA_OPTION,
-				 keeper_cli_getopt_pgdata,
+				 CLI_PGDATA_USAGE,
+				 CLI_PGDATA_OPTION,
+				 cli_getopt_pgdata,
 				 cli_config_check);
 
 static CommandLine config_get =
 	make_command("get",
 				 "Get the value of a given pg_autoctl configuration variable",
-				 "[ section.option ]",
-				 KEEPER_CLI_PGDATA_OPTION,
-				 keeper_cli_getopt_pgdata,
+				 CLI_PGDATA_USAGE "[ section.option ]",
+				 CLI_PGDATA_OPTION,
+				 cli_getopt_pgdata,
 				 cli_config_get);
 
 static CommandLine config_set =
 	make_command("set",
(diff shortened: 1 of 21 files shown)


mtuncer:
we need to know what is included and excluded in **almost**.  I went ahead and tested with `show` command which apparently is not json enabled yet

src/bin/pg_autoctl/cli_config.c
```diff
@@ -110,9 +110,54 @@ cli_config_check(int argc, char **argv)
 
 			(void) cli_config_check_pgsetup(&(mconfig.pgSetup));
 
-			fprintf(stdout, "Current Configuration (includes pgSetup):\n");
-			monitor_config_write(stdout, &mconfig);
-			fprintf(stdout, "\n");
+			if (outputJSON)
+			{
+				JSON_Value *js = json_value_init_object();
```
mtuncer: looks like we are going to need more copy/paste s of this part. Would it make sense to enclose creation in a function and keep the variables  in a struct.

src/bin/pg_autoctl/cli_config.c
```diff
@@ -130,9 +175,54 @@ cli_config_check(int argc, char **argv)
...
+				json_object_set_value(root, "postgres", jsPostgres);
+				json_object_set_value(root, "config", jsKConfig);
+
+				/* output our nice JSON object, pretty printed please */
+				serialized_string = json_serialize_to_string_pretty(js);
+				fprintf(stdout, "%s\n", serialized_string);
+
+				/* free intermediate memory */
+				json_free_serialized_string(serialized_string);
+				json_value_free(js);
```
mtuncer: we are not free in other json values like jsPostgres and jsKConfig

src/bin/pg_autoctl/cli_config.c
DimCitus: I'm not sure we can have something generic enough to support the different use cases (direct JSON object use, and nested-object uses, and we will need to add JSON arrays when we add support for `pg_autoctl show events`). I've simplified the output serialisation and memory freeing though. Thanks!

src/bin/pg_autoctl/cli_config.c
DimCitus: I have worked under the assumptions that as soon as the values are nested into each other, the `json_value_free` will walk the graph... a quick glance at the parson source code seems to agree with that.

DimCitus:
Copy/paste amount reduced. I'm not sure if providing a list of commands with JSON output support is warranted at this point. The goal of this PR  is to simplify the number of `getotps` function we need around after all, not to have JSON output everywhere. I think we have a decent effort in place now and we will be able to add more in the future if people ask for it?

PR #141, opened 2020-01-14
Fix/cli perform destroy

### LINKED ISSUE #94: Permissions missing in monitor database
Hi,

is this a known issue?

```
postgres@pg-af3:/home/postgres/ [af] pg_autoctl show state
Name |   Port | Group |  Node |     Current State |    Assigned State
---------------------------+--------+-------+-------+-------------------+------------------
pg-af1.it.dbi-services.com |   5432 |     0 |     1 |           primary |           primary
pg-af2.it.dbi-services.com |   5432 |     0 |     2 |         secondary |         secondary

17:29:11 postgres@pg-af3:/home/postgres/ [af] psql postgres://autoctl@pg-af3:5432/pg_auto_failover
psql (12.0 dbi services build)
Type "help" for help.

pg_auto_failover=> select pgautofailover.perform_failover(formation_id => 'default', group_id => 0);
ERROR:  permission denied for schema pgautofailover
LINE 1: select pgautofailover.perform_failover(formation_id => 'defa...
^
```

PostgreSQL 12 on CentOS 8.
pg_autoctl version 1.0.5

Not an issue to fix it manually but I still wonder why permissions are missing in the monitor database. Did I miss something?

Best regards
Daniel

DimCitus:
- Add `pg_autoctl perform failover` as a CLI operation
  - Add `pg_autoctl perform switchover` as a synonym to it: when done in a controlled way what we implement is actually a switch-over, thanks to the careful implementation of the FSM on the monitor and to the default to synchronous replication
  - Remove `pg_autoctl destroy` and make it an option to `pg_autoctl drop node --destroy` instead, that fits better with the spirit of our CLI
  - Re-order get/set commands so that they are listed with the command that have sub-commands, before the final commands

In passing, we fix #94 by making `pgautofailover.perform_failover()` a SECURITY DEFINER function.

### DIFF
--- a/src/bin/pg_autoctl/cli_create_drop_node.c
+++ b/src/bin/pg_autoctl/cli_create_drop_node.c
@@ -32,8 +32,12 @@
 #include "pgctl.h"
 #include "primary_standby.h"
 
+/*
+ * Global variables that we're going to use to "communicate" in between getopts
+ * functions and their command implementation. We can't pass parameters around.
+ */
 MonitorConfig monitorOptions;
-bool dropAndDestroy = false;
+static bool dropAndDestroy = false;
 
 static int cli_create_postgres_getopts(int argc, char **argv);
 static void cli_create_postgres(int argc, char **argv);
@@ -792,6 +796,9 @@ cli_drop_node(int argc, char **argv)
 	{
 		/* all we really need now is pg_ctl */
 		set_first_pgctl(&(config.pgSetup));
+		log_info("Configuration file \"%s\" does not exists, using %s",
+				 config.pathnames.config,
+				 config.pgSetup.pg_ctl);
 	}
 
 	/*
--- a/src/bin/pg_autoctl/cli_create_drop_node.c
+++ b/src/bin/pg_autoctl/cli_create_drop_node.c
@@ -47,10 +47,10 @@ static void cli_drop_node(int argc, char **argv);
 static bool discover_nodename(char *nodename, int size,
 							  const char *monitorHostname, int monitorPort);
 static void check_nodename(const char *nodename);
+
 static void stop_postgres_and_remove_pgdata_and_config(
 	ConfigFilePaths *pathnames,
 	PostgresSetup *pgSetup);
-
 
 CommandLine create_monitor_command =
 	make_command("monitor",
@@ -319,7 +319,7 @@ cli_create_monitor_getopts(int argc, char **argv)
 	};
 
 	/* hard-coded defaults */
-	options.pgSetup.pgport = 5432;
+	options.pgSetup.pgport = pgsetup_get_pgport();
 
 	optind = 0;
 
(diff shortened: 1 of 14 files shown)


src/bin/pg_autoctl/cli_create_drop_node.c
```diff
@@ -614,73 +720,171 @@ cli_drop_node(int argc, char **argv)
...
+				 "state file \"%s\" does not exist",
+				 config.pathnames.state);
 	}
 
-	if (pg_ctl_stop(config.pgSetup.pg_ctl, config.pgSetup.pgdata))
+	/*
+	 * Either --destroy the whole Postgres cluster and configuraiton, or leave
+	 * enough behind us that it's possible to re-join a formation later.
+	 */
+	if (dropAndDestroy)
```
mtuncer: destroy also includes drop right ? why don't be make this earlier in the function ? This way we might need less changes in the drop implementation

DimCitus: Well the rest of the actions are common to both `drop node` and `drop node --destroy`, so we don't need the test until way down in the implementation. The new code is reading the configuration file to grab the `postgresql.pg_ctl` parameter and make sure that's the one we use, where before we would just use `set_first_pgctl` and might use `pg_ctl` from Postgres 12 on a Postgres 11 cluster.

src/bin/pg_autoctl/cli_create_drop_node.c
```diff
@@ -33,19 +33,24 @@
 #include "primary_standby.h"
 
 MonitorConfig monitorOptions;
+bool dropAndDestroy = false;
```
mtuncer: not a fun of global stuff. can we at least make it static ?

tests/test_create_run.py
```diff
@@ -16,11 +16,11 @@ def teardown_module():
 
 def test_000_create_monitor():
     global monitor
-    monitor = cluster.create_monitor("/tmp/basic/monitor")
+    monitor = cluster.create_monitor("/tmp/create--run/monitor")
```
mtuncer: maybe `create-run` or `create_run`

tests/pgautofailover_utils.py
```diff
@@ -687,7 +687,7 @@ def stop(self):
 
             try:
                 pgid = os.getpgid(self.run_proc.pid)
-                os.killpg(pgid, signal.SIGTERM)
+                os.killpg(pgid, signal.SIGQUIT)
```
mtuncer: Is this better than SIGTERM ? documentation says it is a signal to dump core

tests/test_create_run.py
DimCitus: Done using `create-run` ; I did like using `--run` in the name of the directory because we are testing `pg_autoctl create --run`, but yeah, that's a tad too much.

tests/pgautofailover_utils.py
DimCitus: We implement our own signal handlers in our code, see https://github.com/citusdata/pg_auto_failover/blob/master/src/bin/pg_autoctl/signals.c#L79 for the SIGQUIT handler. That's an _immediate shutdown_.

src/bin/pg_autoctl/cli_create_drop_node.c
DimCitus: Done, thanks!

Issue #135, opened 2019-12-26
Support debian/ubuntu style PostgreSQLclusters


DimCitus:
When using debian or ubuntu packages, `apt-get install postgresql` creates a cluster in `/var/lib/postgresql/11/main` (or another version), with configuration files in `/etc/` (both `PostgreSQL.conf` and `pg_hba.conf`.

  1. `pg_basebackup` is not copying the files from `/etc` so creating the standby node will fail (see #63) ; so `pg_autoctl create Postgres` could move the files over to `PGDATA`
  2. we want `pg_autoctl run` to be started as a service, and prevent the usual Postgres service to run (in some cases we need to ensure that the local Postgres is NOT running to avoid split-brain), so `pg_autoctl` should change debian configuration:
      - `mv /etc/postgresql/11/main/postgresql.conf` to $PGDATA
      - `mv /etc/postgresql/11/main/pg_hba.conf` to $PGDATA
      - `mv /etc/postgresql/11/main/pg_ident.conf` to $PGDATA
      - rename `/etc/postgresql/11/main/start.conf` to `start.conf.orig`
      - create `/etc/postgresql/11/main/start.conf` with the content `disabled` and a pg_auto_failover comment about DO NOT EDIT THIS FILE
      - issue a log WARN about integration with systemd or something else
  3. we may add symlinks to the 3 configuration files from `/etc/postgresql/11/main` to ease the life of our debian/ubuntu users.
  4. we need to actually edit the `postgresql.conf` file from debian, to comment out the lines where it changes `config_file` and `hba_file` and `ident_file` to use the default values (in PGDATA) again
  5. we should do this operation for any situation where the configuration files are not found in PGDATA, as our step to *own* the cluster.

mtuncer:
This requirement does not address the case when user creates a new data directory via `pg_createcluster` . You can create a database directory as follows

`pg_createcluster -d /tmp/mydb 11 newdb`

This would create 2 directories
`/tmp/mydb` for `$PGDATA`
`/etc/postgresql/11/newdb` for storing configuration files

I haven't found a way to make association between 2 yet.

DimCitus:
I think that the association is the value of the `data_directory` setting in the configuration file, when it is not the default value:

```
#------------------------------------------------------------------------------
# FILE LOCATIONS
#------------------------------------------------------------------------------

# The default values of these variables are driven from the -D command-line
# option or PGDATA environment variable, represented here as ConfigDir.

[...]

PR #144, opened 2020-01-22
Support debian ubuntu style clusters

mtuncer:
Contains items for satisfying  #135 

Did some basic testing

Opening PR to verify the basic idea

### DIFF
--- a/src/bin/pg_autoctl/debian.c
+++ b/src/bin/pg_autoctl/debian.c
@@ -198,7 +198,7 @@ debian_find_postgres_configuration_files(PostgresSetup *pgSetup,
 
 
 /*
- * debian_init_postgres_config_files initialises the given PostgresConfigFiles
+ * debian_init_postgres_config_files initializes the given PostgresConfigFiles
  * structure with the location of existing files as found on-disk given a
  * Postgres configuration kind.
  */
@@ -481,7 +481,7 @@ expandDebianPatternsInDirectoryName(char *pathname,
 }
 
 /*
- * initPostgresConfigFiles initialises PostgresConfigFiles structure with our
+ * initPostgresConfigFiles initializes PostgresConfigFiles structure with our
  * filenames located in given directory pathname.
  */
 static void
--- a/src/bin/pg_autoctl/debian.c
+++ b/src/bin/pg_autoctl/debian.c
@@ -1,0 +1,699 @@
+/*
+ * src/bin/pg_autoctl/debian.c
+ *
+ *   Debian specific code to support registering a pg_autoctl node from a
+ *   Postgres cluster created with pg_createcluster. We need to move the
+ *   configuration files back to PGDATA.
+ *
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the PostgreSQL License.
+ *
+ */
+#include <libgen.h>
+
+#include "postgres_fe.h"
+#include "pqexpbuffer.h"
+
+#include "debian.h"
+#include "keeper.h"
+#include "keeper_config.h"
+#include "parsing.h"
+
+#define EDITED_BY_PG_AUTOCTL "# edited by pg_auto_failover \n"
+
+static bool debian_find_postgres_configuration_files(
+	PostgresSetup *pgSetup,
+	PostgresConfigFiles *pgConfigFiles);
+
+static bool debian_init_postgres_config_files(
+	PostgresSetup *pgSetup,
+	PostgresConfigFiles *pgConfFiles,
+	PostgresConfigurationKind confKind);
+
+static bool buildDebianDataAndConfDirectoryNames(
+	PostgresSetup *pgSetup,
+	DebianPathnames *debPathnames);
+
+static bool expandDebianPatterns(
+	DebianPathnames *debPathnames,
+	const char *dataDirectoryTemplate,
+	const char *confDirectoryTemplate);
+
+static bool expandDebianPatternsInDirectoryName(
+	char *pathname,
+	int pathnameSize,
+	const char *template,
+	const char *versionName,
+	const char *clusterName);
+
+static void initPostgresConfigFiles(
+	const char *dirname,
+	PostgresConfigFiles *pgConfigFiles,
+	PostgresConfigurationKind kind);
+
+static bool postgresConfigFilesAllExist(
+	PostgresConfigFiles *pgConfigFiles);
+
+static bool move_configuration_files(
+	PostgresConfigFiles *src,
+	PostgresConfigFiles *dst);
+
+static bool comment_out_configuration_parameters(
+	const char *srcConfPath,
+	const char *dstConfPath);
+
+static bool disableAutoStart(PostgresConfigFiles *pgConfigFiles);
+
+
+/*
+ * keeper_ensure_pg_configuration_files_in_pgdata checks if postgresql.conf,
+ * pg_hba.conf, pg_ident.conf files exist in $PGDATA, if not it tries to get
+ * them from default location and modifies paths inside copied postgresql.conf.
+ */
+bool
+keeper_ensure_pg_configuration_files_in_pgdata(KeeperConfig *config)
+{
+	PostgresSetup *pgSetup = &(config->pgSetup);
+	PostgresConfigFiles pgConfigFiles = { 0 };
+
+	if (!debian_find_postgres_configuration_files(pgSetup, &pgConfigFiles))
+	{
+		/* errors have already been logged */
+		return false;
+	}
+
+	switch (pgConfigFiles.kind)
+	{
+		case PG_CONFIG_TYPE_POSTGRES:
+		{
+			/* that's it, we're good */
+			return true;
+		}
+
+		case PG_CONFIG_TYPE_DEBIAN:
+		{
+			/*
+			 * So now pgConfigFiles is the debian path for configuration files,
+			 * and we're building a new pgdataConfigFiles for the Postgres
+			 * configuration files in PGDATA.
+			 */
+			PostgresConfigFiles pgdataConfigFiles = { 0 };
+
+			log_info("Found a debian style installation in PGDATA \"%s\" with "
+					 "postgresql.conf located at \"%s\"",
+					 pgSetup->pgdata,
+					 pgConfigFiles.conf);
+
+			initPostgresConfigFiles(pgSetup->pgdata,
+									&pgdataConfigFiles,
+									PG_CONFIG_TYPE_POSTGRES);
+
+			log_info("Moving configuration files back to PGDATA at \"%s\"",
+					 pgSetup->pgdata);
+
+			/* move configuration files back to PGDATA, or die trying */
+			if (!move_configuration_files(&pgConfigFiles,
+										  &pgdataConfigFiles))
+			{
+				char *_dirname = dirname(pgConfigFiles.conf);
+
+				log_fatal("Failed to move the debian configuration files from "
+						  "\"%s\" back to PGDATA at \"%s\"",
+						  _dirname,
+						  pgSetup->pgdata);
+				return false;
+			}
+
+			/* also disable debian auto start of the cluster we now own */
+			if (!disableAutoStart(&pgConfigFiles))
+			{
+				log_fatal("Failed to disable debian auto-start behavior, "
+						  "see above for details");
+				return false;
+			}
+
+			return true;
+		}
+
+		case PG_CONFIG_TYPE_UNKNOWN:
+		{
+			log_fatal("Failed to find the \"postgresql.conf\" file. "
+					  "It's not in PGDATA, and it's not in the debian "
+					  "place we had a look at. See above for details");
+			return false;
+		}
+	}
+}
+
+
+/*
+ * debian_find_postgres_configuration_files finds the Postgres configuration
+ * files following the following strategies:
+ *
+ *  - first attempt to find the files where we expect them, in PGDATA
+ *  - then attempt to find the files in the debian /etc/postgresql/%v/%c place
+ *
+ * At the moment we only have those two strategies, and with some luck that's
+ * all we're ever going to need.
+ */
+static bool
+debian_find_postgres_configuration_files(PostgresSetup *pgSetup,
+										 PostgresConfigFiles *pgConfigFiles)
+{
+	PostgresConfigFiles postgresConfFiles = { 0 };
+	PostgresConfigFiles debianConfFiles = { 0 };
+
+	pgConfigFiles->kind = PG_CONFIG_TYPE_UNKNOWN;
+
+	if (!pg_setup_pgdata_exists(pgSetup))
+	{
+		return PG_CONFIG_TYPE_UNKNOWN;
+	}
+
+	/* is it a Postgres core initdb style setup? */
+	if (debian_init_postgres_config_files(pgSetup,
+										  &postgresConfFiles,
+										  PG_CONFIG_TYPE_POSTGRES))
+	{
+		/* so we're dealing with a "normal" Postgres installation */
+		memcpy(pgConfigFiles, &postgresConfFiles, sizeof(PostgresConfigFiles));
+
+		return true;
+	}
+
+	/* is it a debian postgres-common style setup then? */
+	if (debian_init_postgres_config_files(pgSetup,
+										  &debianConfFiles,
+										  PG_CONFIG_TYPE_DEBIAN))
+	{
+		/* so we're dealing with a "normal" Postgres installation */
+		memcpy(pgConfigFiles, &debianConfFiles, sizeof(PostgresConfigFiles));
+
+		return true;
+	}
+
+	/* well that's all we know how to detect at this point */
+	return false;
+}
+
+
+/*
+ * debian_init_postgres_config_files initialises the given PostgresConfigFiles
+ * structure with the location of existing files as found on-disk given a
+ * Postgres configuration kind.
+ */
+static bool
+debian_init_postgres_config_files(PostgresSetup *pgSetup,
+								  PostgresConfigFiles *pgConfigFiles,
+								  PostgresConfigurationKind confKind)
+{
+	const char *pgdata = pgSetup->pgdata;
+
+	switch (confKind)
+	{
+		case PG_CONFIG_TYPE_UNKNOWN:
+		{
+			/* that's a bug really */
+			log_error("BUG: debian_init_postgres_config_files "
+					  "called with UNKNOWN conf kind");
+			return false;
+		}
+
+		case PG_CONFIG_TYPE_POSTGRES:
+		{
+			initPostgresConfigFiles(pgdata, pgConfigFiles,
+									PG_CONFIG_TYPE_POSTGRES);
+
+			if (postgresConfigFilesAllExist(pgConfigFiles))
+			{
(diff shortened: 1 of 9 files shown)


src/bin/pg_autoctl/keeper_pg_init.c
```diff
@@ -86,6 +87,169 @@ keeper_pg_init_fsm(Keeper *keeper, KeeperConfig *config)
...
+	char searchString[MAXPGPATH];
+
+	snprintf(searchString, MAXPGPATH, "%s =", parameterName);
+
+	while ( (substring = strstr(fileContents + offset, searchString)) != NULL)
+	{
+		offset = substring - fileContents;
+		*substring = '#';
+		memcpy(substring+1, parameterName, strlen(parameterName));
+	}
```
DimCitus: I don't think that loop read the fileContents line by line and matches with the parameter name only when it's the first thing in the line (ignoring white space). It's pretty hard for me to make sure the function implements what we need here.

src/bin/pg_autoctl/keeper_pg_init.c
```diff
@@ -86,6 +87,169 @@ keeper_pg_init_fsm(Keeper *keeper, KeeperConfig *config)
...
+ * It will remove the space between parameter name and = to make room for #
+ * in the beginning in order to perform modification in place.
+ */
+static void disable_configuration_parameter_inplace(char *fileContents, long fileSize, char *parameterName)
+{
+	int offset = 0;
+	char *substring = NULL;
+	char searchString[MAXPGPATH];
+
+	snprintf(searchString, MAXPGPATH, "%s =", parameterName);
```
DimCitus: What if there's a different amount of space around the equal sign?

mtuncer:
@DimCitus  updated the pr addressing most of your comments.

src/bin/pg_autoctl/file_utils.c
```diff
@@ -262,6 +262,124 @@ read_file(const char *filePath, char **contents, long *fileSize)
...
+	struct stat sourceFileStat;
+	bool foundError = false;
+
+	if (!read_file(sourcePath, &fileContents, &fileSize))
+	{
+		/* errors are logged */
+		return false;
+	}
+
+	foundError = !write_file(fileContents, fileSize, destinationPath);
```
DimCitus: This time we might be missing a `file_exists` test, that was placed in the previous function `move_file` where I don't think it belongs.

src/bin/pg_autoctl/keeper_pg_init.c
```diff
@@ -86,6 +88,198 @@ keeper_pg_init_fsm(Keeper *keeper, KeeperConfig *config)
...
+	char *pgdataRelativePath = NULL;
+	char defaultConfigurationPath[MAXPGPATH];
+	bool configurationPathPresent = false;
+	bool postgresqlConfExists = false;
+	char defaultPosgresqlConfPath[MAXPGPATH];
+	char defaultHbaConfPath[MAXPGPATH];
+	char defaultIdentConfPath[MAXPGPATH];
+	char identConfPath[MAXPGPATH];
+
+	get_absolute_path(config->pgSetup.pgdata, pgdata, MAXPGPATH);
```
DimCitus: PGDATA is supposed to exist on-disk already. When it does, we've been using `realpath` to get the absolute path to PGDATA in other places. See `pg_setup_set_absolute_pgdata` in `pgsetup.c` for instance. Why not using the same approach here, or maybe even checking that `pg_setup_set_absolute_pgdata` has been called already (and adding a comment about it), or calling it from here?

src/bin/pg_autoctl/keeper_pg_init.c
```diff
@@ -86,6 +87,169 @@ keeper_pg_init_fsm(Keeper *keeper, KeeperConfig *config)
...
+
+
+	/* check if postgresql.conf exists in pgdata. there is no separate check for pg_hba and pg_ident  */
+	join_path_components(posgresqlConfPath, pgdata, "postgresql.conf");
+
+	postgresqlConfExists = file_exists(posgresqlConfPath);
+
+	if (postgresqlConfExists)
+	{
+		return true;
```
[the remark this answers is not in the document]
DimCitus: I really think we should return true early only after having checked that the three files we are looking for all can be found in PGDATA.

src/bin/pg_autoctl/keeper_pg_init.c
```diff
@@ -86,6 +88,198 @@ keeper_pg_init_fsm(Keeper *keeper, KeeperConfig *config)
...
+	/* check if postgresql.conf exists in pgdata. there is no separate check for pg_hba and pg_ident  */
+	join_path_components(posgresqlConfPath, pgdata, "postgresql.conf");
+	postgresqlConfExists = file_exists(posgresqlConfPath);
+	if (postgresqlConfExists)
+	{
+		return true;
+	}
+
+	/* check if pgdata is in default path. */
+	if (strstr(pgdata, "/var/lib/postgresql") != pgdata)
```
DimCitus: What do we mean by "default path" here? I guess you mean debian packaging default path? but that's very specific to debian style packaging, right? We could maybe `#define` a constant `DEBIAN_PACKAGING_DEFAULT_PGDATA_TOP_DIRECTORY` or something?

src/bin/pg_autoctl/keeper_pg_init.c
```diff
@@ -86,6 +88,198 @@ keeper_pg_init_fsm(Keeper *keeper, KeeperConfig *config)
...
+	{
+		log_error("Cannot determine configuration file location for this instance. pgdata = %s", pgdata);
```
DimCitus: If Postgres is running, we could run a query to discover the configuration file value. We could even run a single query that fetches the file path for the `config_file`, the `hba_file`, and the `pg_ident` file (not sure about that last one) and then fill-in a structure with all the information ready for the rest of this command.

src/bin/pg_autoctl/keeper_pg_init.c
```diff
@@ -86,6 +88,198 @@ keeper_pg_init_fsm(Keeper *keeper, KeeperConfig *config)
...
+		log_error("Cannot find configuration file directory \"%s\"", defaultConfigurationPath);
+		return false;
+	}
+
+	join_path_components(defaultPosgresqlConfPath, defaultConfigurationPath, "postgresql.conf");
+	if (!move_file(defaultPosgresqlConfPath, posgresqlConfPath))
+	{
+		/* errors are already logged */
+		return false;
+	}
```
DimCitus: Could this be done in a specialised function, to keep the main code here as small as possible? Otherwise we're lost in the minutia of the function here and miss the big picture.

src/bin/pg_autoctl/keeper_pg_init.c
```diff
@@ -86,6 +88,198 @@ keeper_pg_init_fsm(Keeper *keeper, KeeperConfig *config)
...
+
+	/* edit postgresql.conf to disable settings */
+	disable_configuration_parameters(posgresqlConfPath);
```
DimCitus: Rather than `disable`, what about using `comment_out` or `edit_..._to_comments` or something more to the point?

src/bin/pg_autoctl/keeper_pg_init.c
```diff
@@ -86,6 +88,198 @@ keeper_pg_init_fsm(Keeper *keeper, KeeperConfig *config)
...
+
+	join_path_components(hbaConfPath, pgdata, "pg_hba.conf");
+	if(!file_exists(hbaConfPath))
+	{
+
+		log_trace("hba file does not exist : %s, need to copy", hbaConfPath);
+		join_path_components(defaultHbaConfPath, defaultConfigurationPath, "pg_hba.conf");
+
+		move_file(defaultHbaConfPath, hbaConfPath);
+	}
```
DimCitus: Please declare the local variable just in the block where you need them, here `defaultHbaConfPath` for instance. Also, the name is not very good, as it's not the default HBA location (which is per Postgres in PGDATA), but the debian pg_createcluster location. Maybe `debianHBAPath` would make more sense?

src/bin/pg_autoctl/keeper_pg_init.c
```diff
@@ -86,6 +88,198 @@ keeper_pg_init_fsm(Keeper *keeper, KeeperConfig *config)
...
+
+	join_path_components(identConfPath, pgdata, "pg_ident.conf");
+	if(!file_exists(identConfPath))
+	{
+
+		log_trace("pg_ident.conf file does not exist : %s, need to copy", identConfPath);
+		join_path_components(defaultIdentConfPath, defaultConfigurationPath, "pg_ident.conf");
+
+		move_file(defaultIdentConfPath, identConfPath);
+	}
```
DimCitus: Same as the block before. Given that we have two blocks that are very similar, maybe we want to build a very small function with a good name and call it instead of copy/pasting the code around that way?

src/bin/pg_autoctl/keeper_pg_init.c
```diff
@@ -86,6 +88,198 @@ keeper_pg_init_fsm(Keeper *keeper, KeeperConfig *config)
...
+
+	/* disable auto start in default configuration */
+	disableAutoStart(defaultConfigurationPath);
```
DimCitus: Again, I believe a name that says `CommentOutDebianAutoStart` would be better than `disable...` for a function that edits a configuration file and prepend comment markers to some lines.

src/bin/pg_autoctl/keeper_pg_init.c
```diff
@@ -86,6 +88,198 @@ keeper_pg_init_fsm(Keeper *keeper, KeeperConfig *config)
...
+	/*
+	 * configuration parameters can appear in any order, and we
+	 * need to check for patterns for NAME = VALUE and NAME=VALUE
+	 */
+	char *targetVariables[] = {
+			"data_directory ", "data_directory =",
+			"hba_file ", "hba_file=",
+			"ident_file ", "ident_file=",
+			"include_dir ", "include_dir=",
+			NULL};
```
DimCitus: Why is `data_directory =` spelled with a space before the = sign where the other parameters have no space before the equal sign? Can we use a regex matching instead? We already have a `regexp_first_match` function in `parsing.c` after all.

A regex as the following should work?

    ^[ ]*(data_directory|hba_file|ident_file|include_dir)[ =]

src/bin/pg_autoctl/keeper_pg_init.c
```diff
@@ -86,6 +88,198 @@ keeper_pg_init_fsm(Keeper *keeper, KeeperConfig *config)
...
+	return true;
+}
+
+
+/*
+ * disable_configuration_parameters read postgresql.conf file and
+ * disables predefined configuration parameters inside.
+ */
+static bool
+disable_configuration_parameters(char *postgresqlConfPath)
```
DimCitus: Per previous comments, please review the function name here.

src/bin/pg_autoctl/keeper_pg_init.c
```diff
@@ -86,6 +88,198 @@ keeper_pg_init_fsm(Keeper *keeper, KeeperConfig *config)
...
+	destroyPQExpBuffer(newConfContents);
+
+	return true;
+}
+
+
+/* disableAutoStart disables auto start in default configuration */
+static void
+disableAutoStart(char *defaultConfigurationPath)
+{
```
DimCitus: Per previous comments, please review the function name here.

src/bin/pg_autoctl/keeper_pg_init.c
```diff
@@ -86,6 +88,198 @@ keeper_pg_init_fsm(Keeper *keeper, KeeperConfig *config)
...
+	char *fileContents = NULL;
+	long fileSize = 0;
+
+	join_path_components(defaultStartConfPath, defaultConfigurationPath, "start.conf");
+	join_path_components(newStartConfPath, defaultConfigurationPath, "start.conf.orig");
+
+	if (read_file(defaultStartConfPath, &fileContents, &fileSize))
+	{
+		write_file(fileContents, fileSize, newStartConfPath);
+	}
```
DimCitus: Is that a manually coded `rename` here?

src/bin/pg_autoctl/keeper_pg_init.c
```diff
@@ -86,6 +88,198 @@ keeper_pg_init_fsm(Keeper *keeper, KeeperConfig *config)
...
+
+	write_file(newStartConfData, strlen(newStartConfData), defaultStartConfPath);
```
DimCitus: It would be nice to add a comment that the file has been edited by `pg_auto_failover`, as we do in all the configuration files we edit.

src/bin/pg_autoctl/keeper_pg_init.c
```diff
@@ -118,6 +312,15 @@ keeper_pg_init_and_register(Keeper *keeper, KeeperConfig *config)
 	bool postgresInstanceExists = pg_setup_pgdata_exists(&pgSetup);
 	bool postgresInstanceIsPrimary = pg_setup_is_primary(&pgSetup);
 
+	if (postgresInstanceExists)
+	{
+		if (!keeper_ensure_pg_configuration_files_in_pgdata(config))
+		{
+			log_fatal("Existing postgresql instance is not supported");
```
DimCitus: We need to give more information to the user here I believe. Just saying that we only support core Postgres setup where the configuration files live in PGDATA might be enough. But "not supported" is not actionable input to the user.

src/bin/pg_autoctl/keeper_pg_init.c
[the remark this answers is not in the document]
mtuncer: ok

src/bin/pg_autoctl/file_utils.c
```diff
@@ -262,6 +262,124 @@ read_file(const char *filePath, char **contents, long *fileSize)
...
+	{
+		log_error("Failed to move file, source file \"%s\" does not exist.", sourcePath);
+		return false;
+	}
+
+	if (file_exists(destinationPath))
+	{
+		log_error("Failed to move file, destination file \"%s\" already exists.", destinationPath);
+		return false;
+	}
```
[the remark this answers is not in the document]
mtuncer: rename would error if the source is not there but does not complain about writing on an existing file

src/bin/pg_autoctl/file_utils.c
mtuncer: added check for destination file

src/bin/pg_autoctl/keeper_pg_init.c
mtuncer: used `realpath`

DimCitus:
I took a stab at solving the points I was raising in the new version I just pushed, please have a look.

src/bin/pg_autoctl/debian.c
```diff
@@ -0,0 +1,699 @@
...
+#include "pqexpbuffer.h"
+
+#include "debian.h"
+#include "keeper.h"
+#include "keeper_config.h"
+#include "parsing.h"
+
+#define EDITED_BY_PG_AUTOCTL "# edited by pg_auto_failover \n"
+
+static bool debian_find_postgres_configuration_files(
```
mtuncer: consider using the same naming convention.

src/bin/pg_autoctl/debian.c
```diff
@@ -0,0 +1,699 @@
...
+
+	if (!debian_find_postgres_configuration_files(pgSetup, &pgConfigFiles))
+	{
+		/* errors have already been logged */
+		return false;
+	}
+
+	switch (pgConfigFiles.kind)
+	{
+		case PG_CONFIG_TYPE_POSTGRES:
```
mtuncer: it is always of type POSTGRES,  could not think of a better name though. DEFAULT ? but we have way too many "default"s around

src/bin/pg_autoctl/debian.c
```diff
@@ -0,0 +1,699 @@
...
+		return true;
+	}
+
+	/* well that's all we know how to detect at this point */
+	return false;
+}
+
+
+/*
+ * debian_init_postgres_config_files initialises the given PostgresConfigFiles
```
mtuncer: s/initialises/initializes

src/bin/pg_autoctl/debian.c
```diff
@@ -0,0 +1,699 @@
...
+ *
+ * Note that debian's /etc/postgresql-common/createcluster.conf defaults to
+ * using the following setup, and that's the only one we support at this
+ * moment.
+ *
+ *   data_directory = '/var/lib/postgresql/%v/%c'
+ *
+ */
+static bool
+buildDebianDataAndConfDirectoryNames(PostgresSetup *pgSetup,
```
mtuncer: we are just translating `/var/lib/postgresql/%VERSION%/%CLUSTER%` to `/etc/postgresql/%VERSION%/%CLUSTER%` 
Looks like this is too much effort.

src/bin/pg_autoctl/file_utils.c
```diff
@@ -518,3 +671,39 @@ set_program_absolute_path(char *program, int size)
 
 	return true;
 }
+
+
+/*
+ * normalize_filename returns the real path of a given filename, resolving
+ * symlinks and pruning double-slashes and other weird constructs.
```
mtuncer: It is worth adding a note that if the file does not exists, the same path is returned without any normalization attempt, or something like that.

tests/pgautofailover_utils.py
```diff
@@ -70,6 +71,30 @@ def create_datanode(self, datadir, port=5432, group=0,
...
+    def pg_createcluster(self, datadir, port=5432):
+        """
+        Initializes a postgresql node using pg_createcluster and returns
+        directory path.
+        """
+        vnode = self.vlan.create_node()
+        nodeid = len(self.datanodes) + 1
+
+        create_command = ["sudo", shutil.which('pg_createcluster'),
+                          "-U", os.getenv("USER"),
```
mtuncer: you mentioned an error earlier `Error: clusters must not be owned by root` this line is the reason for that.

Perhaps we should `postgres` user if  current user is root.

src/bin/pg_autoctl/debian.c
DimCitus: I meant Postgres as in “like the Postgres core code” to differentiate with “like the debian patched version” or nowadays “like the debian packaging”... maybe I should add more comments about that?

DimCitus: Actually debian.h contains the following, I think that's all we need?

```
 * We know how to find configuration files in either PGDATA as per Postgres
 * core, or in the debian cluster configuration directory as per debian
 * postgres-common packaging, implemented in pg_createcluster.
```

src/bin/pg_autoctl/debian.c
DimCitus: Any debian user can change `data_directory` in `/etc/postgresql/createcluster.conf` and use something else that those paths, or in another ordering, or etc. So I though later we might want to parse `createcluster.conf` and use whatever is setup there. Currently we only have their default hard-coded and that might be enough.

tests/pgautofailover_utils.py
DimCitus: The problem was with using "docker" hard-coded there, where we need to use `os.getenv("USER")` like in the other call sites.

PR #146, opened 2020-01-23
Feature/set nodename

### LINKED ISSUE #145: pg_autoctl set node nodename
In some environments the IP address of the nodes are volatile and may change at reboot. It's the case with Kubernetes for instance. It's also the case that Kubernetes does not provide a reverse DNS solution that would comply with Postgres, by default.

To make it easy to use pg_auto_failover in such an environment, it is possible to use the IP address of the node as the nodename, but then it is necessary to be able to update the IP address at container startup.

DimCitus:
Implement a new command to set the node's nodename. We expect users to be using that command when they have to use IP addresses and are in an environment where the IP address can change in between reboots. Postgres being quite picky about the reverse DNS setup in its HBA checks, such environments could be hard to manage without this command.

In passing, review the `pg_autoctl set node|formation` commands to be more like the other ones, with the last level of choice being another nested sub-command. 

Also implement a way to forward the new settings to the primary node when that's relevant, and have the CLI commands know to wait for the setting to have been applied both on the monitor and the primary node in that case, making it easier to integrate the commands in scripts.

Fixes #145.


src/bin/pg_autoctl/monitor.c
```diff
@@ -1801,6 +1792,105 @@ monitor_formation_uri(Monitor *monitor, const char *formation,
...
+								   &context, &parseSingleValueResult))
+	{
+		log_error("Failed to get the synchronous_standby_names setting value "
+				  " from the monitor for formation %s and group %d",
+				  formation, groupId);
+		return false;
+	}
+
+	/* disconnect from PostgreSQL now */
+	pgsql_finish(&monitor->pgsql);
```
mtuncer: we seem to close the connection with the knowledge that this call is the last one ever to run queries on existing pgsql object. Shall we move that burden to whomever  creates this object with `monitor_init_from_pgsetup(&monitor, &config.pgSetup))` ?

src/bin/pg_autoctl/monitor.c
```diff
@@ -1801,6 +1792,105 @@ monitor_formation_uri(Monitor *monitor, const char *formation,
...
+								   paramCount, paramTypes, paramValues,
+								   &parseContext, parseNodeResult))
+	{
+		log_error("Failed to set_node_nodename of node %d from the monitor",
+				  nodeId);
+		return false;
+	}
+
+	/* disconnect from PostgreSQL now */
+	pgsql_finish(&monitor->pgsql);
```
mtuncer: consider moving this out

src/bin/pg_autoctl/monitor.c
```diff
@@ -2031,6 +2115,247 @@ monitor_get_notifications(Monitor *monitor)
...
+
+		sock = PQsocket(connection);
+
+		if (sock < 0)
+			return false;	/* shouldn't happen */
+
+		FD_ZERO(&input_mask);
+		FD_SET(sock, &input_mask);
+
+		if (select(sock + 1, &input_mask, NULL, NULL, NULL) < 0)
```
mtuncer: this does not feel right. We opened the connection using libpq functions.  Can we continue with libpq ?

src/bin/pg_autoctl/monitor.c
```diff
@@ -2031,6 +2115,247 @@ monitor_get_notifications(Monitor *monitor)
...
+			{
+				applySettingsTransitionInProgress = true;
+
+				log_debug("step 3/4: primary node %d (%s:%d) is assigned \"%s\"",
+						  notification.nodeId,
+						  notification.nodeName,
+						  notification.nodePort,
+						  NodeStateToString(notification.goalState));
+			}
+			else if (applySettingsTransitionInProgress
```
mtuncer: This code assume that any of steps 1,2,3 occurs before 4. If we ever loose some notifications and hit 4 before others our logic might break here.

src/bin/pg_autoctl/monitor.c
```diff
@@ -2031,6 +2115,247 @@ monitor_get_notifications(Monitor *monitor)
...
+
+		sock = PQsocket(connection);
+
+		if (sock < 0)
+			return false;	/* shouldn't happen */
+
+		FD_ZERO(&input_mask);
+		FD_SET(sock, &input_mask);
+
+		if (select(sock + 1, &input_mask, NULL, NULL, NULL) < 0)
```
mtuncer: socket  magic. Prefer if we can use PGxxx

src/bin/pg_autoctl/pgsql.c
```diff
@@ -651,6 +651,33 @@ postgres_sprintf_replicationSlotName(int nodeId, char *slotName, int size)
...
+bool
+pgsql_set_synchronous_standby_names(PGSQL *pgsql,
+									char *synchronous_standby_names)
+{
+	char quoted[BUFSIZE] = { 0 };
+	GUC setting = { "synchronous_standby_names", quoted };
+
+	log_info("Enabling synchronous replication");
+
+	if (snprintf(quoted, BUFSIZE, "'%s'", synchronous_standby_names) >= BUFSIZE)
```
mtuncer: nice, I was going to ask about his

src/monitor/version_compat.c
```diff
@@ -0,0 +1,62 @@
...
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the PostgreSQL License.
+ *
+ *-------------------------------------------------------------------------
+ */
+
+#if (PG_VERSION_NUM < 110000)
+
+/*
+ * The list_qsort API was introduced in Postgres 12:
```
mtuncer: I guess you meant 11 here

src/monitor/version_compat.c
```diff
@@ -0,0 +1,62 @@
...
+ *
+ * https://git.postgresql.org/gitweb/?p=postgresql.git;a=commit;h=ab72716778128fb63d54ac256adf7fe6820a1185
+ */
+
+#include "postgres.h"
+#include "nodes/pg_list.h"
+#include "version_compat.h"
+
+/*
```
mtuncer: can we enclose this in #ifdef

src/monitor/pgautofailover.sql
```diff
@@ -579,3 +600,63 @@ comment on function pgautofailover.set_node_replication_quorum(int, text, int, b
 grant execute on function
       pgautofailover.set_node_replication_quorum(int, text, int, bool)
    to autoctl_node;
+
+
+create function pgautofailover.synchronous_standby_names
```
mtuncer: can you reflect changes to upgrade script ?

src/bin/pg_autoctl/monitor.c
DimCitus: I see your point, but every other function does kind of the same way as what you're seeing here. So maybe we could open another issue/PR to review connection closing strategy in the source code?

src/bin/pg_autoctl/monitor.c
DimCitus: See other conversation.

src/bin/pg_autoctl/monitor.c
DimCitus: I am basing this code on the Postgres documentation example code for LISTEN/NOTIFY, as seen at https://www.postgresql.org/docs/current/libpq-example.html#LIBPQ-EXAMPLE-2. I'm going to add a comment in the code that makes this obvious.

src/bin/pg_autoctl/monitor.c
DimCitus: The risk does exist. I have implement a generous 30s timeout after which we exit the loop if we didn't get our notifications.

src/bin/pg_autoctl/monitor.c
DimCitus: See above, that's per Postgres documentation. Now with a comment about it.

src/monitor/version_compat.c
DimCitus: Good catch, fixed.

src/monitor/version_compat.c
DimCitus: How do you mean?

src/monitor/pgautofailover.sql
DimCitus: Done now, thanks.

src/monitor/version_compat.c
mtuncer: implementation is only necessary if we are below PG 11.  We can do something like
```c
#if (PG_VERSION_NUM < 110000)
/* implementation for list_qsort */
#endif 
```

src/bin/pg_autoctl/monitor.c
mtuncer: sounds good

src/monitor/version_compat.c
DimCitus: Well I applied that treatment to the whole file actually...

mtuncer: I take this back. it was already there at top of the file

PR #147, opened 2020-01-24
Ensure current state before reaching the assigned state.

DimCitus:
Even at pg_autoctl start we might be given a transition order by the
monitor, such as SINGLE to WAIT_PRIMARY. We should make sure that Postgres
is running (as per the current state SINGLE) before calling the transition
function, as it might need that pre-condition to be true.

### DIFF
--- a/src/bin/pg_autoctl/keeper.c
+++ b/src/bin/pg_autoctl/keeper.c
@@ -134,29 +134,33 @@ keeper_should_ensure_current_state_before_transition(Keeper *keeper)
 {
 	KeeperStateData *keeperState = &(keeper->state);
 
-	if (keeperState->assigned_role != keeperState->current_role)
+	if (keeperState->assigned_role == keeperState->current_role)
 	{
-		if (keeperState->assigned_role == DEMOTED_STATE
-			|| keeperState->assigned_role == DEMOTE_TIMEOUT_STATE
-			|| keeperState->assigned_role == DEMOTED_STATE)
-		{
-			/* don't ensure Postgres is running before shutting it down */
-			return false;
-		}
-
-		if (keeperState->current_role == DEMOTED_STATE
-			|| keeperState->current_role == DEMOTE_TIMEOUT_STATE
-			|| keeperState->current_role == DEMOTED_STATE)
-		{
-			/* don't ensure Postgres is down before starting it again */
-			return false;
-		}
-
-		/* in all other cases, yes please ensure the current state */
-		return true;
+		/* this function should not be called in that case */
+		log_debug("BUG: keeper_should_ensure_current_state_before_transition "
+				  "called with assigned role == current role == %s",
+				  NodeStateToString(keeperState->assigned_role));
+		return false;
 	}
 
-	return false;
+	if (keeperState->assigned_role == DEMOTED_STATE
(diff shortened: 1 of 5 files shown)


tests/pgautofailover_utils.py
```diff
@@ -331,29 +331,42 @@ def wait_until_state(self, target_state, timeout=STATE_CHANGE_TIMEOUT):
...
+            else:
+                time.sleep(1)
 
             current_state = self.get_state()
 
             # only log the state if it has changed
             if current_state != prev_state:
-                print("state of %s is '%s', waiting for '%s' ..." %
-                    (self.datadir, current_state, target_state))
+                if current_state == target_state:
```
mtuncer: we have the same check twice

can we have something like
```python
if current_state == target_state:
    print("state of %s is '%s', waiting for '%s' ..." %(self.datadir, current_state, target_state))
    return True
```

above the enclosing if ?

DimCitus: I think that if `current_state == target_state` then we are done, and we want to print something only when previous state was either None or different from the current_state. In other words I'm not sure we can simplify the logic down much.

src/bin/pg_autoctl/keeper.c
```diff
@@ -113,6 +113,53 @@ keeper_update_state(Keeper *keeper, int node_id, int group_id,
...
+ * That's because we would then stop Postgres first when going from DEMOTED to
+ * SINGLE, or ensure Postgres is running when going from PRIMARY to DEMOTED.
+ * This last example is a split-brain hazard, too.
+ */
+bool
+keeper_should_ensure_current_state_before_transition(Keeper *keeper)
+{
+	KeeperStateData *keeperState = &(keeper->state);
+
+	if (keeperState->assigned_role != keeperState->current_role)
```
mtuncer: can we something like below to make it more explicit. 
```c
if (keeperState->assigned_role == keeperState->current_role)
  return false
```

DimCitus: Fair enough, done now.

src/bin/pg_autoctl/keeper.c
```diff
@@ -113,6 +113,53 @@ keeper_update_state(Keeper *keeper, int node_id, int group_id,
...
+	{
+		if (keeperState->assigned_role == DEMOTED_STATE
+			|| keeperState->assigned_role == DEMOTE_TIMEOUT_STATE
+			|| keeperState->assigned_role == DEMOTED_STATE)
+		{
+			/* don't ensure Postgres is running before shutting it down */
+			return false;
+		}
+
+		if (keeperState->current_role == DEMOTED_STATE
```
mtuncer: what if assigned_role != <ANY_DEMOTED_STATE> but current role is ?

mtuncer: node is currently demoted, but assigned state is something else

DimCitus: Well then as the comment says: _ don't ensure Postgres is down before starting it again_. What am I missing here?

mtuncer: it made sense when you state it that way, thanks

PR #149, opened 2020-01-26
Fix/json

DimCitus:
Implement more JSON support for the `pg_autoctl show` commands, and fix a buffer shortage problem for JSON output that does not fit in BUFSIZE, which is only 1024 bytes.

  - fix pg_autoctl show state --json
  - implement pg_autoctl show events --json
  - implement pg_autoctl show uri --json

Also in passing make the default output for `pg_autoctl show uri` more useful:

```
$ pg_autoctl show uri 
      Type |    Name | Connection String
-----------+---------+-------------------------------
   monitor | monitor | port=4000 dbname=pg_auto_failover host=/tmp user=autoctl_node
 formation | default | postgres://localhost:4002,localhost:4003,localhost:4001/postgres?target_session_attrs=read-write
```

### DIFF
--- a/src/bin/pg_autoctl/cli_do_monitor.c
+++ b/src/bin/pg_autoctl/cli_do_monitor.c
@@ -223,17 +223,16 @@ cli_do_monitor_get_other_nodes(int argc, char **argv)
 	{
 		char json[BUFSIZE];
 
-		if (!monitor_get_other_nodes_as_json(&monitor,
-											 config.nodename,
-											 config.pgSetup.pgport,
-											 ANY_STATE,
-											 json, BUFSIZE))
+		if (!monitor_print_other_nodes_as_json(&monitor,
+											   config.nodename,
+											   config.pgSetup.pgport,
+											   ANY_STATE,
+											   stdout))
 		{
 			log_fatal("Failed to get the other nodes from the monitor, "
 					  "see above for details");
 			exit(EXIT_CODE_MONITOR);
 		}
-		fprintf(stdout, "%s\n", json);
 	}
 	else
 	{
--- a/src/bin/pg_autoctl/cli_do_monitor.c
+++ b/src/bin/pg_autoctl/cli_do_monitor.c
@@ -195,10 +195,6 @@ cli_do_monitor_get_other_nodes(int argc, char **argv)
 
 	Monitor monitor = { 0 };
 
-	/* arbitrary limit to 12 other nodes */
-	int nodeIndex = 0;
-	NodeAddressArray otherNodesArray;
-
 	bool missingPgdataIsOk = true;
 	bool pgIsNotRunningIsOk = true;
 	bool monitorDisabledIsOk = false;
@@ -224,8 +220,7 @@ cli_do_monitor_get_other_nodes(int argc, char **argv)
 		if (!monitor_print_other_nodes_as_json(&monitor,
 											   config.nodename,
 											   config.pgSetup.pgport,
-											   ANY_STATE,
-											   stdout))
+											   ANY_STATE))
 		{
 			log_fatal("Failed to get the other nodes from the monitor, "
 					  "see above for details");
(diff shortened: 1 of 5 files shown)


src/bin/pg_autoctl/cli_do_monitor.c
```diff
@@ -223,17 +223,16 @@ cli_do_monitor_get_other_nodes(int argc, char **argv)
 	{
 		char json[BUFSIZE];
 
-		if (!monitor_get_other_nodes_as_json(&monitor,
-											 config.nodename,
-											 config.pgSetup.pgport,
-											 ANY_STATE,
-											 json, BUFSIZE))
+		if (!monitor_print_other_nodes_as_json(&monitor,
```
mtuncer: I did not like the idea of passing stdout as file stream.

Can  we switch to old way but instead use a dynamic buffer allocated inside ?  We already allocate buffer in `parseSingleValueResult` function which we did not deallocate. We can just pass/return a pointer to that buffer, print it inside this (`cli_do_monitor_get_other_nodes`) function like we did before and free the buffer afterwards .

src/bin/pg_autoctl/cli_do_monitor.c
```diff
@@ -223,17 +223,16 @@ cli_do_monitor_get_other_nodes(int argc, char **argv)
 	{
 		char json[BUFSIZE];
```
mtuncer: this has become useless after the change

src/bin/pg_autoctl/cli_do_monitor.c
DimCitus: I can understand “I don't like” as a discussion opener, but it's a little short as the whole reason for refusing to approve a PR. Can you please expand on why you do not like using a `FILE *stream` argument to a function, when all the function now does is print out a result on said stream?

src/bin/pg_autoctl/cli_do_monitor.c
DimCitus: Thanks, good catch. Fixed.

src/bin/pg_autoctl/cli_do_monitor.c
mtuncer: agree that "I don't like" was not a good start here.   my reasoning is that, the function is always called with the same parameter that is `stdout` we don't open or create that file descriptor or accept as parameter to calling function.

Since the function name contains `...print...`, it is explicit that we will print something somewhere, if we have to do printing inside the function I would prefer use of stdout is inside.

In the second part of the function. We are separating  data fetching and printing duties. I was thinking about using the same approach in both with and without --json. Perhaps we can merge fetch with print functions and just have 2 function calls for with/without --json.

DimCitus: I can see now that the API could be improved, because we have `print_as_json` functions and could expect similar API for the tabular format. I have done that in the revised version of the PR, thanks!

src/bin/pg_autoctl/monitor.c
```diff
@@ -1793,6 +1922,171 @@ monitor_formation_uri(Monitor *monitor, const char *formation,
...
+	if (!context.parsedOk)
+	{
+		/* errors have already been logged */
+		return false;
+	}
+
+	/* disconnect from PostgreSQL now */
+	pgsql_finish(&monitor->pgsql);
+
+	fprintf(stream, "%s\n", context.strVal);
```
mtuncer: context.strVal is not deallocated

mtuncer: I believe we are going to go over all these in a separate work. so you can ignore this one

src/bin/pg_autoctl/monitor.c
```diff
@@ -1402,12 +1448,12 @@ printCurrentState(void *ctx, PGresult *result)
...
- * monitor_get_state_as_json returns a single string that contains the JSON
- * representation of the current state on the monitor.
+ * monitor_print_state_as_json prints to given stream a single string that
+ * contains the JSON representation of the current state on the monitor.
  */
 bool
-monitor_get_state_as_json(Monitor *monitor, char *formation, int group,
-						  char *json, int size)
+monitor_print_state_as_json(Monitor *monitor, char *formation, int group,
+							FILE *stream)
```
mtuncer: can we drop `*stream` and use `stdout` directly here ?

DimCitus: Done now, good catch, thanks!

PR #153, opened 2020-01-29
Avoid using stdin/stdout/stderr as variable names

RekGRpth:


DimCitus:
See #161 which is the same thing targeting current master's branch.

PR #154, opened 2020-02-04
Feature/application name

DimCitus:
Use application_name in our primary_conninfo setting, so that we can use the computed synchronous_standby_names: the names are checked with the application_name value for the replication connections.

In passing, stop using `pgsql_get_config_file_path()` to discover live where `postgresql.conf` is located and just assume it's located in $PGDATA. After all, we fail to create a standby when that's not the case.

Finally, introduce `pg_autoctl show synchronous_standby_names` command.

### DIFF
--- a/src/bin/pg_autoctl/cli_show.c
+++ b/src/bin/pg_autoctl/cli_show.c
@@ -37,7 +37,6 @@ static void cli_show_events(int argc, char **argv);
 static int cli_show_nodes_getopts(int argc, char **argv);
 static void cli_show_nodes(int argc, char **argv);
 
-static int cli_show_standby_names_getopts(int argc, char **argv);
 static void cli_show_standby_names(int argc, char **argv);
 
 static int cli_show_file_getopts(int argc, char **argv);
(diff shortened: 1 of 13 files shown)


src/bin/pg_autoctl/cli_show.c
```diff
@@ -555,6 +568,184 @@ cli_show_nodes(int argc, char **argv)
...
+		{ "group", required_argument, NULL, 'g' },
+		{ "version", no_argument, NULL, 'V' },
+		{ "verbose", no_argument, NULL, 'v' },
+		{ "quiet", no_argument, NULL, 'q' },
+		{ "help", no_argument, NULL, 'h' },
+		{ NULL, 0, NULL, 0 }
+	};
+
+	/* set default values for our options, when we have some */
+	options.network_partition_timeout = -1;
```
mtuncer: we should probably have a macro to reset these. there are many copy pastes of this block

DimCitus: Also we don't have support for JSON in that command, we should probably add that.

PR #155, opened 2020-02-05
Port the quickstart to azure

jonels-msft:


docs/quickstart.rst
```diff
@@ -61,69 +147,67 @@ instance had already existing in the destination directory, this command
...
+   pg_autoctl create postgres     \
+     --pgdata ha \
+     --pgctl `which pg_ctl` \
+     --nodename `hostname -I` \
+     --monitor postgres://autoctl_node@10.0.1.4/pg_auto_failover
+
+   pg_autoctl run --pgdata ha
+
+It discovers from the monitor that a primary exists, and then switches its own
+state to be a hot standby and begins streaming WAL contents from the primary.
```
DimCitus: I think now is a good time to add a section about the Postgres HBA file and setup, and show the HBA entries created by pg_autoctl in every node. We would have two goals with such a section:

  1. explain what pg_autoctl is doing automatically behind the scene to make everything work
  2. explain that those are only defaults and you might want to use your own setting either on-top of it, or instead of it...

We don't have the `--skip-hba` option yet but we want to get there someday. People might want to use kerberos or LDAP based authentication, or even full SSL certificates with client and server validation. On the cloud, I would argue it's a good idea. We don't automate that, and should encourage users to understand their options and consider doing that for their production installation.

DimCitus: Also, this is again a point where I believe that showing the output of the pg_autoctl command would be enlightening for our readers.

docs/quickstart.rst
```diff
@@ -133,59 +217,61 @@ states it has assigned them:
 
 .. code-block:: text
 
+   # on the monitor virtual machine
+
+   sudo su - postgres
+   export PATH="/usr/lib/postgresql/11/bin/:$PATH"
+
```
DimCitus: I like to export PGDATA here, rather than the PATH. In any case, both of those could be added to the profile of the user that is being used, or maybe that would be another extra step you'd rather bypass in the tutorial?

docs/quickstart.rst
```diff
@@ -133,59 +217,61 @@ states it has assigned them:
...
 This looks good. We can add data to the primary, and watch it get
 reflected in the secondary.
 
 .. code-block:: bash
 
+   # on your local machine
+
    # add data to primary
-   psql -p 6010 \
+   psql TODO -p 6010 \
```
DimCitus: Use the following command to get the connection string to use in the application:

```
$ pg_autoctl show uri --formation default
postgres://localhost:4002,localhost:4003,localhost:4001,localhost:4004/postgres?target_session_attrs=read-write
```

In a future version of the tooling you even have:

```
$ pg_autoctl show uri 
      Type |    Name | Connection String
-----------+---------+-------------------------------
   monitor | monitor | port=4000 dbname=pg_auto_failover host=/tmp user=autoctl_node
 formation | default | postgres://localhost:4003,localhost:4002,localhost:4001,localhost:4004/postgres?target_session_attrs=read-write
```

DimCitus: In that future version you can even have it in JSON. I am currently pondering a 1.0.6 release with all those improvements and bug fixes as an interim release before the next big feature release with multiple standby support.

```
$ pg_autoctl show uri --json
[
    {
        "uri": "port=4000 dbname=pg_auto_failover host=/tmp user=autoctl_node",
        "name": "monitor",
        "type": "monitor"
    },
    {
        "uri": "postgres://localhost:4003,localhost:4002,localhost:4001,localhost:4004/postgres?target_session_attrs=read-write",
        "name": "default",
        "type": "formation"
    }
]
```

I think it's important you have that information now because we will have to refresh the tutorial with some bits from the new version when we have it.

docs/quickstart.rst
```diff
@@ -133,59 +217,61 @@ states it has assigned them:
...
    ---------
     1000000
 
 Cause a failover
 ----------------
 
-This plot is too boring, time to introduce a problem. We’ll run the
-primary out of disk space and watch the secondary get promoted.
+This plot is too boring, time to introduce a problem. We’ll turn off the
+primary and watch the secondary get promoted.
```
DimCitus: That's when having the systemd setup done would be very nice...

docs/quickstart.rst
```diff
@@ -133,59 +217,61 @@ states it has assigned them:
...
+   az vm stop \
+     --resource-group ha-demo \
+     --name ha-demo-a
 
-After a number of failed attempts to restart node A, its keeper signals
-that the node is unhealthy and the node is put into the "demoted" state.
-The monitor promotes node B to be the new primary.
+After a number of failed attempts to talk to node A, the monitor determines
+the node is unhealthy and puts it into the "demoted" state.  The monitor
+promotes node B to be the new primary.
```
DimCitus: Do we want to explain the default values for the health checks (they are found in https://pg-auto-failover.readthedocs.io/en/latest/ref/configuration.html#pg-auto-failover-monitor at the moment) and how they work here?

docs/quickstart.rst
```diff
@@ -133,59 +217,61 @@ states it has assigned them:
...
 
 .. code-block:: bash
 
+   TODO
```
DimCitus: We should show the new state of the formation/group, and then reboot the first node and see it rejoin the cluster as a secondary I believe. Also, again, I think some copy/paste from the actual output of the pg_autoctl command would be nice to include here. There are many things that happen behind the scene and that we can introduce to the user, or at least have them realise they don't want to script all of that themselves and would rather just use our tool here.

docs/quickstart.rst
```diff
@@ -61,69 +147,67 @@ instance had already existing in the destination directory, this command
 would have configured it to serve as a monitor.
 
 In our case, ``pg_autoctl create monitor`` creates a database called
-``pg_auto_failover``, installs the ``pgautofailover`` Postgres extension, and grants access
-to a new ``autoctl_node`` user.
+``pg_auto_failover``, installs the ``pgautofailover`` Postgres extension, and
+grants access to a new ``autoctl_node`` user.
```
DimCitus: We are missing the `pg_autoctl monitor run` command, which now outputs some interesting bits of information, and also should be registered on the system startup scripts (systemd integration). See below for more on that topic and the `pg_autoctl show systemd` command. I believe we should include that topic in our tutorial... what do you think?

docs/quickstart.rst
```diff
@@ -8,51 +8,140 @@ up pg_auto_failover to replicate data between them. We’ll simulate failure in
...
+       --generate-ssh-keys
+   done
+
+List and record the public IP addresses that this command outputs for each
+machine:
+
+.. code-block:: bash
+
+   az vm list-ip-addresses --ids \
+       $(az vm list -g ha-demo --query "[].id" -o tsv)
```
DimCitus: Check that one out! 

```
az vm list-ip-addresses --ids  $(az vm list -g ha-demo --query "[].id" -o tsv) --query '[] [] .virtualMachine .{Name: name, PrivateIP: network.privateIpAddresses[0], PublicIP: network.publicIpAddresses[0].ipAddress}' -o table
```

docs/quickstart.rst
```diff
@@ -8,51 +8,140 @@ up pg_auto_failover to replicate data between them. We’ll simulate failure in
...
+also give ourselves address from our remote IP. We'll do this with a network
+security group and a subnet.
+
+.. code-block:: bash
+
+   az network nsg create \
+       --resource-group ha-demo \
+       --name ha-demo-nsg
+
+   export MY_IP=`dig +short myip.opendns.com @resolver1.opendns.com`
```
DimCitus: This failed for me, I used `curl ifconfig.me` which just worked.

docs/quickstart.rst
mtuncer: I liked this better

docs/quickstart.rst
```diff
@@ -8,51 +8,140 @@ up pg_auto_failover to replicate data between them. We’ll simulate failure in
...
+Create virtual network
+----------------------
+
+Our database machines need to talk to each other and to the monitor node, so
+let's create a virtual network.
+
+.. code-block:: bash
+
+   az group create \
+       --name ha-demo \
```
mtuncer: we have ha-demo repeated everywhere, I used following

```bash
export RESOURCE_GROUP=mt_citusha_1
export VNET_NAME=${RESOURCE_GROUP}_vnet
export SUBNET_NAME=${RESOURCE_GROUP}_subnet
export NSG_NAME=${RESOURCE_GROUP}_nsg
export USERNAME=citus
```

This helped me a lot when using a different resource group than `ha-demo`

docs/quickstart.rst
```diff
@@ -8,51 +8,140 @@ up pg_auto_failover to replicate data between them. We’ll simulate failure in
...
-  curl https://install.citusdata.com/community/rpm.sh | sudo bash
+  ############################################
+  # Run this on the monitor, and nodes a and b
+
+  # The az vm create command also creates ~/.ssh/id_rsa if
+  # that key doesn't already exist, so we'll SSH with that
+  #
+  # Note: substitute the actual IP for "public-ip"
+
+  ssh -i ~/.ssh/id_rsa ha-admin@public-ip
```
mtuncer: this did not work for me

docs/quickstart.rst
mtuncer: `$USERNAME` was a poor choice it was mixed up with system username, now switched to `$ADMIN_USER_NAME`

docs/quickstart.rst
DimCitus: I found how to sort the array, here by virtual machine name. It's possible to sort by privateIp but then `.10` is listed before `.1` because of alphabetic rules, so that's not as useful. I didn't play with trying to map the IP string into a number for the sorting expression...

```
az vm list-ip-addresses \
  --ids  $(az vm list -g ha-demo --query "[].id" -o tsv) \
  -o table \
  --query 'sort_by( [] [] 
             .virtualMachine 
             .{Name: name,
               PrivateIP: network.privateIpAddresses[0], 
               PublicIP: network.publicIpAddresses[0].ipAddress
             }, &Name)'

Name             PrivateIP    PublicIP
---------------  -----------  -------------
ha-demo-a        10.0.1.5     13.92.189.9
ha-demo-app      10.0.1.7     40.80.147.143
ha-demo-b        10.0.1.6     40.71.100.103
ha-demo-c0       10.0.1.8     13.90.42.3
ha-demo-c1       10.0.1.9     13.90.41.111
ha-demo-monitor  10.0.1.4     13.92.243.131
ha-demo-w1       10.0.1.10    13.90.40.182
ha-demo-w2       10.0.1.11    13.90.41.200
ha-demo-w3       10.0.1.12    13.68.225.239
ha-demo-w4       10.0.1.13    13.92.80.219
```

docs/quickstart.rst
```diff
@@ -8,51 +8,140 @@ up pg_auto_failover to replicate data between them. We’ll simulate failure in
...
+----------------------
+
+Our database machines need to talk to each other and to the monitor node, so
+let's create a virtual network.
+
+.. code-block:: bash
+
+   az group create \
+       --name ha-demo \
+       --location eastus
```
DimCitus: I've been desultory reading the output of `az vm list -g ha-demo --query '[0]'` and I discovered the notions of availabilitySet and proximityPlacementGroup. After reading the docs about them I think we should create a PPG for our deployment here, and then put the two Postgres VMs (primary, secondary) in a single availability set.

Reading https://docs.microsoft.com/en-gb/azure/virtual-machines/linux/proximity-placement-groups and https://docs.microsoft.com/en-us/cli/azure/vm/availability-set?view=azure-cli-latest#az-vm-availability-set-create it looks like we could do:

```
az ppg create \
   -n ha-demo-ppg \
   -g ha-demo \
   -l eastus \
   -t standard

az vm availability-set create -n ha-demo-aset -g ha-demo --platform-fault-domain-count 2 --platform-update-domain-count 2 --ppg ha-demo-ppg
```

Then we can use `--ppg ha-demo-ppg` in the `az vm create` command for the monitor, and `--availability-set ha-demo-aset --ppg ha-demo-ppg` when creating the next two VMs.

jonels-msft:
I updated some of the commands.

Looks like the systemd service might not be doing `pg_autoctl run`, since node B says:

```
02:09:12 INFO  Transition complete: current state is now "wait_standby"
02:09:22 INFO  Still waiting for the monitor to drive us to state "catchingup"
02:09:22 WARN  Please make sure that the primary node is currently running `pg_autoctl run` and contacting the monitor.```

**Comment by DimCitus** [comment]
> Looks like the systemd service might not be doing `pg_autoctl run`, since node B says:

Well given `systemctl start pgautfailover` it should, maybe this went wrong somehow? One possible mistake we made would be that when running the `pg_autoctl show systemd` command as `ha-admin` then the `WorkingDirectory` might end up being set to `/home/ha-admin` instead of `/var/lib/potsgresql` ; and it might be that the `postgres` user then doesn't have the rights to be there. Also the `User` parameter might be wrong too.

**Comment by jonels-msft** [comment]
Ah yes, I see from journalctl that it's looking for 

```
/var/lib/postgresql/.config/pg_autoctl/home/ha-admin/ha/pg_autoctl.cfg
```

but that file is actually here:

```
/home/ha-admin/.config/pg_autoctl/home/ha-admin/ha/pg_autoctl.cfg
```

File: docs/quickstart.rst [hunk 13]
```diff
@@ -61,182 +166,269 @@ instance had already existing in the destination directory, this command
...
+want to connect to whichever of these servers supports reads *and* writes.
+That will be the primary server.
 
 .. code-block:: bash
 
-   # add data to primary
-   psql -p 6010 \
-     -c 'create table foo as select generate_series(1,1000000) bar;'
+   # connect to database via psql on the app vm
+   ssh -l ha-admin -t `vm_ip app` psql "\"$APP_DB_URI\""
```
jonels-msft, line 313: @DimCitus actually I get this error:

```
psql: FATAL:  no pg_hba.conf entry for host "10.0.1.5", user "ha-admin", database "appdb", SSL off
FATAL:  no pg_hba.conf entry for host "10.0.1.5", user "ha-admin", database "appdb", SSL off
```

(The `SSL off` business is because this was that cluster we were messing with and it wasn't created uniformly with `--ssl-self-signed`)

docs/quickstart.rst
jonels-msft: Yes, I could still add more detail about that.

docs/quickstart.rst
jonels-msft: Still need to do this, after we review the current functionality

jonels-msft:
@DimCitus I updated the rest of the instructions. Ready for another review!

Can you try my instructions from top to bottom? (Modifying the systemd config by hand of course) My cluster was in a Frankenstein state from our modifying it together, so I'm pretty sure you will uncover issues doing it fresh.

docs/quickstart.rst
```diff
@@ -61,182 +166,269 @@ instance had already existing in the destination directory, this command
...
+
+Within psql:
 
-   # query secondary
-   psql -p 6011 -c 'select count(*) from foo;'
-     count
-   ---------
-    1000000
+.. code-block:: psql
```
jonels-msft, line 317: I would have preferred to send this SQL statement in one step using `psql -c`, but bash was grabbing the `-c`. Perhaps there's a way to quote the ssh command better to fix that.

DimCitus: I guess you could alway `echo 'select ...' | ssh ... sql` if nothing else works?

JelteF: You can probably use `ssh <connection options> -- psql -c "select ..."`. Note the `--` it specifies that no options follow for the command, so everything should be interpreted as normal arguments (even if they start with a `-`).

docs/quickstart.rst
```diff
@@ -8,51 +8,156 @@ up pg_auto_failover to replicate data between them. We’ll simulate failure in
...
+   # allow VMs to resolve each other via internal DNS
+   az network private-dns zone create \
+       --resource-group ha-demo \
+       --name ha-demo.local
+   az network private-dns link vnet create \
+       --name ha-demo-zone-link \
+       --resource-group ha-demo \
+       --virtual-network ha-demo-net \
+       --zone-name ha-demo.local \
+       --registration-enabled true
```
DimCitus, line 81: I think this should be another separate box before we create the VMs. It makes tracking where we are at easier, and also makes more sense (to me at least) logically.

docs/quickstart.rst
```diff
@@ -8,51 +8,156 @@ up pg_auto_failover to replicate data between them. We’ll simulate failure in
...
+function to retrieve their IP addresses:
+
+.. code-block:: bash
+
+  # run this in your local shell as well
+
+  vm_ip () {
+    az vm list-ip-addresses -g ha-demo -n ha-demo-$1 -o tsv \
+      --query '[] [] .virtualMachine.network.publicIpAddresses[0].ipAddress'
+  }
```
DimCitus, line 110: I think now would be a good time to pause and show what we've done, the 4 VMs that are running in a network with known names etc.

DimCitus: Maybe something like this would get the point across (what we did so far) for a quick review?

```
~ az resource list --query "[?resourceGroup=='ha-demo-dim'].{ name: name, flavor: kind, resourceType: type, region: location }" --output table
Name                             ResourceType                                           Region
-------------------------------  -----------------------------------------------------  --------
ha-demo-a                        Microsoft.Compute/virtualMachines                      eastus
ha-demo-app                      Microsoft.Compute/virtualMachines                      eastus
ha-demo-b                        Microsoft.Compute/virtualMachines                      eastus
ha-demo-monitor                  Microsoft.Compute/virtualMachines                      eastus
ha-demo-appVMNic                 Microsoft.Network/networkInterfaces                    eastus
ha-demo-aVMNic                   Microsoft.Network/networkInterfaces                    eastus
ha-demo-bVMNic                   Microsoft.Network/networkInterfaces                    eastus
ha-demo-monitorVMNic             Microsoft.Network/networkInterfaces                    eastus
ha-demo-nsg                      Microsoft.Network/networkSecurityGroups                eastus
ha-demo.local                    Microsoft.Network/privateDnsZones                      global
ha-demo.local/ha-demo-zone-link  Microsoft.Network/privateDnsZones/virtualNetworkLinks  global
ha-demo-a-ip                     Microsoft.Network/publicIPAddresses                    eastus
ha-demo-app-ip                   Microsoft.Network/publicIPAddresses                    eastus
ha-demo-b-ip                     Microsoft.Network/publicIPAddresses                    eastus
ha-demo-monitor-ip               Microsoft.Network/publicIPAddresses                    eastus
ha-demo-net                      Microsoft.Network/virtualNetworks                      eastus
```

docs/quickstart.rst
```diff
@@ -8,51 +8,156 @@ up pg_auto_failover to replicate data between them. We’ll simulate failure in
...
 
-   pg_autoctl create monitor   \
-     --pgdata ./monitor     \
-     --pgport 6000
+   ssh -l ha-admin `vm_ip monitor` -- \
+     pg_autoctl create monitor \
+       --auth trust \
+       --ssl-self-signed \
+       --pgdata monitor \
+       --pgctl /usr/lib/postgresql/11/bin/pg_ctl
```
DimCitus, line 160: For now it times out, and I checked and reinstalled the network rules and it seems good...

```
~ ssh -l ha-admin `vm_ip monitor`
ssh: connect to host 137.135.120.211 port 22: Operation timed out
```

docs/quickstart.rst
jonels-msft: @JelteF weirdly, this is what happens:

```sh
ssh -l ha-admin -t `vm_ip app` -- psql "\"$APP_DB_URI\"" -c 'SELECT count(*) FROM foo;'
```
```
bash: -c: line 0: syntax error near unexpected token `('
bash: -c: line 0: `psql "postgres://etcetc" -c SELECT count(*) FROM foo;'
```

jonels-msft:
@DimCitus I updated it again with your suggestions. Can you check it again, and especially see whether you run into the `FATAL:  no pg_hba.conf entry for host "10.0.1.5"` thing?

docs/quickstart.rst
```diff
@@ -61,198 +199,276 @@ instance had already existing in the destination directory, this command
...
+.. code-block:: ini
+
+   # automatically added to node A
+
+   host "appdb" "ha-admin" ha-demo-a.internal.cloudapp.net trust
+   host all "pgautofailover_monitor" ha-demo-monitor.internal.cloudapp.net trust
+   host replication "pgautofailover_replicator" ha-demo-b.internal.cloudapp.net trust
+   host "appdb" "pgautofailover_replicator" ha-demo-b.internal.cloudapp.net trust
+   host replication "pgautofailover_replicator" ha-demo-a.internal.cloudapp.net trust
+   host "appdb" "pgautofailover_replicator" ha-demo-a.internal.cloudapp.net trust
```
DimCitus, line 292: At this point in the tutorial the last two lines have not been added yet, because we didn't failover. The HBA editing is “lazy” and only happens when needed. I think we should mention that too, and encourage our users to edit the HBA anyway they see fit, because it's going to be needed for the application anyways.

docs/quickstart.rst
```diff
@@ -61,198 +199,276 @@ instance had already existing in the destination directory, this command
...
+                              Name |   Port | Group |  Node |     Current State |    Assigned State
+   --------------------------------+--------+-------+-------+-------------------+------------------
+   ha-demo-a.internal.cloudapp.net |   5432 |     0 |     1 |           primary |           primary
+   ha-demo-b.internal.cloudapp.net |   5432 |     0 |     2 |         secondary |         secondary
+
+
+This looks good. We can add data to the primary, and later see it appear in the
+secondary. We'll connect to the database from inside our "app" virtual machine,
+using a connection string obtained from the monitor. First we'll get the
+connection string and store it in a local environment variable:
```
DimCitus, line 329: Maybe we could show the following output to show what we're dealing with, and then set the APP_DB_URI your way so that we can re-use the connection string?

```
~ ssh -l ha-admin `vm_ip monitor` -- pg_autoctl show uri --pgdata monitor
      Type |    Name | Connection String
-----------+---------+-------------------------------
   monitor | monitor | postgres://autoctl_node@ha-demo-monitor.internal.cloudapp.net:5432/pg_auto_failover?sslmode=require
 formation | default | postgres://ha-demo-a.internal.cloudapp.net:5432,ha-demo-b.internal.cloudapp.net:5432/appdb?target_session_attrs=read-write&sslmode=require
```

docs/quickstart.rst
```diff
@@ -61,198 +199,276 @@ instance had already existing in the destination directory, this command
...
 
-   # query secondary
-   psql -p 6011 -c 'select count(*) from foo;'
-     count
-   ---------
-    1000000
+The connection string contains both our nodes, comma separated, and includes
+the url parameter ``?target_session_attrs=read-write`` telling psql that we
+want to connect to whichever of these servers supports reads *and* writes.
+That will be the primary server.
```
DimCitus, line 341: I think it's fair to say that readers need to have seen the URI before reading that paragraph. Otherwise this requires too much of imagination from them?

docs/quickstart.rst
```diff
@@ -61,198 +199,276 @@ instance had already existing in the destination directory, this command
...
+
+.. code-block:: bash
+
+   # connect to database via psql on the app vm and
+   # create a table with a million rows
+   ssh -l ha-admin -t `vm_ip app` -- \
+     psql "\"$APP_DB_URI\"" \
+       -c "\"CREATE TABLE foo AS SELECT generate_series(1,1000000) bar;\""
```
DimCitus, line 349: Does not work for me:

```
$ ssh -l ha-admin -t `vm_ip app` --   psql "'$APP_DB_URI'" -c '"CREATE TABLE foo AS SELECT generate_series(1,1000000) bar;"'
psql: invalid sslmode value: ""
Connection to 20.188.42.21 closed.

$ ssh -l ha-admin `vm_ip monitor` -- pg_autoctl show uri --pgdata monitor --formation default
postgres://ha-demo-a.internal.cloudapp.net:5432,ha-demo-b.internal.cloudapp.net:5432/appdb?target_session_attrs=read-write&sslmode=

$ export APP_DB_URI='postgres://ha-demo-a.internal.cloudapp.net:5432,ha-demo-b.internal.cloudapp.net:5432/appdb?target_session_attrs=read-write&sslmode=require' 

$ ssh -l ha-admin -t `vm_ip app` -- \
>   psql "\"$APP_DB_URI\"" \
>     -c "\"CREATE TABLE foo AS SELECT generate_series(1,1000000) bar;\""
psql: FATAL:  no pg_hba.conf entry for host "10.0.1.5", user "ha-admin", database "appdb", SSL on
Connection to 20.188.42.21 closed.
```

  1. This looks like a bug in pg_autoctl itself, we should fix it and/or make sure it's fixed in 1.3
  2. the HBA has not been edited for the application VM and its IP address, and that's not done by `pg_autoctl` itself. I think now is a good time in this documentation to include how to achieve that: users will need to.

docs/quickstart.rst
```diff
@@ -61,198 +199,276 @@ instance had already existing in the destination directory, this command
...
-   watch pg_autoctl show events --pgdata ./monitor
+   # initiate failover to node B
+   ssh -l ha-admin -t `vm_ip monitor` \
+     pg_autoctl perform switchover --pgdata monitor
 
-In another terminal we’ll consume node A’s disk space and try to restart
-the database. It will refuse to start up.
+   # watch the progress
+   watch -d \
+     ssh -l ha-admin `vm_ip monitor` pg_autoctl show state --pgdata monitor
```
DimCitus, line 366: Better use watch on the remote instance rather than reconnect using ssh every other second, I think:

```
ssh -t -l ha-admin `vm_ip monitor` -- watch -n 1 pg_autoctl show state --pgdata monitor
```

DimCitus: Also this fails for me on node a when catching-up again with the following, which I need to understand if it's a bug in 1.2 (that was a hasty release) or if we still have it, and then we need to fix it of course.

```
pg_basebackup: could not connect to server: FATAL:  no pg_hba.conf entry for replication connection from host "10.0.1.7", user "pgautofailover_replicator", SSL on
```

docs/quickstart.rst
```diff
@@ -61,198 +199,276 @@ instance had already existing in the destination directory, this command
...
-     'postgres://127.0.0.1:6010,127.0.0.1:6011/?target_session_attrs=read-write'
+   .
+                              Name |   Port | Group |  Node |     Current State |    Assigned State
+   --------------------------------+--------+-------+-------+-------------------+------------------
+   ha-demo-a.internal.cloudapp.net |   5432 |     0 |     1 |      wait_primary |      wait_primary
+   ha-demo-b.internal.cloudapp.net |   5432 |     0 |     2 |           demoted |        catchingup
+
+Node A cannot be considered in full "primary" state since there is no secondary
+present, but it can still serve client requests. It is marked as "wait_primary"
+until a secondary appears, to indicate that it's running without a backup.
```
DimCitus, line 424: Without a secondary. I think using the term “backup” here would be rather confusing to the users. A secondary server can not be considered a backup: imagine running a `TRUNCATE` or a `DROP TABLE` on the primary, then by the time you would `switchover` to the other node the command has been replayed already. So let's not use the term “backup” anywhere here.

docs/quickstart.rst
```diff
@@ -61,198 +199,276 @@ instance had already existing in the destination directory, this command
...
 
-   rm /mnt/node_a/bigfile
+   az vm start \
+     --resource-group ha-demo \
+     --name ha-demo-b
 
-Now the next time the keeper retries, it brings the node back. Node A
-goes through the state "catchingup" while it updates its data to match
-B. Once that's done, A becomes a secondary, and B is now a full primary.
+Now the next time the keeper retries its health check, it brings the node back.
```
DimCitus, line 447: The health check is done only from the monitor. The other way around, we call it the “node active” protocol, the keeper asks the monitor “what now?” basically, and the monitor says “oh you've been demoted”.

docs/quickstart.rst
```diff
@@ -61,198 +199,276 @@ instance had already existing in the destination directory, this command
...
+   ssh -l ha-admin `vm_ip a` -- \
+     pg_autoctl create postgres \
+       --pgdata ha \
+       --auth trust \
+       --ssl-self-signed \
+       --username ha-admin \
+       --dbname appdb \
+       --nodename ha-demo-a.internal.cloudapp.net \
+       --pgctl /usr/lib/postgresql/11/bin/pg_ctl \
+       --monitor postgres://autoctl_node@ha-demo-monitor.internal.cloudapp.net/pg_auto_failover?sslmode=require
```
DimCitus, line 221: I think before we do that we should show where the monitor URI comes from.

docs/quickstart.rst
DimCitus: Another way to quote it:

```
ssh -l ha-admin -t `vm_ip a` --   psql "'$APP_DB_URI'"     -c "'CREATE TABLE foo AS SELECT generate_series(1,1000000) bar;'"
```

docs/quickstart.rst
```diff
@@ -8,51 +8,189 @@ up pg_auto_failover to replicate data between them. We’ll simulate failure in
...
+
+   az network private-dns zone create \
+       --resource-group ha-demo \
+       --name ha-demo.local
+   az network private-dns link vnet create \
+       --name ha-demo-zone-link \
+       --resource-group ha-demo \
+       --virtual-network ha-demo-net \
+       --zone-name ha-demo.local \
+       --registration-enabled true
```
DimCitus, line 78: It so happens that the internal DNS `internal.cloudapp.net` is automatic in Azure, and we don't need to add another one. We can just use this one. Also when we add our own domain then Postgres HBA DNS behaviour stops working, because we now have two names for the same IP address and Postgres does not know how to manage that.

I just did the whole tutorial again without that part and it works better. It's an improvement to delete that.

docs/quickstart.rst
DimCitus: So that's because we have two reserve DNS host names for a given IP address and Postgres is confused, it works better without the extra DNS settings.

DimCitus:
I just updated this branch to the current repository (rebased to master), and then went through all of it again. I had to modify several minor things:

  - we are now using the release 1.3.1 (package 1.3), available as of today,
  - as a result we don't need to `sudo -I -u postgres` for the systemd integration
  - using another DNS on-top of the default one provided with Azure causes trouble, removed
  - some other small things

One thing that remains to be done @begriffs is with using the Postgres service from the `app` VM. For that we need to implement a couple things still:

  - create the application roles: one role `appdba` or something that owns the database and is allowed to create a schema etc, and one role `app` or something that is allowed to use the tables then, as the application code would
  - add an entry for those two roles on the `pg_hba.conf` of the servers for the `ha-demo-app` VM (or maybe the whole network?) ; it might be simpler to do it before creating the standby otherwise it needs to be done manually on both servers, though that's also a good point to illustrate (HBA is not kept in sync in between nodes).

Only then the part with connecting from the `app` VM is possible and make sense.

docs/quickstart.rst
jonels-msft: > This looks like a bug in pg_autoctl itself, we should fix it and/or make sure it's fixed in 1.3

I tried with 1.3 today and ran into the same problem. I was able to run the command through node `vm_ip a` but not `vm_ip app`. The latter says:

```
psql: FATAL:  no pg_hba.conf entry for host "10.0.1.4", user "ha-admin", database "appdb", SSL on
```

How should we proceed? Add instructions to modify pg_hba by hand, make a new release of failover to fix the problem, or run the command ssh'd through node A?

DimCitus: Yes, the HBA needs to be manually edited by our users to add in their application nodes. At the moment this needs to be done on both the primary and the secondary. It is possible to do it when there's only a primary, and then `pg_basebackup` will cary the changes to the secondary when creating it.

PR #157, opened 2020-02-06
Fix include line detection.

DimCitus:
The regexp always failed and we ended-up with the same include line at the
beginning of the postgresql.conf file over and over again. Use the same
technique as in the HBA case, calling strstr() instead.


src/bin/pg_autoctl/pgctl.c
```diff
@@ -271,14 +267,15 @@ pg_include_config(const char *configFilePath,
 	}
 
 	/* find the include 'postgresql-auto-failover.conf' line */
-	includeLine = regexp_first_match(currentConfContents, configIncludeRegex);
-	if (includeLine != NULL)
+	includeLine = strstr(currentConfContents, configIncludeLine);
```
mtuncer: this does not account for extra space after `include` would it be acceptable ?

src/bin/pg_autoctl/pgctl.c
```diff
@@ -271,14 +267,15 @@ pg_include_config(const char *configFilePath,
...
+
+	if (includeLine != NULL && (includeLine ==  currentConfContents ||
+								includeLine[-1] == '\n'))
```
mtuncer: includeLine[-1]  is not valid if `includeLine ==  currentConfContents` whiich is also covered here. some comment would be nice

src/bin/pg_autoctl/pgctl.c
DimCitus: I think yes, because we want to find exactly the line we have added to the configuration file from pg_autoctl, which alway has the same amount of space in there.

src/bin/pg_autoctl/pgctl.c
DimCitus: That's a copy/paste from our file `pghba.c` where there was no specific comment, so I though it would be ok...

PR #169, opened 2020-02-24
Implement an option to skip editing of HBA rules by pg_autoctl

### LINKED ISSUE #62: Feature Request: skip management of pg_hba.conf

DimCitus:
Allow users to `--skip-pg-hba` rules editing by pg_autoctl. In that case, we issue WARNing messages with the rules that pg_autoctl refrains from adding, to make it easy enough for our users to review their own provisioning.

Fixes #91, fixes #62.


src/bin/lib/log/src/log.h
```diff
@@ -24,6 +24,8 @@ enum { LOG_TRACE, LOG_DEBUG, LOG_INFO, LOG_WARN, LOG_ERROR, LOG_FATAL };
 #define log_error(...) log_log(LOG_ERROR, __FILE__, __LINE__, __VA_ARGS__)
 #define log_fatal(...) log_log(LOG_FATAL, __FILE__, __LINE__, __VA_ARGS__)
 
+#define log_level(level, ...) log_log(level, __FILE__, __LINE__, __VA_ARGS__)
```
mtuncer: no need for re-declaration, guess it is fine

mtuncer: wrong comment. taking it back

docs/ref/reference.rst
```diff
@@ -525,9 +541,10 @@ keeper and monitor, use the following command::
 
   $ pg_autoctl drop node --help
   pg_autoctl drop node: Drop a node from the pg_auto_failover monitor
-  usage: pg_autoctl drop node  [ --pgdata ]
+  usage: pg_autoctl drop node  [ --pgdata ----destroy ]
```
mtuncer: extra --

src/bin/pg_autoctl/cli_config.c
```diff
@@ -252,6 +246,101 @@ cli_config_check_pgsetup(PostgresSetup *pgSetup)
...
+		/* errors have already been logged */
+		exit(EXIT_CODE_PGSQL);
+	}
+
+	log_info("Connection to local Postgres ok, using \"%s\"", connInfo);
+
+	/*
+	 * Do not check settings on the monitor node itself.
+	 */
+	if (monitor_pguri != NULL)
```
mtuncer: we are checking this twice and never using it if it is null.

can we have an early exit if monitor_pguri is null ?

src/bin/pg_autoctl/ini_file.c
```diff
@@ -287,6 +287,11 @@ ini_option_to_string(IniOption *option, char *dest, size_t size)
 	{
 		case INI_STRING_T:
 		{
+			if (option->strValue == NULL || *(option->strValue) == NULL)
```
mtuncer: not very trivial to understand. we actually mean is null or empty string. Perhaps consider a macro

src/bin/pg_autoctl/keeper_pg_init.c
```diff
@@ -693,7 +695,7 @@ create_database_and_extension(Keeper *keeper)
 								   pgSetup->dbname,
 								   config->nodename,
 								   pg_setup_get_username(pgSetup),
-								   "trust",
+								   pg_setup_get_auth_method(pgSetup),
```
mtuncer: function returns "trust" as default value

src/bin/pg_autoctl/pgctl.c
```diff
@@ -138,23 +138,25 @@ pg_controldata(PostgresSetup *pgSetup, bool missing_ok)
 	}
 	else
 	{
-		if (!missing_ok)
-		{
-			char *errorLines[BUFSIZE];
-			int lineCount = splitLines(prog.stderr, errorLines, BUFSIZE);
-			int lineNumber = 0;
+		char *errorLines[BUFSIZE];
+		int lineCount = splitLines(prog.stderr, errorLines, BUFSIZE);
```
mtuncer: I know it is an earlier code. but splitLines function is not safe. There is no protection against writing  beyond buffer.

src/bin/pg_autoctl/pgctl.c
```diff
@@ -586,18 +588,32 @@ log_program_output(Program prog)
...
 	}
 
 	if (prog.stderr != NULL)
 	{
-		if (prog.returnCode == 0)
-		{
-			log_info("%s", prog.stderr);
-		}
-		else
+		char *errorLines[BUFSIZE];
```
mtuncer: we have way too many repetitions of the same code block. Would it be possible to encapsulate that into function and call just that ?

src/bin/pg_autoctl/pghba.c
```diff
@@ -68,6 +68,17 @@ pghba_ensure_host_rule_exists(const char *hbaFilePath,
...
 	hbaLineEnd += sprintf(hbaLineEnd, " %s", authenticationScheme);
 
+	/*
+	 * When the authentication method is "skip", the option --skip-pg-hba has
+	 * been used. In that case, we still WARN about the HBA rule that we need,
+	 * so that users can review their HBA settings and provisioning.
+	 */
+	if (SKIP_HBA(authenticationScheme))
+	{
+		log_warn("Skipping HBA edits for rule: %s", hbaLine);
```
mtuncer: consider adding  we relied on user provided hba

src/bin/pg_autoctl/pgsetup.h
```diff
@@ -135,7 +135,12 @@ bool pg_setup_is_running(PostgresSetup *pgSetup);
 bool pg_setup_is_primary(PostgresSetup *pgSetup);
 bool pg_setup_is_ready(PostgresSetup *pgSetup, bool pg_is_not_running_is_ok);
 char *pg_setup_get_username(PostgresSetup *pgSetup);
+
+#define SKIP_HBA(authMethod) (strcmp(authMethod, SKIP_HBA_AUTH_METHOD) == 0)
```
mtuncer: consider `strncmp`

src/bin/pg_autoctl/primary_standby.c
```diff
@@ -432,17 +439,27 @@ primary_create_replication_user(LocalPostgresServer *postgres, char *replication
...
 	char hbaFilePath[MAXPGPATH];
-	char *authMethod =  "trust";
+	char *authMethod = pg_setup_get_auth_method(postgresSetup);
 
-	if (replicationPassword)
+	if (replicationPassword == NULL)
 	{
-		authMethod = pg_setup_get_auth_method(postgresSetup);
+		/* most authentication methods require a password */
+		if (strcmp(authMethod, "trust") != 0
```
mtuncer: consider using a macro like HBA_TRUST, so that we won't have hardcoded "trust" in the c code

tests/test_skip_pg_hba.py
```diff
@@ -0,0 +1,204 @@
...
+    cluster.destroy()
+
+def test_000_create_monitor():
+    monitor = cluster.create_monitor("/tmp/skip/monitor",
+                                     authMethod="skip",
+                                     sslMode="verify-ca")
+    monitor.wait_until_pg_is_running()
+
+    with open(os.path.join("/tmp/skip/monitor", "pg_hba.conf"), 'a') as hba:
+        hba.write("hostssl all all 172.27.1.0/24 cert\n")
```
mtuncer: consider defining the network prefix in pautofailover_utils.py

tests/test_skip_pg_hba.py
```diff
@@ -0,0 +1,204 @@
+import pgautofailover_utils as pgautofailover
```
mtuncer: test case naming could be test_cert_usage. Most of the file contents is related to certificate creation and usage.

src/bin/pg_autoctl/keeper_pg_init.c
DimCitus: Indeed. And when the user said `--auth cert` or `--auth scram-sha-256`, that's what we're going to use here when opening the LAN for a whole formation of nodes.

.travis.yml
```diff
@@ -24,8 +24,8 @@ before_install:
   - nuke_pg
   - python --version
   - pyenv versions
-  - pyenv install -s 3.6.9
-  - pyenv global 3.6.9
+  - pyenv install -s 3.7.6
```
mtuncer: can we just use 3.7 ?

docs/ref/reference.rst
DimCitus: Fixed, thanks.

src/bin/pg_autoctl/cli_config.c
DimCitus: Fixed, thanks. I though about that code as two separate units, but you're right that it makes sense to implement an early return, it's now easier to read.

src/bin/pg_autoctl/ini_file.c
DimCitus: Well we have a `char **` so that's two pointers, any of those could be NULL. I added a comment to that effect.

src/bin/pg_autoctl/pgctl.c
DimCitus: I added the needed buffer size protection in splitLines.

src/bin/pg_autoctl/pgctl.c
DimCitus: I have improved our `log_program_output` to make it usable from more places, reducing the code copying.

src/bin/pg_autoctl/pghba.c
DimCitus: Done!

src/bin/pg_autoctl/pgsetup.h
DimCitus: Done.

src/bin/pg_autoctl/primary_standby.c
DimCitus: I don't think we would gain anything that way. Maybe we could have a quick parser for HBA authentication methods and use an enum internally? I didn't do that, as I'm not sure it's necessary to go that far at the moment.

tests/test_skip_pg_hba.py
DimCitus: Done.

tests/test_skip_pg_hba.py
DimCitus: We actually test `--skip-pg-hba` here, which requires making sure a customer can come up with their own full implementation of the authentication needs for pg_auto_failover, from within their own provisioning system. I chose to implement SSL based authentication with the `cert` authentication method, because I think that's the most secure alternative and having an example how to do that might be good for the customers.

.travis.yml
DimCitus: Good point, done.

docs/ref/reference.rst
```diff
@@ -125,10 +133,14 @@ commands are dealing with the monitor:
...
-     If this option is used, please make sure password is manually set on
-     the monitor, and appropriate setting is added to `.pgpass` file on data node.
+     The ``--auth`` option allows setting up authentication method to be
+     used for connections from data nodes with ``autoctl_node`` user. When
+     testing pg_auto_failover for the first time using ``--auth trust``
+     makes things easier. When getting production ready, review your options
+     here and choose at least ``--auth scram-sha-256`` and make sure
+     password is manually set on the monitor, and appropriate setting is
+     added to `.pgpass` file on data node. You could also use some of the
+     advanced Postgres authentication mechanism such as SSL certificates.
```
mtuncer: we should mention if we use either of  --auth or --skip-pg-hba options we have to setup autoctl_node user in the monitor and setup client access mechanism before creating any pg_autoctl node. 

Otherwise whole system might be in an unstable state.  pg_autoctl create succeeds without registering the node to the monitor. following pg_autoctl run command fails because the node is node registered

mtuncer: I can't reproduce the problem locally anymore. But  still believe we mention it

PR #171, opened 2020-02-27
Implement support for SSL connections

DimCitus:
In this PR we implement SSL support for all pg_auto_failover connections.
There are multiple ways to do it, from self-signed certificates to full
blown CA signed certificates with authentication.


docs/security.rst
```diff
@@ -0,0 +1,214 @@
...
+
+When using ``--auth trust`` pg_autoctl adds new HBA rules in the monitor and
+the Postgres nodes to enable connections as seen above.
+  
+Authentication with passwords
+-----------------------------
+
+  - use ``--auth scram-sha-256`` or ``--auth md5``
+  - then ``alter user autoctl_node set password ...``
+  - then set PGPASSWORD or edit ~/.pgpass
```
JelteF:
```suggestion
  - then set the PGPASSWORD evironment variable or edit ~/.pgpass
```

JelteF:
```suggestion
  - then set the PGPASSWORD evironment variable or edit ~/.pgpass
```

src/bin/pg_autoctl/pgsetup.c
```diff
@@ -1122,3 +1134,199 @@ pgsetup_get_pgport()
...
+
+		return true;
+	}
+
+	/* --no-ssl is ok */
+	if (ssl->active == 0)
+	{
+		return true;
+	}
+
```
JelteF: There should be an error here that it is required to specify an ssl flag

docs/security.rst
```diff
@@ -0,0 +1,277 @@
...
+  $ pg_autoctl create postgres --ssl-self-signed ...      \
+                               --auth scram-sha-256 ...   \
+                               ...
+
+In that example we setup SSL connections to encrypt the network traffic, and
+we still have to setup an authentication mechanism exactly as in the
+previous sections of this document. Here ``scram-sha-256`` has been
+selected, and the password will be sent over an encrypted channel.
+
+When using the ``--ssl`` option without providing any certificate,
```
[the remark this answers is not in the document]
DimCitus: Thanks, done.

docs/security.rst
```diff
@@ -0,0 +1,277 @@
...
+Encryption of network communications
+------------------------------------
+
+Postgres knows how to use SSL to enable network encryption of all
+communications, including authentication with passwords and the whole data
+set when streaming replication is used.
+
+To enable SSL on the server an SSL certificate is needed. It could be as
+simple as a self-signed certificate, and ``pg_autoctl`` will create such a
+certificate for you when setup with ssl support and without other
```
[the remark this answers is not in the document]
DimCitus: Done.

docs/security.rst
```diff
@@ -0,0 +1,277 @@
...
+                               ...
+
+In that example we setup SSL connections to encrypt the network traffic, and
+we still have to setup an authentication mechanism exactly as in the
+previous sections of this document. Here ``scram-sha-256`` has been
+selected, and the password will be sent over an encrypted channel.
+
+When using the ``--ssl`` option without providing any certificate,
+``pg_autoctl`` creates a self-signed certificate, as per the Postgres
+documentation at the `Creating Certificates`__ page.
```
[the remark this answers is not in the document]
DimCitus: Added, good point!

src/bin/pg_autoctl/pgsetup.c
```diff
@@ -1122,3 +1134,199 @@ pgsetup_get_pgport()
...
+/*
+ * pgsetup_sslmode_to_string returns the string representation of the enum.
+ */
+char *
+pgsetup_sslmode_to_string(SSLMode sslMode)
+{
+	switch (sslMode)
+	{
+		case SSL_MODE_UNKNOWN:
+			return "unknown ssl mode";
```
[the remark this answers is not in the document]
DimCitus: Changed it to `unknown`.

src/bin/pg_autoctl/pgsetup.c
```diff
@@ -1122,3 +1134,199 @@ pgsetup_get_pgport()
...
+							SSL_MODE_VERIFY_CA,
+							SSL_MODE_VERIFY_FULL };
+
+	char *sslModeArray[] = {"disable", "allow", "prefer", "require",
+							"verify-ca", "verify-full", NULL};
+
+	int sslModeArrayIndex = 0;
+
+	for (sslModeArrayIndex = 0;
+		 sslModeArray[sslModeArrayIndex] != NULL;
```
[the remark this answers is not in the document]
DimCitus: We use NULL terminated arrays in many other places in pg_auto_failover code, so I'd rather leave it like that... it allows to use dynamic array sizes without having to refactor the code if that's needed later...

src/bin/pg_autoctl/pgsetup.c
```diff
@@ -1122,3 +1134,199 @@ pgsetup_get_pgport()
...
+	}
+
+	/*
+	 * When --ssl-self-signed is used, we default to using sslmode=require.
+	 * Setting higher than that are wrong, false sense of security.
+	 */
+	if (ssl->createSelfSignedCert)
+	{
+		/* in that case we want an sslMode of prefer at most */
+		if (ssl->sslMode > SSL_MODE_PREFER)
```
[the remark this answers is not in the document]
DimCitus: I think you're right. Changed.

src/bin/pg_autoctl/cli_create_drop_node.c
```diff
@@ -448,6 +467,81 @@ cli_create_monitor_getopts(int argc, char **argv)
 				break;
 			}
 
+			case 's':
```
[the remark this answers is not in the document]
DimCitus: Unfortunately I don't know how to do that with the libc `getopt` API. We might consider switching to `argp` instead, as documented at https://www.gnu.org/software/libc/manual/html_node/Argp.html#Argp ; but I must admit I didn't have time to look into code sharing with this API.

src/bin/pg_autoctl/monitor_pg_init.c
```diff
@@ -128,6 +133,21 @@ monitor_pg_init(Monitor *monitor, MonitorConfig *config)
 	 */
 	join_path_components(configFilePath, pgSetup.pgdata, "postgresql.conf");
 
+	/*
+	 * When --ssl has been used without SSL certificates being given, now is
```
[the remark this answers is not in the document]
DimCitus: Fixed.

src/bin/pg_autoctl/pgsetup.c
DimCitus: Good spot! It's not where to implement the error catching, because we're lacking the information at this place, so I made it happen in the main argument parsing function... which means in two places, unfortunately.

src/bin/pg_autoctl/pgsetup.c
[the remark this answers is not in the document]
JelteF: Alright if it's done like that in many more places I think it's better to stay consistent indeed. I don't think I understand your other point completely though, using `sizeof` you should also be able to change the array sizes without refactoring afaict.

src/bin/pg_autoctl/cli_common.c
```diff
@@ -309,6 +317,81 @@ cli_create_node_getopts(int argc, char **argv,
 				break;
 			}
 
+			case 's':
+			{
+				/* { "ssl-self-signed", no_argument, NULL, 's' }, */
+				if (sslOrNoSSLCount > 0)
```
mtuncer: we can get into this state with having more than one -s. Error is not really accurate in that way

src/bin/pg_autoctl/cli_common.c
```diff
@@ -309,6 +317,81 @@ cli_create_node_getopts(int argc, char **argv,
...
+						{
+							errors++;
+							log_error(
+								"Using either --no-ssl or --ssl-self-signed "
+								"with user-provided SSL certificates "
+								"is not supported");
+						}
+
+						sslUserCertificates = true;
+						++sslOrNoSSLCount;
```
mtuncer: this expands the meaning of the variable. it becomes basically any ssl setting count

src/bin/pg_autoctl/cli_common.c
```diff
@@ -401,6 +513,79 @@ cli_create_node_getopts(int argc, char **argv,
...
+ * { "server-crt", required_argument, &ssl_flag, SSL_SERVER_CRT_FLAG }
+ * { "server-key", required_argument, &ssl_flag, SSL_SERVER_KEY_FLAG }
+ * { "ssl-mode", required_argument, &ssl_flag, SSL_MODE_FLAG },
+ *
+ * As those options are not using any short option (one-char) variant, they all
+ * fall in the case 0, and we can process them thanks to the global variable
+ * ssl_flag, an int.
+ */
+bool
+cli_getopt_ssl_flags(PostgresSetup *pgSetup)
```
mtuncer: can we have ssl_flag passed as argument instead of being used as global variable ?

mtuncer: and optarg too

src/bin/pg_autoctl/cli_common.c
DimCitus: Good point. I stopped being lazy about it and introduced an enum type which allows for better tracking of the user choice here.

src/bin/pg_autoctl/cli_common.c
DimCitus: Fixed thanks to the new enum for tracking user choice for SSL setup.

src/bin/pg_autoctl/cli_common.c
DimCitus: I could do some of it as suggested, but my C compiler wants to know the size of the ssl_flag pointer at compile time and that means global variable as far as I can see.

src/bin/pg_autoctl/cli_create_drop_node.c
```diff
@@ -245,12 +249,19 @@ cli_create_postgres_getopts(int argc, char **argv)
...
+		{ "ssl-crl-file", required_argument, &ssl_flag, SSL_CRL_FILE_FLAG },
+		{ "server-cert", required_argument, &ssl_flag, SSL_SERVER_CRT_FLAG },
+		{ "server-key", required_argument, &ssl_flag, SSL_SERVER_KEY_FLAG },
 		{ NULL, 0, NULL, 0 }
 	};
 
 	int optind =
 		cli_create_node_getopts(argc, argv,
-								long_options, "C:D:H:p:l:U:A:Sd:n:f:m:MRVvqhx",
+								long_options, "C:D:H:p:l:U:A:Sd:n:f:m:MRVvqhxs",
```
mtuncer: option `N` is missing

src/bin/pg_autoctl/cli_create_drop_node.c
```diff
@@ -326,7 +344,7 @@ cli_create_monitor_getopts(int argc, char **argv)
 
 	optind = 0;
 
-	while ((c = getopt_long(argc, argv, "C:D:p:n:l:A:SVvqhx",
+	while ((c = getopt_long(argc, argv, "C:D:p:n:l:A:SVvqhxs",
```
mtuncer: option `N` is missing

src/bin/pg_autoctl/pgctl.c
```diff
@@ -1224,3 +1380,109 @@ pg_is_running(const char *pg_ctl, const char *pgdata)
...
+ * https://www.postgresql.org/docs/current/ssl-tcp.html#SSL-CERTIFICATE-CREATION
+ *
+ * openssl req -new -x509 -days 365 -nodes -text -out server.crt \
+ *             -keyout server.key -subj "/CN=dbhost.yourdomain.com"
+ */
+bool
+pg_create_self_signed_cert(PostgresSetup *pgSetup, const char *nodename)
+{
+	Program program;
+	char subject[BUFSIZE] = { 0 };
```
mtuncer: we would have to change this to PQExpBufffer after the merge. Can we do it here instead ?
this way we can get rid of snprintfs too

src/bin/pg_autoctl/pghba.c
```diff
@@ -53,7 +54,15 @@ pghba_ensure_host_rule_exists(const char *hbaFilePath,
 	char *includeLine = NULL;
 	PQExpBuffer newHbaContents = NULL;
 
-	hbaLineEnd += sprintf(hbaLineEnd, "host ");
+	if (ssl)
+	{
+		hbaLineEnd += sprintf(hbaLineEnd, "hostssl ");
```
mtuncer: whole string usage is being changed in another PR.

src/bin/pg_autoctl/cli_create_drop_node.c
DimCitus: Thanks, added back!

src/bin/pg_autoctl/cli_create_drop_node.c
DimCitus: Thanks, added.

src/bin/pg_autoctl/pghba.c
DimCitus: Yeah but we don't know yet which PR is making it first, and it might be the SSL one anyway.

DimCitus:
> `N` option was not evaluated at all. Do we test this code path ?

All the tests are using `--no-ssl` except for the two new SSL related one, `test_ssl_cert.py` and `test_ssl_self_signed.py`. So yeah the code path has been tested...

PR #172, opened 2020-02-27
{WIP} make string and buffer usages safer

mtuncer:
We are replacing unsafe usages of string functions and memory operations with safer ones.


src/bin/pg_autoctl/cli_common.c
```diff
@@ -197,8 +197,8 @@ cli_create_node_getopts(int argc, char **argv,
 			case 'g':
 			{
 				/* { "group", required_argument, NULL, 'g' } */
-				int scanResult = sscanf(optarg, "%d", &LocalOptionConfig.groupId);
-				if (scanResult == 0)
+				LocalOptionConfig.groupId = strtol(optarg, NULL, 10);
+				if (errno != 0)
```
JelteF: afaict this error will not be hit in the following cases:
```
--group ''
--group '123somethingelse'
```

This is why I used these function in citus: https://github.com/citusdata/citus/pull/3602

I think it makes sense to have a similar function in this codebase, so small mistakes like this don't happen:
```c
bool StringToInt(const char *str, int *result)
```

mtuncer: I did not change the sscanf behavior there. We seem to print different error messages depending on the context

agree on our own StringToInt function.

src/bin/pg_autoctl/main.c
```diff
@@ -13,6 +13,7 @@
 #include "postgres_fe.h"
 
 #include "cli_root.h"
+#include "env_utils.h"
```
JelteF: Seems this is not included in this PR

mtuncer: added

JelteF:
Replaced by a lot of separate PRs

PR #174, opened 2020-02-28
Auto format all files with citus_indent

JelteF:
@DimCitus can you check if this also works for you? I only had to add some
lines to .gitattributes to make it work for me.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,17 +1,21 @@
 ## Contributing
 
-This project welcomes contributions and suggestions. Most contributions require you to
-agree to a Contributor License Agreement (CLA) declaring that you have the right to,
-and actually do, grant us the rights to use your contribution. For details, visit
-https://cla.microsoft.com.
+This project welcomes contributions and suggestions. Most contributions
+require you to agree to a Contributor License Agreement (CLA) declaring that
+you have the right to, and actually do, grant us the rights to use your
+contribution. For details, visit https://cla.microsoft.com.
 
-When you submit a pull request, a CLA-bot will automatically determine whether you need
-to provide a CLA and decorate the PR appropriately (e.g., label, comment). Simply follow the
-instructions provided by the bot. You will only need to do this once across all repositories using our CLA.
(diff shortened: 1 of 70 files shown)


DimCitus:
Can you add a Makefile target to cover the indenting process with out tooling, and a note in either our README or maybe in a new CONTRIBUTING file that explains how to setup the tool chain here?

I think that's needed as part of a PR that re-indents the code-base automatically, and to make it possible to maintain the indenting long-term.

JelteF:
Updated CONTRIBUTING and Makefile. CI seems to be broken because of some weird python issue.

CONTRIBUTING.md
```diff
@@ -0,0 +1,50 @@
...
+cd ../..
+
+git clone https://github.com/citusdata/tools.git
+cd tools/uncrustify
+make install
+```
+
+After installing like this you can run the following before committing:
+```bash
+citus_indent
```
DimCitus: Also we should instruct the user to get back in their directory where they have the pg_auto_failover sources to run `citus_indent`, as I suppose that's where it should run. Not in the `tools/uncrustify` directory, right?

CONTRIBUTING.md
```diff
@@ -0,0 +1,50 @@
...
+```
+
+You can also run the following to automatically format all the files that you
+have changed before committing.
+```bash
+echo '#!/bin/bash' > .git/hooks/pre-commit
+echo 'citus_indent --check --diff || { citus_indent --diff; exit 1; }' >> .git/hooks/pre-commit
```
DimCitus: Let's use a here doc instead?

```bash
cat > .git/hooks/pre-commit << __EOF__
#!/bin/bash
citus_indent --check --diff || { citus_indent --diff; exit 1; }
__EOF__
chmod +x .git/hooks/pre-commit
```

DimCitus:
> Updated CONTRIBUTING and Makefile. CI seems to be broken because of some weird python issue.

Thanks! Yeah Travis peen does not have 3.6.9 anymore, so we're switching to Python 3.7.6 on Travis now. I have had to update several other PRs that way already.

PR #175, opened 2020-02-28
Ignore compile_commands.json

JelteF:
This file is generated by some build tools.


DimCitus:
If that file is the result of a build tool that is more “personal” rather than part of the included build system for the project, then I believe the right place for this ignore rule would be `.git/info/exclude` rather than `.gitignore`. What do you think?

JelteF:
Makes sense I guess, added it to my global git ignore.

PR #176, opened 2020-02-28
Specify only major python version in travis

mtuncer:
Travis sometimes change available python minor version. This is the protect against that


DimCitus:
In the `--skip-pg-hba` pull request we now depend on Python 3.7 in the unit tests, because I've been using `subprocess.run` with the `text=True` option.

mtuncer:
closing the PR then. We should probably use 3.7 instead

PR #179, opened 2020-03-06
Implement pg_autoctl drop monitor and drop node --nodename --nodeport.

DimCitus:
This allows removing a node from the monitor when it's been declared dead
already.

### DIFF
--- a/src/bin/pg_autoctl/cli_create_drop_node.c
+++ b/src/bin/pg_autoctl/cli_create_drop_node.c
@@ -107,11 +107,11 @@ CommandLine drop_monitor_command =
 CommandLine drop_node_command =
 	make_command("node",
 				 "Drop a node from the pg_auto_failover monitor",
-				 "[ --pgdata --destroy --nodename --nodeport ]",
+				 "[ --pgdata --destroy --nodename --pgport ]",
 				 "  --pgdata      path to data directory\n"
 				 "  --destroy     also destroy Postgres database\n"
 				 "  --nodename    nodename to remove from the monitor\n"
-				 "  --nodeport    Postgres port of the node to remove",
+				 "  --pgport      Postgres port of the node to remove",
 				 cli_drop_node_getopts,
 				 cli_drop_node);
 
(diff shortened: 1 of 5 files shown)


src/bin/pg_autoctl/cli_create_drop_node.c
```diff
@@ -792,6 +806,26 @@ cli_drop_node_getopts(int argc, char **argv)
 				break;
 			}
 
+			case 'n':
```
mtuncer: user should not be able to specidy --pgdata when --nodename or --nodeport is specified, also we should probably require --nodeport if --nodename is specified

src/bin/pg_autoctl/cli_create_drop_node.c
```diff
@@ -764,6 +776,8 @@ cli_drop_node_getopts(int argc, char **argv)
 	static struct option long_options[] = {
 		{ "pgdata", required_argument, NULL, 'D' },
 		{ "destroy", no_argument, NULL, 'd' },
+		{ "nodename", required_argument, NULL, 'n' },
+		{ "nodeport", required_argument, NULL, 'p' },
```
mtuncer: we used --pgport when creating nodes. We should stick to that naming scheme.

src/bin/pg_autoctl/cli_create_drop_node.c
```diff
@@ -865,179 +908,202 @@ cli_drop_node_getopts(int argc, char **argv)
...
 			}
-		}
-	}
 
-	/* only keeper_remove when we still have a state file around */
-	if (file_exists(config.pathnames.state))
-	{
-		bool ignoreMonitorErrors = true;
+			/* expose the pgSetup in the given KeeperConfig */
+			memcpy(&(config.pgSetup),
```
mtuncer: what about `config.pgSetup = mconfig.pgSetup`

src/bin/pg_autoctl/cli_create_drop_node.c
DimCitus: We still need the PGDATA of the local monitor node, when dropping a distant node. So PGDATA is either the one of the node that is being removed or destroyed, or the PGDATA of the monitor where we want to call `select pgautofailover.remove_node(nodename, pgport)`.

src/bin/pg_autoctl/cli_create_drop_node.c
DimCitus: Good idea, done that way now.

src/bin/pg_autoctl/cli_create_drop_node.c
DimCitus: Done, thanks!

PR #180, opened 2020-03-09
Update show uri install instructions with pgdata

JelteF:
Otherwise I get the following error:
```
16:19:47 FATAL Failed to get PGDATA either from the environment or from --pgdata
```


DimCitus:
See #155 where it should be fixed.

PR #182, opened 2020-03-11
Use our own secure version of fprintf.

DimCitus:
There are several security issues with fprintf that we want to avoid:
  - the FILE * stream might be NULL,
  - the format string might be NULL,
  - any %s argument might be NULL.

What happens in those cases is undefined, so we'd rather make sure those
cases are not happening. This PR introduces pgaf_fprintf() where we add the
first two checks, and then import pg_fprintf() from Postgres port directory
to fix the third issue.


DimCitus:
> Let's rename all the `pg_xxx` functions to `pg12_xxx`, so they don't cause linking conflicts with functions defined in postgres its `port.h`, which is included by `postgres_fe.h`.

Instead, from our discussion offline, I'm now using the USE_REPL_SNPRINTF symbol to decide if we should include our local copy or if we can rely on Postgres to provide it for us in libpq.

> One thing that I think also makes sense (but could also be done in a separate PR) is change the final `strerror` usages where we don't do `strerror(errrno)` but `strerror(somevar)` to:

Done. I was reluctant, but I don't think we need to save errno away in our single use case left.

src/bin/lib/pg/snprintf.c
```diff
@@ -31,6 +31,8 @@
  * src/port/snprintf.c
  */
 
+#ifndef USE_REPL_SNPRINTF
```
JelteF: This needs to be done after including the postgres header that sets this, either `port.h` (PG12) or `confdefs.h` (PG11/PG10). It could be these are both included in `postgres_fe.h` and/or `c.h`, so doing this check after including those might be easier. The way this check is done now I would expect it to always be true. Same for the cases below. 

One way to confirm that the check is working is to put the `#error "should not be reached"` in it and confirm that it still compiles on PG12.

DimCitus: Thanks for the careful review, done that way now. I can confirm the `#error` verification on a manual build test here.

PR #183, opened 2020-03-11
Fix issues found by static analysis

JelteF:

PR #184, opened 2020-03-11
Second batch of static analysis fixes

JelteF:
This fixes some more issues found by static analysis tools. They are separate
from the fixes in #183. Since that PR can almost be merged these are done in a
new PR.

### DIFF
--- a/src/bin/pg_autoctl/cli_common.c
+++ b/src/bin/pg_autoctl/cli_common.c
@@ -494,12 +494,6 @@ cli_create_node_getopts(int argc, char **argv,
 		exit(EXIT_CODE_BAD_ARGS);
 	}
 
-	if (errors > 0)
-	{
-		commandline_help(stderr);
-		exit(EXIT_CODE_BAD_ARGS);
-	}
-
 	/* publish our option parsing now */
 	memcpy(options, &LocalOptionConfig, sizeof(KeeperConfig));
 
--- a/src/bin/pg_autoctl/pgsql.c
+++ b/src/bin/pg_autoctl/pgsql.c
@@ -852,6 +852,7 @@ pgsql_get_current_setting(PGSQL *pgsql, char *settingName, char **currentValue)
 	if (!pgsql_execute_with_params(pgsql, sql, paramCount, paramTypes, paramValues,
 								   &context, &parseSingleValueResult))
 	{
+		/* errors have already been logged */
 		return false;
 	}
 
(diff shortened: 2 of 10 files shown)


src/bin/pg_autoctl/cli_common.c
```diff
@@ -494,12 +494,6 @@ cli_create_node_getopts(int argc, char **argv,
 		exit(EXIT_CODE_BAD_ARGS);
 	}
 
-	if (errors > 0)
-	{
-		commandline_help(stderr);
-		exit(EXIT_CODE_BAD_ARGS);
-	}
-
```
DimCitus: That comes as a surprise, at first re-read of the code I would say we still need this...

src/bin/pg_autoctl/pgsql.c
```diff
@@ -1178,8 +1178,11 @@ pgsql_has_replica(PGSQL *pgsql, char *userName, bool *hasReplica)
 	context.resultType = PGSQL_RESULT_BOOL;
 	context.parsedOk = false;
 
-	pgsql_execute_with_params(pgsql, sql, paramCount, paramTypes, paramValues,
-							  &context, &parseSingleValueResult);
+	if (!pgsql_execute_with_params(pgsql, sql, paramCount, paramTypes, paramValues,
+								   &context, &parseSingleValueResult))
+	{
+		return false;
```
DimCitus: What about logs? at least a comment saying that `/* errors have already been logged */` after verifying that this is indeed the case?

src/bin/pg_autoctl/pgsql.c
```diff
@@ -1416,8 +1419,11 @@ pgsql_get_postgres_metadata(PGSQL *pgsql, const char *slotName,
 	const char *paramValues[1] = { slotName };
 	int paramCount = 1;
 
-	pgsql_execute_with_params(pgsql, sql, paramCount, paramTypes, paramValues,
-							  &context, &parsePgMetadata);
+	if (!pgsql_execute_with_params(pgsql, sql, paramCount, paramTypes, paramValues,
+								   &context, &parsePgMetadata))
+	{
+		return false;
```
DimCitus: Same thing with logs.

src/bin/pg_autoctl/cli_common.c
JelteF: This exact check is already done in line 398 and after that errors is not incremented in this function anymore.

PR #185, opened 2020-03-12
Replace getenv calls with safer ones

mtuncer:
introduces a new function get_env_variable and replaces each getenv call with this one


src/bin/pg_autoctl/env_utils.c
```diff
@@ -0,0 +1,69 @@
...
+	 */
+	envvalue = getenv(name); /* IGNORE-BANNED */
+
+	if (envvalue == NULL)
+	{
+		return -1;
+	}
+
+	valueLength = (int) strlen(envvalue);
+	if (valueLength == 0)
```
JelteF: I don't think this needs to be a special case. It seems to actually cause a bug in this case, because now result won't necessarily contain a '\0' char. Removing this check should fix that.

src/bin/pg_autoctl/env_utils.c
```diff
@@ -0,0 +1,69 @@
...
+#include "log.h"
+
+/*
+ * get_env_variable checks for the environment variable and copies
+ * its value into provided buffer if the buffer is not null.
+ *
+ * Function returns the length of the environment variable.
+ *
+ * Return values
+ * - (-1) if variable is not found
```
JelteF: I think it makes sense to add another return code here:
-2 Error while getting environment variable

That way this function can actually check if the environment variable fits into the buffer, and fail otherwise. It's always nice to take away responsibility from the caller as much as possible. (-2 can also be returned in the error check at the start of the function)

I think it's also useful for readability to give these values a name using `#define`. e.g.:
```c
#define ENV_ERROR -2
#define ENV_NOT_FOUND -1
#define ENV_EMPTY 0
```

src/bin/pg_autoctl/cli_common.c
```diff
@@ -406,16 +407,13 @@ cli_create_node_getopts(int argc, char **argv,
 	 */
 	if (IS_EMPTY_STRING_BUFFER(LocalOptionConfig.pgSetup.pgdata))
 	{
-		char *pgdata = getenv("PGDATA");
-
-		if (pgdata == NULL)
+		int pgdatalen = get_env_variable("PGDATA", LocalOptionConfig.pgSetup.pgdata, MAXPGPATH);
+		if (pgdatalen <= 0 || pgdatalen >= MAXPGPATH)
```
JelteF: This is a behavioral change, since `PGDATA=""` will now fail (maybe this is fine)

src/bin/pg_autoctl/config.c
```diff
@@ -34,10 +35,13 @@ build_xdg_path(char *dst,
...
-	char *home = getenv("HOME");
-	char *xdg_topdir;
+	char home[MAXPGPATH];
+	int homePathLength = 0;
+	char xdg_topdir[MAXPGPATH];
+	int xdg_topdir_length = 0;
 
-	if (home == NULL)
+	homePathLength = get_env_variable("HOME", home, MAXPGPATH);
+	if (homePathLength <= 0 || homePathLength >= MAXPGPATH)
```
JelteF: This is a behavioral change, since `HOME=""` will now fail (maybe this is fine)

src/bin/pg_autoctl/env_utils.c
DimCitus: If we need to do that, then we should return a boolean and handle the variable content and size separately, and then maybe a copy of the environment variable name.

``` c
typedef struct EnvVar
{
    char name[BUFSIZE];
    char value[BUFSIZE];
    size_t valueSize;
} EnvVar;

...

    EnvVar pg_regress_sock_dir = { "PG_REGRESS_SOCK_DIR", { 0 }, 0 };
    
    if (!getEnvVar(EnvVar))
    {
        return false;
    }    
```

mtuncer: added but not sure if it is helpful.  anything < 0 is error case.

src/bin/pg_autoctl/env_utils.c
mtuncer: good catch, thanks

src/bin/pg_autoctl/cli_common.c
```diff
@@ -406,16 +407,13 @@ cli_create_node_getopts(int argc, char **argv,
 	 */
 	if (IS_EMPTY_STRING_BUFFER(LocalOptionConfig.pgSetup.pgdata))
 	{
-		char *pgdata = getenv("PGDATA");
-
-		if (pgdata == NULL)
+		int pgdatalen = get_env_variable("PGDATA", LocalOptionConfig.pgSetup.pgdata, MAXPGPATH);
```
JelteF: We fetch PG_DATA a lot of times, all with this same error. I think it makes sense to put this in a function instead, e.g. `get_env_PGDATA_or_fatal(char** result)`

src/bin/pg_autoctl/env_utils.c
```diff
@@ -0,0 +1,63 @@
...
+	{
+		return GETENV_NOT_FOUND;
+	}
+
+	if (result == NULL)
+	{
+		return strlen(envvalue);;
+	}
+
+	return strlcpy(result, envvalue, maxLength);
```
JelteF: Still missing the check here if everything could be copied, else log_error and return -2

src/bin/pg_autoctl/ipaddr.c
```diff
@@ -375,8 +376,7 @@ fetchIPAddressFromInterfaceList(char *localIpAddress, int size)
 static bool
 isTestEnv(void)
 {
-	const char *socketDir = getenv("PG_REGRESS_SOCK_DIR");
-	return socketDir != NULL && socketDir[0] == '\0';
+	return get_env_variable("PG_REGRESS_SOCK_DIR", NULL, 0) == 0;
```
JelteF: Could maybe be made a function too (is done 3 times), e.g. `bool unix_sockets_enabled()`

src/bin/pg_autoctl/pgctl.c
```diff
@@ -1419,10 +1438,17 @@ pg_create_self_signed_cert(PostgresSetup *pgSetup, const char *nodename)
...
+	int n = 0;
 
 	char openssl[MAXPGPATH];
 	char **opensslSearchList = NULL;
-	int n = search_pathlist(getenv("PATH"), "openssl", &opensslSearchList);
+
+	if (get_env_variable("PATH", envPath, MAXPATHSIZE) > 0)
+	{
+		n = search_pathlist(envPath, "openssl", &opensslSearchList);
+	}
```
JelteF: This is done often. I think it makes sense to make this a function, e.g. 
```c
int search_PATH_for_command(char *command, char **path_list)

File: src/bin/pg_autoctl/pgsetup.c [hunk 9]
```diff
@@ -160,12 +155,7 @@ pg_setup_init(PostgresSetup *pgSetup,
 	}
 	else
 	{
-		char *pguser = getenv("PGUSER");
-
-		if (pguser)
-		{
-			strlcpy(pgSetup->username, pguser, NAMEDATALEN);
-		}
+		(void) get_env_variable("PGUSER", pgSetup->username, NAMEDATALEN);
```
JelteF: Why is there no check for errors here? If that's on purpose, let's add a comment.

src/bin/pg_autoctl/monitor.c
```diff
@@ -1727,15 +1728,17 @@ monitor_ensure_extension_version(Monitor *monitor,
 								 MonitorExtensionVersion *version)
 {
 	const char *extensionVersion = PG_AUTOCTL_EXTENSION_VERSION;
+	char envExtensionVersion[MAXPGPATH];
 
-	/* in test environement, we can export any target version we want */
-	if (getenv(PG_AUTOCTL_DEBUG) != NULL)
+	/* in test environment, we can export any target version we want */
+	if (get_env_variable(PG_AUTOCTL_DEBUG, NULL, 0) >= 0L)
```
JelteF: It might be useful to have a function that just checks if the env is set or not, e.g:
```c
bool is_env_variable_set(char *name)
```

Seems like a common case.

src/bin/pg_autoctl/env_utils.c
```diff
@@ -0,0 +1,63 @@
...
+	envvalue = getenv(name); /* IGNORE-BANNED */
+
+	if (envvalue == NULL)
+	{
+		return GETENV_NOT_FOUND;
+	}
+
+	if (result == NULL)
+	{
+		return strlen(envvalue);;
```
JelteF:
```suggestion
		return strlen(envvalue);
```

src/bin/pg_autoctl/cli_common.c
DimCitus: If we create a function such as `getPGDATA(char *pgdata, int size, bool error_if_not_found)` we should then probably test that we can actually use the pgdata that was found. Empty value does not make sense for pgdata, we need to be able to `mkdir -p` the directory, but also we need the directory to NOT exists yet when we do `initdb`. We could remove the directory before `initdb`, that said.

src/bin/pg_autoctl/cli_common.c
DimCitus: Maybe something like `getPGDATA(char *pgdata, int size, bool error_if_not_found)`.

src/bin/pg_autoctl/env_utils.c
```diff
@@ -0,0 +1,63 @@
...
+ * get_env_variable checks for the environment variable and copies
+ * its value into provided buffer if the buffer is not null.
+ *
+ * Function returns the length of the environment variable.
+ *
+ * Return values
+ * - GETENV_ERROR (-2)  : incorrect variable name is provided
+ * - GETENV_NOT_FOUND (-1) : variable is not found
+ * - GETENV_EMPTY (0)  : variable is found but no value is set
+ * - (1+) length of the environment variable
```
DimCitus: That's too complex. Let's simplify the API with returning a bool and storing the other bits safely in a struct, like in the other comment I made about that.

src/bin/pg_autoctl/monitor.c
DimCitus: The EnvVar struct could have a `bool found` slot to keep the information, and this function would then return the value from the slot when it's been initialised. So another slot `bool initialised` would be good too I guess.

src/bin/pg_autoctl/cli_common.c
JelteF: I think it makes sense to do these checks, but more in a separate PR. Always nice to keep refactor changes and behavioural changes mostly separate.

src/bin/pg_autoctl/config.c
mtuncer: home is used later. it better be not empty

src/bin/pg_autoctl/cli_common.c
mtuncer: get_env_pgdata, however without the error message

src/bin/pg_autoctl/env_utils.c
mtuncer: added

src/bin/pg_autoctl/pgctl.c
```diff
@@ -776,6 +781,9 @@ pg_ctl_start(const char *pg_ctl,
 	char *args[12];
 	int argsIndex = 0;
 
+	char env_pg_regress_sock_dir[MAXPGPATH];
+	int env_pg_regress_sock_dir_len = 0;
```
DimCitus: Do we use that?

src/bin/pg_autoctl/pgsetup.c
```diff
@@ -641,33 +630,37 @@ bool
...
+	}
 	/*
 	 * When PG_REGRESS_SOCK_DIR is set and empty, we force the connection
 	 * string to use "localhost" (TCP/IP hostname for IP 127.0.0.1 or ::1,
 	 * usually), even when the configuration setup is using a unix directory
 	 * setting.
 	 */
-	if (pg_regress_sock_dir != NULL
-		&& strcmp(pg_regress_sock_dir, "") == 0
+	if (pg_regress_sock_dir_exists && strlen(pg_regress_sock_dir) == 0
```
DimCitus: Can we maybe use `env_empty()` here?

src/bin/pg_autoctl/cli_common.c
```diff
@@ -809,7 +792,7 @@ void
 set_first_pgctl(PostgresSetup *pgSetup)
 {
 	char **pg_ctls = NULL;
-	int n = search_pathlist(getenv("PATH"), "pg_ctl", &pg_ctls);
+	int n = search_path("pg_ctl", &pg_ctls);
```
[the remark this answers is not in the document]
JelteF: Multiple entries are handled differently everywhere:
- Error out that there are multiple `pg_ctl` 
- Take the first one

Since grabbing the first one is very simple, i.e. `results[0]` I don't think there's much advantage to to this.

src/bin/pg_autoctl/file_utils.c
```diff
@@ -469,26 +470,30 @@ path_in_same_directory(const char *basePath, const char *fileName,
 
 
 /*
- * Searches all the directories in the ':' separated pathlist for
+ * Searches all the directories in the PATH environment variable for
```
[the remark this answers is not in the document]
JelteF: done

src/bin/pg_autoctl/cli_common.c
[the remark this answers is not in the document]
JelteF: Actually I did this anyway for the last case. But I kept the old function as well for more advanced handling.

src/bin/pg_autoctl/file_utils.c
```diff
@@ -469,26 +470,51 @@ path_in_same_directory(const char *basePath, const char *fileName,
...
 
 /*
- * Searches all the directories in the ':' separated pathlist for
+ * search_path_one copies the first entry found in PATH to result. result
+ * should be a buffer of (at least) MAXPGPATH size.
+ * The function returns false and logs an error when it cannot find the command
+ * in PATH.
+ */
+bool
+search_path_one(const char *filename, char *result)
```
DimCitus: I think the name of the function should be `search_path_first` here. I though we would need `search_path_single` too with an error case for more than one item found. We might need both and `search_path_one` is confusing? does it mean `first` or `single`?

src/bin/pg_autoctl/env_utils.c
```diff
@@ -0,0 +1,163 @@
...
+#include "log.h"
+
+
+/*
+ * env_empty returns true if the passed environment variable is the empty
+ * string. It returns false when the environment variable is not set or if it
+ * set but is something else than the empty string.
+ */
+bool
+env_empty(const char *name)
```
DimCitus: Should the name be `env_found_empty` or something that conveys that the variable is SET to the known value _empty string_ in the environment? when reading the call sites it's not as obvious as I think it would be with the current name.

src/bin/pg_autoctl/pgctl.c
```diff
@@ -173,7 +174,11 @@ int
 config_find_pg_ctl(PostgresSetup *pgSetup)
 {
 	char **pg_ctls = NULL;
-	int n = search_pathlist(getenv("PATH"), "pg_ctl", &pg_ctls);
+	/*
+	 * path variable could be arbitrarily large. We put an upper limit on 10K here.
+	 * In this case we only consider the part we can read
+	 */
```
DimCitus: Does this comment still apply?

src/bin/pg_autoctl/pgctl.c
```diff
@@ -797,12 +804,15 @@ pg_ctl_start(const char *pg_ctl,
 		args[argsIndex++] = (char *) listen_addresses_option;
 	}
 
-	if (getenv("PG_REGRESS_SOCK_DIR") != NULL)
-	{
+	if (env_exists("PG_REGRESS_SOCK_DIR")) {
+		if (!get_env_copy("PG_REGRESS_SOCK_DIR", env_pg_regress_sock_dir, MAXPGPATH)) {
```
DimCitus: Style issue: brackets on their own lines.

src/bin/pg_autoctl/env_utils.c
```diff
@@ -0,0 +1,163 @@
...
+	 * getenv is safe here because we never provide null argument,
+	 * and copy out the result immediately.
+	 */
+	envvalue = getenv(name); /* IGNORE-BANNED */
+	if (envvalue == NULL)
+	{
+		envvalue = fallback;
+		if (envvalue == NULL)
+		{
+			log_error("Failed to get %s environment setting. It was not set", name);
```
DimCitus: I think we can improve this error message to “Failed to get value for environment variable \"%s\", which is  unset”.

src/bin/pg_autoctl/pgsetup.c
```diff
@@ -160,11 +156,10 @@ pg_setup_init(PostgresSetup *pgSetup,
 	}
 	else
 	{
-		char *pguser = getenv("PGUSER");
-
-		if (pguser)
+		if (!get_env_copy_with_fallback("PGUSER", pgSetup->username, NAMEDATALEN, ""))
```
DimCitus: I think we need a comment to explain why it's okay to use "" as the fallback here, rather than NULL and have an error.

src/bin/pg_autoctl/pgsetup.c
```diff
@@ -641,33 +630,37 @@ bool
...
-	char *pg_regress_sock_dir = getenv("PG_REGRESS_SOCK_DIR");
+	char pg_regress_sock_dir[MAXPGPATH];
 	char *connStringEnd = connectionString;
+	bool pg_regress_sock_dir_exists = env_exists("PG_REGRESS_SOCK_DIR");
 
 	connStringEnd += sprintf(connStringEnd, "port=%d dbname=%s",
 							 pgSetup->pgport, pgSetup->dbname);
 
+	if (pg_regress_sock_dir_exists &&
+			!get_env_copy("PG_REGRESS_SOCK_DIR", pg_regress_sock_dir, MAXPGPATH)) {
```
DimCitus: Style issue: brackets on their own lines, and usually && at the beginning of the line makes it easier to read the code.

src/bin/pg_autoctl/pgsetup.c
```diff
@@ -867,7 +860,7 @@ pg_setup_get_username(PostgresSetup *pgSetup)
 {
 	uid_t uid;
 	struct passwd *pw;
-	char *userEnv;
+	char userEnv[NAMEDATALEN];
```
DimCitus: Now we have a double copy of the userEnv name, local buffer and then strdup() to the caller. It it worth changing the API to make a single copy somehow?

JelteF: I think it's fine for now. I never like changing an API for only a single usecase, because you don't know what a good abstraction level is at that point. If this happens more often in the future we can add a helper then.

PR #186, opened 2020-03-12
Implement sformat, a secured version of snprintf.

DimCitus:
Replace all our calls to snprintf and sprintf with sformat, which is based on Postgres
pg_snprintf code and adds a couple of checks against NULL strings. Some of the more
dynamic use cases are now using PQExpBuffer API instead.


src/bin/pg_autoctl/file_utils.c
```diff
@@ -764,3 +764,24 @@ fformat(FILE *stream, const char *fmt, ...)
...
+ * sformat is a secured down version of pg_snprintf
+ */
+int
+sformat(char *str, size_t count, const char *fmt, ...)
+{
+	int			len;
+	va_list		args;
+
+	if (str == NULL || fmt == NULL)
+	{
```
JelteF: probably good to log here, these should both never happen

DimCitus: Log added, thanks!

PR #189, opened 2020-03-13
Replace ctime with ctime_r

JelteF:
ctime is a banned function

### DIFF
--- a/src/bin/pg_autoctl/state.c
+++ b/src/bin/pg_autoctl/state.c
@@ -501,17 +501,12 @@ NodeStateFromString(const char *str)
 const char *
 epoch_to_string(uint64_t seconds, char *buffer)
 {
-	char * result;
+	char *result = NULL;
 	if (seconds <= 0)
 	{
 		strlcpy(buffer, "0", MAXCTIMESIZE);
 		return buffer;
 	}
-	/*
-	 * ctime_r returns a string that ends with \n, which we don't want.
-	 * Replace it with a null string terminator, even if that means we lose
-	 * a precious byte of memory here..
-	 */
 	result = ctime_r((time_t *) &seconds, buffer);
 
 	if (result == NULL)
@@ -520,7 +515,14 @@ epoch_to_string(uint64_t seconds, char *buffer)
 				  seconds);
 		return NULL;
 	}
-	result[strlen(result) - 1] = '\0';
+	if (strlen(result) != 0 && result[strlen(result) - 1] == '\n')
+	{
+		/*
+		 * ctime_r normally returns a string that ends with \n, which we don't
+		 * want. We strip it by replacing it with a null string terminator.
+		 */
+		result[strlen(result) - 1] = '\0';
+	}
 	return buffer;
 }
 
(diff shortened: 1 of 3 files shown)


src/bin/pg_autoctl/state.c
```diff
@@ -492,29 +494,32 @@ NodeStateFromString(const char *str)
...
+ *
+ * This string is stored in buffer and for ease of use of the function this
+ * buffer is also returned. The buffer should (at least) be MAXCTIMESIZE large.
  */
 const char *
-epoch_to_string(uint64_t seconds)
+epoch_to_string(uint64_t seconds, char *buffer)
 {
-	if (seconds > 0)
+	if (seconds <= 0) {
```
DimCitus: Style issue: brackets on their own line.

src/bin/pg_autoctl/state.c
```diff
@@ -492,29 +494,32 @@ NodeStateFromString(const char *str)
...
-			return NULL;
-		}
-		str[strlen(str) - 1] = '\0';
-		return str;
+		log_error("Failed to convert epoch %" PRIu64 " to string: %m",
+				  seconds);
+		return NULL;
 	}
-	return "0";
+	buffer[strlen(buffer) - 1] = '\0';
```
DimCitus: Can the buffer be an empty string or something else that would trip us here? It looks suspicious the way we address the memory here.

src/bin/pg_autoctl/state.c
```diff
@@ -492,29 +494,32 @@ NodeStateFromString(const char *str)
...
+	if (seconds <= 0) {
+		strlcpy(buffer, "0", MAXCTIMESIZE);
+		return buffer;
+	}
+	/*
+	 * ctime_r returns a string that ends with \n, which we don't want.
+	 * Replace it with a null string terminator, even if that means we lose
+	 * a precious byte of memory here..
+	 */
+	ctime_r((time_t *) &seconds, buffer);
```
DimCitus: When discarding the return value of a function call, we often write `(void) funcall(...)` to make it obvious to the readers: we are discarding it on purpose.

src/bin/pg_autoctl/state.c
DimCitus: I now see that we used to do it that way already, so I guess it's fine. Still looks suspicious... but it does not belong to the current PR to solve it I guess.

src/bin/pg_autoctl/state.c
JelteF: I'm actually using the result now, since it indicates if it errors.

src/bin/pg_autoctl/state.c
JelteF: Added some sanity checks there.

src/bin/pg_autoctl/state.c
```diff
@@ -492,29 +494,35 @@ NodeStateFromString(const char *str)
...
+	}
+	result = ctime_r((time_t *) &seconds, buffer);
+
+	if (result == NULL)
 	{
+		log_error("Failed to convert epoch %" PRIu64 " to string: %m",
+				  seconds);
+		return NULL;
+	}
+	if (strlen(result) != 0 && result[strlen(result) - 1] == '\n') {
```
DimCitus: Style issue: brackets on their own line.

PR #190, opened 2020-03-13
Open files in a safe way

JelteF:
And reads are moved to a dedicated function for easy ignoring fopen as a banned
function there.

### DIFF
--- a/src/bin/pg_autoctl/file_utils.c
+++ b/src/bin/pg_autoctl/file_utils.c
@@ -153,6 +153,21 @@ fopen_with_umask(const char *filePath, const char* modes, int flags, mode_t umas
 
 
 /*
+ * fopen_read_only opens the file as a read only stream.
+ */
+FILE *
+fopen_read_only(const char *filePath)
+{
+	/*
+	 * Explanation of IGNORE-BANNED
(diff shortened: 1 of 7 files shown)


src/bin/pg_autoctl/file_utils.c
```diff
@@ -161,7 +176,7 @@ bool
 write_file(char *data, long fileSize, const char *filePath)
 {
 
-	FILE *fileStream = fopen_with_umask(filePath, "wb", O_WRONLY | O_CREAT, 0644);
+	FILE *fileStream = fopen_with_umask(filePath, "wb", O_WRONLY | O_CREAT | O_EXCL, 0644);
```
DimCitus: I think we should introduce a #define in defaults.h for FOPEN_FLAGS_W and FOPEN_FLAGS_A and then use the symbols rather than re-use always the same magic list. Also that gives a good place to comment on our O_EXCL usage here.

JelteF: O_EXCL actually doesn't make sense (which is why the tests failed). I thought it meant that only one process could open it, however it means that the file should not exist yet. Which we don't care about.

PR #198, opened 2020-03-16
QA work following the docs quickstart

DimCitus:
Follow the Quickstart and fix everything that we broke in the release making.

### DIFF
--- a/src/bin/pg_autoctl/monitor_config.c
+++ b/src/bin/pg_autoctl/monitor_config.c
@@ -578,11 +578,10 @@ monitor_config_set_setting(MonitorConfig *config,
 bool
 monitor_config_update_with_absolute_pgdata(MonitorConfig *config)
 {
-	PostgresSetup pgSetup = config->pgSetup;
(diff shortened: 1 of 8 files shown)


src/bin/pg_autoctl/monitor_config.c
```diff
@@ -578,11 +578,10 @@ monitor_config_set_setting(MonitorConfig *config,
 bool
 monitor_config_update_with_absolute_pgdata(MonitorConfig *config)
 {
-	PostgresSetup pgSetup = config->pgSetup;
+	PostgresSetup *pgSetup = &(config->pgSetup);
```
mtuncer: change of behavior, instead of copy of the data we are working on the actual config. Is this intentional ?

DimCitus: It is yeah, otherwise we don't use the absolute path and then we fail to use the proper PATH to the self-signed certificate that has been created when the users has given us a relative path to pgdata, such as `--pgdata ./monitor` as in the documentation.

PR #202, opened 2020-03-17
QA fixes.

DimCitus:


src/bin/pg_autoctl/primary_standby.c
```diff
@@ -64,7 +64,7 @@ GUC postgres_default_settings[] = {
 
 GUC citus_default_settings[] = {
 	DEFAULT_GUC_SETTINGS_FOR_PG_AUTO_FAILOVER,
-	{ "shared_preload_libraries", "citus" },
+	{ "shared_preload_libraries", "'citus'" },
```
mtuncer: I did not get why we need this

src/bin/pg_autoctl/pgsetup.c
```diff
@@ -411,6 +411,16 @@ get_pgpid(PostgresSetup *pgSetup, bool pg_is_not_running_is_ok)
 
 	join_path_components(pidfile, pgSetup->pgdata, "postmaster.pid");
 
+	if (!file_exists(pidfile))
```
mtuncer: is it because we want to have a different error message ? I suppose `read_file` returns false when file does not exist

src/bin/pg_autoctl/primary_standby.c
DimCitus: I saw the configuration file in the output and it didn't have any quotes around the shared library name and I though that was not right...

src/bin/pg_autoctl/pgsetup.c
DimCitus: It's because `read_file` always outputs an ERROR line when the file does not exists, and the `get_pgpid` might just be probing if the file exists or not, and it's very fine for it to not exists... so that we don't want an ERROR message to be unconditionally part of the output in that case.

PR #205, opened 2020-03-19
Merge/release 1.2

DimCitus:
Merge our release 1.2 back into the master development branch.


src/bin/pg_autoctl/cli_do_fsm.c
```diff
@@ -299,18 +299,15 @@ cli_do_fsm_assign(int argc, char **argv)
 			goalState = NodeStateFromString(argv[1]);
 
 			/* now prepare id, host, and port in keeper.otherNodes */
-			otherNode->nodeId = strtol(argv[2], NULL, 10);
-			if (otherNode->nodeId == 0 && errno == EINVAL)
+			if (!stringToInt(argv[2], &otherNode->nodeId))
```
mtuncer: I could not put a marker on the case label `case 4` was `case 3` in the source. Please review for correctness

mtuncer: There was probably a bug in the master branch. Please revisit this part

DimCitus: In the master branch we identify a node with its id/name/port, where before we would not ask for the id. Which means that `case 3` in the `release-1.2` branch is now `case 4` in the `master` branch. I had another look and it looks good to me...

Issue #181, opened 2020-03-09
How to stop the monitor


JelteF:
This error is not very useful
```
$ pg_autoctl stop --pgdata /mnt/monitor/
15:09:38 FATAL The command `pg_autoctl stop` does not apply to a monitor node.
```

DimCitus:
Fixed in https://github.com/citusdata/pg_auto_failover/pull/241.

Issue #128, opened 2019-12-24
Does or doesn't the monitor own its postgres instance?


Bouke:
Currently the ownership relationship between the monitor and postgres instance is inconsistent. It will start the instance if it is not running, but doesn't take ownership over the instance. As a result:

- stopping the monitor will keep the postgres instance running
- killing the instance will put the monitor in an infinite loop (#86)
- managing the instance through an external system (e.g. systemd) becomes difficult

I would either expect the monitor to:
- Not own the instance at all. This gives me the option to run/manage/monitor the instance through systemd.
- Own the instance.
   - When the monitor is stopped, the instance is stopped as well.
      Additionally I found that running `pg_autoctl run` through systemd, will stop the postgres instance when the unit is stopped. I assume this is a feature of systemd.
   - When the monitor fails, the instance is also taken down.
   - When the instance fails:
      - the monitor fails and restarting the monitor is handled by systemd, or:
      - the monitor restarts the instance.

Bouke:
See also #143, which is related to the same underlying problem. If the monitor owns the postgres instance, it should also handle stderr from the instance.

PR #206, opened 2020-03-24
WIP Feature/supervise postmaster

DimCitus:
We now have a process tree with two sub-processes, one running the monitor's
listener client and one is running `postgres` directory, bypassing pg_ctl
start.

This allows to control the life-cycle of Postgres through controlling the life-cycle of pg_autoctl.

See #124 
See #181 
See #128


JelteF:
I think this can be closed, now that we we have split the PRs and the final one is #284

PR #214, opened 2020-04-02
Change default TLS cipherlist

JelteF:
This is the Microsoft approved cipherlist for OpenSSL applications.


DimCitus:
I updated the master branch with the changes allowing to release again from it, and update this PR to target the master's branch. We have two commits from the release-1.2 that apparently did not make it to master, so they are visible in that PR.

If everything is okay, then I think we should rebase & merge so that we have the three commits visible on the master's branch as in this PR.

PR #216, opened 2020-04-03
Add autoformatting for python

JelteF:
This adds autoformatting of test files with [black](https://github.com/psf/black)


DimCitus (approved):
I am not fond of some of the changes, but having something automated to take care of it is a good thing!

DimCitus (approved):
I could checkout your PR and then `make indent` locally and see that there is no change added on-top of your PR, which is good. FWIW my installing of the tooling is different from the CONTRIBUTING.md file: `brew install black`. I do not think that we have to maintain how to install packages for every possible OS though, so we could merge that way.

```
black .
All done! ✨ 🍰 ✨
21 files left unchanged.
```

PR #217, opened 2020-04-03
Make CC generate dependency information for Makefile

JelteF:
This makes sure you don't have to run "make clean" anymore to get a
correct build when you change header files. To get the same support for
the monitor extension, you should compile your postgres with --enable-depend.

### DIFF
--- a/src/bin/pg_autoctl/Makefile
+++ b/src/bin/pg_autoctl/Makefile
@@ -3,6 +3,8 @@
 
 PG_AUTOCTL = ./pg_autoctl
 
+DEPDIR = .deps
+
 SRC_DIR := $(dir $(abspath $(lastword $(MAKEFILE_LIST))))
 
 INCLUDES  = $(patsubst ${SRC_DIR}%.h,%.h,$(wildcard ${SRC_DIR}*.h))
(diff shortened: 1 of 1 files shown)


DimCitus:
I always compile with the following line, which is very fast for me, so I don't feel a need for enhanced make dependencies. Then it looks complex, but I guess that's what it is. 

```
$ time make -s clean && make -s -j12 install

real	0m0.087s
user	0m0.035s
sys	0m0.044s
```

src/bin/pg_autoctl/Makefile
```diff
@@ -47,23 +49,35 @@ LIBS += -lpq
 
 all: $(PG_AUTOCTL) ;
 
+# Based on Postgres Makefile for automatic dependency generation
+# https://github.com/postgres/postgres/blob/1933ae629e7b706c6c23673a381e778819db307d/src/Makefile.global.in#L890-L924
+%.o : %.c
+	@if test ! -d $(DEPDIR); then mkdir -p $(DEPDIR); fi
+	$(CC) $(CFLAGS) -c -MMD -MP -MF$(DEPDIR)/$(*F).Po -o $@ $<
```
DimCitus, line 56: Can we add the new flags to a variable and then to CFLAGS, as we use it everywhere we use CFLAGS I believe?

JelteF: I tried to do that, but since these flags contain variables which depend on being in the target it doesn't work, i.e  `$(*F)`. I'm fine with adding the -MMD and -MP flags to CFLAGS though, but maybe it's nicer to keep these flags together.

PR #218, opened 2020-04-03
Always error out on warnings indicating obvious mistakes

JelteF:
Implicit function declarations and wrong return types cause confusing errors
during development and are never behave like you want.


DimCitus:
I think those are already activated in `clang` ; at least I usually see warnings about those error conditions and agree it's very useful.

JelteF:
Now they will become errors instead of warnings, so you won't get some confusing error later.

PR #219, opened 2020-04-03
Speed up tests a bit

JelteF:
Sleeping with smaller resolution makes sure that when something takes 150ms we
don't wait a full second, but instead only 200ms.

### DIFF
--- a/tests/pgautofailover_utils.py
+++ b/tests/pgautofailover_utils.py
@@ -362,16 +362,16 @@ class PGNode:
         Waits until the underlying Postgres process is running.
         """
         wait_until = dt.datetime.now() + dt.timedelta(seconds=timeout)
+        time.sleep(1)
         while wait_until > dt.datetime.now():
+            if self.pg_is_running():
+                time.sleep(1)
+                return True
             time.sleep(1)
 
-            if self.pg_is_running():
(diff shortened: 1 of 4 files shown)


JelteF:
This is now based on the #222 to make sure it has some of the fixes for reliable tests from there. Now it seems very stable and makes the full test run ~40-50 seconds faster on travis.

tests/pgautofailover_utils.py
```diff
@@ -362,16 +362,16 @@ def wait_until_pg_is_running(self, timeout=STATE_CHANGE_TIMEOUT):
...
         """
         wait_until = dt.datetime.now() + dt.timedelta(seconds=timeout)
+        time.sleep(1)
         while wait_until > dt.datetime.now():
-            time.sleep(1)
-
             if self.pg_is_running():
+                time.sleep(1)
                 return True
+            time.sleep(1)
```
DimCitus, line 370: That's a lot of manual sleeping around here. I think we should add some comments about why we came to add so many sleep calls, and in particular one after we know that pg_is_running.

JelteF: Reverted that change, it's not necessary anymore now that tests are working correctly.

PR #220, opened 2020-04-06
Speed up tests more by decreasing keeper sleep time

JelteF:
Tests are failing because of this, which is a problem. This change should not
impact correctness at all.


DimCitus:
From a quick reading of the tests output, it seems like the `test_ensure` is now failing because when the primary reaches the `DEMOTE_TIMEOUT` state then to does not get assigned the `DEMOTED` state by the monitor. This should happen when the secondary is reporting `STOP_REPLICATION` and the monitor finds that the primary has been in DEMOTED_TIMEOUT state for 30s or more (draining timeout expired), as we can read in `src/monitor/group_state_machine.c`.

``` C
	if (IsCurrentState(activeNode, REPLICATION_STATE_STOP_REPLICATION) &&
		(IsCurrentState(primaryNode, REPLICATION_STATE_DEMOTE_TIMEOUT) ||
		 IsDrainTimeExpired(primaryNode)))
```

Now I wonder if the keeper loop logic is still waiting long enough for the situation to kick-in as expected with a delay of 1s instead of 5s... or what else could impact the FSM to get blocked?

JelteF:
This test shows the problem in the worst way https://travis-ci.com/github/citusdata/pg_auto_failover/jobs/315021995:
```
eventtime                        id:nodename                   state/goal state        repl st        lsn event

2020-04-06 09:01:36.548707+00:00  1:172.27.1.3                  init/single              async        0/0 Setting goal state of 172.27.1.3:5432 to single as there is no other node.

2020-04-06 09:01:37.678936+00:00  1:172.27.1.3                single/single            unknown  0/1662198 Node 172.27.1.3:5432 reported new state single

2020-04-06 09:01:37.678936+00:00  1:172.27.1.3                  init/single              async        0/0 Setting goal state of 172.27.1.3:5432 to single as there is no other node.

2020-04-06 09:01:38.400711+00:00  2:172.27.1.4          wait_standby/wait_standby      unknown        0/0 Node 172.27.1.4:5432 reported new state wait_standby

2020-04-06 09:01:39.037028+00:00  1:172.27.1.3                single/wait_primary      unknown  0/1662198 Setting goal state of 172.27.1.3:5432 to wait_primary after 172.27.1.4:5432 joined.

2020-04-06 09:01:39.078019+00:00  1:172.27.1.3          wait_primary/wait_primary      unknown  0/167AE88 Node 172.27.1.3:5432 reported new state wait_primary

2020-04-06 09:01:39.407776+00:00  2:172.27.1.4          wait_standby/catchingup        unknown        0/0 Setting goal state of 172.27.1.4:5432 to catchingup after 172.27.1.3:5432 converged to wait_primary.

2020-04-06 09:01:42.618013+00:00  2:172.27.1.4            catchingup/catchingup        unknown  0/3000000 Node 172.27.1.4:5432 reported new state catchingup

2020-04-06 09:01:57.203583+00:00  2:172.27.1.4            catchingup/secondary         unknown  0/3000060 Setting goal state of 172.27.1.3:5432 to primary and 172.27.1.4:5432 to secondary after 172.27.1.4:5432 caught up.

2020-04-06 09:01:57.203583+00:00  1:172.27.1.3          wait_primary/primary             async  0/3000060 Setting goal state of 172.27.1.3:5432 to primary and 172.27.1.4:5432 to secondary after 172.27.1.4:5432 caught up.

2020-04-06 09:01:57.220108+00:00  2:172.27.1.4             secondary/secondary         unknown  0/3000060 Node 172.27.1.4:5432 reported new state secondary

2020-04-06 09:01:57.411178+00:00  1:172.27.1.3               primary/primary              sync  0/3000060 Node 172.27.1.3:5432 reported new state primary

2020-04-06 09:02:22.630264+00:00  2:172.27.1.4             secondary/prepare_promotion unknown  0/3000060 Setting goal state of 172.27.1.3:5432 to draining and 172.27.1.4:5432 to prepare_promotion after 172.27.1.3:5432 became unhealthy.

2020-04-06 09:02:22.630264+00:00  1:172.27.1.3               primary/draining             sync  0/3000060 Setting goal state of 172.27.1.3:5432 to draining and 172.27.1.4:5432 to prepare_promotion after 172.27.1.3:5432 became unhealthy.

2020-04-06 09:02:22.649413+00:00  2:172.27.1.4     prepare_promotion/prepare_promotion unknown  0/3000060 Node 172.27.1.4:5432 reported new state prepare_promotion

2020-04-06 09:02:23.667711+00:00  2:172.27.1.4     prepare_promotion/stop_replication  unknown  0/3000060 Setting goal state of 172.27.1.3:5432 to demote_timeout and 172.27.1.4:5432 to stop_replication after 172.27.1.4:5432 converged to prepare_promotion.

2020-04-06 09:02:23.667711+00:00  1:172.27.1.3               primary/demote_timeout       sync  0/3000060 Setting goal state of 172.27.1.3:5432 to demote_timeout and 172.27.1.4:5432 to stop_replication after 172.27.1.4:5432 converged to prepare_promotion.

2020-04-06 09:02:24.943384+00:00  2:172.27.1.4      stop_replication/stop_replication  unknown  0/3000218 Node 172.27.1.4:5432 reported new state stop_replication

2020-04-06 09:02:27.728376+00:00  1:172.27.1.3        demote_timeout/demote_timeout    unknown        0/0 Node 172.27.1.3:5432 reported new state demote_timeout

```

The log shows the secondary being in stop_replication and the primary being in demote_timeout, both in state and in goal state. So in this case the check you show should not even need to check the timeout part, since both the actual states are already in this place.

So it definitely seems like it gets stuck, especially because it waits for ~1.5 minutes after this last state change (based on the timings from the events and the logs above). 

Could it be because `activeNode` and `primaryNode` are swapped?

DimCitus:
Well the nodes aren't swapped really, the code moves the FSM when the secondary server reports to the monitor... I think because we want to make progress on the secondary node itself. The draining timeout is useful in case of async replication to ensure that the secondary has a chance to receive the WAL from the primary. 

Now if the primary has reported DEMOTE_TIMEOUT we know that Postgres is not running anymore on the primary, so I wonder what this expiration timeout means in this case.

In any case, this used to work fine with a 5s delay, and it does not work anymore with a 1s delay, and I'm not sure why that would the case. It might be the test framework itself that misses a state change? Can you reproduce the bug interactively?

JelteF:
Yes running this test locally with this branch it reproduces 80% of the time.

DimCitus:
When running locally and having a repro, can you share what's the secondary logs look like? is it confused or just looping around its STOP_REPLICATION state happily waiting for the monitor to take a decision to move forward?

DimCitus:
So do we want to change the PG_AUTOCTL_KEEPER_SLEEP_TIME from 5s down to 1s for everyone or just for the test suite? it could depend on PG_AUTOCTL_DEBUG being set in the environment if we wanted to, or even be controlled by some other mechanism.

I wonder if querying the monitor every 1s instead for every 5s when everything is fine is something we want to do. It reduces our reaction time, though, and I think that's good...

JelteF:
> I wonder if querying the monitor every 1s instead for every 5s when everything is fine is something we want to do. It reduces our reaction time, though, and I think that's good...

You only really know that "everything is fine" by querying the monitor, right? I don't really think there should be a problem with polling every second instead of every 5. Even a very slow server should be able to handle 1 query per node per second. And if you have such a big postgres cluster that it doesn't, you should probably just get a slightly bigger monitor node.

tests/pgautofailover_utils.py
```diff
@@ -551,13 +565,22 @@ def wait_until_state(self, target_state,
...
+            else:
+                other_pglogs = ""
 
-            if self.pg_autoctl and self.pg_autoctl.run_proc:
+            if self.running():
                 out, err = self.stop_pg_autoctl()
+                if other_node and other_node.running():
+                    other_out, other_err = other_node.stop_pg_autoctl()
+                else:
+                    other_out, other_err = "", ""
```
DimCitus, line 578: Let's use None here?

tests/pgautofailover_utils.py
```diff
@@ -551,13 +565,22 @@ def wait_until_state(self, target_state,
...
+
                 raise Exception("%s failed to reach %s after %d attempts: " \
-                                "\n%s\n%s\n%s\n%s" %
+                                "\n%s\n%s\n%s\n%s\n%s\n%s\n%s" %
```
DimCitus, line 581: I think we need a label (with empty lines before and after) that describe which node the next output belongs to...

PR #222, opened 2020-04-06
Feature/replication slots

DimCitus:
Enhance management of replication slots:

  - maintain replication slots on every node rather than just the primary, using the new in Postgres 11 function `pg_replication_slot_advance`
  - refrain from dropping replication slots at failover time, keeping the primary slot on the secondary ensures `pg_rewind` compatibility

The maintenance of replication slots depend on the full list of nodes on the monitor, and the monitor now has the LSN of every node, which enables us to advance to a recent-enough known location everywhere.


src/bin/pg_autoctl/pgsql.c
```diff
@@ -654,43 +721,265 @@ pgsql_drop_replication_slots(PGSQL *pgsql)
...
+												  "  FROM pg_replication_slots JOIN nodes USING(slot_name), "
+												  "       LATERAL pg_replication_slot_advance(slot_name, lsn) a"
+												  " WHERE nodes.lsn <> '0/0'"
+												  "), \n"
+												  "created as ("
+												  "SELECT c.slot_name, c.lsn "
+												  "  FROM nodes LEFT JOIN pg_replication_slots pgrs USING(slot_name), "
+												  "       LATERAL pg_create_physical_replication_slot(slot_name, true) c"
+												  " WHERE pgrs.slot_name IS NULL "
+												  ") \n"
```
JelteF, line 908: I think it makes sense to make these CTEs all look as similar as possible. That way the differences can be seen quickly and it makes it clear that they are all doing a very similar thing.

1. You should be able to remove the LATERAL join by doing either:
```sql
SELECT slot_name, (pg_replication_slot_advance(slot_name, lsn)).end_lsn ...
```
or 
```sql
SELECT (pg_replication_slot_advance(slot_name, lsn)).* ...
```
I'm fine with either of these.



2. Let's keep the tables in the same order in `created` and make it a RIGHT JOIN.

tests/pgautofailover_utils.py
```diff
@@ -405,6 +408,41 @@ def get_postgres_logs(self):
...
+    def pgversion(self):
+        """
+        Query local Postgres for its version. Cache the result.
+        """
+        if self._pgversion:
+            return self._pgversion
+
+        # server_version_num is 110005 for 11.5
+        self._pgversion = int(self.run_sql_query("show server_version_num")[0][0])
+        self._pgmajor = int(self._pgversion / 10000)
```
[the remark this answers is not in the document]
DimCitus: Done with //, thanks!

src/bin/pg_autoctl/pgsql.c
```diff
@@ -595,27 +596,96 @@ pgsql_check_monitor_settings(PGSQL *pgsql, bool *settings_are_ok)
...
+{
+	int bytesWritten =
+		sformat(slotName, size, "%s_%d", REPLICATION_SLOT_NAME_DEFAULT, nodeId);
+
+	return bytesWritten <= size;
+}
+
+
+/*
+ * pgsql_set_synchronous_standby_names set synchronous_standby_names ton the
```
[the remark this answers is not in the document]

src/bin/pg_autoctl/pgsql.c
```diff
@@ -654,43 +721,265 @@ pgsql_drop_replication_slots(PGSQL *pgsql)
...
+												  "  FROM nodes LEFT JOIN pg_replication_slots pgrs USING(slot_name), "
+												  "       LATERAL pg_create_physical_replication_slot(slot_name, true) c"
+												  " WHERE pgrs.slot_name IS NULL "
+												  ") \n"
+												  "SELECT 'create', slot_name, lsn FROM created "
+												  " union all "
+												  "SELECT 'drop', slot_name, NULL::pg_lsn FROM dropped "
+												  " union all "
+												  "SELECT 'advance', slot_name, end_lsn FROM advanced ";
+
```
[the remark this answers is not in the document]
DimCitus: Added the magic comment, hoped that maybe we could have a better option there...

src/bin/pg_autoctl/pgsql.c
```diff
@@ -654,43 +721,265 @@ pgsql_drop_replication_slots(PGSQL *pgsql)
...
+		"WITH nodes(slot_name, lsn) as ("
+		" SELECT '" REPLICATION_SLOT_NAME_DEFAULT "_' || id, lsn"
+												  "   FROM (%s) as sb(id, lsn) "
+												  "), \n"
+												  "dropped as ("
+												  " SELECT slot_name, pg_drop_replication_slot(slot_name) "
+												  "   FROM pg_replication_slots pgrs LEFT JOIN nodes USING(slot_name) "
+												  "  WHERE nodes.slot_name IS NULL "
+												  ") \n"
+												  "SELECT 'drop', slot_name, NULL::pg_lsn FROM dropped";
```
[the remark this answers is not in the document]
DimCitus: Added the magic comment again.

src/bin/pg_autoctl/pgsql.c
```diff
@@ -654,43 +721,265 @@ pgsql_drop_replication_slots(PGSQL *pgsql)
...
+
+		valuesIndex += snprintf(buffer + valuesIndex, BUFSIZE - valuesIndex,
+								"%s($%d, $%d%s)",
+								valuesIndex == 0 ? "" : ",",
+
+		                        /* we begin at $1 here: intentional off-by-one */
+								paramIndex - 1, paramIndex,
+
+		                        /* cast only the first row */
+								valuesIndex == 0 ? "::pg_lsn" : "");
```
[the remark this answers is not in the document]
DimCitus: Good point! It's much easier to follow through that way, done.

src/bin/pg_autoctl/pgsql.c
```diff
@@ -654,43 +721,265 @@ pgsql_drop_replication_slots(PGSQL *pgsql)
...
+		log_error("Query returned %d columns, expected 3", PQnfields(result));
+		context->parsedOK = false;
+		return;
 	}
 
-	return pgsql_alter_system_set(pgsql, setting);
+	for (rowNumber = 0; rowNumber < PQntuples(result); rowNumber++)
+	{
+		char *operation = PQgetvalue(result, rowNumber, 0);
+		char *slotName = PQgetvalue(result, rowNumber, 1);
```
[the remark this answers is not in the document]
DimCitus: Not the way we build the query no. The operation is static in the query string itself, the slot name being NULL would mean the function call would have failed / errored out.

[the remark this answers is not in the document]
DimCitus: I added a quick comment...

src/bin/pg_autoctl/pgsql.c
DimCitus: The notation `(funcall()).*` has a better avoided side-effect; it calls the function as many times as there are parameters. LATERAL JOIN is wonderful because of that, much better than using the SELECT projection...

The main problem I have with that code now, after the fact, is that in case we fail to advance, create, or drop a single replication slot, then we stall progress for all the others. I think I might go on and have an explicit loop for create or advance and then a specific call for drop ; where we don't have a list to walk anyway...

src/bin/pg_autoctl/primary_standby.c
```diff
@@ -132,6 +135,27 @@ local_postgres_finish(LocalPostgresServer *postgres)
 }
 
 
+static bool
+local_postgres_update(LocalPostgresServer *postgres, bool postgresNotRunningIsOk)
```
JelteF, line 139: Missing comment

src/bin/pg_autoctl/primary_standby.c
```diff
@@ -591,8 +640,31 @@ standby_init_database(LocalPostgresServer *postgres,
...
 	 * PostgreSQL instance. We can safely proceed with pg_basebackup.
 	 */
-	if (!pg_basebackup(pgSetup->pgdata, pgSetup->pg_ctl, replicationSource))
+	if (!pg_basebackup(pgSetup->pgdata,
+					   pgSetup->pg_ctl,
+					   replicationSource))
+	{
+		return false;
+	}
+
```
JelteF, line 649: Let's keep the `normalize_filename` call I added here, so that pgdata is fully qualified in all your calls too.

src/bin/pg_autoctl/pgctl.c
```diff
@@ -1447,6 +1460,61 @@ pg_write_standby_signal(const char *configFilePath,
...
+ * pg_cleanup_standby_mode cleans-up the replication settings for the local
+ * instance of Postgres found at pgdata.
+ *
+ *  - remove either recovery.conf or standby.signal
+ *
+ *  - when using Postgres 12 also make postgresql-auto-failover-standby.conf an
+ *    empty file, so that we can still include it, but it has no effect.
+ */
+bool
+pg_cleanup_standby_mode(uint32_t pg_control_version,
```
JelteF, line 1473: Is this function still needed now that we don't use `--write-recovery-conf` anymore?

src/bin/pg_autoctl/pgsql.c
```diff
@@ -654,43 +721,277 @@ pgsql_drop_replication_slots(PGSQL *pgsql)
...
- * pgsql_set_synchronous_standby_names set synchronous_standby_names ton the
- * local Postgres to the value computed on the pg_auto_failover monitor.
+ * pgsql_replication_slot_maintain maintains replication slots that belong to
+ * other standby nodes. We call that function on the standby nodes, where the
+ * slots are maintained manually just in case we need them at failover.
  */
 bool
-pgsql_set_synchronous_standby_names(PGSQL *pgsql,
-									char *synchronous_standby_names)
+pgsql_replication_slot_maintain(PGSQL *pgsql, NodeAddressArray *nodeArray)
```
JelteF, line 890: The name of this function and it's comment confused me a bit. I assumed that it only maintained existing slots, but it actually also creates and drops slots. I cannot really think of a better name, but the comment should be a little bit more clear on that, e.g.: 
```
pgsql_replication_slot_maintain creates, updates and drops replication slots that 
belong to other standby nodes...
```

src/bin/pg_autoctl/pgsql.c
```diff
@@ -654,43 +721,277 @@ pgsql_drop_replication_slots(PGSQL *pgsql)
...
+	return pgsql_execute_with_params(pgsql, sql,
+									 paramCount, paramTypes, paramValues,
+									 &context,
+									 parseReplicationSlotMaintain);
+}
 
-	if (sformat(quoted, BUFSIZE, "'%s'", synchronous_standby_names) >= BUFSIZE)
+
+/*
+ * parsePgMetadata parses the result from a PostgreSQL query fetching
```
JelteF, line 953:
```suggestion
 * parseReplicationSlotMaintain parses the result from a PostgreSQL query fetching
```

src/bin/pg_autoctl/primary_standby.c
DimCitus: Added, thanks.

src/bin/pg_autoctl/primary_standby.c
DimCitus: In my copy this call is there still after the rebase, so I'm a little surprised by your comment. That said, I have moved the call to `normalize_filename` to the `local_postgres_update` function that I think it's a better place.

src/bin/pg_autoctl/pgctl.c
DimCitus: We still need to remove `recovery.conf` or `standby.signal` depending on the version of Postgres when promoting a standby to now be a primary. Otherwise at next restart it's not going to be a primary anymore. I don't believe `pg_ctl promote` would do the cleanup for us, and if it did, it would not clean-up the `postgresql-auto-failover-standby.conf` file where we keep the `primary_conninfo` and other replication parameters.

src/bin/pg_autoctl/pgsql.c
DimCitus: Improved the comments, thanks.

src/bin/pg_autoctl/pgsql.c
DimCitus: Done, thanks!

src/bin/pg_autoctl/primary_standby.c
JelteF: What I meant was move it back to this location. I think your move of the call to `local_postgres_update` is fine, but we need to move the `join_path_components` call for the config path after it too. Otherwise it still has the relative path.

DimCitus: Oh I see. Yet tests did pass, should we review the tests (in another PR) related to relative pathnames impacts?

Meanwhile I have refactored where we compute the configFilePath and done that in the `pg_setup_standby_mode` function itself: after all we have to pass PGDATA down to it, and the configFilePath is a direct compute from PGDATA.

PR #224, opened 2020-04-07
Actually run the linting scripts we have in ci

JelteF:
Checks both for correct indentation and disallows use of banned APIs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ matrix:
     - env: PGVERSION=12
     - env: LINTING=true
 before_install:
-  - git clone -b v0.7.12 --depth 1 https://github.com/citusdata/tools.git
+  - git clone -b update-uncrustify-on-travis --depth 1 https://github.com/citusdata/tools.git
   - sudo make -C tools install
   - '[ "$LINTING" == true ] || setup_apt'
   - '[ "$LINTING" == true ] || nuke_pg'
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,8 +21,8 @@ matrix:
 before_install:
   - git clone -b update-uncrustify-on-travis --depth 1 https://github.com/citusdata/tools.git
   - sudo make -C tools install
-  - 'if [ "$LINTING" == "" ]; then setup_apt; fi'
-  - 'if [ "$LINTING" == "" ]; then nuke_pg; fi'
+  - 'if [ -z "$LINTING" ]; then setup_apt; fi'
+  - 'if [ -z "$LINTING" ]; then nuke_pg; fi'
   - python --version
   - pyenv versions
   - pyenv install -s 3.7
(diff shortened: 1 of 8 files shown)


.travis.yml
```diff
@@ -17,11 +17,12 @@ matrix:
...
     - env: PGVERSION=11
     - env: PGVERSION=12
+    - env: LINTING=true
 before_install:
   - git clone -b v0.7.12 --depth 1 https://github.com/citusdata/tools.git
   - sudo make -C tools install
-  - setup_apt
-  - nuke_pg
+  - '[ "$LINTING" == true ] || setup_apt
+  - '[ "$LINTING" == true ] || nuke_pg
```
DimCitus, line 25: That reads strange. If listing is true then, oh wait, no, if it's not true then. Can we use `&&` instead, or at least add comments about why we need to use this convoluted form? Also I'm not familiar with the first single quote in the line but that might just be the YAML support at Travis to escape to the shell?

JelteF: It was because all the lines need to return true for Travis to pass CI and that doesn't work when you do `&&`. I changed it to `if [ "$LINTING" == "" ]; then setup_apt; fi` now, for better reability.

The quotes are their to indeed just make the whole line a string and not let yaml iterpret characters as things that they are not. (The single quote was simply a mistake on my end, I didn't put one at the end of the line.)

.travis.yml
```diff
@@ -17,11 +17,12 @@ matrix:
...
     - env: PGVERSION=12
+    - env: LINTING=true
 before_install:
-  - git clone -b v0.7.12 --depth 1 https://github.com/citusdata/tools.git
+  - git clone -b update-uncrustify-on-travis --depth 1 https://github.com/citusdata/tools.git
   - sudo make -C tools install
-  - setup_apt
-  - nuke_pg
+  - 'if [ "$LINTING" == "" ]; then setup_apt; fi'
+  - 'if [ "$LINTING" == "" ]; then nuke_pg; fi'
```
DimCitus, line 25: Another way to write that is to use the built-in test `-z` and `-n` operations, which you might prefer?

In a shell here where $LINTING is not defined here's what I have:

``` shell
$ help test
...
        -z STRING      True if string is empty.    
        -n STRING
        STRING         True if string is not empty.
...

$ test -n "$LINTING" || echo ok
ok

$ [ -n "$LINTING" ] || echo ok
ok
```

ci/banned.h.sh
```diff
@@ -12,7 +12,7 @@
 
 set -eu
 
-files=$(find src -iname '*.[ch]' -not -path "src/bin/lib/*")
+files=$(find src -iname '*.[ch]' -type f | git check-attr --stdin citus-style | grep -v ': unset$' | sed 's/: citus-style: set$//')
```
DimCitus, line 15: Each time I see `grep -v | another-filter` I cringe internally and want to use `awk` instead. It would look like the following if we did use awk. What do you think?

``` shell
find src -iname '*.[ch]' -type f | git check-attr --stdin citus-style | awk -F ':' '! /: unset$/ {print $1}'
```

.travis.yml
DimCitus: I see, thanks for the details! I added some more comments...

PR #225, opened 2020-04-07
Check result code of sformat

JelteF:
Although we are using `snprintf` to make sure we don't overflow a buffer, we
don't actually check if the whole string was written in most cases. If this
would occur, this would cause very weird bugs and possibly buffer overflows
later in the code.  This PR makes sure we actually check the result of sformat.
It uses macros to limit clutter on the callsite for an edge case that almost
certainly will not happen.


DimCitus:
Alternative thinking: is there a way to get the caller of the function in the logs when that happens? When using `-vv` or `-vvv` we output __FILE__ and __LINE__ in the logs, and that's the information we want here. What about turning `sformat` into a macro to have the caller information in the trace then?

JelteF:
I put an extra `log_debug` in the macro itself. This way file/line info should be shown there when using `-vv`.

PR #226, opened 2020-04-08
Poll with 100ms resolution in wait_until_pg_is_running

JelteF:
Same as #219, this causes a lot of instability in the tests so it's split out
of that PR.

In the end it doesn't seem to make the suite much (or really any) faster. 

It does fix a lot of hidden problems in the test suite though. So I think it's definetely still worth it to get this merged, to make sure in the future we don't have these problems. Things this fixes:

1. `PGAutoCtl.execute()` didn't flush output of pg_autoctl while waiting, so that could deadlock because the pipe was not being read.
2. `NSPopen` objects were not being released. This could cause out of memory issues on travis.
3. Timeout errors in `PGAutoCtl.execute()` didn't have any useful information on the state of the cluster. This makes debugging those issues very hard and time consuming.


tests/network.py
```diff
@@ -11,6 +12,18 @@
...
+
+@contextmanager
+def managed_nspopen(*args, **kwds):
+    # Code to acquire resource, e.g.:
+    proc = NSPopen(*args, **kwds)
+    try:
+        yield proc
+    finally:
+        # Code to release resource, e.g.:
+        proc.release()
```
DimCitus, line 24: This "e.g." looks like a leftover from a copy paste. Also I think we need to `proc.wait()` before we `proc.release` given the NSPopen API.

tests/pgautofailover_utils.py
```diff
@@ -246,10 +248,10 @@ def set_user_password(self, username, password):
...
                           '-d', self.database,
                           '-c', alter_user_set_passwd_command]
-        passwd_proc = self.vnode.run(passwd_command)
-        wait_or_timeout_proc(passwd_proc,
-                         name="user passwd",
-                         timeout=COMMAND_TIMEOUT)
+        with self.vnode.run(passwd_command) as passwd_proc:
+            wait_or_timeout_proc(passwd_proc,
+                                 name="user passwd",
+                                 timeout=COMMAND_TIMEOUT)
```
DimCitus, line 254: That would read better as:

``` Python
        with self.vnode.run(passwd_command, name="user passwd") as passwd_proc:
             passwd_proc.wait(timeout=COMMAND_TIMEOUT)
```

tests/pgautofailover_utils.py
```diff
@@ -279,93 +281,98 @@ def stop_postgres(self):
...
+                    if stop_proc.returncode > 0:
+                        print("stopping postgres for '%s' failed, out: %s\n, err: %s"
+                              % (self.vnode.address, out, err))
+                        return False
+                    elif stop_proc.returncode is None:
+                        print("stopping postgres for '%s' timed out")
+                        return False
+                    return True
             except subprocess.TimeoutExpired:
+                stop_proc.release()
```
DimCitus, line 296: I'm surprised that stop_proc is defined here, but maybe I'm reading the diff wrong?

tests/pgautofailover_utils.py
```diff
@@ -279,93 +281,98 @@ def stop_postgres(self):
...
     def wait_until_pg_is_running(self, timeout=STATE_CHANGE_TIMEOUT):
         """
         Waits until the underlying Postgres process is running.
         """
         wait_until = dt.datetime.now() + dt.timedelta(seconds=timeout)
         while wait_until > dt.datetime.now():
-            time.sleep(1)
             if self.pg_is_running():
                 return True
+            time.sleep(0.1)
```
DimCitus, line 375: Maybe our sleep time here should now be a constant?

tests/pgautofailover_utils.py
```diff
@@ -484,13 +492,36 @@ def show_uri(self, json=False):
...
         """
-        command = PGAutoCtl(self.vnode, self.datadir)
+        command = PGAutoCtl(self)
         if json:
             out, err = command.execute("show uri", 'show', 'uri', '--json')
         else:
             out, err = command.execute("show uri", 'show', 'uri')
         return out
 
+    def detailed_exception(self, error_msg, other_node=None):
```
DimCitus, line 502: I think this could be better named `get_detailed_exception` so that we don't have to second-guess the choice of returning the object compared to raising the exception. Or maybe `build_detailed_exception` or even `make_detailed_exception`?

DimCitus: Reading it another time, it should return only the exception error message. So what about `build_detailed_error_message` or `make_error_report` or something like that?

tests/pgautofailover_utils.py
```diff
@@ -628,30 +660,9 @@ def wait_until_state(self, target_state,
...
-            if other_node.running():
-                out, err = other_node.stop_pg_autoctl()
-                error_msg += f"STDOUT OF PG_AUTOCTL FOR OTHER NODE:\n{out}\n"
-                error_msg += f"STDERR OF PG_AUTOCTL FOR OTHER NODE:\n{err}\n"
-
-            pglogs = other_node.get_postgres_logs()
-            error_msg += f"POSTGRES LOGS FOR OTHER NODE:\n{pglogs}\n"
-
-        raise Exception(error_msg)
+        raise self.detailed_exception(error_msg)
```
DimCitus, line 665: As per above, this could be better as something in the spirit of:

``` Python
        raise Exception("%s" % self.make_full_error_report())
```

tests/pgautofailover_utils.py
```diff
@@ -975,10 +988,10 @@ def create_formation(self, formation_name,
...
                 formation_command += ['--disable-secondary']
 
-        formation_proc = self.vnode.run(formation_command)
-        wait_or_timeout_proc(formation_proc,
-                             name="create formation",
-                             timeout=COMMAND_TIMEOUT)
+        with self.vnode.run(formation_command) as formation_proc:
+            wait_or_timeout_proc(formation_proc,
+                                 name="create formation",
+                                 timeout=COMMAND_TIMEOUT)
```
DimCitus, line 994: Again, what about:

``` Python
        with self.vnode.run(formation_command, name="create formation") as formation_proc:
              formation_proc.wait(timeout=COMMAND_TIMEOUT)
```

tests/pgautofailover_utils.py
```diff
@@ -1060,44 +1073,43 @@ def run(self, level='-vv'):
...
-            raise Exception("%s timed out after %d seconds.\n%s\n%s\n%s"%
-                            (name,
-                             COMMAND_TIMEOUT,
-                             " ".join(self.command),
-                             self.out,
-                             self.err))
-
-        return self.out, self.err
+            proc.kill()
+            proc.wait()
```
DimCitus, line 1106: Yeah I guess the `proc.wait()` call here should be part of the managed bits.

tests/pgautofailover_utils.py
JelteF: yes that's that's definitely wrong, removing it. I guess it works because `with` doesn't actually create a separate scope (just like `if`/`for`/`while` etc). And `stop_proc.release()` is probably a no-op if it's already been called (which it is because that happens when the `with` block is exited)

tests/network.py
JelteF: > This "e.g." looks like a leftover from a copy paste. 
Yes it is

> Also I think we need to proc.wait() before we proc.release given the NSPopen API.

I'll take a look if it's necessary. It already fixed the out of memory issues I was haveing without it.

JelteF:
So, this PR is finally passing. It doesn't seem to make the suite much (or really any) faster. 

It does fix a lot of hidden problems in the test suite though. So I think it's definetely still worth it to get this merged, to make sure in the future we don't have these problems. Things this fixes:

1. `PGAutoCtl.execute()` didn't flush output of pg_autoctl while waiting, so that could deadlock because the pipe was not being read.
2. `NSPopen` objects were not being released. This could cause out of memory issues on travis.
3. Timeout errors in `PGAutoCtl.execute()` didn't have any useful information on the state of the cluster. This makes debugging those issues very hard and time consuming.

I'll clean it up a bit next week, so it can be merged.

tests/network.py
```diff
@@ -123,14 +141,48 @@ def destroy(self):
...
+        NSOpen has the same API as subprocess.POpen.
+        """
+        sudo_command = ['sudo', '-E', '-u', user,
+                        'env', 'PATH=' + os.getenv("PATH")] + command
+        return managed_nspopen(self.namespace, sudo_command,
+                               stdin=subprocess.PIPE, stdout=subprocess.PIPE,
+                               stderr=subprocess.PIPE, universal_newlines=True,
+                               start_new_session=True)
+
+    def wait_or_timeout_command(self, command, name, timeout):
```
DimCitus, line 154: What about naming this just `wait` or something in that spirit? The timeout raises an exception so I don't think it's very good to have it in the name... (in the name means it's expected, means it's not an exception?)

tests/network.py
```diff
@@ -10,6 +11,23 @@
...
+    try:
+        yield proc
+    finally:
+        if proc.poll() is None:
+            # send SIGKILL to the process and wait for it to die if it's still
+            # running
+            proc.kill()
+            proc.communicate()
+        # release proxy process resourecs
+        proc.release()
```
DimCitus, line 29: Given https://docs.pyroute2.org/netns.html?highlight=nspopen#pyroute2.netns.process.proxy.NSPopen I think we should do 

``` Python
    proc.wait()
    proc.release()
```

What do you think?

JelteF: Doesn't seem very useful given that the process is known not be running at this point: https://docs.python.org/3/library/subprocess.html#subprocess.Popen.wait

DimCitus: I see it now, thanks...

tests/pgautofailover_utils.py
```diff
@@ -93,19 +94,18 @@ def pg_createcluster(self, datadir, port=5432):
...
-        create_proc = vnode.run(create_command)
-        out, err = create_proc.communicate(timeout=COMMAND_TIMEOUT)
-        if create_proc.returncode > 0:
-            raise Exception("pg_createcluster failed, out: %s\n, err: %s" %
-                            (out, err))
+        with vnode.run(create_command) as create_proc:
+            out, err = create_proc.communicate(timeout=COMMAND_TIMEOUT)
+            if create_proc.returncode > 0:
+                raise Exception("pg_createcluster failed, out: %s\n, err: %s" %
+                                (out, err))
```
DimCitus, line 108: Could this be `self.vnode.run_and_wait(create_command, name="pg_createcluster")`?

tests/pgautofailover_utils.py
```diff
@@ -115,11 +115,11 @@ def pg_createcluster(self, datadir, port=5432):
...
-        chmod_proc = vnode.run(chmod_command)
-        out, err = chmod_proc.communicate(timeout=COMMAND_TIMEOUT)
-        if chmod_proc.returncode > 0:
-            raise Exception("chmod failed, out: %s\n, err: %s" %
-                            (out, err))
+        with vnode.run(chmod_command) as chmod_proc:
+            out, err = chmod_proc.communicate(timeout=COMMAND_TIMEOUT)
+            if chmod_proc.returncode > 0:
+                raise Exception("chmod failed, out: %s\n, err: %s" %
+                                (out, err))
```
DimCitus, line 122: Same as before, looks like a single call to `self.vnode.run_and_wait()` now, right?

tests/pgautofailover_utils.py
```diff
@@ -1060,44 +1060,40 @@ def run(self, level='-vv'):
...
             self.command = [self.program, 'run', '--pgdata', self.datadir, level]
 
-        self.run_proc = self.vnode.run(self.command)
+        if self.run_proc:
+            self.run_proc.release()
+        self.run_proc = self.vnode.run_unmanaged(self.command)
         print("pg_autoctl run [%d]" % self.run_proc.pid)
 
-    def execute(self, name, *args):
+    def execute(self, name, *args, timeout=COMMAND_TIMEOUT, other_node=None):
```
DimCitus, line 1068: I now wonder if we could give the list of other nodes at `__init__` time automatically from the node method where I think we have access to the monitor which has the list of all the data nodes anyway. Adding this `other_node=` parameter everywhere manually is becoming tiring. That said, that's material for another PR.

JelteF: Sounds good but I think indeed this fits better in a separate PR. I was thinking something along the lines of adding keeping a reference to the cluster in a node and doing all the "other_node" stuff for all of the nodes in the cluster. This would make it work for multiple stanbys too.

PR #227, opened 2020-04-08
Stop postgres more reliably by trying again in case pg_autoctl restarted it

JelteF:


tests/pgautofailover_utils.py
```diff
@@ -260,10 +261,19 @@ def stop_postgres(self):
...
           pg_ctl -D ${self.datadir} --wait --mode fast stop
         """
-        stop_command = [shutil.which('pg_ctl'), '-D', self.datadir,
-                        '--wait', '--mode', 'fast', 'stop']
-        stop_proc = self.vnode.run(stop_command)
-        out, err = stop_proc.communicate(timeout=10)
+        for i in range(60):
+            stop_command = [shutil.which('pg_ctl'), '-D', self.datadir,
+                            '--wait', '--mode', 'fast', 'stop']
+            stop_proc = self.vnode.run(stop_command)
```
DimCitus, line 267: Let's add some comments about how `pg_ctl stop` is racy when another process is starting Postgres for you in the background. That explains why we do all that dance here.

PR #228, opened 2020-04-08
Fix ssl_cert_file/ssl_key_file being incorrect on standby

JelteF:
Fixes problems with SSL settings not being applied correctly on the standby.

### DIFF
--- a/src/bin/pg_autoctl/file_utils.c
+++ b/src/bin/pg_autoctl/file_utils.c
@@ -743,24 +743,10 @@ set_program_absolute_path(char *program, int size)
 
 
 /*
- * normalize_filename_inplace returns the real path of a given filename that
- * belongs to an existing file on-disk, resolving symlinks and pruning
- * double-slashes and other weird constructs. It stores it back in filename,
- * which is assumed to be a buffer of size MAXPGPATH.
- */
-bool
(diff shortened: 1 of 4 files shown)


src/bin/pg_autoctl/file_utils.c
```diff
@@ -742,6 +742,21 @@ set_program_absolute_path(char *program, int size)
...
 
 
+/*
+ * normalize_filename_inplace returns the real path of a given filename that
+ * belongs to an existing file on-disk, resolving symlinks and pruning
+ * double-slashes and other weird constructs. It stores it back in filename,
+ * which is assumed to be a buffer of size MAXPGPATH.
+ */
+bool
+normalize_filename_inplace(char *filename)
```
DimCitus, line 752: Can we instead make it so that `normalise_filename` works well when given the same filename twice, same as `join_path_components` that we use that way (in e.g. src/bin/pg_autoctl/config.c):

``` c
	/* and finally add the configuration file name */
	join_path_components(filename, filename, name);

	/* normalize the path to the configuration file, if it exists */
	if (!normalize_filename(filename, dst, MAXPGPATH))
	{
		return false;
	}
```

JelteF: Modified, it actually already supported that.

PR #233, opened 2020-04-16
Implement a way to update to SSL when it's not been installed yet.

DimCitus:


docs/security.rst
```diff
@@ -280,3 +280,50 @@ If you have your own HBA provisioning solution, you can include the rules
...
+``postgresql-auto-failover.conf`` Postgres configuration file to match the
+command line arguments given and enable SSL as instructed, and then updates
+the pg_autoctl configuration.
+
+The HBA settings are not edited, irrespective of the ``--skip-pg-hba`` that
+has been used at creation time. That's because the ``host`` records match
+either SSL or non-SSL connection attempts in Postgres HBA file, so the
+pre-existing setup will continue to work. To enhance the SSL setup, you can
+manually edit the HBA files and change the existing lines from ``host`` to
+``hostssl``.
```
JelteF, line 317:
```suggestion
``hostssl`` to only allow encrypted traffic.
```

docs/security.rst
```diff
@@ -280,3 +280,50 @@ If you have your own HBA provisioning solution, you can include the rules
...
+
+When updating a node's SSL setup, we connect to the monitor and check if the
+SSL setup has been made there. When that's the case, the monitor URI in the
+pg_autoctl configuration file is not edited for you, because it might be
+using complex local connection options. Please take time to edit your
+connection string to the monitor using::
```
JelteF, line 323: So we do check something, but we don't actually change anything?
I think it's good if it starts with, something similar to first sentence of the HBA section. e.g. "The connection string to connect to the monitor is not automatically updated when using this command"

I also think this section should be moved before the HBA setting paragraph, since it seems more important than that.

DimCitus: If I'm understanding this right, it's now done!

JelteF: I think the change you made is good, but I simply meant updating the documentation.
The HBA changes are more like a bonus, but the monitor connection string changes are pretty important.

I meant two things should be changed in the docs:
1. this section should be moved up above the HBA hostssl changes
2. this section should start with a clear sentence stating that the monitor connection string should be updated manually

DimCitus: I see, good point. I have now update the documentation that way and added a clear operative summary of the list of tasks needed to implement SSL.

docs/security.rst
```diff
@@ -280,3 +280,65 @@ If you have your own HBA provisioning solution, you can include the rules
...
+variants, depending if you want self-signed certificates or fully verified
+ssl certificates::
+
+  $ pg_autoctl enable ssl --ssl-self-signed --ssl-mode required
+
+  $ pg_autoctl enable ssl --ssl-ca-file root.crt   \
+                          --ssl-crl-file root.crl  \
+                          --server-crt server.crt  \
+                          --server-key server.key  \
+                          --ssl-mode validate-full
```
JelteF, line 305:
```suggestion
                          --ssl-mode verify-full
```

docs/security.rst
```diff
@@ -280,3 +280,65 @@ If you have your own HBA provisioning solution, you can include the rules
...
+Installing Self-Signed certificates on-top of an already existing
+pg_auto_failover setup is done with one of the following pg_autoctl command
+variants, depending if you want self-signed certificates or fully verified
+ssl certificates::
+
+  $ pg_autoctl enable ssl --ssl-self-signed --ssl-mode required
+
+  $ pg_autoctl enable ssl --ssl-ca-file root.crt   \
+                          --ssl-crl-file root.crl  \
+                          --server-crt server.crt  \
```
JelteF, line 303: Throughout the documentation we use `--server-crt` but the option is `--server-cert`. This should be fixed in many more places.
```suggestion
                          --server-cert server.crt  \
```

JelteF:
I've checked off above which problems are now indeed fixed according to my manual testing:
The one that's not is that the new monitor connection string is not actually used right away in my setup because of the following two errors:

This one on the primary:
```
16:52:18 WARN  Received signal Hangup
16:52:18 ERROR Attempt to change replication.slot from "pgautofailover_standby" to "pgautofailover_standby_1"
16:52:18 WARN  Failed to read configuration file "/home/jelte/.config/pg_autoctl/home/jelte/work/pg_auto_failover/node1/pg_autoctl.cfg", continuing with the same configuration.
```

And this one on the secondary:
```
16:53:10 WARN  Received signal Hangup
16:53:10 ERROR Attempt to change postgresql.pgdata from "node2" to "/home/jelte/work/pg_auto_failover/node2"
16:53:10 WARN  Failed to read configuration file "/home/jelte/.config/pg_autoctl/home/jelte/work/pg_auto_failover/node2/pg_autoctl.cfg", continuing with the same configuration.
```

I think it's good to fix these two issues and also note in the docs that it is required to look at the pg_autoctl process logs to confirm that it has reloaded successfully, not just at the command output. Since there's no indication there that the command failed. Obviously if we can see that the command failed it would be even better, but I think that's probably hard to do and not worth it.

docs/security.rst
```diff
@@ -280,3 +280,65 @@ If you have your own HBA provisioning solution, you can include the rules
...
+is not edited for you, because it might be using complex local connection
+options. Please take time to edit your connection string to the monitor
+using::
+
+  $ pg_autoctl config set pg_autoctl.monitor
+
+Before that, you can verify your current configuration with::
+
+  $ pg_autoctl config get pg_autoctl.monitor
+
```
JelteF, line 325: Something like this should be placed here:

An already running `pg_autoctl` deamon will try to reload it's configuration after `pg_autoctl config set` has finished. In some cases this is not possible to do without a restart. So be sure to check the logs from a running daemon to confirm that the reload succeeded. If it did not you need to restart the daemon to ensure the new connection string is used.

DimCitus: Done now, thanks!

src/bin/pg_autoctl/cli_enable_disable.c
```diff
@@ -402,3 +432,593 @@ cli_disable_maintenance(int argc, char **argv)
...
+			{
+				log_info("  Self-Signed certificates have been created and "
+						 "deployed in Postgres configuration settings "
+						 "ssl_key_file and ssl_cert_file");
+			}
+
+			log_info("  Health-check connections to Postgres nodes now "
+					 "use sslmode=%s",
+					 pgsetup_sslmode_to_string(pgSetup->ssl.sslMode));
+
```
JelteF: Healt checks never use SSL at the moment, so let's remove this log as it's simply incorrect. We can add it again if we change this.

```suggestion
```

src/bin/pg_autoctl/cli_enable_disable.c
```diff
@@ -402,3 +432,593 @@ cli_disable_maintenance(int argc, char **argv)
...
+}
+
+
+/*
+ * update_monitor_connection_string connects to the monitor to see if ssl is
+ * active on the server. When that's the case, the function edits the monitor
+ * URI in the given KeeperConfig.
+ */
+static bool
+update_monitor_connection_string(KeeperConfig *config)
```
JelteF, line 880: This function name and comment suggest that this function updates the connection string, but actually it only logs.

src/bin/pg_autoctl/cli_enable_disable.c
```diff
@@ -402,3 +432,593 @@ cli_disable_maintenance(int argc, char **argv)
...
+
+			if (pgSetup->ssl.createSelfSignedCert)
+			{
+				log_info("  Self-Signed certificates have been created and "
+						 "deployed in Postgres configuration settings "
+						 "ssl_key_file and ssl_cert_file");
+			}
+
+			log_warn("  Postgres connection string to the monitor "
+					 "has not been changed, see above for details");
```
JelteF, line 789: This log seems unnecessary given all the info shown by `update_monitor_connection_string`.

```suggestion
```

src/bin/pg_autoctl/cli_common.c
```diff
@@ -1279,3 +1279,25 @@ stop_postgres_and_remove_pgdata_and_config(ConfigFilePaths *pathnames,
...
+ * cli_pg_autoctl_reload signals the pg_autoctl process to reload its
+ * configuration by sending it the SIGHUP signal.
+ */
+bool
+cli_pg_autoctl_reload(const char *pidfile)
+{
+	pid_t pid;
+
+	if (read_pidfile(pidfile, &pid))
+	{
```
JelteF, line 1294: Maybe do a sanity check that the pid is positive otherwise you can accidently send SIGHUP to all processes (I remember seeing a check like this somewhere else in the codebase):
https://rachelbythebay.com/w/2014/08/19/fork/

DimCitus: Done now!

src/bin/pg_autoctl/keeper_config.c
```diff
@@ -799,6 +799,22 @@ keeper_config_accept_new(KeeperConfig *config, KeeperConfig *newConfig)
...
 	}
 
+	if (config->pgSetup.ssl.active != newConfig->pgSetup.ssl.active)
+	{
+		log_info("Reloading configuration: ssl is now %s; used to be %s",
+				 newConfig->pgSetup.ssl.active ? "active" : "disabled",
+				 config->pgSetup.ssl.active ? "active" : "disabled");
+	}
+
+	if (config->pgSetup.ssl.sslMode != newConfig->pgSetup.ssl.sslMode)
```
JelteF, line 809: For consistency I guess this check should also be done for the other ssl fields, i.e. `caFile` and `crlFile`.

src/bin/pg_autoctl/cli_enable_disable.c
JelteF: Actually never mind, I think this log is good now that I have seen it in the output.

src/bin/pg_autoctl/cli_common.c
```diff
@@ -1292,12 +1292,21 @@ cli_pg_autoctl_reload(const char *pidfile)
...
 		if (kill(pid, SIGHUP) != 0)
 		{
 			log_error("Failed to send SIGHUP to the pg_autoctl's pid %d: %m",
 					  pid);
 			return false;
 		}
 	}
-	return true;
+
+	return false;
```
JelteF, line 1311: This change seems incorrect, since now it always fails.

src/bin/pg_autoctl/cli_enable_disable.c
DimCitus: I kind of hope to fix that someday with actually doing the work, leaving in here a clue that we  would like to be able to. Maybe that warrants a better comment though...

src/bin/pg_autoctl/keeper_config.c
DimCitus: Done now.

src/bin/pg_autoctl/cli_common.c
DimCitus: Good catch, fixed now!

src/bin/pg_autoctl/fsm_transition.c
```diff
@@ -195,12 +195,20 @@ fsm_init_primary(Keeper *keeper)
...
 					  "see above for details");
 			return false;
 		}
+
+		/* update our configuration with ssl server.{key,cert} */
+		if (!keeper_config_write_file(config))
+		{
+			/* errors have already been logged */
+			return false;
+		}
```
JelteF, line 211: This needs to be done in `standby_init_database` too. Otherwise secondary doesn't have the self signed certificates in there.

DimCitus: I think the `pg_autoclt` configuration management is done at other places, in `standby_init_database` we are only re-creating the certificates so that at the same PATH as on the primary we find new files, not the same server key as on the primary, that `pg_basebackup` sent over to us...

tests/test_enable_ssl.py
```diff
@@ -0,0 +1,91 @@
...
+    print()
+    print("Enabling maintenance on node2")
+    node2.enable_maintenance()
+    assert node2.wait_until_state(target_state="maintenance")
+
+def test_005_enable_ssl_monitor():
+    monitor.enable_ssl(sslSelfSigned=True, sslMode="require")
+    monitor.sleep(2) # we signaled, wait some time
+
+    assert monitor.config_get("ssl.sslmode") == "require"
```
JelteF, line 63: Let's do this check with `eq_` to, so it's easier to debug when they are ever wrong. (same for similar checks below)

DimCitus: Done now.

docs/security.rst
```diff
@@ -280,3 +280,63 @@ If you have your own HBA provisioning solution, you can include the rules
...
+In summary, to upgrade an existing pg_auto_failover setup to enable SSL:
+
+  1. run the ``pg_autoctl enable ssl`` command on your monitor and then all
+     the Postgres nodes,
+
+  2. on the Postgres nodes, review your pg_autoctl logs to make sure that
+     the reload operation has been effective, and review your Postgres
+     settings to verify that you have the expected result,
+
+  3. review youur HBA rules setup to change the pg_auto_failover rules from
```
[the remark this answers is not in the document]
DimCitus: Done.

docs/security.rst
```diff
@@ -280,3 +280,63 @@ If you have your own HBA provisioning solution, you can include the rules
...
+the pg_autoctl configuration.
+
+The connection string to connect to the monitor is also automatically
+updated by the ``pg_autoctl enable ssl`` command. You can verify your new
+configuration with::
+
+  $ pg_autoctl config get pg_autoctl.monitor
+
+Note that an already running pg_autoctl deamon will try to reload its
+configuration after ``pg_autoctl config set`` has finished. In some cases
```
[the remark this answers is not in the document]
DimCitus: Done.

src/bin/pg_autoctl/cli_config.c
```diff
@@ -580,6 +581,12 @@ cli_config_set(int argc, char **argv)
 		case PG_AUTOCTL_ROLE_KEEPER:
 		{
 			(void) cli_keeper_config_set(argc, argv);
+
+			if (!cli_pg_autoctl_reload(config.pathnames.pid))
+			{
+				/* errors have already been logged */
+				exit(EXIT_CODE_INTERNAL_ERROR);
+			}
```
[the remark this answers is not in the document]
DimCitus: Done.

src/bin/pg_autoctl/monitor.c
```diff
@@ -3060,3 +3062,13 @@ prepare_connection_to_current_system_user(Monitor *source, Monitor *target)
...
+
+
+/*
+ * monitor_ssl_active checks if SSL is active on the monitor.
+ */
+bool
+monitor_ssl_active(Monitor *monitor)
+{
+	return pgsql_ssl_active(&(monitor->pgsql));
+}
```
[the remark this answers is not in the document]
DimCitus: Good point, removed.

src/bin/pg_autoctl/pgsql.c
```diff
@@ -1926,3 +1933,30 @@ pgsql_alter_extension_update_to(PGSQL *pgsql,
 
 	return true;
 }
+
+
+/*
+ * monitor_ssl_active checks if SSL is active on the monitor.
+ */
+bool
+pgsql_ssl_active(PGSQL *pgsql)
```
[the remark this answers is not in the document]
DimCitus: Got rid of it now!

tests/test_enable_ssl.py
```diff
@@ -0,0 +1,97 @@
...
+        file_path = os.path.join(node.datadir, f)
+        assert os.path.isfile(file_path)
+        eq_(node.pg_config_get(setting), file_path)
+
+
+def test_000_create_monitor():
+    global monitor
+    monitor = cluster.create_monitor("/tmp/enable/monitor")
+    monitor.run()
+    monitor.wait_until_pg_is_running()
```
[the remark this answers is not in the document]
DimCitus: Done now.

tests/test_enable_ssl.py
```diff
@@ -0,0 +1,97 @@
...
+    # without too much delay/sleep hacking; when doing the `pg_autoctl
+    # enable ssl` online we need to make sure the signal made it to the
+    # running process and then was acted upon
+    node1.stop_pg_autoctl()
+    node1.enable_ssl(sslSelfSigned=True, sslMode="require")
+    node1.run()
+    node1.sleep(2)
+
+    eq_(node1.config_get("ssl.sslmode"), "require")
+    eq_(node1.pg_config_get('ssl'), "on")
```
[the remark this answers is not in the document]
DimCitus: Done.

tests/test_enable_ssl.py
```diff
@@ -0,0 +1,97 @@
...
+def test_007_enable_ssl_secondary():
+    node2.enable_ssl(sslSelfSigned=True, sslMode="require")
+    node2.sleep(5)
+
+    eq_(node2.config_get("ssl.sslmode"), "require")
+
+    node2.wait_until_pg_is_running()
+
+    eq_(node2.pg_config_get('ssl'), "on")
+    check_ssl_files(node2)
```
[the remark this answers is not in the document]
DimCitus: I didn't do that, because it's more work. You can't query that from Postgres before version 12, and it was in `recovery.conf` ; since Postgres 12 we put it in `postgresql-auto-failover-standby.conf`. And then we would need to parse the file...

tests/test_enable_ssl.py
```diff
@@ -0,0 +1,97 @@
...
+
+def test_008_disable_maintenance():
+    print("Disabling maintenance on node2")
+    node2.disable_maintenance()
+    assert node2.wait_until_pg_is_running()
+    assert node2.wait_until_state(target_state="secondary")
+    assert node1.wait_until_state(target_state="primary")
```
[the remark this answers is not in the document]
DimCitus: Can we add that in another PR? That would allow closing this one...

src/bin/pg_autoctl/cli_enable_disable.c
```diff
@@ -402,3 +432,566 @@ cli_disable_maintenance(int argc, char **argv)
 		exit(EXIT_CODE_MONITOR);
 	}
 }
+
+
+/*
+ * cli_create_monitor_getopts parses the command line options necessary to
```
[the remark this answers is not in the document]
DimCitus: Done.

src/bin/pg_autoctl/cli_enable_disable.c
```diff
@@ -402,3 +432,566 @@ cli_disable_maintenance(int argc, char **argv)
...
+	{
+		commandline_help(stderr);
+		exit(EXIT_CODE_BAD_ARGS);
+	}
+
+	/*
+	 * We're not using pg_setup_init() here: we are following a very different
+	 * set of rules. We just want to check:
+	 *
+	 *   - PGDATA is set and the directory does not exists
```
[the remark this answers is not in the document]
DimCitus: Fixed.

src/bin/pg_autoctl/cli_enable_disable.c
```diff
@@ -402,3 +432,566 @@ cli_disable_maintenance(int argc, char **argv)
...
+
+	return optind;
+}
+
+
+/*
+ * cli_enable_ssl enables SSL setup on this node.
+ *
+ *  - edit our Postgres configuration with the given SSL files and options
+ *  - when run on a keeper, checks that the monitor accepts ssl connections
```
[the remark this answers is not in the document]
DimCitus: Done.

src/bin/pg_autoctl/cli_enable_disable.c
```diff
@@ -402,3 +432,566 @@ cli_disable_maintenance(int argc, char **argv)
...
+				log_info("  Self-Signed certificates have been created and "
+						 "deployed in Postgres configuration settings "
+						 "ssl_key_file and ssl_cert_file");
+			}
+
+			if (reloadedService)
+			{
+				log_info("  pg_autoctl service has been signaled to reload "
+						 "its configuration");
+			}
```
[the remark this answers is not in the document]
DimCitus: Done.

src/bin/pg_autoctl/cli_enable_disable.c
```diff
@@ -402,3 +432,566 @@ cli_disable_maintenance(int argc, char **argv)
...
+		if (!pg_create_self_signed_cert(pgSetup, nodename))
+		{
+			log_error("Failed to create SSL self-signed certificate, "
+					  "see above for details");
+			return false;
+		}
+	}
+
+	/* HBA rules for hostssl are not edited */
+	sformat(hbaFilePath, MAXPGPATH, "%s/pg_hba.conf", pgSetup->pgdata);
```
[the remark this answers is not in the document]
DimCitus: Done that way, good point.

src/bin/pg_autoctl/keeper.c
```diff
@@ -702,6 +706,226 @@ keeper_ensure_postgres_is_running(Keeper *keeper, bool updateRetries)
...
+	return true;
+}
+
+
+/*
+ * keeper_ensure_configuration updates the Postgres settings to match the
+ * pg_autoctl configuration file, if necessary.
+ *
+ * This includes making sure that the SSL server.{key,cert} files are used in
+ * the Postgres configuration, and on a secondary server, that means updateing
```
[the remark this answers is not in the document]
DimCitus: Done.

tests/test_enable_ssl.py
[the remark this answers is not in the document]
JelteF: Okay, makes sense. Let's only do it for PG12 then with a simple `node2.pg_config_get('primary_conninfo')`. I have been testing this myself using `SHOW primary_conninfo` (on PG12 indeed, but maybe it works on PG11/PG10 too).

tests/test_enable_ssl.py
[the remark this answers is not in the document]
JelteF: Yes sounds good

src/bin/pg_autoctl/cli_enable_disable.c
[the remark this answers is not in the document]
JelteF: The change you did is not what I meant. I think it's good too and we should keep it. But it doesn't cover one important case, the case where the service is not running at all. Especially on the monitor this seems like a situation that could occur, since it does not do much there.

What I meant is add an error line to the summary. That if it was not reloaded the service should be (re)started, to apply the changes.

tests/test_enable_ssl.py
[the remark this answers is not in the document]
DimCitus: Done now, thanks!

tests/test_enable_ssl.py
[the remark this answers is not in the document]
DimCitus: Here's the issue to track progress on that front, next week: https://github.com/citusdata/pg_auto_failover/issues/248

src/bin/pg_autoctl/cli_enable_disable.c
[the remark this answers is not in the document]
DimCitus: Done now.

PR #235, opened 2020-04-17
Normalize pgdata in a better place

JelteF:
The normalize we had didn't actually work correctly, since it was doing it on
`pgSetup` instead of `newPgSetup`. So the result was basically thrown away.
This moves the normalization to `pg_setup_init`, which seems reasonable because
`normalize_filename` is a no-op when the directory does not exist.


DimCitus:
We could call `normalise_filename` in `pg_setup_init` just after we make sure that the PGDATA directory exists, around where we call `pg_controldata` to fetch more information about the running server?

PR #240, opened 2020-04-21
Improve "show uri" command

### LINKED ISSUE #238: Add a --monitor flag to "show uri"
Right now you can get the uri for a specific formation using `--formation`, but for the monitor you cannot. In that case you have to use `--json` and parse it using `jq`.

JelteF:
This does three things:
1. Make "show uri" output consistent for keeper and monitor
2. Add --monitor flag to get only the monitor uri
3. Fix sslmode in a uri being empty on old clusters

NOTE: Before this change on a keeper `pg_autoctl show uri` would output the
string for the formation that the keeper was a part of. This is not possible
anymore now without passing `--formation the-formation-string`. If it is deemed
useful to have this old behaviour behind a flag, it would be quite easy to add.
Another option is to have it stay the default, but add an `--all` flag that can
be used to show all formation strings on the keeper.

Fixes #238

### DIFF
--- a/src/bin/pg_autoctl/cli_show.c
+++ b/src/bin/pg_autoctl/cli_show.c
@@ -857,7 +857,6 @@ cli_show_uri(int argc, char **argv)
 		}
 	}
 
-
 	if (showUriOptions.monitorOnly)
 	{
 		(void) print_monitor_uri(&monitor, stdout);
--- a/src/bin/pg_autoctl/cli_show.c
+++ b/src/bin/pg_autoctl/cli_show.c
@@ -47,18 +47,22 @@ static bool fprint_file_contents(const char *filename);
 
 static int cli_show_uri_getopts(int argc, char **argv);
 static void cli_show_uri(int argc, char **argv);
-static void cli_show_all_uri(int argc, char **argv);
-static void cli_show_formation_uri(int argc, char **argv);
 
-static void print_monitor_and_formation_uri(KeeperConfig *config,
-											Monitor *monitor,
-											FILE *stream);
+static void print_monitor_uri(Monitor *monitor, FILE *stream);
+static void print_formation_uri(SSLOptions *ssl,
+								Monitor *monitor,
+								const char *formation,
+								FILE *stream);
+static void print_all_uri(SSLOptions *ssl,
+						  Monitor *monitor,
+						  FILE *stream);
 
 CommandLine show_uri_command =
 	make_command("uri",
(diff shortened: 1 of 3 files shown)


src/bin/pg_autoctl/keeper_config.h
```diff
@@ -48,6 +48,9 @@ typedef struct KeeperConfig
 	int prepare_promotion_walreceiver;
 	int postgresql_restart_failure_timeout;
 	int postgresql_restart_failure_max_retries;
+
+	/* pg_autoctl show uri */
+	bool monitorUriOnly;
```
DimCitus, line 53: We don't add command line option switches to our internal state data structures. That's why we have global variables for some switches (like outputJSON for instance, and keeperOptions and monitorOptions. It would be great to have a different design with a command line specific structure (that could still have a PostgresSetup field), but that's not what we have now and it might be a little too much to ask for this PR to refactor the whole CLI option business...

Still, I think it's important that we refrain from adding command-line processing features to the keeper state data structures.

JelteF: Created a `ShowUriOptions` struct to store the options now.

src/bin/pg_autoctl/cli_show.c
```diff
@@ -152,6 +156,14 @@ typedef struct ShowFileOptions
 
 static ShowFileOptions showFileOptions;
 
+typedef struct ShowUriOptions
+{
+	bool monitorOnly;
+	char formation[NAMEDATALEN];
+} ShowUriOptions;
```
DimCitus, line 163: What about also having a `PostgresSetup pgSetup` slot in there and then using it in the option parsing function, so that we can later initialise easily either a monitor or a keeper setup depending on what we find from the PGDATA, rather than having to transform the `keeperOptions` into a monitor setting?

If that works well, I'm thinking we should at some point prepare a refactoring for all the other “dual mode” commands to use this technique rather than the current `monitor_config_init_from_pgsetup` way.

DimCitus: We would still use `monitor_config_init_from_pgsetup` I guess, but from a "clean" pgSetup rather than from the keeperOptions ; it would allow to decouple the option parsing data structures from the keeper and monitor internal state data structures, and I think it would be good to decouple those.

DimCitus: Off line discussion shows that this kind of refactoring needs more design still so that we can do a full command option parsing facility rather than try to solve it partially just here.

PR #243, opened 2020-04-23
Avoid requiring a restart of Postgres during our initialization steps.

DimCitus:
If we create the postgresql.conf before starting Postgres, we can include
the citus extension in the shared_preload_libraries and be done with that
step without having to later pg_ctl restart.

### DIFF
--- a/src/bin/pg_autoctl/keeper_pg_init.c
+++ b/src/bin/pg_autoctl/keeper_pg_init.c
@@ -648,6 +648,33 @@ create_database_and_extension(Keeper *keeper)
 	local_postgres_init(&initPostgres, &initPgSetup);
 
 	/*
+	 * Add pg_autoctl PostgreSQL settings, including Citus extension in
+	 * shared_preload_libraries when dealing with a Citus worker or coordinator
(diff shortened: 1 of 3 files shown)


src/bin/pg_autoctl/keeper_pg_init.c
```diff
@@ -774,6 +758,12 @@ create_database_and_extension(Keeper *keeper)
 		 * Connect to pgsql as the system user to create extension: Same
 		 * user as initdb with superuser privileges.
 		 */
+		if (!keeper_update_pg_state(keeper))
```
JelteF, line 761: Let's move this before the CREATE EXTENSION log line 5 lines above to keep clear logging. Also let's add a comment why this call is needed, because it's not immediately clear to me. What's the reason this is only needed for Citus nodes?

DimCitus: We need to re-init our connection string to do the `create extension` as the proper user, and calling that function is doing that for us, IIRC. I added a comment to that end.

PR #244, opened 2020-04-23
Cleanup, fix, refactor pgsetup Postgres pidfile utilities.

DimCitus:
Also make our functions exported in the PG_AUTOCTL_DEBUG facilities, to make
it easy to test and fix them. Since they're published there now, also use
them in our test suite rather than re-implementing a special-purpose version
of them in Python.

### DIFF
--- a/src/bin/pg_autoctl/cli_do_misc.c
+++ b/src/bin/pg_autoctl/cli_do_misc.c
@@ -253,7 +253,7 @@ keeper_cli_pgsetup_is_ready(int argc, char **argv)
 		exit(EXIT_CODE_PGCTL);
 	}
 
-	log_debug("Initialized pgSetup, now calling pg_setup_wait_until_is_ready()");
+	log_debug("Initialized pgSetup, now calling pg_setup_is_ready()");
 
 	pgIsReady = pg_setup_is_ready(&pgSetup, pgIsNotRunningIsOk);
 
@@ -310,7 +310,7 @@ keeper_cli_pgsetup_startup_logs(int argc, char **argv)
 		exit(EXIT_CODE_PGCTL);
 	}
 
-	log_debug("Initialized pgSetup, now calling pg_setup_wait_until_is_ready()");
+	log_debug("Initialized pgSetup, now calling pg_log_startup()");
 
 	if (!pg_log_startup(pgSetup.pgdata, LOG_INFO))
 	{
--- a/src/bin/pg_autoctl/cli_do_misc.c
+++ b/src/bin/pg_autoctl/cli_do_misc.c
@@ -215,11 +215,11 @@ keeper_cli_add_standby_to_hba(int argc, char **argv)
 
 
 /*
- * keeper_cli_discover_pg_setup implements the CLI to discover a PostgreSQL
+ * keeper_cli_pgsetup_discover implements the CLI to discover a PostgreSQL
  * setup thanks to PGDATA and other environment variables.
  */
 void
-keeper_cli_discover_pg_setup(int argc, char **argv)
+keeper_cli_pgsetup_discover(int argc, char **argv)
 {
 	PostgresSetup pgSetup = { 0 };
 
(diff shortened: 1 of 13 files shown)


src/bin/pg_autoctl/pgsetup.c
```diff
@@ -153,6 +158,13 @@ pg_setup_init(PostgresSetup *pgSetup,
 		}
 		else
 		{
+			/* get the real path of PGDATA now */
+			if (!normalize_filename(pgSetup->pgdata, pgSetup->pgdata, MAXPGPATH))
+			{
+				/* errors have already been logged */
+				return false;
+			}
```
JelteF, line 166: Why was the move of this block needed?

src/bin/pg_autoctl/pgsetup.c
```diff
@@ -367,33 +375,13 @@ pg_setup_init(PostgresSetup *pgSetup,
...
-		 * of problems during the `pg_autoctl create` command.
-		 */
-		strlcpy(dbname, pgSetup->dbname, NAMEDATALEN);
-		strlcpy(pgSetup->dbname, "template1", NAMEDATALEN);
-
-		/* initialise a SQL connection to the local postgres server */
-		pg_setup_get_local_connection_string(pgSetup, connInfo);
-		pgsql_init(&pgsql, connInfo, PGSQL_CONN_LOCAL);
-
-		if (!pgsql_is_in_recovery(&pgsql, &pgSetup->is_in_recovery))
```
JelteF, line 388: Why do we not need this anymore?

src/bin/pg_autoctl/pgsetup.c
```diff
@@ -499,7 +498,20 @@ read_pg_pidfile(PostgresSetup *pgSetup, bool pg_is_not_running_is_ok)
...
+		 */
+		if (maxRetries > 0)
+		{
+			log_trace("read_pg_pidfile: \"%s\" does not exists [%d]",
+					  pidfile, maxRetries);
+			pg_usleep(250 * 1000); /* wait for 250ms and try again */
+			return read_pg_pidfile(pgSetup, pgIsNotRunningIsOk, maxRetries - 1);
+		}
+
+		if (!pgIsNotRunningIsOk)
```
JelteF, line 514: This way if pgIsNotRunningIsOk is false we will first wait `maxRetries * 250ms` before continueing. Is that on on purpose?

src/bin/pg_autoctl/pgsetup.c
DimCitus: Not a big deal, could be done otherwise. I had it that way in the other branch where this comes from: we normalise the path after making sure that `pg_controldata` is happy with the PGDATA in this version...

src/bin/pg_autoctl/pgsetup.c
DimCitus: Exactly. It's been there since about the beginning, but we moved a long time ago already to using other means to update that parts, see `pgsql_get_postgres_metadata` that combines several pieces of information that we need in a single query: https://github.com/citusdata/pg_auto_failover/blob/master/src/bin/pg_autoctl/pgsql.c#L1664

src/bin/pg_autoctl/pgsetup.c
DimCitus: Yeah, we actually retry after 250ms rather than returning an error immediately. We retry up to maxRetries time before telling to the caller that we're out of luck.

src/bin/pg_autoctl/cli_do_misc.c
```diff
@@ -237,6 +237,91 @@ keeper_cli_discover_pg_setup(int argc, char **argv)
...
+keeper_cli_pgsetup_startup_logs(int argc, char **argv)
+{
+	PostgresSetup pgSetup = { 0 };
+
+	if (!pg_setup_init(&pgSetup, &keeperOptions.pgSetup, true, true))
+	{
+		exit(EXIT_CODE_PGCTL);
+	}
+
+	log_debug("Initialized pgSetup, now calling pg_setup_wait_until_is_ready()");
```
JelteF, line 313:
```suggestion
	log_debug("Initialized pgSetup, now calling pg_log_startup()");
```

src/bin/pg_autoctl/pgctl.c
```diff
@@ -920,6 +919,156 @@ pg_ctl_start(const char *pg_ctl,
...
+	 * rather short.
+	 *
+	 * Also we setup log_directory to be "log" so that's where we are looking
+	 * into.
+	 */
+
+	/* prepare PGDATA/log directory path */
+	join_path_components(pgLogDirPath, pgdata, "log");
+
+	if (directory_exists(pgLogDirPath))
```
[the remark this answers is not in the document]
DimCitus: Good point. I liked the ability to declare some of the local variables in a narrower scope, but decreasing indentation level on a large part of this function's code makes for easier code to maintain. Done now.

src/bin/pg_autoctl/pgsetup.c
```diff
@@ -534,30 +568,38 @@ read_pg_pidfile(PostgresSetup *pgSetup, bool pg_is_not_running_is_ok)
...
 			{
-				log_error("Postgres pidfile does not contain a valid pid %s", line);
+				log_error("Postgres pidfile does not contain a valid pid %s",
+						  line);
 
 				return false;
 			}
 
-			pgSetup->pidFile.pid = pid;
+			pgSetup->pidFile.pid = abs(pid);
```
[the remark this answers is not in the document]
DimCitus: Added.

PR #245, opened 2020-04-23
Drop removed replication slots when disabling replication

JelteF:
Tests were failing randomly because replication slots were not in sync when a
node was reporting as "single". This also makes the comment of
fsm_disable_replication correct again.

This also disables logging of connections and disconnections by default,
because this filled up the logs too much.


tests/pgautofailover_utils.py
```diff
@@ -892,11 +892,13 @@ def has_needed_replication_slots(self):
         if set(expected_slots) == set(current_slots):
             print("slots list on %s is %s, as expected" %
                   (self.datadir, current_slots))
-        else:
-            print("slots list on %s is %s, expected %s" %
-                  (self.datadir, current_slots, expected_slots))
+            return True
 
-        return set(expected_slots) == set(current_slots)
+        self.print_debug_logs()
```
DimCitus, line 897: I guess that's just for the development phase, not intended for commit?

JelteF: It is intended for commit. This is only printed when this check fails, so the logs are very useful for investigating.

JelteF:
`fsm_resume_as_primary` already called `keeper_drop_replication_slots_for_removed_nodes` indirectly since it calls `fsm_disable_replication`. I pushed changes such that `fsm_promote_standby` does the same.

src/bin/pg_autoctl/fsm_transition.c
```diff
@@ -353,7 +355,7 @@ fsm_disable_replication(Keeper *keeper)
  *
  *    start_postgres
  * && disable_synchronous_replication
- * && drop_replication_slot
+ * && keeper_drop_replication_slots_for_removed_nodes
```
DimCitus, line 358: The name of the action here does not have to match the name of the function in the source code, but that's fine too.

src/bin/pg_autoctl/fsm_transition.c
```diff
@@ -917,7 +926,7 @@ fsm_promote_standby(Keeper *keeper)
 		return false;
 	}
 
-	if (!primary_disable_synchronous_replication(postgres))
+	if (!fsm_disable_replication(keeper))
```
DimCitus, line 929: Mmm, it seems ok to share code that way for now, yes.

JelteF: I guess the main reason it was not like that before was because the standby would not have a replication slot, so there would be nothing to delete anyway.

Issue #213, opened 2020-04-01
Problems encountered when trying to set up local cluster


JelteF:
This was the log of me trying to set up a local pg_auto_failover cluster:
```bash
# Start monitor
pg_autoctl create monitor --pgdata monitor --auth trust --ssl-self-signed
# Thought: Start node1 on same port as monitor
# Error: fails because port is already taken
# Problem 1: node is added to monitor even though it cannot start correctly
pg_autoctl create postgres --pgdata node1 --auth trust --ssl-self-signed --monitor 'postgres://autoctl_node@172.24.132.206:5432/pg_auto_failover?sslmode=require'
# Thought: Makes sense, let's try adding it with a different --pgport
# Error: Monitor ERROR:  node 172.24.132.206:5433 is not registered
# Problem 2: Updates the port of a node that's already registered to the monitor
pg_autoctl create postgres --pgdata node1 --auth trust --ssl-self-signed --monitor 'postgres://autoctl_node@172.24.132.206:5432/pg_auto_failover?sslmode=require' --pgport 5433
# Thought: Okay, let's remove it and try again
# Problem 3: Node is not removed from monitor because port has changed
pg_autoctl drop node --destroy --pgdata node1
# Thought: Let's try again with the right port
# Works, but node is added as wait_standby and that is never changed because
# single does not exist anymore
pg_autoctl create postgres --pgdata node1 --auth trust --ssl-self-signed --monitor 'postgres://autoctl_node@172.24.132.206:5432/pg_auto_failover?sslmode=require' --pgport 5433
# Thought: Okay let's remove the node with the old port then
# Problem 4: Node is not changed from wait_standby to single
pg_autoctl drop node --pgdata monitor/ --nodename 172.24.132.206 --pgport 5432
# Thought: Let's start over completely and use the right port right away
# Works, but took very long
pg_autoctl drop node --pgdata monitor/ --nodename 172.24.132.206 --pgport 5433
pg_autoctl create postgres --pgdata node1 --auth trust --ssl-self-signed --monitor 'postgres://autoctl_node@172.24.132.206:5432/pg_auto_failover?sslmode=require' --pgport 5433
```

DimCitus:
Now fixed thanks to #340.

PR #249, opened 2020-04-27
Refactor --nodename to be --hostname, and add-in a user-defined nodename

DimCitus:
This refactoring allows pg_auto_failover to manage a user-defined public
node name for the nodes separate from the hostname used to connect to the
Postgres service on nodes. It appears that people like naming things.

Next-up in this PR: allowing the edit the metadata of a node at any point,
in particular change the name or hostname or port at pg_autoctl run, for
easy k8s integration (--hostname may change at each pod reboot).

Fixes https://github.com/citusdata/pg_auto_failover/issues/213.


src/bin/pg_autoctl/monitor_config.c
```diff
@@ -599,13 +599,13 @@ monitor_config_accept_new(MonitorConfig *config, MonitorConfig *newConfig)
...
+	/* changing the hostname online is supported */
+	if (strneq(newConfig->hostname, config->hostname))
 	{
-		log_info("Reloading configuration: nodename is now \"%s\"; "
+		log_info("Reloading configuration: hostname is now \"%s\"; "
 				 "used to be \"%s\"",
-				 newConfig->nodename, config->nodename);
-		strlcpy(config->nodename, newConfig->nodename, _POSIX_HOST_NAME_MAX);
+				 newConfig->hostname, config->hostname);
+		strlcpy(config->hostname, newConfig->hostname, _POSIX_HOST_NAME_MAX);
```
JelteF, line 608: Isn't changing the nodename online also supported? (If not let's not implement it in this PR)

DimCitus: The nodename is now only used for display, and can be changed at any time.

The hostname is used to connect to a node, so when it changes on a standby the primary needs to edit its HBA file, which we do now in the current state of the PR. When changed on the primary, we can expect the replication to stop working and then we need to make sure that the next time we set it up (through the CATCHING UP state) we fetch the current primary information again. That's an open ticket in this PR, I didn't take the time to make sure this happens yet. I know I've done that already in the multiple-standbys PR.

The port is used to connect to a node and also in the `postgres ...` command line when starting the Postgres service. This can not be changed while Postgres is running, obviously. We error out when that is attempted. If Postgres is not running though, we can change the port and at next Postgres start we'll use the new one.

src/bin/pg_autoctl/cli_common.c
```diff
@@ -1311,3 +1321,168 @@ cli_pg_autoctl_reload(const char *pidfile)
...
+	static struct option long_options[] = {
+		{ "pgdata", required_argument, NULL, 'D' },
+		{ "nodename", required_argument, NULL, 'n' },
+		{ "hostname", required_argument, NULL, 'h' },
+		{ "pgport", required_argument, NULL, 'p' },
+		{ "json", no_argument, NULL, 'J' },
+		{ "version", no_argument, NULL, 'V' },
+		{ "verbose", no_argument, NULL, 'v' },
+		{ "quiet", no_argument, NULL, 'q' },
+		{ "help", no_argument, &help_flag, 0 },
```
JelteF, line 1346: For consistency and usability sake. I think it's nice to keep `-h` as an alias for `--help`. We can use `H` for hostname I guess.

DimCitus: Well that's the thing, `'H'` is already `--pghost`, the hostname we use to connect to the local instance of Postgres (typically we use Unix sockets in either `/var/run/postgres` or `/tmp` depending on the platform). But we don't document the short options anyway, and even mentioned getting rid of them at some point maybe?

JelteF: I think it's better to keep `'h'` for help then, and use a random other letter (like `'N'`) for hostname? As long as we don't expose it through the shortflag string I think that's fine.

JelteF:
I just did some basic testing of this feature. I have two notes:
- [ ] I think one thing that would very much improve it is making `--nodename` optional by making it a nullable column. That way people can choose if they want to name their nodes or not. 
- [ ] There should be an upgrade path for current users.

src/bin/pg_autoctl/cli_service.c
```diff
@@ -41,9 +41,12 @@ static void cli_service_stop(int argc, char **argv);
 CommandLine service_run_command =
 	make_command("run",
 				 "Run the pg_autoctl service (monitor or keeper)",
-				 CLI_PGDATA_USAGE,
-				 CLI_PGDATA_OPTION,
-				 cli_getopt_pgdata,
+				 " [ --pgdata --nodename --hostname --pgport ] ",
```
JelteF, line 44: CLI_PGDATA_USAGE contains `--json`  is that ommited here on purpose?

src/bin/pg_autoctl/cli_common.c
DimCitus: Fair, doing that soon, thanks!

DimCitus:
I'm now thinking the default `--nodename` value could be `sformat("node_%d", nodeId)` ; if we do that we solve both remarks. There's a default, and an easy upgrade path. And the column can remain not-null.

JelteF:
> I'm now thinking the default --nodename value could be sformat("node_%d", nodeId) ; if we do that we solve both remarks. There's a default, and an easy upgrade path. And the column can remain not-null.

That's also a good way of solving it indeed.

DimCitus:
Superseded by PR #340

PR #250, opened 2020-04-28
Use log_warn instead of log_info when pg_autoctl is not running

JelteF:
This makes it extra clear to users that additional actions are needed to
complete the setup.

### DIFF
--- a/src/bin/pg_autoctl/cli_enable_disable.c
+++ b/src/bin/pg_autoctl/cli_enable_disable.c
@@ -708,8 +708,8 @@ cli_enable_ssl(int argc, char **argv)
 
 				if (!reloadedService)
 				{
-					log_error("Failed to reload the pg_autoctl, consider "
-							  "restarting it to implement the SSL changes");
+					log_warn("Failed to reload the pg_autoctl, consider "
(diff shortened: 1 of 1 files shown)


src/bin/pg_autoctl/cli_enable_disable.c
```diff
@@ -840,8 +840,8 @@ cli_enable_ssl(int argc, char **argv)
 			}
 			else
 			{
-				log_info("  pg_autoctl service is not running, changes "
-						 "will only apply at next start of pg_autoctl");
+				log_error("  pg_autoctl service is not running, changes "
```
DimCitus, line 843: Should this one be a WARN message really? I mean there's a use case to changing the SSL settings while in maintenance or something...

src/bin/pg_autoctl/cli_enable_disable.c
```diff
@@ -732,8 +732,8 @@ cli_enable_ssl(int argc, char **argv)
...
+						  "will only apply at next start of pg_autoctl");
```
DimCitus, line 736: Same as the other comment actually.

Issue #248, opened 2020-04-24
Add test covering upgrade from self-signed SSL cert to full certs


DimCitus:
See comments in there:
https://github.com/citusdata/pg_auto_failover/pull/233#discussion_r414607966

PR #251, opened 2020-04-28
Add SSL tests to upgrade from require to verify-ca.

DimCitus:
Work in progress, refactoring only at the moment.

See #248.

### DIFF
--- a/tests/pgautofailover_utils.py
+++ b/tests/pgautofailover_utils.py
@@ -610,6 +610,38 @@ class PGNode:
         command = PGAutoCtl(self, argv=ssl_args)
         out, err = command.execute("enable ssl")
 
+    def get_monitor_uri(self):
+        """
+        pg_autoctl show uri --monitor
+        """
+        command = PGAutoCtl(self)
+        out, err = command.execute("show uri --monitor",
+                                   'show', 'uri', '--monitor')
+        return out
+
+    def get_formation_uri(self, formationName='default'):
+        """
+        pg_autoctl show uri --monitor
+        """
+        command = PGAutoCtl(self)
+        out, err = command.execute("show uri --formation",
+                                   'show', 'uri', '--formation', formationName)
+        return out
+
+    def check_conn_string_ssl(self, conn_string, sslmode):
+        """
+        Asserts that given connection string embeds expected SSL settings.
+        """
+        crl = None
+        rootCert = self.sslCAFile
+
+        print("checking connstring =", conn_string)
+        assert f"sslmode={sslmode}" in conn_string
+        if rootCert:
+            assert f"sslrootcert={rootCert}" in conn_string
+        if crl:
+            assert f"sslcrl={crl}" in conn_string
+
     def check_ssl(self, ssl, sslmode, monitor=False, primary=False):
         """
         Checks if ssl settings match how the node is set up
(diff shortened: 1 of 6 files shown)


tests/pgautofailover_utils.py
```diff
@@ -225,10 +232,19 @@ def connection_string(self):
...
+            sslMode = self.sslMode
+        elif self.sslSelfSigned:
+            sslMode = "require"
+        else:
+            sslMode = "prefer"
+
+        dsn += "?sslmode=%s" % sslMode
+
+        if self.sslCAFile:
+            dsn += f"&sslrootcert={self.sslCAFile}"
```
DimCitus, line 247: Do we need to add the client root certificate and revocation list when those exist?

tests/pgautofailover_utils.py
```diff
@@ -594,6 +610,71 @@ def enable_ssl(self, sslMode=None, sslSelfSigned=None,
...
+            "ECDHE-RSA-AES128-SHA256:"
+            "ECDHE-RSA-AES256-SHA384"
+        )
+
+        # TODO: also do this for monitor once we can have superuser access to
+        # the monitor
+        if not monitor:
+            eq_(self.pg_config_get("ssl_ciphers"), expected_ciphers)
+
+        def check_conn_string(conn_string):
```
DimCitus, line 644: I think this function should be defined as a first-class method rather than dynamically again every time we call check_ssl ; it does not seem to close over any local definitions in a way that we need this nested construct...

tests/pgautofailover_utils.py
```diff
@@ -594,6 +610,71 @@ def enable_ssl(self, sslMode=None, sslSelfSigned=None,
...
+            print("checking connstring =", conn_string)
+            assert f"sslmode={sslmode}" in conn_string
+            if rootCert:
+                assert f"sslrootcert={rootCert}" in conn_string
+            if crl:
+                assert f"sslcrl={crl}" in conn_string
+
+        conn_string, _ = PGAutoCtl(self)\
+            .execute("show uri --monitor", 'show', 'uri', '--monitor')
```
DimCitus, line 653: What about using `self.monitor_uri()` instead, to clean-up the code and make it easier to reuse?

tests/pgautofailover_utils.py
```diff
@@ -594,6 +610,71 @@ def enable_ssl(self, sslMode=None, sslSelfSigned=None,
...
+        check_conn_string(conn_string)
+        if not monitor:
+            check_conn_string(self.config_get("pg_autoctl.monitor"))
+            conn_string, _ = PGAutoCtl(self)\
+                .execute("show uri --monitor", 'show', 'uri', '--formation', 'default')
```
DimCitus, line 658: Same with `self.formation_uri()` maybe?

tests/pgautofailover_utils.py
```diff
@@ -594,6 +610,71 @@ def enable_ssl(self, sslMode=None, sslSelfSigned=None,
...
+                ("ssl_ca_file", "ssl.ca_file", rootCert)]:
+            if file_path is None:
+                continue
+            assert os.path.isfile(file_path)
+            print("checking", pg_setting)
+            eq_(self.pg_config_get(pg_setting), file_path)
+            eq_(self.config_get(autoctl_setting), file_path)
+
+        if monitor or primary:
+            return
```
DimCitus, line 673: Then maybe the parameter, instead of primary, could be `checkPrimaryConninfo` or something like this?

tests/pgautofailover_utils.py
DimCitus: Done.

tests/pgautofailover_utils.py
DimCitus: Done.

tests/pgautofailover_utils.py
DimCitus: Done.

tests/pgautofailover_utils.py
```diff
@@ -610,6 +610,38 @@ def enable_ssl(self, sslMode=None, sslSelfSigned=None,
...
+        pg_autoctl show uri --monitor
+        """
+        command = PGAutoCtl(self)
+        out, err = command.execute("show uri --monitor",
+                                   'show', 'uri', '--monitor')
+        return out
+
+    def get_formation_uri(self, formationName='default'):
+        """
+        pg_autoctl show uri --monitor
```
JelteF, line 624:
```suggestion
        pg_autoctl show uri --formation {formationName}
```

PR #256, opened 2020-04-30
Make sure old pgautofailover files are removed before install

JelteF:
We do the same in the citus repo. This makes sure that in dev environments (or
building from source normally) old sql migration files are not accidentally
left behind.

### DIFF
--- a/src/monitor/Makefile
+++ b/src/monitor/Makefile
@@ -20,7 +20,14 @@ PG_CONFIG ?= pg_config
 PGXS = $(shell $(PG_CONFIG) --pgxs)
 USE_PGXS = 1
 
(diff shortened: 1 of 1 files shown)


src/monitor/Makefile
```diff
@@ -20,7 +20,14 @@ PG_CONFIG ?= pg_config
 PGXS = $(shell $(PG_CONFIG) --pgxs)
 USE_PGXS = 1
 
+.PHONY: clenaup-before-install
```
DimCitus, line 23: Typo in cleanup ; also, what about naming it `maintainer-clean` as in Postgres upstream?

PR #257, opened 2020-05-04
Prevent manual failovers when there's no node currently "primary".

DimCitus:
Otherwise our state machine gets confused. Also it does not make sense to
allow for manual failover when the current state of the cluster is not a
stable "primary/secondary" state anyway.

### DIFF
--- a/src/monitor/node_active_protocol.c
+++ b/src/monitor/node_active_protocol.c
@@ -943,8 +943,7 @@ perform_failover(PG_FUNCTION_ARGS)
 	AutoFailoverNode *primaryNode = NULL;
 	AutoFailoverNode *secondaryNode = NULL;
 
-	List *primaryStates = list_make2_int(REPLICATION_STATE_PRIMARY,
-										 REPLICATION_STATE_WAIT_PRIMARY);
+	List *primaryStates = list_make1_int(REPLICATION_STATE_PRIMARY);
 	List *secondaryStates = list_make2_int(REPLICATION_STATE_SECONDARY,
 										   REPLICATION_STATE_CATCHINGUP);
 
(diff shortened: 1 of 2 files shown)


src/monitor/node_active_protocol.c
```diff
@@ -943,8 +943,7 @@ perform_failover(PG_FUNCTION_ARGS)
 	AutoFailoverNode *primaryNode = NULL;
 	AutoFailoverNode *secondaryNode = NULL;
 
-	List *primaryStates = list_make2_int(REPLICATION_STATE_PRIMARY,
-										 REPLICATION_STATE_WAIT_PRIMARY);
+	List *primaryStates = list_make1_int(REPLICATION_STATE_PRIMARY);
 	List *secondaryStates = list_make2_int(REPLICATION_STATE_SECONDARY,
 										   REPLICATION_STATE_CATCHINGUP);
```
JelteF, line 948: When is it now possible that the secondary is "catchingup"? Only in the future with multiple standbys? If it's not possible even with multiple standbys I guess we should  remove this state too.

JelteF: Thinking about this more, I think we should actually remove this. For multiple standbys we should not failover to a "catchingup" node, but to an up to date secondary.

DimCitus: Because of the way the node_active protocol works it may be that once the monitor has assigned primary and secondary to both nodes, then the primary reports first, and later the secondary. In that case we can have primary/secondary assigned but primary/catchingup reported, and we may still proceed with the failover.

DimCitus: Per discussion, this is now commented properly (I believe) and simplified down to something that's easy to make sense of.

PR #266, opened 2020-05-13
Remove useless lines caused by declaration-after-statement warning

JelteF:
This is done using the same automated technique introduced for the Citus
repo in https://github.com/citusdata/citus/pull/3181. The main idea is
to replace:

```c
int a = 0;
// some code
a = 1;
```

with:

```c
// some code
int a = 1;
```

For details of how this works see the above mentioned Citus PR.

This comment about not doing early an early return to limit the scope of
variables caused me to add this here too:
https://github.com/citusdata/pg_auto_failover/pull/244#discussion_r424306388
With this commit merged, scope of variables will always be minimal in
cases where an early return is possible.

To make sure merge conflict handling is easy the following steps should be done:
1. Before merging this, all open PRs should be rebased on master.
2. Then this can be merged.
3. Then all PRs should be rebased on master again.
4. If merge conflicts occur (which is likely), they can be resolved trivially
   by doing `git checkout --theirs src` to ignore the changes caused by this PR.
   Then running `ci/remove_useless_declarations.sh` will redo all the changes.

To further limit impact, this PR can be merged right before a release, to make
sure most open PRs have been merged. This PR can be trivially updated as well by doing:
```
git rebase
# if merge conflict run
git checkout --ours src
```

### DIFF
--- a/src/bin/pg_autoctl/Makefile
+++ b/src/bin/pg_autoctl/Makefile
@@ -39,6 +39,7 @@ DEFAULT_CFLAGS += -Wall
 DEFAULT_CFLAGS += -Werror=implicit-int
 DEFAULT_CFLAGS += -Werror=implicit-function-declaration
 DEFAULT_CFLAGS += -Werror=return-type
+DEFAULT_CFLAGS += -Wno-declaration-after-statement
 
 # Needed for OSX
 DEFAULT_CFLAGS += -Wno-missing-braces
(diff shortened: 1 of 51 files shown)


DimCitus (approved):
I think it would be best if `src/tools/remove_useless_declarations.sh` would just work from the top-level directory, so we might want to either add a Makefile target or use `git rev-parse --show-toplevel` or such so that the script always work, whatever the current working directory happens to be when it's invoked.

Otherwise we need to remember more things about it, and we just won't use it, and when we want to it won't work.

LGTM otherwise: as mentioned in conversation, many of the changes in the PR are not very useful, and I'm quite neutral about it. That said, some of the changes are net improvements, and I don't feel like blocking the PR because of the neutral changes: they are neutral after all.

PR #271, opened 2020-05-19
Make sure teardown does not get stuck

JelteF:
This is done by putting some more (aggressive) timeouts in places that
should not take long. Without this teardown would get stuck on test failure,
because then our cleanup commands would get stuck.

### DIFF
--- a/tests/network.py
+++ b/tests/network.py
@@ -24,7 +24,9 @@ def managed_nspopen(*args, **kwds):
             # send SIGKILL to the process and wait for it to die if it's still
             # running
             proc.kill()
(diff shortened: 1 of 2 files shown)


JelteF:
They are a little randomly chosen indeed, if there's issues we can change them. It's mainly to give it more time than reasonably needed for when travis is having a bad day, but not get stuck for minutes/forever on something that will never finish. Both the destroy command and the `kill` command should be quite quick. Especcially the `kill`, more than half a second is already quite a bit more than expected there. `kill -9` should be basically instant.

PR #273, opened 2020-05-25
Rename --nodename to --hostname, and nodename to nodehost.

DimCitus:
This allows to later introduce a --nodename that is a user given name to
nodes in the cluster, and would default to `node_{id}`. This first step only
renames internal variables, column names, and option names.


JelteF:
A few things:
1. Can you split the completely automated regex replace and other changes in separate commits
2. Can you share the commands that you used for the automated replacement
3. Running `rg nodename -i` (case insensitive [ripgrep](https://github.com/BurntSushi/ripgrep)) still returns quite some matches. Some in `.rst` files, but also in `src/bin/pg_autoctl/monitor.c` and some others. Can you either modify the automation command to either also replace in those places, or can you explain why we still want nodename there?

JelteF:
Also, what's the reason sometimes it uses `nodeHost` and sometimes `hostname` in the replacement?

JelteF:
For future reference, this replaces all usages in code and most in the docs:
```bash
#!/bin/bash
sed -i 's/NodeName/NodeHost/g'   src/monitor/{*.{c,h,sql},expected/*.out,sql/monitor.sql}
sed -i 's/nodeName/nodeHost/g'   src/monitor/{*.{c,h,sql},expected/*.out,sql/monitor.sql}
sed -i 's/nodename/nodehost/g'   src/monitor/{*.{c,h,sql},expected/*.out,sql/monitor.sql}
sed -i 's/NODENAME/NODEHOST/g'   src/monitor/{*.{c,h,sql},expected/*.out,sql/monitor.sql}
sed -i 's/NODE_NAME/NODE_HOST/g' src/monitor/{*.{c,h,sql},expected/*.out,sql/monitor.sql}
sed -i 's/nodeName/hostName/g'   tests/*.py src/bin/pg_autoctl/*.{c,h}
sed -i 's/nodename/hostname/g'   tests/*.py src/bin/pg_autoctl/*.{c,h}
sed -i 's/NodeName/HostName/g'   tests/*.py src/bin/pg_autoctl/*.{c,h}
sed -i 's/NODENAME/HOSTNAME/g'   tests/*.py src/bin/pg_autoctl/*.{c,h}
sed -i 's/NODE_NAME/HOST_NAME/g' tests/*.py src/bin/pg_autoctl/*.{c,h}
sed -i 's/--nodename/--hostname/g' docs/{,ref/}*.rst README.md
```
This only leaves ~15 `nodename` usages in the docs, but I think it's safer to replace those by hand, since some might have to be changed to `nodehost` instead of `hostname`.

DimCitus:
> For future reference, this replaces all usages in code and most in the docs:

Thanks for taking the time to write the script at the time. I have now rebased this PR on top of the current main branch, keeping only the manual documentation changes, and then I prepared the `discover_nodename` function manually so that we could apply your script and still compile the project, and then applied your script. The current state of this PR is the result of those actions.

PR #274, opened 2020-05-27
Implement semaphores to lock access to stderr when logging.

DimCitus:


jonels-msft:
Any reason to prefer the implementation in this PR over using [POSIX semaphores](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/semaphore.h.html)? Are those not available on all platforms that we're targeting?

Also if the goal is just locking stderr, would the [stdio locking functions](https://pubs.opengroup.org/onlinepubs/9699919799/functions/flockfile.html) suffice?

DimCitus:
Finally we're using SysV semaphores in #267 directly, thanks to SEM_UNDO and IPC_PRIVATE uses that allow for better resource management and clean-up.

PR #275, opened 2020-05-28
Use a private semaphore

JelteF:
The semaphore ID is passed to child processes using a environment variable.
This way we never conflict with other semaphores.

We need to either remove the semaphore cleanup or we should store the semaphore
id in the pidfile too.

### DIFF
--- a/src/bin/pg_autoctl/defaults.h
+++ b/src/bin/pg_autoctl/defaults.h
@@ -23,8 +23,8 @@
 #define PG_AUTOCTL_DEBUG "PG_AUTOCTL_DEBUG"
 #define PG_AUTOCTL_EXTENSION_VERSION_VAR "PG_AUTOCTL_EXTENSION_VERSION"
 
-/* environment variable for logging semaphore and locks */
-#define PG_AUTOCTL_SERVICE "PG_AUTOCTL_SERVICE"
+/* environment variable for containing the id of the logging semaphore */
+#define PG_AUTOCTL_LOG_SEMAPHORE "PG_AUTOCTL_LOG_SEMAPHORE"
 
 /* default values for the pg_autoctl settings */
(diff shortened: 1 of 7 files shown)


DimCitus:
Using IPC_PRIVATE rather than the pid seems like an improvement. In that case I would rather place the semaphore id in the pid file and use the pid file (second line) as a way of communicating the identifier to the sub-processes. After all, the pid file exists when we start the sub-processes.

In the same vein we might also be able to use `ftok(config.pathnames.pid, <proj_id>)` to protect access to the pidfile for creation / removal, if we have a way to always release the semaphore when the process dies.

JelteF:
If we would communicate to the subprocesseses through the pidfile instead of environment variable, we would have to parse the command line arguments (at least `--pgdata`). Which you noted we don't want, because we to have the logging synchronized before that already.

JelteF:
My storing it in the pidfile idea was more for cleanup of the semaphore after unclean exit and restart.

DimCitus:
Ah yeah, we don't want to have to parse the command line to decide which semaphore to use, you're right of course. We should still publish to the pidfile and use it at clean-up, I believe. Exposing the semaphore key to the pid file also allows the users to use the commands `ipcs` and `ipcrm` should they need to.

So we might want to improve our `pg_autoctl show file --pid --content [ --json ]` command to that effect, make it more obvious what is what in that file.

DimCitus:
Reviewing this idea again, I don't see how to attach the existing semaphore created with IPC_PRIVATE. We need to use the same key from all processes. Before this PR the key is the parent pid, after this process the key is unknown. At create time `semget(2)` returns the semaphore set identifier, not the key.

DimCitus:
Well it seems calling `semget(2)` when you already have the semaphore set id is not necessary and we can proceed directly.

Issue #101, opened 2019-11-08
Register an existing standby node


DimCitus:
At the moment our command `pg_autoctl create postgres` knows how to register a pre-existing PostgreSQL instance when it's a primary server. This allows users to provision as usual and only then register to the monitor.

There are two main cases when registering an existing standby:

  1. we only have a pre-existing PGDATA directory with a base backup copy ready for recovery and connection
  2. the standby is already running and in recovery

In the first case we should check that the `system_identifier` is the same as in the primary server and then continue with the standby initialisation as usual, only skipping the `pg_basebackup` step.

In the second case, we should still check that the `system_identifier` is the same as in the primary server, and then verify that the setup is done according to pg_auto_failover expectations:

  1. list all the setup that we do when creating a standby (replication slot, primary_conninfo, password, HBA, PostgreSQL configuration, pg_stat_wal_receiver, etc)
  2. either implement a way to check that everything is in place in a way that allows us to implement a failover, or override the existing settings with the pg_auto_failover settings for a standby and restart the standby
  3. register the standby to the monitor either as WAIT_STANDBY (if the primary is not ready yet, which is visible from the standby by the lack of an entry in pg_stat_wal_receiver) or in CATCHINGUP state

This is a brain dump, the actual implementation might differ to a large extend with the thinking here, though I think this is the general idea we should pursue here.

MichaelDBA:
To me, this is a showstopper.  We simply cannot use pg_basebackup to create a standby since our size is about 30TB.  It would take pg_basebackup forever practically speaking!  The only tool that I know of that can restore VLDBs is pgbackrest.  Will consider this tool only when that practicality is available.  I don't know why this "enhancement" is not prioritized.

DimCitus:
Hi @MichaelDBA ; thanks for your feedback and interest in pg_auto_failover. This item has been prioritised to our next release, 1.4. We are currently finishing the packages for release 1.3. So be sure to try again soon!

MichaelDBA:
Thanks for the update.  Will check back in 1.4.  Will also peruse the code to see if I feel capable.

MichaelDBA:
Guess I can't close this issue, so feel free to do so.

DimCitus:
> Guess I can't close this issue, so feel free to do so.

[...]

PR #276, opened 2020-05-28
Allow adding a standby node from an exiting PGDATA directory.

DimCitus:
This allows users to use their favorite backup and restore options when
preparing to add a standby node to pg_auto_failover.

We still have a restriction that we don't know how to accept an already
running standby: we would need to check the primary conninfo authentication
options in use, the replication slot, etc. So we still bail out in that
case.

See #101.

### DIFF
--- a/src/bin/pg_autoctl/fsm_transition.c
+++ b/src/bin/pg_autoctl/fsm_transition.c
@@ -112,6 +112,11 @@ fsm_init_primary(Keeper *keeper)
 	if (initState->pgInitState == PRE_INIT_STATE_EMPTY &&
 		!postgresInstanceExists)
 	{
+		Monitor *monitor = &(keeper->monitor);
+		PostgresSetup newPgSetup = { 0 };
+		bool missingPgdataIsOk = false;
+		bool postgresNotRunningIsOk = true;
+
 		if (!pg_ctl_initdb(pgSetup->pg_ctl, pgSetup->pgdata))
 		{
 			log_fatal("Failed to initialise a PostgreSQL instance at \"%s\""
@@ -119,6 +124,18 @@ fsm_init_primary(Keeper *keeper)
 
 			return false;
 		}
+
+		if (!pg_setup_init(&newPgSetup,
+						   pgSetup,
+						   missingPgdataIsOk,
+						   postgresNotRunningIsOk))
+		{
+			/* errors have already been logged */
+			log_error("pg_setup_wait_until_is_ready: pg_setup_init is false");
+			return false;
+		}
+
+		*pgSetup = newPgSetup;
 
 		/*
 		 * We managed to initdb, refresh our configuration file location with
@@ -128,6 +145,18 @@ fsm_init_primary(Keeper *keeper)
 		if (!keeper_config_update_with_absolute_pgdata(&(keeper->config)))
 		{
 			/* errors have already been logged */
+			return false;
+		}
+
+		/*
+		 * We have a new system_identifier, we need to publish it now.
+		 */
+		if (!monitor_set_node_system_identifier(
+				monitor,
+				keeper->state.current_node_id,
+				pgSetup->control.system_identifier))
+		{
+			log_error("Failed to update the new node system_identifier");
 			return false;
 		}
 	}
@@ -670,6 +699,8 @@ fsm_init_standby(Keeper *keeper)
 	int groupId = keeper->state.current_group;
 	NodeAddress *primaryNode = NULL;
 
+	bool skipBaseBackup = false;
+
 	/* get the primary node to follow */
 	if (!config->monitorDisabled)
 	{
@@ -701,7 +732,10 @@ fsm_init_standby(Keeper *keeper)
 		return false;
 	}
 
-	if (!standby_init_database(postgres, config->nodename))
+	skipBaseBackup = file_exists(keeper->config.pathnames.init) &&
+					 keeper->initState.pgInitState == PRE_INIT_STATE_EXISTS;
+
+	if (!standby_init_database(postgres, config->nodename, skipBaseBackup))
 	{
 		log_error("Failed initialise standby server, see above for details");
 		return false;
@@ -766,10 +800,12 @@ fsm_rewind_or_init(Keeper *keeper)
 
 	if (!primary_rewind_to_standby(postgres))
 	{
+		bool skipBaseBackup = false;
+
 		log_warn("Failed to rewind demoted primary to standby, "
 				 "trying pg_basebackup instead");
 
-		if (!standby_init_database(postgres, config->nodename))
+		if (!standby_init_database(postgres, config->nodename, skipBaseBackup))
 		{
 			log_error("Failed to become standby server, see above for details");
 			return false;
--- a/src/bin/pg_autoctl/fsm_transition.c
+++ b/src/bin/pg_autoctl/fsm_transition.c
@@ -751,6 +751,18 @@ fsm_init_standby(Keeper *keeper)
 		return false;
 	}
 
+	/*
+	 * Publish our possibly new system_identifier now.
+	 */
+	if (!monitor_set_node_system_identifier(
+			monitor,
+			keeper->state.current_node_id,
+			postgres->postgresSetup.control.system_identifier))
+	{
+		log_error("Failed to update the new node system_identifier");
+		return false;
+	}
+
 	/* ensure the SSL setup is synced with the keeper config */
 	if (!keeper_create_self_signed_cert(keeper))
 	{
--- a/src/bin/pg_autoctl/fsm_transition.c
+++ b/src/bin/pg_autoctl/fsm_transition.c
@@ -732,6 +732,16 @@ fsm_init_standby(Keeper *keeper)
 		return false;
 	}
 
+	/*
+	 * At pg_autoctl create time when PGDATA already exists and we were
+	 * successful in registering the node, then we can proceed without a
+	 * pg_basebackup: we already have a copy of PGDATA on-disk.
+	 *
+	 * The existence of PGDATA at pg_autoctl create time is tracked in our init
+	 * state as the PRE_INIT_STATE_EXISTS enum value. Once init is finished, we
+	 * remove our init file: then we need to pg_basebackup again to init a
+	 * standby.
+	 */
 	skipBaseBackup = file_exists(keeper->config.pathnames.init) &&
 					 keeper->initState.pgInitState == PRE_INIT_STATE_EXISTS;
 
--- a/src/bin/pg_autoctl/keeper_pg_init.c
+++ b/src/bin/pg_autoctl/keeper_pg_init.c
@@ -7,6 +7,7 @@
  *
  */
 
+#include <inttypes.h>
 #include <stdbool.h>
 #include <unistd.h>
 
@@ -43,6 +44,7 @@ bool keeperInitWarnings = false;
 
 static bool keeper_pg_init_fsm(Keeper *keeper);
 static bool keeper_pg_init_and_register(Keeper *keeper);
+static bool keeper_pg_init_and_register_primary(Keeper *keeper);
 static bool reach_initial_state(Keeper *keeper);
 static bool wait_until_primary_is_ready(Keeper *config,
 										MonitorAssignedState *assignedState);
@@ -117,9 +119,11 @@ keeper_pg_init_and_register(Keeper *keeper)
 	 * The initial state we may register in depend on the current PostgreSQL
 	 * instance that might exist or not at PGDATA.
 	 */
-	PostgresSetup pgSetup = config->pgSetup;
-	bool postgresInstanceExists = pg_setup_pgdata_exists(&pgSetup);
-	bool postgresInstanceIsPrimary = pg_setup_is_primary(&pgSetup);
+	PostgresSetup *pgSetup = &(config->pgSetup);
+	bool postgresInstanceExists = pg_setup_pgdata_exists(pgSetup);
+	bool postgresInstanceIsRunning = pg_setup_is_running(pgSetup);
+	PostgresRole postgresRole = pg_setup_role(pgSetup);
+	bool postgresInstanceIsPrimary = postgresRole == POSTGRES_ROLE_PRIMARY;
 
 	if (postgresInstanceExists)
 	{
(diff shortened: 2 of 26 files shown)


JelteF:
This is missing a test

src/monitor/node_active_protocol.c
```diff
@@ -108,6 +110,11 @@ register_node(PG_FUNCTION_ARGS)
 
 	checkPgAutoFailoverVersion();
 
+	if (!PG_ARGISNULL(4))
```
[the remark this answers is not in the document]
DimCitus: Yes that's why I got it back to being STRICT finally, and use zero instead of NULL to show that we don't have the system identifier yet. Not ideal, but allowed to solve all the other things to solve in that PR, might revisit later.

This is a left-over.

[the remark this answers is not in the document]
DimCitus: Fixed now.

DimCitus:
> This is missing a test

Test added, we have test_auth.py now create its standby from a `cp -a` of the primary's PGDATA.

JelteF:
I found a weird bug here. Reproducing can be done this way:

Shell 1 (monitor):
```bash
pg_autoctl create monitor --pgdata monitor --ssl-self-signed --auth trust --run
```
Shell 2 (primary):
```bash
pg_autoctl create postgres --pgdata node1 --monitor $(pg_autoctl show uri --pgdata monitor --monitor) --pgport 5433 --auth trust --ssl-self-signed --run
```
Shell 3 (secondary):
```bash
cp node1 -R node2
rm node2/postmaster.pid
pg_autoctl create postgres --pgdata node2 --monitor $(pg_autoctl show uri --pgdata monitor --monitor) --pgport 5434 --run --auth trust --ssl-self-signed
# successfully uses directory
Ctrl+C
pg_autoctl drop node --destroy --pgdata node2
initdb node2
# Without this next command I get the following error in node2/log/*.log
# FATAL:  hot standby is not possible because max_wal_senders = 4 is a lower setting than on the master server (its value was 10)
# I think this might be a separate bug
echo 'max_wal_senders = 10' >> node2/postgresql.conf
pg_autoctl create postgres --pgdata node2 --monitor (pg_autoctl show uri --pgdata monitor --monitor) --pgport 5434 --run --auth trust --ssl-self-signed
```
At this point `node2` is a secondary according to the monitor, but has this error in the logs:
```
2020-06-02 14:44:28.102 CEST [19079] DETAIL:  The primary's identifier is 6833727098633262131, the standby's identifier is 6833728377217151671.
2020-06-02 14:44:33.107 CEST [19134] FATAL:  database system identifier differs between the primary and standby
```

DimCitus:
> I found a weird bug here. Reproducing can be done this way:

Thanks for this extra testing!

> # Without this next command I get the following error in node2/log/*.log
> # FATAL:  hot standby is not possible because max_wal_senders = 4 is a lower setting than on the master server (its value was 10)
> # I think this might be a separate bug

Well our DEFAULT_GUC_SETTINGS_FOR_PG_AUTO_FAILOVER contains `{ "max_wal_senders", "4" }` and the default in Postgres itself has been raised to 10 recently. I think that in the Multiple Standby PR we should revisit our default to be something like 12 at least.

> At this point `node2` is a secondary according to the monitor, but has this error in the logs:
> 
> ```
> 2020-06-02 14:44:28.102 CEST [19079] DETAIL:  The primary's identifier is 6833727098633262131, the standby's identifier is 6833728377217151671.
> 2020-06-02 14:44:33.107 CEST [19134] FATAL:  database system identifier differs between the primary and standby
> ```

This should never happen because of the constraint we have in the schema. And that happened anyway because I went too fast this morning when cleaning-up the `PG_ARGISNULL()` call: the system identifier was initialised to 0 (zero) always instead of `PG_GETARG_INT64(4)`. Fixed now.

Here's what it looks like at the moment:

```
15:21:24 81897 INFO  Registering Postgres system 6833739975658110910 found at "/private/tmp/plop/b"
15:21:24 81897 ERROR Monitor ERROR:  conflicting key value violates exclusion constraint "node_groupid_sysidentifier_excl"
15:21:24 81897 ERROR Monitor DETAIL:  Key (groupid, sysidentifier)=(0, 6833739975658110910) conflicts with existing key (groupid, sysidentifier)=(0, 6833738207436515421).
15:21:24 81897 ERROR Monitor CONTEXT:  SQL statement "INSERT INTO pgautofailover.node (formationid, groupid, nodename, nodeport, sysidentifier,   goalstate, reportedstate, candidatepriority, replicationquorum) VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9) RETURNING nodeid"
15:21:24 81897 ERROR SQL query: SELECT * FROM pgautofailover.register_node($1, $2, $3, $4, $5, $6, $7::pgautofailover.replication_state, $8, $9, $10)
15:21:24 81897 ERROR SQL params: 'default', 'localhost', '5002', 'postgres', '6833739975658110910', '-1', 'init', 'standalone', '100', 'true'
15:21:24 81897 ERROR Failed to register node localhost:5002 in group -1 of formation "default" with initial state "init", see previous lines for details
15:21:24 81897 ERROR Failed to register the existing local Postgres node "localhost:5002" running at "/private/tmp/plop/b"to the pg_auto_failover monitor at postgres://autoctl_node@localhost:5000/pg_auto_failover?sslmode=prefer, see above for details
```

src/monitor/pgautofailover.sql
```diff
@@ -97,6 +97,7 @@ CREATE TABLE pgautofailover.node
     groupid              int not null,
     nodename             text not null,
     nodeport             int not null,
+    sysidentifier        bigint,
```
JelteF, line 100: Can we add a simple CHECK constraint here that this can only be `null` whenever the `reportedstate` is `init` or something? In all other cases this should never be `NULL` afaict.

src/monitor/pgautofailover.sql
```diff
@@ -111,6 +112,7 @@ CREATE TABLE pgautofailover.node
     replicationquorum	 bool not null default true,
 
     UNIQUE (nodename, nodeport),
+    EXCLUDE USING gist(groupid with =, sysidentifier with <>),
```
JelteF, line 115: Can we add a comment explaining the goal of this constraint? You need quite some background to understand this without context. Maybe giving the constraint a user friendly name is also nice, since it is displayed in the error message.

src/bin/pg_autoctl/pgsetup.h
```diff
@@ -56,6 +56,24 @@ typedef enum
 	POSTMASTER_STATUS_STANDBY
 } PostmasterStatus;
 
+/*
+ * When discovering Postgres we try to determine if the local $PGDATA directory
+ * belons to a primary or a secondary server. If the server is running, it's
```
JelteF, line 61:
```suggestion
 * belongs to a primary or a secondary server. If the server is running, it's
```

src/bin/pg_autoctl/pgsetup.c
```diff
@@ -1157,22 +1157,101 @@ pg_setup_wait_until_is_stopped(PostgresSetup *pgSetup, int timeout, int logLevel
  * pg_setup_is_primary returns true when the local PostgreSQL instance is known
  * to not be recovery.
  */
-bool
-pg_setup_is_primary(PostgresSetup *pgSetup)
+PostgresRole
+pg_setup_role(PostgresSetup *pgSetup)
```
JelteF, line 1161: Function comment is outdated

src/monitor/pgautofailover.sql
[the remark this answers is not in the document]
DimCitus: Yes. We have other changes coming with an upgrade script. Do we want to start the effort in this PR of consolidate with a later one? I though a later one would do (`--hostname --nodename` is what I had in mind).

src/bin/pg_autoctl/fsm_transition.c
```diff
@@ -701,7 +732,10 @@ fsm_init_standby(Keeper *keeper)
 		return false;
 	}
 
-	if (!standby_init_database(postgres, config->nodename))
+	skipBaseBackup = file_exists(keeper->config.pathnames.init) &&
+					 keeper->initState.pgInitState == PRE_INIT_STATE_EXISTS;
```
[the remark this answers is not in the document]
DimCitus: Done now.

src/bin/pg_autoctl/keeper_pg_init.c
```diff
@@ -284,6 +274,74 @@ keeper_pg_init_and_register(Keeper *keeper)
...
+	 * which the user has to give us permission for with the command line
+	 * option --allow-removing-pgdata.
+	 *
+	 * When allowRemovingPgdata is true, we accept any role and thus register
+	 * with the expected state INIT (which is, we don't have any expectation
+	 * really).
+	 *
+	 * When allowRemovingPgdata is false, we expect to be registed in the
+	 * SINGLE state and publish that to the register_node() API call, so that
+	 * it fails on the server side when that's not possible.
```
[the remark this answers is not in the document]
DimCitus: We're not removing the feature, it should just work as before. There's quite some refactoring around it though, so the diff might not be easy to read?

src/monitor/pgautofailover.sql
DimCitus: Done.

src/bin/pg_autoctl/pgsetup.h
DimCitus: Done.

src/bin/pg_autoctl/pgsetup.c
DimCitus: Fixed.

DimCitus:
Now added better testing coverage and some bug fixes too.

src/bin/pg_autoctl/keeper_pg_init.c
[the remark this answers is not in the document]
JelteF: I might be missing a use case where this functionality is activated then. AFAICT this flag could be used in two cases before that we handle differently now:
1. When adding a standby that already had data with the same systemid. Now we simply use the data directory
2. When adding a standby that had data with a different systemid. Now we error out in this case.

If these are indeed the only use cases of this feature, I think it's safe to completely remove it. If not, can you explain which use case I'm missing?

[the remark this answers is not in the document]
DimCitus: The option `--allow-removing-pgdata` currently applies only to the case when you have a primary server (running or not) and you want to register it to the monitor in a group where we are expecting a standby. Are we allowed to remove the PGDATA directory and take it from there?

At the moment I had not though you could want to allow removing PGDATA on the standby when the system identifier does not match with the local one on a standby. We might want to allow for that indeed, and then try again I suppose. Within this PR it's the job of the user to `rm -rf` and try again.

tests/test_create_standby_with_pgdata.py
```diff
@@ -0,0 +1,86 @@
...
+    except Exception as e:
+        # we want to see the failure here
+        print(e)
+
+        print("Failed as expected, cleaning up")
+        print("rm -rf /tmp/sb-from-pgdata/node2")
+        p = subprocess.Popen(["sudo", "-E", '-u', os.getenv("USER"),
+                          'env', 'PATH=' + os.getenv("PATH"),
+                          "rm", "-rf", "/tmp/sb-from-pgdata/node2"])
+        assert(p.wait() == 0)
```
JelteF, line 58: Right now this test always succeeds, because of the `try` ... `except`. Something like this should be better, because then it actually fails when `node2.create()` doesn't raise an exception.
```suggestion
@raises(Exception)
def test_004_create_raises_error():
    try:
        node2.create()
    except Exception as e:
        # we want to see the failure here
        print(e)
        raise
        
def test_005_cleanup_after_failure():
        print("Failed as expected, cleaning up")
        print("rm -rf /tmp/sb-from-pgdata/node2")
        p = subprocess.Popen(["sudo", "-E", '-u', os.getenv("USER"),
                          'env', 'PATH=' + os.getenv("PATH"),
                          "rm", "-rf", "/tmp/sb-from-pgdata/node2"])
        assert(p.wait() == 0)
```

tests/test_create_standby_with_pgdata.py
```diff
@@ -0,0 +1,86 @@
...
+
+def test_004_init_secondary():
+    global node3
+
+    # create node2 from a manual copy of node1 to test creating a standby
```
JelteF, line 63:
```suggestion
    # create node3 from a manual copy of node1 to test creating a standby
```

tests/test_create_standby_with_pgdata.py
DimCitus: Good point. I played around with it but failed to come to the try/except/raise idea. Thanks!

src/bin/pg_autoctl/keeper_pg_init.c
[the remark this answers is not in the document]
DimCitus: After more discussion and thinking I just got rid of `--allow-removing-pgdata` which we can't really target anymore, thanks to the new monitor's constraints around system_identifier.

tests/test_create_standby_with_pgdata.py
```diff
@@ -0,0 +1,90 @@
...
+        
+def test_005_cleanup_after_failure():
+        print("Failed as expected, cleaning up")
+        print("rm -rf /tmp/sb-from-pgdata/node2")
+        p = subprocess.Popen(["sudo", "-E", '-u', os.getenv("USER"),
+                          'env', 'PATH=' + os.getenv("PATH"),
+                          "rm", "-rf", "/tmp/sb-from-pgdata/node2"])
+        assert(p.wait() == 0)
+
+def test_004_init_secondary():
```
JelteF, line 64:
```suggestion
def test_006_init_secondary():
```

tests/test_create_standby_with_pgdata.py
```diff
@@ -0,0 +1,90 @@
...
+
+    os.remove("/tmp/sb-from-pgdata/node3/postmaster.pid")
+
+    node3 = cluster.create_datanode("/tmp/sb-from-pgdata/node3")
+    node3.create()
+    node3.run()
+    assert node3.wait_until_state(target_state="secondary")
+    assert node1.wait_until_state(target_state="primary")
+
+def test_005_failover():
```
JelteF, line 85:
```suggestion
def test_007_failover():
```

harry-sallinggroup:
I am getting this error :

15:21:24 81897 INFO  Registering Postgres system 6833739975658110910 found at "/private/tmp/plop/b"
15:21:24 81897 ERROR Monitor ERROR:  conflicting key value violates exclusion constraint "node_groupid_sysidentifier_excl"
15:21:24 81897 ERROR Monitor DETAIL:  Key (groupid, sysidentifier)=(0, 6833739975658110910) conflicts with existing key (groupid, sysidentifier)=(0, 6833738207436515421).
15:21:24 81897 ERROR Monitor CONTEXT:  SQL statement "INSERT INTO pgautofailover.node (formationid, groupid, nodename, nodeport, sysidentifier,   goalstate, reportedstate, candidatepriority, replicationquorum) VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9) RETURNING nodeid"
15:21:24 81897 ERROR SQL query: SELECT * FROM pgautofailover.register_node($1, $2, $3, $4, $5, $6, $7::pgautofailover.replication_state, $8, $9, $10)
15:21:24 81897 ERROR SQL params: 'default', 'localhost', '5002', 'postgres', '6833739975658110910', '-1', 'init', 'standalone', '100', 'true'
15:21:24 81897 ERROR Failed to register node localhost:5002 in group -1 of formation "default" with initial state "init", see previous lines for details
15:21:24 81897 ERROR Failed to register the existing local Postgres node "localhost:5002" running at "/private/tmp/plop/b"to the pg_auto_failover monitor at postgres://autoctl_node@localhost:5000/pg_auto_failover?sslmode=prefer, see above for details

JelteF:
That means that the standby you're trying to add doesn't have a PGDATA that's a backup from the primary. Otherwise the system identifier of the database directory would be the same. 

Some background on the system identifier: https://www.2ndquadrant.com/en/blog/support-for-postgresqls-system-identifier-in-barman/

harry-sallinggroup:
Yes thanks for the quick reply.. Manage to get pass this issue :)

PR #280, opened 2020-06-03
Improve handling of SIGQUIT in pg_autoctl processes.

DimCitus:
The supervisor process should relay the urgency of the SIGQUIT signal down
to its sub-processes and it's now the case. Also, the intermediate Postgres
controller service should handle SIGQUIT properly: terminate Postgres, exit
with EXIT_CODE_QUIT (zero), and let the supervisor handle the situation.

WIP: will add `atexit()` calls to manage the resource cleanup (pidfile, semaphore).


DimCitus:
> Let's create a separate PR for `atexit`. The SIGQUIT change is pretty much separate and I think this is good to merge.

I think we need to first clean-up the SIGQUIT semantics and what we do there before being able to just register clean-up function at exit time, so that's not that separate, but we can still finish this one while I prepare the `atexit` one on-top of it.

Issue #272, opened 2020-05-25
Supervision handling of SIGQUIT.


DimCitus:
The supervisor does not handle SIGQUIT any different than SIGINT right now. So this only matters if a SIGQUIT is sent manually to the monitor subprocess. Maybe we should change the supervisor process to forward SIGQUIT to the child processes.

SIDENOTE: I just tried sending SIGQUIT manually to the postgres start/stop process, but that killed the pg_auto_ctl process and orphaned the actual postgres subprocess. Which seems like undesireable behaviour TBH.

DimCitus:
Also implement `atexit` semantics for pidfile removal and other cleanup (semaphore removal) on the supervisor process. Upon receiving SIGQUIT we call `exit()` which means that the `atexit` registered functions should be called, if my reading of the man pages is good.

PR #281, opened 2020-06-03
Move the log semaphore clean-up function to atexit().

DimCitus:
Note: We can't move the pidfile cleanup to the atexit stage easily enough because of using a local supervisor variable, and we don't really need to do that anyway because of the way we handle SIGQUIT and stale pidfiles anyway.

See #272.

### DIFF
--- a/src/bin/pg_autoctl/pgctl.c
+++ b/src/bin/pg_autoctl/pgctl.c
@@ -803,7 +803,7 @@ pg_ctl_initdb(const char *pg_ctl, const char *pgdata)
 
 	/* initdb takes time, so log about the operation BEFORE doing it */
 	log_info("Initialising a PostgreSQL cluster at \"%s\"", pgdata);
-	log_info("%s initdb -s -D %s", pg_ctl, pgdata);
+	log_info("%s initdb -s -D %s --option '--auth=trust'", pg_ctl, pgdata);
 
 	program = run_program(pg_ctl, "initdb",
 						  "--silent",
--- a/src/bin/pg_autoctl/pgctl.c
+++ b/src/bin/pg_autoctl/pgctl.c
@@ -803,6 +803,7 @@ pg_ctl_initdb(const char *pg_ctl, const char *pgdata)
 
 	/* initdb takes time, so log about the operation BEFORE doing it */
 	log_info("Initialising a PostgreSQL cluster at \"%s\"", pgdata);
+	log_info("%s initdb -s -D %s", pg_ctl, pgdata);
 
 	program = run_program(pg_ctl, "initdb",
 						  "--silent",
@@ -811,7 +812,6 @@ pg_ctl_initdb(const char *pg_ctl, const char *pgdata)
 	                      /* avoid warning message */
 						  "--option", "'--auth=trust'",
 						  NULL);
-	log_info("%s initdb -s -D %s", pg_ctl, pgdata);
 
 	success = program.returnCode == 0;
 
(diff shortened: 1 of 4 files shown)


src/bin/pg_autoctl/service_monitor_init.c
```diff
@@ -146,6 +146,17 @@ service_monitor_init_start(void *context, pid_t *pid)
...
+				 * We are in a sub-process and didn't call exec() on our
+				 * pg_autoctl do service listener program yet we do not want to
+				 * clean-up the semaphore just yet. Publish that we are a
+				 * sub-process and only then quit, avoiding to call the
+				 * atexit() semaphore clean-up function.
+				 */
+				IntString semIdString = intToString(log_semaphore.semId);
+
+				setenv(PG_AUTOCTL_LOG_SEMAPHORE, semIdString.strValue, 1);
+
```
JelteF, line 159: I think it makes sense to do this right away after every `fork` that returns `0` in the codebase.

src/bin/pg_autoctl/pgctl.c
```diff
@@ -803,6 +803,7 @@ pg_ctl_initdb(const char *pg_ctl, const char *pgdata)
 
 	/* initdb takes time, so log about the operation BEFORE doing it */
 	log_info("Initialising a PostgreSQL cluster at \"%s\"", pgdata);
+	log_info("%s initdb -s -D %s", pg_ctl, pgdata);
```
JelteF, line 806: Maybe also add `--option '-auth=trust'` here

JelteF:
Looks good overal. Regarding not adding the pidfile cleanup to `atexit`. I think it would be nice if we could modify the code to do it anyway. You can register multiple functions by calling `atexit` multiple times, so it's possible to only register it somewhere else in the code. If that is not the problem can you explain the problem a bit more then, possibly linking to the relevant code?

src/bin/pg_autoctl/service_monitor_init.c
DimCitus: That's another way to do it, but when `createAndRun` is true we are going to call `service_monitor_runprogram` which takes care of it anyway. And doing it in the child arm of the fork begs the question of always doing that in other fork call sites, and I don't think we should.

Sending a code revision following your comments so that we can see if we like it better that way or not.

JelteF: > And doing it in the child arm of the fork begs the question of always doing that in other fork call sites, and I don't think we should.

Why should we not do that? It seems to me that only the parent process should kill the semaphore `atexit`, so doing it after every `fork` call that indicates it's a child seems a good thing.

Maybe it would be clearer to set a global indicating if it is the parent process or not, and only do the semaphore unlinking in that case.

DimCitus:
> Looks good overal. Regarding not adding the pidfile cleanup to `atexit`. I think it would be nice if we could modify the code to do it anyway. You can register multiple functions by calling `atexit` multiple times, so it's possible to only register it somewhere else in the code. If that is not the problem can you explain the problem a bit more then, possibly linking to the relevant code?

The function registered to `atexit` does not take a parameter, and the context we need to handle the pidfile is not global, it's in a local variable. I'm not sure how to best refactor the code to make it possible to use `atexit` and not sure it is worth it.

src/bin/pg_autoctl/service_monitor_init.c
DimCitus: The problem we have here is that the signals and globals and many things are shared at `fork()` time and lost at `exec()` time. The `atexit` registrations have that policy, so that when we do our run program (which calls `exec`) we don't have the `atexit` functions anymore. 

There's only one place where we `fork()` and then we don't `exec()`, it's when doing `pg_autoctl create monitor` without the `--run` option, because we don't `exec()` the monitor installer.

PR #285, opened 2020-06-05
pg_autoctl show file ... --content --json

DimCitus:
And by default pg_autoctl show file now outputs a human-readable tabular format.

### DIFF
--- a/src/bin/pg_autoctl/cli_show.c
+++ b/src/bin/pg_autoctl/cli_show.c
@@ -57,6 +57,8 @@ static void print_all_uri(SSLOptions *ssl,
 						  Monitor *monitor,
 						  FILE *stream);
 
+static void fprint_pidfile_as_json(const char *pidfile);
+
 CommandLine show_uri_command =
 	make_command("uri",
 				 "Show the postgres uri to use to connect to pg_auto_failover nodes",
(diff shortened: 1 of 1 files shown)


src/bin/pg_autoctl/cli_show.c
```diff
@@ -1186,35 +1196,114 @@ cli_show_file(int argc, char **argv)
...
 
+				fformat(stdout, "%6s | %s\n", "Config", config.pathnames.config);
+
+				if (role == PG_AUTOCTL_ROLE_KEEPER)
+				{
+					fformat(stdout, "%6s | %s\n", "State", config.pathnames.state);
+					fformat(stdout, "%6s | %s\n", "Init", config.pathnames.init);
+				}
+				fformat(stdout, "%6s | %s\n", "Pid", config.pathnames.pid);
+				fformat(stdout, "\n");
```
JelteF, line 1235: nitpick: I think the table could look slightly nicer. Instead of this:
```
  File | Path
-------+----------------
Config | /home/jelte/.config/pg_autoctl/home/jelte/work/pg_auto_failover/monitor/pg_autoctl.cfg
   Pid | /tmp/pg_autoctl/home/jelte/work/pg_auto_failover/monitor/pg_autoctl.pid
```
It would look like this (note the extra space in front of "Config"):
```
   File | Path
--------+----------------
 Config | /home/jelte/.config/pg_autoctl/home/jelte/work/pg_auto_failover/monitor/pg_autoctl.cfg
    Pid | /tmp/pg_autoctl/home/jelte/work/pg_auto_failover/monitor/pg_autoctl.pid
```
Even nicer would be if the header line of `Path` would match up with the contents, i.e. getting longest length. But I do not think this is worth the effort tbh.

DimCitus: Fair enough, done! Well not the Path parts which I agree with you doesn't seem worth the effort, and also could be ugly if the header line spans larger than the terminal COLUMNS etc.

PR #286, opened 2020-06-05
Arrange to forward pg_basebackup output while it's running.

DimCitus:
This can be a pretty long operation and being in the dark for its whole
duration is not a great user experience.

### DIFF
--- a/src/bin/pg_autoctl/pgctl.c
+++ b/src/bin/pg_autoctl/pgctl.c
@@ -658,8 +658,8 @@ pg_basebackup(const char *pgdata,
 
 	/*
 	 * We do not want to call setsid() when running this program, as the
-	 * postgres subprogram is not intended to be its own session leader, but
-	 * remain a sub-process in the same group as pg_autoctl.
+	 * pg_basebackup subprogram is not intended to be its own session leader,
+	 * but remain a sub-process in the same group as pg_autoctl.
 	 */
 	program = initialize_program(args, false);
 	program.processBuffer = &processBufferCallback;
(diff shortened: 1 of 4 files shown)


src/bin/pg_autoctl/pgctl.c
```diff
@@ -633,29 +639,45 @@ pg_basebackup(const char *pgdata,
...
+	args[argsIndex++] = "--wal-method=stream";
+	args[argsIndex++] = "--slot";
+	args[argsIndex++] = replicationSource->slotName;
+	args[argsIndex] = NULL;
 
-	/* pg_basebackup uses stderr for all of its output */
-	(void) log_program_output(program, LOG_INFO, LOG_INFO);
+	/*
+	 * We do not want to call setsid() when running this program, as the
+	 * postgres subprogram is not intended to be its own session leader, but
```
JelteF, line 661:
```suggestion
	 * pg_basebackup subprogram is not intended to be its own session leader, but
```

src/bin/pg_autoctl/string_utils.c
```diff
@@ -443,3 +443,29 @@ splitLines(char *errorMessage, char **linesArray, int size)
...
+{
+	char *outLines[BUFSIZE] = { 0 };
+	int lineCount = splitLines((char *) buffer, outLines, BUFSIZE);
+	int lineNumber = 0;
+
+	for (lineNumber = 0; lineNumber < lineCount; lineNumber++)
+	{
+		/*
+		 * pg_basebackup and other utilities write their progress output on
+		 * stderr, we don't want to have ERROR message when it's all good.
```
JelteF, line 464: This comment seems to have nothing to do with the code below. It seems that simply ignores empty lines. It's good to note somewhere that processBufferCallback is called both for `stderr` and `stdout`, but this does not seem like the right place.

src/bin/pg_autoctl/pgctl.c
DimCitus: Fixed.

src/bin/pg_autoctl/string_utils.c
DimCitus: Ah yeah. Good point. Revised now, thanks!

PR #288, opened 2020-06-09
Fix pgautofailover extension upgrade script from 1.2 to 1.3.

DimCitus:
When using Postgres < 12 we can't ALTER an ENUM TYPE within a transaction,
and the extension upgrade script runs in a transaction. Which means we can't
upgrade pgautofailover when using Postgres 12 at the moment.

The fix requires dropping the enum type entirely, keeping the data in its
text form, and casting the data again to the new enum type after it's been
created. It's the long and painful way, but it is required for our support
of Postgres 10 and 11.

### DIFF
--- a/src/monitor/pgautofailover--1.2--1.3.sql
+++ b/src/monitor/pgautofailover--1.2--1.3.sql
@@ -4,8 +4,77 @@
 -- complain if script is sourced in psql, rather than via CREATE EXTENSION
 -- \echo Use "ALTER EXTENSION pgautofailover UPDATE TO 1.3" to load this file. \quit
 
-ALTER TYPE pgautofailover.replication_state ADD VALUE 'join_primary';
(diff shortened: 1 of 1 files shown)


JelteF:
According to my own answer on stackoverflow from 4.5 years ago you can do this a bit simpler:
https://stackoverflow.com/a/33617845/2570866

```sql
ALTER TYPE pgautofailover.replication_state RENAME TO pgautofailover.old_replication_state;
CREATE TYPE pgautofailover.replication_state AS ENUM (.....);
ALTER TABLE
      ALTER COLUMN goalstate
              TYPE pgautofailover.replication_state
             USING goalstate::text::pgautofailover.replication_state,
      ALTER COLUMN reportedstate
              TYPE pgautofailover.replication_state
             USING goalstate::pgautofailover.replication_state;
DROP pgautofailover.old_replication_state;
```

PR #289, opened 2020-06-09
Improve pg_autoctl perform failover/switchover default --group.

DimCitus:
Refuse to proceed with default group 0 when there is more than one group
registered in the target formation on the monitor, ask the user to specify
the target group with using the --group option.

Otherwise, suprises might ensue.


JelteF:
I approved this, since I think it's better than the current state. However, maybe one further improvement would be to default to the group of the node you trigger the failover on.

DimCitus:
See #291 for the second PR. It is indeed a good idea!

PR #290, opened 2020-06-09
Command pg_autoctl perform failover now waits until failover is done.

DimCitus:
We use the same approach as when applying a new setting: listen to
notification messages from the monitor and parse them until we see that we
have reached the expected target state.

In this case it's easy, we now we reached the target state when we see the
notification of a node reaching state primary.


JelteF:
I like the table a lot, maybe at the end log again to say that failover was successful.

DimCitus:
> I like the table a lot, maybe at the end log again to say that failover was successful.

I tried that but it kind of gets in the way visually. I didn't like the output with a log line after the end of the table...

PR #294, opened 2020-06-11
Tighten when to accept to perform a failover even more.

DimCitus:
In order to safely proceed we need one of the following situations:

  - primary node has both reported and goal state PRIMARY
  - primary node is still in progress (assigned PRIMARY, not reported yet)
    and the secondary node has already reported SECONDARY

Any other situation does not warrant for a failover to be performed.


src/bin/pg_autoctl/fsm.c
```diff
@@ -169,13 +169,17 @@ KeeperFSMTransition KeeperFSM[] = {
 	{ PRIMARY_STATE, DEMOTED_STATE, COMMENT_PRIMARY_TO_DEMOTED, &fsm_stop_postgres },
 	{ PRIMARY_STATE, DEMOTE_TIMEOUT_STATE, COMMENT_PRIMARY_TO_DEMOTED, &fsm_stop_postgres },
 
+	{ WAIT_PRIMARY_STATE, DRAINING_STATE, COMMENT_PRIMARY_TO_DRAINING, &fsm_stop_postgres },
+	{ WAIT_PRIMARY_STATE, DEMOTED_STATE, COMMENT_PRIMARY_TO_DEMOTED, &fsm_stop_postgres },
+	{ WAIT_PRIMARY_STATE, DEMOTE_TIMEOUT_STATE, COMMENT_PRIMARY_TO_DEMOTED, &fsm_stop_postgres },
+
```
JelteF, line 175: I guess we don't need these now anymore, right? Because these state transition can only happen from PRIMARY now.
```suggestion
```

src/bin/pg_autoctl/primary_standby.c
```diff
@@ -984,3 +984,53 @@ check_postgresql_settings(LocalPostgresServer *postgres, bool *settings_are_ok)
...
+			return false;
+		}
+
+		if (!reachedLSN)
+		{
+			log_info("Standby reached LSN %s, waiting for LSN %s",
+					 standbyCurrentLSN, postgres->currentLSN);
+			pg_usleep(AWAIT_PROMOTION_SLEEP_TIME_MS * 1000);
+		}
+	}
```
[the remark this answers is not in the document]
DimCitus: Good point. The local currentLSN is updated at every keeper loop when we fetch the local Postgres metadata on the primary, so it's recent enough on each loop. Will see about adding a (short) timeout in that loop and allow the main FSM loop logic to kick-in.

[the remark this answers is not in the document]
JelteF: > so it's recent enough on each loop

What I meant is that if on the first try:
Enable sync replication
LSN is 100 after enabling sync replication

Secondary LSN is 50 
Primary goes back to FSM

On second try:
Enable sync replication (no-op)
currentLSN is now 150

Secondary LSN is now 120

At that point we secondary has caught up enough, but because we updated `currentLSN` it's we do not notice. So again we go back to the FSM.

So it would be nice if we save the LSN right after we enable sync replication and save that throughout multiple FSM loops (and unset it when we enable async replication or something).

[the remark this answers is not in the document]
DimCitus: I see. Yes, indeed, that's how it should be done.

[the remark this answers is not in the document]
DimCitus: Done now.

src/bin/pg_autoctl/fsm.c
DimCitus: Good point, removed.

src/bin/pg_autoctl/primary_standby.c
```diff
@@ -992,45 +992,44 @@ check_postgresql_settings(LocalPostgresServer *postgres, bool *settings_are_ok)
...
-	log_info("Standby reached LSN %s, thus advanced past LSN %s",
-			 standbyCurrentLSN, postgres->currentLSN);
+	if (reachedLSN)
+	{
+		log_info("Standby reached LSN %s, thus advanced past LSN %s",
+				 standbyCurrentLSN, postgres->standbyTargetLSN);
 
-	return true;
+		/* cache invalidation */
+		bzero((void *) postgres->standbyTargetLSN, PG_LSN_MAXLENGTH);
```
JelteF, line 1024: This should also be done when going from `WAIT_PRIMARY` to `SINGLE`.

DimCitus: Fair, thanks! cache invalidation is hard!

DimCitus: (done now)

src/bin/pg_autoctl/pgsql.c
```diff
@@ -2260,6 +2269,13 @@ parsePgReachedTargetLSN(void *ctx, PGresult *result)
 		return;
 	}
 
+	if (PQntuples(result) == 0)
+	{
+		log_warn("Query returned no rows");
```
JelteF, line 2274: This vague `log_warn` seems a bit unnecessary, given we already have the clearer later in the code: "No standby nodes are connected at the moment"

DimCitus: Good point, going to make it a debug or trace message.

PR #296, opened 2020-06-12
Feature/restart and upgrade

### LINKED ISSUE #293: Expose a way to restart pg_autoctl without restarting Postgres
Thanks to our work in #284 and previous PRs pg_autoctl is now the parent process of Postgres. When installing a new version of pg_autoctl for an upgrade we want to be able to restart pg_autoctl without interrupting the Postgres service.

One way to expose the feature we already have to the users would be:

```
$ pg_autoctl restart
pg_autoctl restart: expected a command

Available commands:
  pg_autoctl restart
    postgres                  Restart the pg_autoct Postgres service
    node-active            Restart the pg_autoctl node-active service

$ pg_autoctl restart node-active
```

DimCitus:
Implement `pg_autoctl restart` subcommands and document how to upgrade pg_auto_failover when it's running in production. When using the new capabilities of the system the upgrade procedure is a little surprising, but quite flexible and we can restart our `node-active` services without restarting Postgres.

Fixes #293


docs/operations.rst
```diff
@@ -14,42 +14,169 @@ Provisioning
...
+pg_auto_failover monitor. The ``pg_autoctl create`` command honors the
+``PGDATA`` environment variable, and checks whether PostgreSQL is already
+running. If Postgres is detected, the new node is registered in SINGLE mode,
+bypassing the monitor's role assignment policy.
 
-.. _pg_auto_failover_security:
+Upgrading pg_auto_failover
+--------------------------
 
-Security
```
JelteF, line 24: I guess this could be deleted, because security now has it's own complete page?

docs/operations.rst
```diff
@@ -14,42 +14,169 @@ Provisioning
...
+		  from pgautofailover.node
+		 where reportedstate = 'secondary'
+		   and not exists
+			  (
+			   select nodeid
+				 from pgautofailover.node
+				where goalstate <> reportedstate
+				   or reportedstate not in ('primary', 'secondary')
+			  )
+		order by nodeid;
```
JelteF, line 70: It would be nice if we had a command line tool to do this. Something like:
```
pg_autoctl status --is-stable
```

And for the enable/disable maintainance queries we could have:
```
pg_autoctl enable maintainance --all
pg_autoctl disable maintainance --all
```

Or maybe something like:
```
pg_autoctl upgrade prepare
pg_autoctl upgrade finish
```

That seems nicer than telling people to connect manually to the monitor database, which usually they shouldn't do.

This might be better in a separate PR though.

docs/operations.rst
```diff
@@ -14,42 +14,169 @@ Provisioning
...
+
+	   sudo systemctl status pgautofailover
+
+	 At this point all the Postgres nodes in the system are going to error
+	 out every second when connecting to the monitor. In case of error to
+	 connect to the monitor, the node-active process retries every second,
+	 and will be able to continue normally as soon as the monitor is back
+	 online.
+
+  3. Upgrade the pg_auto_failover package on the monitor.
```
JelteF, line 123: I thought this step should be after upgrading (and restarting) pg_auto_failover on the postgres nodes. Otherwise the old pg_autoctl keeper processes will communicate with the new monitor extension. Sounds like that could mess stuff up.

docs/operations.rst
```diff
@@ -14,42 +14,169 @@ Provisioning
...
+	 Use the same command as on the monitor in step 4 above.
+
+  6. Restart the ``node-service`` service on all the Postgres nodes.
+
+	 When using the systemd integration, restarting the ``pgautofailover``
+	 service will also restart Postgres, which is not always possible. If
+	 you did implement step 2, your current service is not running, so you
+	 can restart it now::
+
+	   pg_autoctl restart node-active
```
JelteF, line 159: inconsistent use of `node-service`/`node-active`

docs/operations.rst
DimCitus: Good ideas, even if the exact CLI/API needs to be refined. On a first approach I'm thinking it might make sense to have those operations part of the `pg_autoctl perform` sub-commands, the main (only?) place where we initiate cluster wide operations at the moment.

Let's open a separate issue/PR for this!

docs/operations.rst
DimCitus: Yes.

docs/operations.rst
DimCitus: It's problematic in both cases in different ways. We don't guarantee API compatibility in either direction. 

What I now think we should do, given your feedback here, is to check for the monitor extension version to be as expected by the client each time we open a connection to the monitor, as the first query we do. When the version is not as expected, close the connection and return false to indicate we failed to open it, and let the main loop take over from there.

I believe that if we implement that, then you can upgrade the client pg_autoctl nodes at anytime without consequences: we would only call the API when we're on the expected version anyway, and keep retrying otherwise.

We might even be able to run `pg_autoctl version --json` and parse the result when we see that the server version is different from the current in-memory expected version, and when we see that the on-disk version is the same as the monitor version then just restart automatically?

For that we might want to add the supported monitor version in the JSON output, it's not there yet.

DimCitus: Fixed in #318 then #316.

docs/operations.rst
DimCitus: Thanks to restarting the "node-active" sub-process on extension version mismatch on the monitor, the upgrade procedure can be simplified down to two steps, and none of the preparation queries are needed anymore.

  1. upgrade the OS packages on every node
  2. restart the monitor

docs/operations.rst
DimCitus: Fixed.

docs/operations.rst
```diff
@@ -14,42 +14,79 @@ Provisioning
...
+  1. Upgrade the pg_auto_failover package on the all the nodes, monitor
+     included.
 
-Operations
-----------
+	 When using a debian based OS, this looks like the following command for
+	 pg_auto_failover 1.3.1::
+
+	   sudo apt-get remove pg-auto-failover-cli-enterprise-1.0 postgresql-11-auto-failover-enterprise-1.0
+	   sudo apt-get install -q -y pg-auto-failover-cli-enterprise-1.3 postgresql-11-auto-failover-enterprise-1.3
```
JelteF, line 58: This upgrade procedure only works for upgrades from 1.4 to future 1.5 right? I think it's good to make that clear somewhere. Pretending like it works to upgrade from 1.0 to 1.3 seems like a recipe for problems. I'm not sure if we need to write down the pre-1.4 upgrade procedure or not, but making clear that this procedure is not the right one for pre-1.4 seems critical.

docs/operations.rst
```diff
@@ -14,42 +14,79 @@ Provisioning
...
+	 pgautofailover UPDATE TO ...`` command. The monitor is ready with the
+	 new version of pg_auto_failover.
+
+When the Postgres nodes ``pg_autoctl`` process connects to the new monitor
+version, the check for version compatibility fails, and the "node-active"
+sub-process exits. The main ``pg_autoctl`` process supervisor then restart
+the "node-active" sub-process from its on-disk binary executable file, which
+has been upgraded to the new version. That's why we first install the new
+packages for pg_auto_failover on every node, and only then restart the
+monitor.
```
JelteF, line 83: Probably good to note here (and/or at the top of this step) that not updating pg_autoctl on Postgres nodes will result in postgres going down.

docs/operations.rst
DimCitus: Thinking more about it, I think we should still have a short version of the pre-1.4 procedure. The extra SQL queries we used to have in this document were more confusing than useful, so I re-installed something simpler where you have to run `pg_autoctl enable maintenance` on every standby node.

docs/operations.rst
DimCitus: Added a note about that, thanks!

docs/operations.rst
```diff
@@ -14,42 +14,144 @@ Provisioning
...
+  1. Upgrade the pg_auto_failover package on the all the nodes, monitor
+     included.
 
-Operations
-----------
+	 When using a debian based OS, this looks like the following command for
+	 pg_auto_failover 1.3.1::
+
+	   sudo apt-get remove pg-auto-failover-cli-enterprise-1.0 postgresql-11-auto-failover-enterprise-1.0
+	   sudo apt-get install -q -y pg-auto-failover-cli-enterprise-1.3 postgresql-11-auto-failover-enterprise-1.3
```
JelteF, line 58:
```suggestion
	   sudo apt-get remove pg-auto-failover-cli-enterprise-1.4 postgresql-11-auto-failover-enterprise-1.4
	   sudo apt-get install -q -y pg-auto-failover-cli-enterprise-1.5 postgresql-11-auto-failover-enterprise-1.5
```

docs/operations.rst
```diff
@@ -14,42 +14,144 @@ Provisioning
...
-__ https://www.postgresql.org/docs/current/libpq-pgpass.html
+As a result, here is the standard upgrade plan for pg_auto_failover:
 
+  1. Upgrade the pg_auto_failover package on the all the nodes, monitor
+     included.
 
-Operations
-----------
+	 When using a debian based OS, this looks like the following command for
+	 pg_auto_failover 1.3.1::
```
JelteF, line 55:
```suggestion
	 When using a debian based OS, this looks like the following command when 
	 from 1.4 to 1.5::
```

PR #298, opened 2020-06-15
Skip network DNS probes on pg_autoctl create monitor --skip-pg-hba.

DimCitus:


src/bin/pg_autoctl/pghba.c
```diff
@@ -308,6 +308,12 @@ pghba_enable_lan_cidr(PGSQL *pgsql,
 	char ipAddr[BUFSIZE];
 	char cidr[BUFSIZE];
 
+	if (SKIP_HBA(authenticationScheme))
+	{
+		log_warn("Skipping HBA edits (per --skip-pg-hba) for local network");
```
JelteF, line 313: `log_warn` seems a bit excessive, since the user requested this. `log_info` seems a better fit.

DimCitus: The only other use case we have for `SKIP_HBA` issues a warn too, with the following rationale:

``` c
	/*
	 * When the authentication method is "skip", the option --skip-pg-hba has
	 * been used. In that case, we still WARN about the HBA rule that we need,
	 * so that users can review their HBA settings and provisioning.
	 */
```

I'm inclined to push as it is.

JelteF: Alright that reasoning makes sense. Can we make the log a bit more actionable for the user in that case? Something like:
```
Skipping HBA edits (per --skip-pg-hba), please make sure you have allowed the required access
```

DimCitus: In the other branch where we use `SKIP_HBA` we have the rule and display it in full. It used to be the case that we would also do that here, but we also want to avoid displaying errors of network detection.

So I just pushed an alternative version where we consider the network detection errors as non-fatal in case when `--skip-pg-hba` has been used, and still go as far as computing the HBA rule and displaying the rule we skip otherwise:

```
15:59:48 62881 INFO  Granting connection privileges on ::1/128
15:59:48 62881 WARN  Skipping HBA edits (per --skip-pg-hba) for rule: host "pg_auto_failover" "autoctl_node" ::1/128 skip```

File: src/bin/pg_autoctl/pghba.c [hunk 2]
```diff
@@ -311,22 +311,30 @@ pghba_enable_lan_cidr(PGSQL *pgsql,
...
+		int logLevel = SKIP_HBA(authenticationScheme) ? LOG_WARN : LOG_FATAL;
+
+		log_level(logLevel,
+				  "Failed to find IP address for hostname \"%s\", "
 				  "see above for details",
 				  hostname);
-		return false;
+
+		/* when --skip-pg-hba is used, we don't mind the failure here */
+		return SKIP_HBA(authenticationScheme) ? true : false;
```
JelteF, line 322:
```suggestion
		return SKIP_HBA(authenticationScheme);
```

src/bin/pg_autoctl/pghba.c
```diff
@@ -311,22 +311,30 @@ pghba_enable_lan_cidr(PGSQL *pgsql,
...
 	}
 
 	if (!fetchLocalCIDR(ipAddr, cidr, BUFSIZE))
 	{
 		log_warn("Failed to determine network configuration for "
 				 "IP address \"%s\", skipping HBA settings", ipAddr);
-		return false;
+
+		/* when --skip-pg-hba is used, we don't mind the failure here */
+		return SKIP_HBA(authenticationScheme) ? true : false;
```
JelteF, line 331:
```suggestion
		return SKIP_HBA(authenticationScheme);
```

PR #301, opened 2020-06-17
Make monitor internal db

### LINKED ISSUE #270: Use UNIX sockets to connect to local postgres instead of "localhost"
Right now the only way to run pg_auto_ctl is to completely trust the node it runs on, because any connection made from "localhost" is allowed to authenticate using "trust". Even when usig `--skip-pg-hba`. This is because we do not change the default `--auth` argument when we run `initdb`. https://www.postgresql.org/docs/12/app-initdb.html 
https://github.com/citusdata/pg_auto_failover/blob/master/src/bin/pg_autoctl/pgctl.c#L755
I tried changing it to `--auth=reject --auth-local=trust`, but this causes connection errors on the `pg_auto_ctl` side.

It would be nice to use UNIX socket connections from pg_auto_ctl to postgres instead of TCP connections over localhost. This way we can make sure that only the `postgres` user (or whatever user used `pg_auto_ctl create` can connect to `postgres` without auth).

onderkalaci:
Fixes #270

With this PR, we  suggest to treat the monitor database as an internal/embedded database for pg_auto_failover. To do that, we do few things:
- Ignore PGHOST env. variable
- Only allow unix domain socket connections from local host
- Only allow the user who created the database to be able to connect via the unix domain socket.

TODO:
- [ ] I still see the following on the pg_hba.conf (`hostssl "pg_auto_failover" "autoctl_node" 192.168.2.0/24 trust # Auto-generated by pg_auto_failover`). So, `replace_monitor_pg_hba ` is probably in the wrong way
- [ ] I get some errors while trying to create postgres instances (even on master branch). So, I wasn't able to test the changes in detail yet locally
```
15:07:34 68157 TRACE pgsetup.c:1375: nodeKindFromString: "standalone" ➜ 1
68157 Failed to acquire a lock with semaphore 4325417: Invalid argument
68157 Failed to acquire a lock with semaphore 4325417: Invalid argument
```

### DIFF
--- a/src/bin/pg_autoctl/cli_create_drop_node.c
+++ b/src/bin/pg_autoctl/cli_create_drop_node.c
@@ -723,7 +723,7 @@ cli_create_monitor(int argc, char **argv)
 	 * 	unix socket directories, now --pghost is mandatory, but unset.
 	 *
 	 */
-	if (env_found_empty("PG_REGRESS_SOCK_DIR") && unsetenv("PGHOST") != 0)
+	if (!env_found_empty("PG_REGRESS_SOCK_DIR")  && unsetenv("PGHOST") != 0)
 	{
 		exit(EXIT_CODE_BAD_STATE);
 	}
--- a/src/bin/pg_autoctl/cli_create_drop_node.c
+++ b/src/bin/pg_autoctl/cli_create_drop_node.c
@@ -711,17 +711,14 @@ cli_create_monitor(int argc, char **argv)
 
 
 	/*
-	 * We prefer to ignore PGHOST variable for the monitor. The reason is that we
-	 * qualify monitor database as an internal/embedded database for pg_auto_failover.
-	 * In other words, we don't want anyone else to be able to connect to the database,
-	 * except for pg_auto_failover itself. To do that, we hard code host to be the
-	 * unix domain socket, where we can restrict the access.
+	 * The pg_auto_failover monitor Postgres database needs to allow
+	 * connections from the pg_autoctl monitor binary itself (this code) and
+	 * later from other pg_autoctl nodes that are using this monitor instance.
 	 *
-	 * TODO: Not allowing to set PGHOST fails regression tests with:
-	 *
-	 * 	ERROR pgsetup.c:269 PG_REGRESS_SOCK_DIR is set to "" to disable
-	 * 	unix socket directories, now --pghost is mandatory, but unset.
-	 *
+	 * In any case, we don't use --pghost nor PGHOST to build our connection
+	 * string to the Postgres monitor database in this `pg_autoctl create
+	 * monitor` program. Instead we always use the Unix Domain Socket directory
+	 * used by Postgres.
 	 */
 	if (!env_found_empty("PG_REGRESS_SOCK_DIR")  && unsetenv("PGHOST") != 0)
 	{
--- a/src/bin/pg_autoctl/cli_create_drop_node.c
+++ b/src/bin/pg_autoctl/cli_create_drop_node.c
@@ -709,22 +709,6 @@ cli_create_monitor(int argc, char **argv)
 
 	monitor.config = monitorOptions;
 
-
-	/*
-	 * The pg_auto_failover monitor Postgres database needs to allow
-	 * connections from the pg_autoctl monitor binary itself (this code) and
-	 * later from other pg_autoctl nodes that are using this monitor instance.
-	 *
-	 * In any case, we don't use --pghost nor PGHOST to build our connection
-	 * string to the Postgres monitor database in this `pg_autoctl create
-	 * monitor` program. Instead we always use the Unix Domain Socket directory
-	 * used by Postgres.
-	 */
-	if (!env_found_empty("PG_REGRESS_SOCK_DIR")  && unsetenv("PGHOST") != 0)
-	{
-		exit(EXIT_CODE_BAD_STATE);
-	}
-
 	/*
 	 * We support two modes of operations here:
 	 *   - configuration exists already, we need PGDATA
--- a/src/bin/pg_autoctl/cli_create_drop_node.c
+++ b/src/bin/pg_autoctl/cli_create_drop_node.c
@@ -710,6 +710,18 @@ cli_create_monitor(int argc, char **argv)
 	monitor.config = monitorOptions;
 
 	/*
+	 * We prefer to ignore PGHOST variable for the monitor. The reason is that we
+	 * qualify monitor database as an internal/embedded database for pg_auto_failover.
+	 * In other words, we don't want anyone else to be able to connect to the database,
+	 * except for pg_auto_failover itself. To do that, we hard code host to be the
+	 * unix domain socket, where we can restrict the access.
+	 */
+	if (unsetenv("PGHOST") != 0)
+	{
+		exit(EXIT_CODE_BAD_STATE);
+	}
+
+	/*
 	 * We support two modes of operations here:
 	 *   - configuration exists already, we need PGDATA
 	 *   - configuration doesn't exist already, we need PGDATA, and more
--- a/src/bin/pg_autoctl/cli_create_drop_node.c
+++ b/src/bin/pg_autoctl/cli_create_drop_node.c
@@ -709,14 +709,21 @@ cli_create_monitor(int argc, char **argv)
 
 	monitor.config = monitorOptions;
 
+
 	/*
 	 * We prefer to ignore PGHOST variable for the monitor. The reason is that we
 	 * qualify monitor database as an internal/embedded database for pg_auto_failover.
 	 * In other words, we don't want anyone else to be able to connect to the database,
 	 * except for pg_auto_failover itself. To do that, we hard code host to be the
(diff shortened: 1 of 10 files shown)


src/bin/pg_autoctl/cli_create_drop_node.c
```diff
@@ -753,6 +753,25 @@ cli_create_monitor(int argc, char **argv)
...
+	 *
+	 * 	ERROR pgsetup.c:269 PG_REGRESS_SOCK_DIR is set to "" to disable
+	 * 	unix socket directories, now --pghost is mandatory, but unset.
+	 *
+	 */
+	if (!env_found_empty("PG_REGRESS_SOCK_DIR")  && unsetenv("PGHOST") != 0)
+	{
+		exit(EXIT_CODE_BAD_STATE);
+	}
+
```
DimCitus, line 774: I think this comment and code does not reflect the decision we made entirely. On top of that, the decision we made is not compatible with the current coding of the test-suite, and that needs to be taken care of in the test suite itself, as you started to discover.

A better comment here might be:

``` c
	/*
	 * The pg_auto_failover monitor Postgres database needs to allow
	 * connections from the pg_autoctl monitor binary itself (this code) and
	 * later from other pg_autoctl nodes that are using this monitor instance.
	 *
	 * In any case, we don't use --pghost nor PGHOST to build our connection
	 * string to the Postgres monitor database in this `pg_autoctl create
	 * monitor` program, instead we always use the Unix Domain Socket directory
	 * used by Postgres.
	 */
```

Then in our testing environment we set PG_REGRESS_SOCK_DIR to an empty string. In that we imitate what `pg_regress` does at https://github.com/postgres/postgres/blob/master/src/test/regress/pg_regress.c#L802, for much the same reasons: allow to start several Postgres instances on the same port on the same host. In our test suite, we use network namespaces so that each Postgres instance has their own IP address to listen to, and a single file system, so that we can't use unix domain sockets at the moment: we would try to create the same files for all the nodes (same default Postgres directory, same port).

I think we should mimic what `pg_regress` is doing in `make_temp_sockdir` at https://github.com/postgres/postgres/blob/master/src/test/regress/pg_regress.c#L318 when PG_REGRESS_SOCK_DIR is empty and use the temporary directory created both in the `postmaster -k` option and then in our connection strings, so in our `pgSetup.pghost` structure too. We could implement that part by forcing the `PGHOST` value to the temp sockdir we just created and then `pgsetup.c` will take it from there.

Given that, we should be able to connect to the monitor Postgres database from the test suite without exposing a unix socket directory at the default place, which would then be re-used by every other Postgres instance, and because they all use the same port in our test-suite, that would fail.

src/bin/pg_autoctl/monitor_pg_init.c
```diff
@@ -117,6 +126,28 @@ monitor_pg_init(Monitor *monitor)
 }
 
 
+/*
+ * replace_monitor_pg_hba gets the path of pgData of the monitor, and
+ * overrides the pg_hba rules.
```
DimCitus, line 131: I think this comment is a good place to explain two things:

  1. why we don't like the default rules in pg_hba.conf and how the different --auth-local and --auth-host options can't help us there,
  2. what we expect to see as our own default rules in pg_hba.conf in the monitor instead

src/bin/pg_autoctl/pghba.c
```diff
@@ -172,6 +172,59 @@ pghba_ensure_host_rule_exists(const char *hbaFilePath,
 }
 
 
+/*
+ * write_trust_local_hba_rule gets the hbaFilePath and overrides it with
+ * 		"local all all  trust"
+ * 	which means allow all connections via unix domain sockets.
```
DimCitus, line 178: Again, we should explain why we do things...

src/bin/pg_autoctl/pghba.c
```diff
@@ -172,6 +172,59 @@ pghba_ensure_host_rule_exists(const char *hbaFilePath,
...
+
+	append_database_field(hbaLineBuffer, HBA_DATABASE_ALL, NULL);
+	appendPQExpBufferStr(hbaLineBuffer, " ");
+
+	appendPQExpBufferStr(hbaLineBuffer, userName);
+	appendPQExpBufferStr(hbaLineBuffer, " ");
+
+	appendPQExpBuffer(hbaLineBuffer, " %s", authenticationScheme);
+
+	appendPQExpBufferStr(hbaLineBuffer, HBA_LINE_COMMENT "\n");
```
DimCitus, line 203: Please re-use `pghba_ensure_host_rule_exists` and extend it with a new parameter `HBAConnectionType` (enum with 6 entries for `local`, `host`, `hostssl`, `hostnossl`, `hostgssenc`, `hostnogssenc` per https://www.postgresql.org/docs/current/auth-pg-hba-conf.html) much like `HBADatabaseType`. This parameter would replace `bool ssl` obviously.

src/bin/pg_autoctl/pghba.c
```diff
@@ -172,6 +172,59 @@ pghba_ensure_host_rule_exists(const char *hbaFilePath,
...
+
+	log_debug("generated monitor pg_hba rule for unix domain socket: \"%s\"", hbaLineBuffer->data);
+
+	/* write the resulting content at the destination path */
+	if (!write_file(hbaLineBuffer->data, hbaLineBuffer->len, hbaFilePath))
+	{
+		/* error is already logged */
+		destroyPQExpBuffer(hbaLineBuffer);
+		return false;
+	}
```
DimCitus, line 220: What would be good here would be to follow Jelte's remark from the other PR and allow us to keep the useful comments from the main pg_hba.conf file. I just had a look at it seems to me that there's a rather easy way to do that: as long as the lines start with a `#` character, keep it. First time it's not the case, stop copying from the existing HBA file, add an empty line, then our own stuff.

If we do that we lose the following table-like header, but we could always manually re-add it before we add our own lines, and then we could use `%25s` or whatever to respect the table like formatting of the HBA rules:

```
# TYPE  DATABASE        USER            ADDRESS                 METHOD
```

Then the default HBA file could be kept under the name `pg_hba.default.conf` or something.

DimCitus: See the `splitLines` function in `string_utils.c` and its call sites for an example how to easily achieve that.

src/bin/pg_autoctl/cli_create_drop_node.c
onderkalaci: There are few things that I cannot easily follow here which makes it hard for me to implement

> Then in our testing environment we set PG_REGRESS_SOCK_DIR to an empty string. In that we imitate what pg_regress does at https://github.com/postgres/postgres/blob/master/src/test/regress/pg_regress.c#L802, for much the same reasons:


When I read the Postgres  code/commit that introduces the env. variable, the main motivation seems different: https://github.com/postgres/postgres/commit/31c6e54ec9abab0c63d709e492ef18a701b02641

>  allow to start several Postgres instances on the same port on the same host.

Is that possible with this? I'm trying to do that locally, but I cannot see a way for doing _that other than setting `listen_addresses ` to empty list.


> We could implement that part by forcing the PGHOST value to the temp sockdir we just created and then pgsetup.c will take it from there.

Could you please clarify this a little further? I though we're trying to unset `PGHOST`, how can we force it?

DimCitus: 1. ah right it was security for `pg_regress`, for the build farm animals I suppose
  2. that's possible if you have ethernet interface aliases or network namespace, the latter is what we do in our test suite ; then each Postgres instance can listen to its own ethernet interface (which just happens even with `listen_addresses = '*'` with network namespaces)
  3. we never said we wanted to unset `PGHOST`, only that we don't care about the value that the user has set in their environment

onderkalaci: > we never said we wanted to unset PGHOST, only that we don't care about the value that the user has set in their environment

Hmm, my fear is that if we don't unset it, we might end-up with a state where in some  parts of the code we rely on `PGHOST`, in some it doesn't. That way, we might lead to a discrepancy which would be hard to understand by  anyone.  Sort of creating a hard to track dependency. 

That's why I considered this as we should "unset" it. One  of the reasons for me to go with "unset" approach is that I don't have a good enough visibility of the code-base at this moment. But, if you think it is safe just ignore here and let other parts rely on it (if any), I can  do that.

DimCitus: Thanks to this discussion what I now have in mind is that the `pg_autoctl create monitor` command, when PG_REGRESS_SOCK_DIR is set empty in the environment, would then create a temporary socket directory and use it in the `-k` option used when starting Postgres in either `pg_ctl_start` or `pg_ctl_postgres`.

Once we've done that though, we still need to generate a local connection string that uses that temp socket directory as our `host` option, as in `psql -h /tmp -p 5001 -d postgres` if you have a local server running on port 5001 with unix sockets created in the `/tmp` directory.

The easiest way to achieve that given our current code-base (see `pgsetup.c`) might be to force set the PGHOST environment variable to that temporary directory internally, and make sure to call `pg_setup_init` again after that.

PR #302, opened 2020-06-17
Implement different Postgres connection retry policies.

DimCitus:
When the keeper loop is running, we should not retry connecting to the
monitor and let the main loop logic take care of potential network split
situations, and take split-brain protective steps.

When running interactive commands, we might want to retry for up to a
minute, as a service to our users.

When provisionning a cluster, some nodes might get ready to connect to the
monitor before that one is live, so it warrants yet another retry policy.

In all cases, we use an Exponential Backoff with Decorrelated Jitter
approach to decide how much time to sleep in between retries, to avoid
hammering the monitor all at once when it's ready / back from the dead.


src/bin/pg_autoctl/pgsql.c
```diff
@@ -270,65 +519,107 @@ pgsql_retry_open_connection(PGSQL *pgsql)
...
 			{
-				int logLevel = attempts % 30 == 0 ? LOG_INFO : LOG_DEBUG;
-
-				log_level(logLevel,
-						  "PQping: no response from server at \"%s\" "
-						  "after %d attempts",
-						  pgsql->connectionString,
-						  attempts);
-				retry = true;
+				if (lastWarningMessage != PQPING_NO_RESPONSE ||
```
JelteF, line 580: I was wondering when we would get this response. Turns out it's because of the `connect_timeout` connection string param or the `PGCONNECT_TIMEOUT` env var. We don't set either of these. On my machine this means the FSM is blocked for 30 seconds when I drop all traffic to the monitor using:
```bash
sudo iptables -A INPUT -p tcp --dport 5432 -m conntrack --ctstate NEW,ESTABLISHED -j DROP
```
This seems like a serious problem as this will impact the 30s timeout semantics of `DEMOTE_TIMEOUT`

src/bin/pg_autoctl/pgsql.c
```diff
@@ -117,6 +123,158 @@ pgsql_init(PGSQL *pgsql, char *url, ConnectionType connectionType)
...
+}
+
+
+/*
+ * pgsql_set_interactive_retry_policy sets the retry policy to 1 minute of
+ * total retrying time, unbounded number of attempts, and up to 2 seconds of
+ * sleep time in between attempts.
+ */
+void
+pgsql_set_interactive_retry_policy(PGSQL *pgsql)
```
JelteF, line 201: I think it's good to have this retry behaviour, but I think people should opt-in to it. For interactive usage fail-fast behaviour is usually what you want. So I think a flag like `--retry-connect` would be good to have to opt-in to it.

Also, do we want this retry behaviour for every interactive command or just a specific set? Currently it's not done for a command like:
```
pg_autoctl show state --pgdata node1
```

src/bin/pg_autoctl/pgsql.c
```diff
@@ -230,32 +428,83 @@ static PGconn *
...
+		 * all of our attempts count budget, whichever comes first.
+		 *
+		 * maxR = 0 (zero) means no retry at all, checked before the loop
+		 * maxR < 0 (zero) means unlimited number of retries
+		 */
+		if ((now - startTime) >= pgsql->retryPolicy.maxT ||
+			(pgsql->retryPolicy.maxR > 0 &&
+			 pgsql->retryPolicy.attempts >= pgsql->retryPolicy.maxR))
+		{
+			log_warn("Failed to connect to \"%s\" "
```
JelteF, line 474: I think this should be a `log_error` instead. 

Also, this error doesn't give any indication what the problem with connecting is. The previous error that contained all this information:
```
11:57:14 18700 ERROR Connection to database failed: could not connect to server: Connection refused
11:57:14 18700 ERROR 	Is the server running on host "172.24.28.118" and accepting
11:57:14 18700 ERROR 	TCP/IP connections on port 5432?
11:57:14 18700 ERROR Failed to connect to monitor database at "postgres://autoctl_node@172.24.28.118:5432/pg_auto_failover?sslmode=require", see above for details
```

Can we add this to the final fail in some way?

src/bin/pg_autoctl/loop.c
```diff
@@ -171,6 +176,19 @@ keeper_node_active_loop(Keeper *keeper, pid_t start_pid)
...
 
+		if (!couldContactMonitor && couldContactMonitorThisRound && !firstLoop)
+		{
+			/*
+			 * Last message the user saw in the output is the following, and so
+			 * we should say that we're back to the expected situation:
+			 *
+			 * Failed to get the goal state from the monitor
+			 */
+			log_info("Could get the goal state from the monitor");
```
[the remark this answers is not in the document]
DimCitus: Ah yes thanks, I didn't like my own version very much but was not sure how to improve it. That's an improvement ;-)

src/bin/pg_autoctl/pgsql.c
DimCitus: In other places of the code, like for calling `pg_basebackup`, we force the environment variable to 1 second. I have now done the same for establishing all our connections, always, so that we have better control.

src/bin/pg_autoctl/pgsql.c
DimCitus: So I have now made the interactive retry policy the default with a 2 seconds max time. Also, I made it so that users can set `PGCONNECT_TIMEOUT` in their environment to influence if they want to. We still manage that with our own retry strategy, though, and internally we use a timeout of 1 second only.

DimCitus: Also all the commands are now using the interactive retry policy, only the main loop is having something different, and that's done in the main loop code itself.

src/bin/pg_autoctl/pgsql.c
DimCitus: Added now, thanks!

DimCitus:
So. I changed our value of `POSTGRES_CONNECT_TIMEOUT` to 2, per your reading, it was 5. Sorry about that... Now, I also made a change related to when we start counting time and attempts so that the first attempt is part of what we count. I think it was a fair off-by-one otherwise. 

It also means that the strange error message can't happen anymore, it will be at least 2 seconds and 2 attempts always now...

src/bin/pg_autoctl/pgsql.c
```diff
@@ -227,35 +438,87 @@ pgsql_open_connection(PGSQL *pgsql)
...
+
+		/* we have milliseconds, pg_usleep() wants microseconds */
+		(void) pg_usleep(sleep * 1000);
+
+		log_debug("PQping(%s): slept %d ms on attempt %d",
+				  pgsql->connectionString,
+				  pgsql->retryPolicy.sleepTime,
+				  pgsql->retryPolicy.attempts);
+
 		switch (PQping(pgsql->connectionString))
```
JelteF, line 511: After doing PQping we `now` should be calculated again, otherwise the below log messages use the now from before connecting. Another option is to not use `now` in those error mesasges, but use `time(NULL)` directly.

DimCitus: Done!

PR #305, opened 2020-06-19
Don't show postgres startup messages on succesful startup

JelteF:
Like the comment says we would get an error message like this on startup
sometimes:
```
16:31:53 21530 ERROR 2020-06-19 16:31:51.463 CEST [21801] ERROR:  database "postgres" already exists
```


DimCitus:
> I believe we also need to change https://github.com/citusdata/pg_auto_failover/blob/master/src/bin/pg_autoctl/pgctl.c#L1114 and the following to avoid hard-coding `log_fatal` and `log_error` calls.

It'a actually good the way it is, skipping the output entirely unless with high verbosity. Then raising the Postgres errors and fatal message to our own ERROR and FATAL level makes senes in the context.

PR #309, opened 2020-06-25
Review the registering process & transaction.

DimCitus:
Ensure that when we could register successfully to the monitor but then
failed to create our local state file, we cancel (ROLLBACK) the
registration. Otherwise we have a node on the monitor and no way to relate
that node to the current pg_autoctl process.

### DIFF
--- a/src/bin/pg_autoctl/keeper.c
+++ b/src/bin/pg_autoctl/keeper.c
@@ -1081,6 +1081,10 @@ keeper_register_and_init(Keeper *keeper, NodeState initialState)
 	Monitor *monitor = &(keeper->monitor);
 	MonitorAssignedState assignedState = { 0 };
 
+	char *sqlBegin = "BEGIN";
+	char *sqlCommit = "COMMIT";
+	char *sqlRollback = "ROLLBACK";
+
 	/*
 	 * First try to create our state file. The keeper_state_create_file function
 	 * may fail if we have no permission to write to the state file directory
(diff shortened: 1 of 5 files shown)


JelteF:
These kind of things are always hard to make work correctly in all failure cases. What happens now when the commit fails, but we have written the state file? The state file will then contain an id that the monitor does not know about. Can you try out what happens in that case?

DimCitus:
> These kind of things are always hard to make work correctly in all failure cases. What happens now when the commit fails, but we have written the state file? The state file will then contain an id that the monitor does not know about. Can you try out what happens in that case?

I can't think of another way than injecting `exit(0)` at different places in the function and then recompiling without the exit() and see what happens then. I didn't run those, but I would think that the following makes it clear what happens when we fail to commit the transaction: we remove the local file...

``` c
	if (!pgsql_execute(&(monitor->pgsql), "COMMIT"))
	{
		log_error("Failed to COMMIT register_node transaction on the "
				  "monitor, see above for details");

		/* we can't send a ROLLBACK when a COMMIT failed */
		unlink_file(config->pathnames.state);
		return false;
	}
```

src/bin/pg_autoctl/monitor.c
```diff
@@ -615,9 +614,6 @@ monitor_register_node(Monitor *monitor, char *formation, char *host, int port,
 		return false;
 	}
 
-	/* disconnect from PostgreSQL now */
-	pgsql_finish(&monitor->pgsql);
```
JelteF, line 619: We don't call this anywhere anymore now. Shouldn't these calls be placed after `COMMIT`/`ROLLBACK` now?

src/bin/pg_autoctl/keeper.c
```diff
@@ -1200,8 +1224,7 @@ keeper_register_and_init(Keeper *keeper, NodeState initialState)
 		 * prevent trying to init again and cause strange errors.
 		 */
 		unlink_file(config->pathnames.state);
```
JelteF, line 1226: This call is now unnecessary. Let's move the comment to the `rollback` section.

src/bin/pg_autoctl/keeper.c
```diff
@@ -1165,6 +1165,30 @@ keeper_register_and_init(Keeper *keeper, NodeState initialState)
...
+	 */
+	if (!pgsql_execute(&(monitor->pgsql), "BEGIN"))
+	{
+		log_error("Failed to open a SQL transaction to register this node");
+
+		unlink_file(config->pathnames.state);
+		return false;
+	}
+
 	if (!monitor_register_node(monitor,
```
JelteF, line 1192: I think this error case should also `goto rollback`.

src/bin/pg_autoctl/monitor.c
DimCitus: That's a good point, fixed. In the branch where we refactor the main keeper look, I have also added some connection management, and so I guess that's also what I had in mind here, but well, this branch didn't have it. I have added it now, and kept the idea to close the connection after either COMMIT or ROLLBACK of the registration call.

src/bin/pg_autoctl/keeper.c
DimCitus: Good point, done that way now.

src/bin/pg_autoctl/keeper.c
DimCitus: Yeah that's a fair point, done that way now.

PR #311, opened 2020-06-25
Edit the HBA file as soon as a new node is added to the group.

DimCitus:
We don't have to wait until a failover happens to install the HBA rules we
need. Also installing those at node creation would help users review and
edit them, if needed.


src/bin/pg_autoctl/service_keeper.c
```diff
@@ -402,77 +372,12 @@ keeper_node_active_loop(Keeper *keeper, pid_t start_pid)
...
-							  "replication slot name, see above for details");
-					return false;
-				}
-			}
-		}
-		else
+		if (keeperState->assigned_role != keeperState->current_role)
 		{
-			log_error("Failed to get the goal state from the monitor");
+			needStateChange = true;
```
JelteF, line 377: There is a behavioural change here. When going to `DEMOTE_TIMEOUT_STATE`, we actually did not set `needStateChange` before. But we do now, because we set it regardless of `couldContactMonitor`.

src/bin/pg_autoctl/service_keeper.c
```diff
@@ -590,6 +499,144 @@ keeper_node_active_loop(Keeper *keeper, pid_t start_pid)
...
+		bool postgresNotRunningIsOk = false;
+
+		if (!keeper_config_set_groupId_and_slot_name(config,
+													 assignedState.nodeId,
+													 assignedState.groupId))
+		{
+			log_error("Failed to update the configuration file "
+					  "with groupId %d and replication.slot \"%s\"",
+					  assignedState.groupId, expectedSlotName);
+			return false;
```
JelteF, line 624: Here's another behavioral change in the code caused by this refactoring. Before, when this failed we would error out right away and not do any state change. Now wo de not, but instead we set `couldContactMonitorThisRound` to false. 

It seems better not to error out, but try to do the state change instead. Setting `couldContactMonitorThisRound` to false, seems at least confusing. I think it makes more sense to return true in these cases, since the monitor could be contacted.

src/bin/pg_autoctl/pghba.c
```diff
@@ -155,15 +202,176 @@ pghba_ensure_host_rule_exists(const char *hbaFilePath,
...
+			hbaLineReplicationBuffer,
+			hbaLineDatabaseBuffer,
+			NULL
+		};
+
+		if (hbaLineReplicationBuffer == NULL ||
+			hbaLineDatabaseBuffer == NULL)
+		{
+			log_error("Failed to allocate memory");
+			return false;
```
JelteF, line 278: On this `return false` and the ones below, we leak `hbaLineReplicationBuffer`, `hbaLineDatabaseBuffer` and `newHbaContents`

src/bin/pg_autoctl/pghba.c
```diff
@@ -155,15 +202,176 @@ pghba_ensure_host_rule_exists(const char *hbaFilePath,
...
+			 * If the rule was found and it starts on a new line. We can
+			 * skip adding it.
+			 */
+			if (includeLine != NULL && (includeLine == currentHbaContents ||
+										includeLine[-1] == '\n'))
+			{
+				log_debug("Line already exists in %s, skipping %s",
+						  hbaFilePath, hbaLineBuffer->data);
+
+				continue;
```
JelteF, line 330: On this `continue` and the one below we leak `hbaLineReplicationBuffer` and `hbaLineDatabaseBuffer`. I think it would be easiest and least error-prone to future changes to define the variables outside of the loop and free them at the start before allocating.

src/bin/pg_autoctl/pghba.c
```diff
@@ -36,76 +42,101 @@ static int escape_hba_string(char *destination, const char *hbaString);
...
+
+	if (!pghba_append_rule_to_buffer(hbaLineBuffer,
+									 ssl,
+									 databaseType,
+									 database,
+									 username,
+									 host,
+									 authenticationScheme))
+	{
+		/* errors have already been logged */
```
JelteF, line 138: Leaks `hbaLineBuffer`, same for all the early returns below.

src/bin/pg_autoctl/pghba.c
```diff
@@ -36,76 +42,101 @@ static int escape_hba_string(char *destination, const char *hbaString);
...
 	}
 
-	/*
-	 * When the authentication method is "skip", the option --skip-pg-hba has
-	 * been used. In that case, we still WARN about the HBA rule that we need,
-	 * so that users can review their HBA settings and provisioning.
-	 */
-	if (SKIP_HBA(authenticationScheme))
+	/* memory allocation could have failed while building string */
+	if (PQExpBufferBroken(buffer))
```
JelteF, line 93: This is duplicate check from the one above.

src/bin/pg_autoctl/keeper.c
```diff
@@ -1333,3 +1318,198 @@ keeper_state_as_json(Keeper *keeper, char *json, int size)
...
+	 * We have a new list of other nodes, update the HBA file. We only update
+	 * the nodes that we didn't know before, or that have a new host property.
+	 */
+	if (!keeper_update_group_hba(keeper, &diffNodesArray))
+	{
+		log_error("Failed to update the HBA entries for the new "
+				  "elements in the our formation \"%s\" and group %d",
+				  keeper->config.formation,
+				  keeper->state.current_group);
+		return false;
```
JelteF, line 1444: If this fails, we won't ever retry adding the HBA entry. Because `keeper->otherNodes` is already modified by `monitor_get_other_nodes`. We should only update keeper->otherNodes  once this succeeded, or we exited early because of `otherNodesArray->count == 0 || diffNodesArray.count == 0`

src/bin/pg_autoctl/keeper.c
```diff
@@ -1333,3 +1318,198 @@ keeper_state_as_json(Keeper *keeper, char *json, int size)
...
+	log_trace("keeper_refresh_other_nodes");
+
+	if (!monitor_get_other_nodes(monitor, host, port, ANY_STATE, otherNodesArray))
+	{
+		log_error("Failed to get_other_nodes() on the monitor");
+		return false;
+	}
+
+	/* compute nodes that need an HBA change (new ones, new hostnames) */
+	if (forceCacheInvalidation)
```
JelteF, line 1412: `forceCacheInvalidation` is not persistent across failures, e.g. `monitor_get_other_nodes` failed. I'm not sure we want that, but if we do we could achieve it easily by setting `keeper->otherNodes` to empty.

src/bin/pg_autoctl/fsm_transition.c
```diff
@@ -399,124 +397,13 @@ fsm_resume_as_primary(Keeper *keeper)
  * fsm_prepare_replication is used when a new standby was added.
  *
  * add_standby_to_hba && create_replication_slot
+ *
+ * Those operations are now done eagerly rather than just in time. So it's been
+ * taken care of aready, nothing to do within this state transition.
```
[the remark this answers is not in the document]
DimCitus: It would be possible yes. At the moment having a function that only does `return true` or a NULL function pointer in the KeeperFSM structure bears the same meaning. I though that we might want to do more things in that transition later though, and keeping that around does not sound like a maintenance burden...

src/bin/pg_autoctl/keeper.c
```diff
@@ -1333,3 +1318,198 @@ keeper_state_as_json(Keeper *keeper, char *json, int size)
...
+	}
+
+	return true;
+}
+
+
+/*
+ * merge_new_node_array merges the node Array just received from the monitor
+ * with the existing one. The nodes are all sorted on nodeId so we can
+ * implement a merge diff here.
```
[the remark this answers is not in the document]
DimCitus: Fixed, thanks.

src/bin/pg_autoctl/service_keeper.c
DimCitus: Fixed, thanks.

src/bin/pg_autoctl/service_keeper.c
DimCitus: I think it's better the way it is now. After thinking about it for a while, the main problem I see is the name of the boolean “CouldContactMonitor”... we kind of mean something else now. Will think about a better name, kept it that way for now.

src/bin/pg_autoctl/pghba.c
DimCitus: Fixed, thanks.

src/bin/pg_autoctl/pghba.c
DimCitus: Fixed, thanks.

src/bin/pg_autoctl/pghba.c
DimCitus: Fixed, thanks.

src/bin/pg_autoctl/pghba.c
DimCitus: Fixed too, thanks.

src/bin/pg_autoctl/keeper.c
DimCitus: Fixed now by re-installing the “current” version of the array when we fail to update the HBA, and some comments about why it appears safe to proceed that way.

src/bin/pg_autoctl/keeper.c
DimCitus: When we fail to get the other nodes we fail early in the `keeper_refresh_other_nodes` and then we fail in the transition function, so we will retry it on the next run of the loop. I think that's good.

src/bin/pg_autoctl/keeper.c
```diff
@@ -1333,3 +1318,211 @@ keeper_state_as_json(Keeper *keeper, char *json, int size)
...
+			 * We might have entries in previousNodesArray that are not found
+			 * in currentNodesArray anymore, but we don't know how to clean-up
+			 * the HBA file entries at the moment anyway, so we just skip them.
+			 */
+			diffNodesArray[diffIndex++] = currentNodesArray[currIndex];
+			break;
+		}
+		else
+		{
+			log_error("BUG in diff_nodesArray!");
```
onderkalaci, line 1524: Minor suggestion:
given that these type of functions are always prone to bugs, one suggestion could be to follow Postgres `list_difference`. And, an example of specific implementation in Citus code can be found [here](https://github.com/citusdata/citus/blob/004e0e4617081205dc48f7f6c58a3a7d4b954f20/src/backend/distributed/planner/multi_physical_planner.c#L5200)  

The  above  suggestion is certainly not  as optimized as this code, but looks  simpler/safer. Given that we don't expect 100s of nodes, this seems like a useful trade-off for me

DimCitus: I think I see what you mean, but the simplification you're proposing would still be O(n²), right? I find the current code simple enough at the moment to avoid getting there...

src/bin/pg_autoctl/fsm_transition.c
```diff
@@ -399,124 +397,13 @@ fsm_resume_as_primary(Keeper *keeper)
...
-		if (!postgres_sprintf_replicationSlotName(otherNode->nodeId,
-												  replicationSlotName, BUFSIZE))
-		{
-			/* that's highly unlikely... */
-			log_error("Failed to snprintf replication slot name for node %d",
-					  otherNode->nodeId);
-			return false;
-		}
-
-		if (!primary_create_replication_slot(postgres, replicationSlotName))
```
onderkalaci, line 511: I'm a little confused. We're also removing `primary_create_replication_slot`, and I couldn't see where we create replication slots other than this (other than cli).

Shouldn't we need to create `primary_create_replication_slot ` at some point?

DimCitus: That's a good point. The function that ensures that is badly named: `pgsql_replication_slot_drop_removed` is called by `postgres_replication_slot_drop_removed` and does both the `pg_drop_replication_slot` calls and also the `pg_create_physical_replication_slot` calls... Maybe this is a good time to rename that function.

src/bin/pg_autoctl/keeper.c
```diff
@@ -1317,3 +1302,211 @@ keeper_state_as_json(Keeper *keeper, char *json, int size)
...
+	NodeAddressArray diffNodesArray = { 0 };
+
+	char *host = config->nodename;
+	int port = postgresSetup->pgport;
+
+	log_trace("keeper_refresh_other_nodes");
+
+	if (!monitor_get_other_nodes(monitor, host, port, ANY_STATE, otherNodesArray))
+	{
+		log_error("Failed to get_other_nodes() on the monitor");
```
JelteF, line 1391: Maybe good to also reset `otherNodesArray` here like is done in the other error case. I don't think it's strictly necessary, because of the details of the parsing in `monitor_get_other_nodes`. But I think that's a fairly specific implementation detail, that's best not to be relied upon.

Another way would be to it the other way around and only change `keeper->otherNodes` when we finish successfully.
Instead of `otherNodesArray` you could have something like:
```c
NodeAddressArray newNodesArray = { 0 };
NodeAddressArray oldNodesArray = keeper->otherNodes;
NodeAddressArray diffNodesArray = { 0 };
if (!monitor_get_other_nodes(monitor, host, port, ANY_STATE, newNodesArray)) {
    return false;
}
// do things
keeper->otherNodes = newNodesArray;
return true
```

I think I like this second approach a bit better, since I think it makes the code easier to understand.

DimCitus: Fair point, done that way now.

src/bin/pg_autoctl/service_keeper.c
```diff
@@ -596,6 +500,144 @@ keeper_node_active_loop(Keeper *keeper, pid_t start_pid)
...
+		 * Check whether we're likely to be in a network partition.
+		 * That will cause the assigned_role to become demoted.
+		 */
+		if (keeperState->current_role == PRIMARY_STATE)
+		{
+			log_warn("Checking for network partitions...");
+
+			if (!is_network_healthy(keeper))
+			{
+				keeperState->assigned_role = DEMOTE_TIMEOUT_STATE;
```
JelteF, line 574: NOTE: This behaviour described in this comment also applies the code on the `master` branch. However, it seems wrong or at least very confusing there also.

After reading and understanding this code better I realised that this assigned role will never actually become the `current_role`. The reason for this is that `needsStateChange` is only set when `couldContactMonitor` is `true`. Which is always `false` since this code is only executed when the monitor could not be reached. This means postgres is never actually turned off when `DEMOTE_TIMEOUT_STATE` is set.

I tried this out locally:
1. Start monitor
2. Start primary
3. Start secondary
4. kill both monitor and secondary
5. Wait 20 seconds

Expected behaviour:
6. the primary postgres is turned off 
Actual behaviour:
6. we start getting these log messages:
```
16:11:59 31917 WARN  Checking for network partitions...
16:11:59 31917 INFO  Failed to contact the monitor or standby in 18446744073709551581 seconds, at 20 seconds we shut down PostgreSQL to prevent split brain issues
16:11:59 31917 INFO  Network in not healthy, switching to state demote_timeout
```
However `psql -p {primaryPort}` still works. Write queries are blocked though, since the secondary cannot be reached and it is in `sync_stanbys`. 

I think the main problem that this behaviour causes is that clients will still connect to this primary, since `target_session_attrs=read-write` is still true for this server. 

Side-issue, the seconds in the log message are clearly incorrect:
```
Failed to contact the monitor or standby in 18446744073709551581 seconds
```

src/bin/pg_autoctl/service_keeper.c
JelteF: Can this cause any issues, where if calling `keeper_config_set_groupId_and_slot_name` fails we do not demote the primary when we should? Something, like what is happening now with `DEMOTE_TIMEOUT_STATE`.

src/bin/pg_autoctl/service_keeper.c
DimCitus: It looks to me that the culprit is a change that (IIRC) comes with this PR, where we protected FSM changes when we failed to contact the monitor, which is the wrong thing to do as you're showing. So to me, on a first reading, it boils down to the following change that I have just added:

``` diff
modified   src/bin/pg_autoctl/service_keeper.c
@@ -372,13 +372,21 @@ keeper_node_active_loop(Keeper *keeper, pid_t start_pid)
 
 		couldContactMonitor = couldContactMonitorThisRound;
 
-		if (couldContactMonitor &&
-			keeperState->assigned_role != keeperState->current_role)
+		if (keeperState->assigned_role != keeperState->current_role)
 		{
 			needStateChange = true;
 
```

Please have a look at this change.

DimCitus: I see that it's also broken on master in a different way now, so the breakage has been introduced earlier. It still looks to me that the current approach should fix it in this PR.

src/bin/pg_autoctl/service_keeper.c
```diff
@@ -372,13 +372,21 @@ keeper_node_active_loop(Keeper *keeper, pid_t start_pid)
...
-			log_info("Monitor assigned new state \"%s\"",
-					 NodeStateToString(keeperState->assigned_role));
+			if (couldContactMonitor)
+			{
+				log_info("Monitor assigned new state \"%s\"",
+						 NodeStateToString(keeperState->assigned_role));
+			}
+			else
+			{
+				/* if network is not healthy we might self-assign a state */
```
JelteF, line 386: This comment is not strictly correct, since `couldContactMonitor` can be false because either `keeper_refresh_other_nodes`, `keeper_config_set_groupId_and_slot_name` or `keeper_ensure_configuration` failed.

JelteF: To be clear, in that case the new value would come from the monitor. Just some work after it would have failed.

DimCitus: That's right. That said the comment is still valid, it's just incomplete. The other cases you mention are that we could contact the monitor, but then failed to sync the new values received with our local state/config. In my notes I have also seen that we should remove the replication slot name from the config, as we now automate it from the node id. We rely in more and more places on the name being `pgautofailover_standby_%d`. So I will open a PR to clean that up and include some more comments on why we don't want to apply the monitor state change when we failed to sync local state or local config from the given values.

PR #297, opened 2020-06-14
Add support for building against Postgres 13

lfittl:
Depends on https://github.com/citusdata/tools/pull/78 for Postgres 13 testing on Travis.

PR #312, opened 2020-06-25
Add support for building against Postgres 13

lfittl-msft:
Re-opening #297 with my MSFT account.

### DIFF
--- a/src/bin/pg_autoctl/supervisor.c
+++ b/src/bin/pg_autoctl/supervisor.c
@@ -226,8 +226,7 @@ supervisor_loop(Supervisor *supervisor)
 				}
 				else
 				{
-					log_fatal("Failed to call waitpid(): %m");
-					return false;
+					log_debug("Failed to call waitpid(): %m");
 				}
 			}
 
--- a/src/bin/pg_autoctl/supervisor.c
+++ b/src/bin/pg_autoctl/supervisor.c
@@ -220,6 +220,9 @@ supervisor_loop(Supervisor *supervisor)
 						log_info("Internal subprocesses are done, stopping");
 						return true;
 					}
+
+					log_fatal("Unexpected ECHILD error from waitpid()");
+					return false;
 				}
 				else
 				{
--- a/src/bin/pg_autoctl/supervisor.c
+++ b/src/bin/pg_autoctl/supervisor.c
@@ -228,6 +228,8 @@ supervisor_loop(Supervisor *supervisor)
 				{
 					log_debug("Failed to call waitpid(): %m");
 				}
+
+				break;
 			}
 
 			case 0:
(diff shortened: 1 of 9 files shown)


src/bin/pg_autoctl/supervisor.c
```diff
@@ -219,6 +219,9 @@ supervisor_loop(Supervisor *supervisor)
 						log_info("Internal subprocesses are done, stopping");
 						return true;
 					}
+
+					log_fatal("Unexpected ECHILD error from waitpid()");
+					return false;
```
DimCitus, line 224: It could be that someone did `kill -TERM` or `kill -QUIT` all our child processes without signaling the supervisor process. Now, we could restart every sub-process then, or yeah, bail out to the caller (systemd in many cases) for them to do something about it (restart the main `pg_autoctl` service in many cases).

src/monitor/node_metadata.c
```diff
@@ -360,7 +360,12 @@ FindFailoverNewStandbyNode(List *groupNodeList)
...
-pgautofailover_node_candidate_priority_compare(const void *a, const void *b)
+#if (PG_VERSION_NUM >= 130000)
+pgautofailover_node_candidate_priority_compare(const union ListCell *a,
+											   const union ListCell *b)
+#else
+pgautofailover_node_candidate_priority_compare(const void * a, const void * b)
+#endif
 {
 	AutoFailoverNode *node1 = (AutoFailoverNode *) lfirst(*(ListCell **) a);
 	AutoFailoverNode *node2 = (AutoFailoverNode *) lfirst(*(ListCell **) b);
```
DimCitus, line 371: I'm surprised this works both in Postgres 12 and Postgres 13, where I read the function prototype in a way that we are now doing one pointer dereference too many. I am not sure that we exercise this code path yet in the test suite, it's part of the Multiple Standby PR building blocks... which might explain that it could compile but it's going to fail.

DimCitus:
Let's review that `lfirst(*(ListCell **) a)` once more, that's the only part of the PR I'm not convinced about yet...

src/monitor/node_metadata.c
DimCitus: Fixed now.

src/bin/pg_autoctl/supervisor.c
DimCitus: So I don't think we should return false in that case, because we don't know if we still have child process running or not. When we quit and child process are running, they get re-assigned to init (pid 1) on Unix, and having a stale running Postgres in the background is not going to help anyone.

The error message is now a `log_debug()` and we continue looping.

lfittl-msft:
@DimCitus Let me know if you'd like me to merge this - would be great to have for 1.4 release :)

PR #314, opened 2020-06-28
Add guards on memory allocation failure

danielgustafsson:
Make sure to catch in case the memory allocation, and exit gracefully in case of memory exhaustion. This falls in the "unlikely to happen" bucket but makes static analysis easier and is generally good hygiene to follow.


DimCitus:
Thanks @danielgustafsson!

PR #318, opened 2020-07-01
Check for monitor extension version every time we connect.

DimCitus:


src/bin/pg_autoctl/service_keeper.c
```diff
@@ -534,6 +525,32 @@ keeper_node_active(Keeper *keeper)
 	bool forceCacheInvalidation = false;
 	bool reportPgIsRunning = ReportPgIsRunning(keeper);
 
+	/*
+	 * First, connect to the monitor and check we're compatible with the
+	 * extension there. An upgrade on the monitor might have happened in
+	 * between loops here.
+	 */
+	if (!keeper_check_monitor_extension_version(keeper))
```
JelteF, line 533: I think to be really safe this should be done in the same transaction with the `node_active` call. Otherwise the version might change inbetween the version check and the actual `node_active` call. To really get that behaviour we might even need to use `SELECT FOR SHARE` or something.

src/bin/pg_autoctl/monitor.c
```diff
@@ -2998,9 +2998,6 @@ monitor_get_extension_version(Monitor *monitor, MonitorExtensionVersion *version
 		return false;
 	}
 
-	/* disconnect from PostgreSQL now */
-	pgsql_finish(&monitor->pgsql);
```
JelteF, line 3002: This removal also has effect on `monitor_ensure_extension_version`, because this function is called there twice too.

src/bin/pg_autoctl/service_keeper.c
DimCitus: I don't think we need that strong a guarantee: the extension on the monitor is protected against running a version of the `node_active` (or any other) function that does not match with the SQL level version. Then, if we changed the API without changing the arguments, that means we changed what we may return. We are protected against changes in number of return values, so we're left with changes within the columns themselves. Basically that's a new state that we don't know how to handle. In that case we're going to fail to parse it, and at next attempt we're going to catch up with the new version number.

All in all, it seems to me that the worst case is going to be one extra call before we restart node active process, and an extra error message in the logs during the live upgrade of pg_auto_failover.

JelteF: Sounds good, maybe add this reasoning to the comment.

src/bin/pg_autoctl/monitor.c
DimCitus: The `monitor_ensure_extension_version` in turn is only called in `monitor_service_run` which implements a LISTEN/NOTIFY client loop, where we keep the same connection for a long time anyway. I think it's okay even in that case that we're not closing the monitor connection anymore in the lower-level functions, the code would just re-open the connection at the next step anyway...

PR #321, opened 2020-07-06
Only use replication slots on standby nodes for good Postgres versions.

### LINKED ISSUE #283: WAL files not recycled on archive=on replicas
Hi,

It is quite possible that I am missing something but I think the following PR introduced a regression  (or at least a behavior change) in pg_auto_failover v1.3  : https://github.com/citusdata/pg_auto_failover/pull/222

Because pg_autoctl would automatically (re)create a replication slot on a replica and having the archive=on config option enabled, postgresql will *not* recycle WAL files that are no longer used after a CHECKPOINT like it normally should and keep those old WAL files indefinitely, in the eventuality that another replica will later attach to this replication slot and replay the transaction backlog.

The outcome is that the disk used by WAL files in the $PGDATA/pg_wal folder on the replica will grow very quickly, eventually resulting in a disk-full situation.

The primary doesn't behave like this and WAL recycling is working fine because the replica is consuming data on its replication slot.

Then, when a replica reach a disk-full situation, the primary reach a point were new non-readonly transactions will hang thanks to the synchronous replication setting and the replica not able to COMMIT on its full filesystem... and you can imagine what's come next :-)

This setup was working fine with pg_auto_failover v1.1 but not anymore.
As a temporary workaround and to ensure I always have an up-to-date replica, I had to enable maintenance mode on my replica (pg_autoctl won't recreate the replication slot when in maintenance mode) and manually delete the unused replication slot for the database service to be functional for more than a few hours.

But this is obviously not ideal as the autofailover feature is not functional.

I guess there is good reason to create the replication slot on all replicas but, if my observation are correct, this is not really compatible with archive=on.

Thoughts ?

PS: thanks for this project, it's awesome :-)

DimCitus:
A bug in Postgres prevents WAL recycling on standby nodes when we use the
function pg_replication_slot_advance() to maintain their LSN position. In
this patch we work around the bug to avoid it by just not installing our
replication slots on the standby nodes when we're running a Postgres version
with the bug. That's 11.0 to 11.8 included, and 12.0 to 12.3 included.

Fixes #283.

### DIFF
--- a/src/bin/pg_autoctl/cli_config.c
+++ b/src/bin/pg_autoctl/cli_config.c
@@ -334,6 +334,15 @@ cli_config_check_connections(PostgresSetup *pgSetup,
 				 "please review your Postgres configuration");
 	}
 
+	if (pg_setup_standby_slot_supported(pgSetup, LOG_WARN))
+	{
+		int major = pgSetup->control.pg_control_version / 100;
+		int minor = pgSetup->control.pg_control_version % 100;
+
+		log_info("Postgres version %d.%d allows using replication slots "
+				 "on the standby nodes", major, minor);
+	}
+
 	/*
 	 * Now, on Postgres nodes, check that the monitor uri is valid and that we
 	 * can connect to the monitor just fine. This requires having setup the
--- a/src/bin/pg_autoctl/config.c
+++ b/src/bin/pg_autoctl/config.c
@@ -105,7 +105,7 @@ build_xdg_path(char *dst,
 		 * yet, precluding the use of realpath(3) to get the absolute name
 		 * here.
 		 */
-		char currentWorkingDirectory[MAXPGPATH];
+		char currentWorkingDirectory[MAXPGPATH] = { 0 };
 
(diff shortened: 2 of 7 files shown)


DimCitus:
> Something is not working in this PR. I'm runing PG 12.3 and a replication slot is still created on the standby.

We still create the replication slots when PG_AUTOCTL_DEBUG is set, so that we can test our logic for creating the slots and using them in Travis and interactively as developers. It is not expected that we create them when PG_AUTOCTL_DEBUG is not set though. Was it set in your environment when testing?

JelteF:
> Was it set in your environment when testing?

No, it was not.

DimCitus:
> > Was it set in your environment when testing?
> 
> No, it was not.

Okay, I found the issue. We force the variable to true always nowadays because we run the sub-processes via the DEBUG command line: `pg_autoctl do service ...`. We now use `PG_REGRESS_SOCK_DIR` in addition to `PG_AUTOCTL_DEBUG` to detect that we're in the test suite environment, where we still want to check for replication slots management.

PR #323, opened 2020-07-06
Handle empty pgdata during creating monitor

danielgustafsson:
In case the --pgdata parameter to "create monitor" is an existing, but empty directory, the process would fail with a rather hard to interpret error message.  Make sure to properly support existing empty directories and initdb them just like how a non-existing directory is handled.

### DIFF
--- a/src/bin/pg_autoctl/file_utils.c
+++ b/src/bin/pg_autoctl/file_utils.c
@@ -7,6 +7,7 @@
  *
  */
 
+#include <dirent.h>
 #include <limits.h>
 #include <sys/stat.h>
 #include <unistd.h>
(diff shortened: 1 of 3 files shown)


DimCitus:
Hey @danielgustafsson ; I'm not so sure this is the right/obvious way to fix it, but I'm curious. What is the error you're seeing without your patch? The one I have locally kind of doesn't make sense with the fix I'm seeing:

```
ERROR file_utils.c:252: Failed to open file "/private/tmp/daniel/monitor/postgresql.conf": No such file or directory
ERROR monitor_pg_init.c:330: Failed to add default settings to "/private/tmp/daniel/monitor/postgresql.conf": couldn't write the new postgresql.conf, see above for details
```

This error comes from something we changed very recently:

```
DEBUG pgsetup.c:382: Found PostgreSQL system 0 at "/private/tmp/daniel/monitor", version 0, catalog version 0
```

And I think we should fix that discovery thing and export the information that the directory does not exists or is empty in a better way in our internal representation of the PostgresSetup.

danielgustafsson:
The error I'm getting is:
```
$ pg_autoctl create monitor --auth trust --ssl-self-signed --pgdata ./monitor --pgport 6000
00:25:24 88744 INFO  Using default --ssl-mode "require"
00:25:24 88744 INFO  Using --ssl-self-signed: pg_autoctl will create self-signed certificates, allowing for encrypted network traffic
00:25:24 88744 WARN  Self-signed certificates provide protection against eavesdropping; this setup does NOT protect against Man-In-The-Middle attacks nor Impersonation attacks.
00:25:24 88744 WARN  See https://www.postgresql.org/docs/current/libpq-ssl.html for details
00:25:24 88744 INFO   /Users/dgustafsson/dev/postgresql/pg_auto_failover/obj/openssl/bin/openssl req -new -x509 -days 365 -nodes -text -out /Users/dgustafsson/dev/postgresql/pg_auto_failover/data/monitor/server.crt -keyout /Users/dgustafsson/dev/postgresql/pg_auto_failover/data/monitor/server.key -subj "/CN=daniels-mbp"
00:25:24 88744 ERROR Failed to open file "/Users/dgustafsson/dev/postgresql/pg_auto_failover/data/monitor/postgresql.conf": No such file or directory
00:25:24 88744 ERROR Failed to add default settings to "/Users/dgustafsson/dev/postgresql/pg_auto_failover/data/monitor/postgresql.conf": couldn't write the new postgresql.conf, see above for details
00:25:24 88744 FATAL Failed to initialize our Postgres settings, see above for details
```
..which isn't all too helpful for users. Since the certificates were successfully created in the empty directory, the monitor node should be too. I'll have another look at conveying it via `PostgresSetup` in the morning.

DimCitus:
We might have introduced some breaking changes in https://github.com/citusdata/pg_auto_failover/commit/2f27724d2d82dacd38c4b159d7b954f5e86d9e9f recently, have a look maybe there. It might not be related, that said, but I think reviewing that merged PR might help you here.

PR #325, opened 2020-07-06
Add alt text to all images

jonels-msft:


jonels-msft:
I rebased out all changes except the alt tags.

The theme upgrade didn't gain us any fixes, the current version is recent enough.

Converting tabular output to actual tables obscured the fact that the output came from the command line tools, as well as rendering poorly in PDF.

PR #331, opened 2020-07-13
Fix elapsed time display and keeper state update.

### LINKED ISSUE #330: Log of no-contact time is clearly wrong
These seconds are clearly incorrect
```
Failed to contact the monitor or standby in 18446744073709551581 seconds
```

DimCitus:
Fixes #330


DimCitus:
We can't obtain the LSN of a Postgres node that's not running... at restart it might do some crash recovery and the LSN will change too. Or a `pg_rewind` has happened while it was down, and the LSN changes then too.

Issue #277, opened 2020-06-02
Fix FATAL authentication errors in logs


JelteF:
Since pg_auto_failover 1.3 (because of #247) there's been these errors in server logs:
```
FATAL: no pg_hba.conf entry for host ‘172.xxx.xxx.xxx’, user ‘pgautofailover_monitor’, database ‘postgres’, SSL off
```

NOTE: These errors are not actually a problem. The healtcheck works even if authentication fails.

We should fix this as it's confusing to users and is flooding the logs.

I think the easiest way is to change to a TCP ping, instead of doing a postgres level ping. Essentially we are doing this anyway already.

PR #332, opened 2020-07-16
Decrease healtcheck period to 1 second for faster alive detection

### LINKED ISSUE #329: Transition from catchingup to secondary takes a long time
For some reason on an empty cluster, when adding a standby for the first time the transition from `catchingup` to `secondary` takes 15 seconds. 
```
12:46:30 7863 INFO  The version of extension "pgautofailover" is "1.3" on the monitor
12:46:30 7863 INFO  Reloaded the new configuration from "/home/jelte/.config/pg_autoctl/home/jelte/work/pg_auto_failover/node2/pg_autoctl.cfg"
12:46:30 7863 INFO  pg_autoctl service is running, current state is "catchingup"
12:46:45 7863 INFO  Monitor assigned new state "secondary"
12:46:45 7863 INFO  FSM transition from "catchingup" to "secondary": Convinced the monitor that I'm up and running, and eligible for promotion again
12:46:45 7863 INFO  Creating replication slot "pgautofailover_standby_1"
12:46:45 7863 INFO  Transition complete: current state is now "secondary"
```

I'm not sure what's happening here. Maybe it is expected. However, it's a bit weird that it takes 15 seconds while seemingly nothing happens. Note, `pg_basebackup` has already completed at that point. Also when looking at `pgautofailover.node` both nodes report the same LSN for quite some time already.

JelteF:
This seemed to locally fix #329 for me. I'm mainly openining this as a draft PR to check if it improves test run times in CI as well.

### DIFF
--- a/src/bin/pg_autoctl/cli_config.c
+++ b/src/bin/pg_autoctl/cli_config.c
@@ -334,6 +334,15 @@ cli_config_check_connections(PostgresSetup *pgSetup,
 				 "please review your Postgres configuration");
 	}
 
+	if (pg_setup_standby_slot_supported(pgSetup, LOG_WARN))
+	{
+		int major = pgSetup->control.pg_control_version / 100;
+		int minor = pgSetup->control.pg_control_version % 100;
+
+		log_info("Postgres version %d.%d allows using replication slots "
+				 "on the standby nodes", major, minor);
+	}
+
 	/*
 	 * Now, on Postgres nodes, check that the monitor uri is valid and that we
 	 * can connect to the monitor just fine. This requires having setup the
--- a/src/bin/pg_autoctl/config.c
+++ b/src/bin/pg_autoctl/config.c
@@ -105,7 +105,7 @@ build_xdg_path(char *dst,
 		 * yet, precluding the use of realpath(3) to get the absolute name
 		 * here.
 		 */
-		char currentWorkingDirectory[MAXPGPATH];
+		char currentWorkingDirectory[MAXPGPATH] = { 0 };
 
 		if (getcwd(currentWorkingDirectory, MAXPGPATH) == NULL)
 		{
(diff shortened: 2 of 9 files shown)


DimCitus:
I think we have other GUCs that drive the health check work with timeouts and retry policy, and I'm not sure that 1 second only is currently a wise choice given the other settings there:

``` c
/* GUC variables */
int HealthCheckPeriod = 20 * 1000;
int HealthCheckTimeout = 5 * 1000;
int HealthCheckMaxRetries = 2;
int HealthCheckRetryDelay = 2 * 1000;
```

That said changing the default from 20s down to something faster sounds reasonable to me, unless it introduces flaps where we would failover too often for no strong reasons. Maybe we will need to re-think the health check process entirely at some point and allow other nodes to contribute to the health check tests in some ways.

DimCitus:
Meanwhile what we could/should do is maybe change the GUC in the test suite. That sounds easy enough to install our own version of that setting when creating a monitor node there?

JelteF:
Yes, I completely agree this change should not be merged like this, especially given all the settings that we have. I merely opened it to see what the impact was on test runs.
Turns out runtime of individual jobs goes down by 2-3 minutes, so pretty significant but not too crazy ([before](https://travis-ci.com/github/citusdata/pg_auto_failover/builds/175588088) and [after](https://travis-ci.com/github/citusdata/pg_auto_failover/builds/175865977)).

Some options I see:
1. Rethink the values of the settings you mention with 1 second HealthCheckPeriod in mind.
2. Fix this specific issue would be to do the health check once per second whenever a node is considered down. 

One thing to keep in mind if we go for option 1 is that we would currently increase the impact of https://github.com/citusdata/pg_auto_failover/issues/277, because we would get even more authentication errors in the logs.

I like your idea of changing the GUC in the test suite to get the benefits to our test suite sooner, but on the other hand it's good to use the default settings that users use as well.

PR #333, opened 2020-07-16
Fix/various bugs

DimCitus:


src/bin/pg_autoctl/keeper.c
```diff
@@ -1568,7 +1566,8 @@ diff_nodesArray(NodeAddressArray *previousNodesArray,
 				log_debug("Node %d has a new hostname \"%s\"",
 						  currNode->nodeId, currNode->host);
 
-				diffNodesArray[diffIndex++] = currentNodesArray[currIndex];
+				diffNodesArray->count++;
+				diffNodesArray->nodes[diffIndex++] = *currNode;
 			}
 			prevIndex++;
```
JelteF, line 1572: If we increase `prevIndex` further than `previousNodesArray->count` we would get uninitialized/zerod out nodes on the next iteration of this loop. I believe this can easily be fixed by doing something like this instead:
```suggestion
            if (previousNodesArray->count > prevIndex) {
			    prevIndex++;
			}
```

DimCitus: Thanks, done now. Well I did the test the other way round, when that makes sense with the code I would rather read numbers from left to right with the smaller on the left. I have found it easier to read that way. In that case also we can see prevIndex “moving” closer to the wall that the count represents, and then being stopped there. Anyway, fixed.

PR #335, opened 2020-07-17
Feature/pgtune

DimCitus:
In many cases `pg_autoctl` is going to do the `pg_ctl initdb` installation of a new Postgres database cluster.
We then have the opportunity to compute some dynamic Postgres tuning, and in this PR we do that for a couple
of easy to compute and highly rewarding settings, such as memory related settings.

### DIFF
--- a/src/bin/pg_autoctl/pgtuning.c
+++ b/src/bin/pg_autoctl/pgtuning.c
@@ -1,0 +1,338 @@
+/*
+ * src/bin/pg_autoctl/pgtuning.c
+ *     Adjust some very basic Postgres tuning to the system properties.
+ *
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the PostgreSQL License.
+ *
+ */
+
+#include "postgres_fe.h"
+#include "pqexpbuffer.h"
+
+#include "config.h"
+#include "file_utils.h"
+#include "log.h"
+#include "pgctl.h"
+#include "pgtuning.h"
+#include "system_utils.h"
+
+/*
+ * In most cases we are going to initdb a Postgres instance for our users, we
+ * might as well introduce some naive Postgres tuning. In the static array are
+ * selected Postgres default values and static values we always set.
+ *
+ * Dynamic code is then used on the target systems to compute better values
+ * dynamically for some parameters: work_mem, maintenance_work_mem,
+ * effective_cache_size, autovacuum_max_workers.
+ */
+GUC postgres_tuning[] = {
+	{ "shared_preload_libraries", "'pg_stat_statements'" },
+	{ "track_io_timing", "on" },
+	{ "track_functions", "pl" },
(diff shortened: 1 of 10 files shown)


nathanhere:
This is great! 👏 👏 👏

src/bin/pg_autoctl/pgtuning.c
```diff
@@ -0,0 +1,335 @@
...
+ * In most cases we are going to initdb a Postgres instance for our users, we
+ * might as well introduce some naive Postgres tuning. In the static array are
+ * selected Postgres default values and static values we always set.
+ *
+ * Dynamic code is then used on the target systems to compute better values
+ * dynamically for some parameters: work_mem, maintenance_work_mem,
+ * effective_cache_size, autovacuum_max_workers.
+ */
+GUC postgres_tuning[] = {
+	{ "track_io_timing", "on" },
```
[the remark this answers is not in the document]
DimCitus: I guess we could be conservative and turn it off then. The reason why I went ahead and activated it is because of those data points:

> Finally there are blk_read_time and blk_write_time. Usually those fields are empty unless track_io_timing is turned on. The idea here is to be able to measure the amount of time a certain type of query spends on I/O. It will allow you to answer the question whether your system is I/O bound or CPU bound. In most cases it is a good idea to turn on I/O timing because it will give you vital information.

That's from https://www.cybertec-postgresql.com/en/detecting-performance-problems-easily-in-postgresql/. And then http://pgconfigurator.cybertec.at sets it on too.

Reading https://www.postgresql.org/docs/current/pgtesttiming.html again, I though we could maybe parse the output of that tool to decide if we can set track_io_timings on, but it's complex enough that I now believe we should not mess with that thing. Going to push a revision where it's off.

src/bin/pg_autoctl/system_utils.c
```diff
@@ -0,0 +1,136 @@
...
+
+#endif
+
+
+/*
+ * pretty_print_bytes pretty prints bytes in a human readable form. Given
+ * 17179869184 it places the string "16 GB" in the given buffer.
+ */
+void
+pretty_print_bytes(char *buffer, size_t size, uint64_t bytes)
```
JelteF, line 113: This rounds down `1.9999 GB` as a byte number to `1 GB`. Is that really what we want here?

DimCitus: I just pushed a change that should now display `'1999 MB'`, and display `4096 MB` where it would before display `4 GB`. Please try the changes?

JelteF: Cool on my machine I now get `1989 MB` shared_buffers

DimCitus: I'm now thinking that we might need a way to instruct pg_autoctl to not do the tuning... can you still create a monitor and 3 postgres or 4 nodes locally for testing? ;-)

PR #338, opened 2020-07-22
Fix bug in pg_setup_controldata.

DimCitus:
We need to copy over at pgSetup.pgdata and pgSetup.pg_ctl before calling
into the pg_controldata binary.

### DIFF
--- a/src/bin/pg_autoctl/pgsetup.c
+++ b/src/bin/pg_autoctl/pgsetup.c
@@ -451,6 +451,9 @@ pg_setup_controldata(PostgresSetup *pgSetup, bool missingPgDataIsOk)
 	/*
 	 * Now fetch new control values from running pg_controldata on PGDATA.
 	 */
+	strlcpy(newPgSetup.pgdata, pgSetup->pgdata, MAXPGPATH);
+	strlcpy(newPgSetup.pg_ctl, pgSetup->pg_ctl, MAXPGPATH);
+
 	if (!pg_controldata(&newPgSetup, missingPgDataIsOk))
 	{
 		/* errors have already been logged */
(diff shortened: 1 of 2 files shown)


src/bin/pg_autoctl/pgsetup.c
```diff
@@ -451,6 +451,9 @@ pg_setup_controldata(PostgresSetup *pgSetup, bool missingPgDataIsOk)
 	/*
 	 * Now fetch new control values from running pg_controldata on PGDATA.
 	 */
+	strlcpy(newPgSetup.pgdata, pgSetup->pgdata, MAXPGPATH);
```
onderkalaci, line 454: I think this makes sense and fixes the  problem that  I've observed (already tested).

Do you  think would using `pg_setup_init(pgSetup, pgSetup , ..`  make it more robust? 

Either ways works, but it seems slightly preferable to use `pg_setup_init`  as  it does  some extra checks and make the remaining of the  code path a  little more robust for future changes

DimCitus: Short answer: we can't `pg_setup_init` with same pgSetup as both source and target. Also, it could be that we don't actually need to call `pg_setup_controldata` at all when Postgres is not running, that might be the long term approach.

Longer answer: I would like to refactor this function's prototype and have an “alias” of the structure `PostgresSetup` that would be named `PostgresSetupTemplate` or something like that. The idea would then to distinguish a template from a probed instance, and then the prototype would be:

``` c
bool pg_setup_init(PostgresSetup *pgSetup,
				   PostgresSetupTemplate *options,
```

That would allow the compiler to help us distinguish between a forged pgSetup in memory that we have not verified with the current on-disk situation from the probed actual Postgres Setup we have discovered.

PR #340, opened 2020-07-23
Implement a node name.

DimCitus:
Nodes already have an id, and then technical information to connect to them
such as the hostname and port. In this PR we add the notion of a node name.

This PR replaces #249.

### DIFF
--- a/src/bin/pg_autoctl/cli_create_drop_node.c
+++ b/src/bin/pg_autoctl/cli_create_drop_node.c
@@ -92,7 +92,8 @@ CommandLine create_postgres_command =
 		"  --listen          PostgreSQL's listen_addresses\n"
 		"  --username        PostgreSQL's username\n"
 		"  --dbname          PostgreSQL's database name\n"
-		"  --hostname        pg_auto_failover node\n"
+		"  --name            pg_auto_failover node name\n"
+		"  --hostname        hostname used to connect from the other nodes\n"
 		"  --formation       pg_auto_failover formation\n"
 		"  --monitor         pg_auto_failover Monitor Postgres URL\n"
 		"  --auth            authentication method for connections from monitor\n"
@@ -218,6 +219,7 @@ cli_create_postgres_getopts(int argc, char **argv)
 		{ "auth", required_argument, NULL, 'A' },
 		{ "skip-pg-hba", no_argument, NULL, 'S' },
 		{ "dbname", required_argument, NULL, 'd' },
+		{ "name", required_argument, NULL, 'a' },
 		{ "hostname", required_argument, NULL, 'n' },
 		{ "formation", required_argument, NULL, 'f' },
 		{ "monitor", required_argument, NULL, 'm' },
@@ -242,7 +244,7 @@ cli_create_postgres_getopts(int argc, char **argv)
 
 	int optind =
 		cli_create_node_getopts(argc, argv, long_options,
-								"C:D:H:p:l:U:A:Sd:n:f:m:MRVvqhP:r:xsN",
+								"C:D:H:p:l:U:A:Sd:a:n:f:m:MRVvqhP:r:xsN",
 								&options);
 
 	/* publish our option parsing in the global variable */
--- a/src/bin/pg_autoctl/cli_create_drop_node.c
+++ b/src/bin/pg_autoctl/cli_create_drop_node.c
@@ -144,7 +144,10 @@ cli_create_config(Keeper *keeper, KeeperConfig *config)
 	 */
 	if (file_exists(config->pathnames.config))
 	{
+		Monitor *monitor = &(keeper->monitor);
 		KeeperConfig options = *config;
+		KeeperConfig oldConfig = { 0 };
+		PostgresSetup optionsFullPgSetup = { 0 };
 
 		if (!keeper_config_read_file(config,
 									 missingPgdataIsOk,
@@ -156,6 +159,23 @@ cli_create_config(Keeper *keeper, KeeperConfig *config)
 			exit(EXIT_CODE_BAD_CONFIG);
 		}
 
+		oldConfig = *config;
+
+		/*
+		 * Before merging command line options into the (maybe) pre-existing
+		 * configuration file, we should also mix in the environment variables
+		 * values in the command line options.
+		 */
+		if (!pg_setup_init(&optionsFullPgSetup,
+						   &(options.pgSetup),
+						   missingPgdataIsOk,
+						   pgIsNotRunningIsOk))
+		{
+			exit(EXIT_CODE_BAD_ARGS);
+		}
+
+		options.pgSetup = optionsFullPgSetup;
+
 		/*
 		 * Now that we have loaded the configuration file, apply the command
 		 * line options on top of it, giving them priority over the config.
@@ -164,6 +184,27 @@ cli_create_config(Keeper *keeper, KeeperConfig *config)
 		{
 			/* errors have been logged already */
 			exit(EXIT_CODE_BAD_CONFIG);
+		}
+
+		/*
+		 * If we have registered to the monitor already, then we need to check
+		 * if the user is providing new --nodename, --hostname, or --pgport
+		 * arguments. After all, they may change their mind of have just
+		 * realized that the --pgport they wanted to use is already in use.
+		 */
+		if (!monitor_init(monitor, config->monitor_pguri))
+		{
+			/* errors have already been logged */
+			exit(EXIT_CODE_BAD_ARGS);
+		}
+
+		if (file_exists(config->pathnames.state) && !config->monitorDisabled)
+		{
+			if (!keeper_set_node_metadata(keeper, &oldConfig))
+			{
+				/* errors have already been logged */
+				exit(EXIT_CODE_MONITOR);
+			}
 		}
 	}
 	else
--- a/src/bin/pg_autoctl/cli_get_set_properties.c
+++ b/src/bin/pg_autoctl/cli_get_set_properties.c
@@ -19,12 +19,11 @@ static void cli_get_formation_number_sync_standbys(int argc, char **argv);
 
 static void cli_set_node_replication_quorum(int argc, char **argv);
 static void cli_set_node_candidate_priority(int argc, char **argv);
-static void cli_set_node_hostname(int argc, char **argv);
+static void cli_set_node_metadata(int argc, char **argv);
 static void cli_set_formation_number_sync_standbys(int arc, char **argv);
 
 static bool set_node_candidate_priority(Keeper *keeper, int candidatePriority);
 static bool set_node_replication_quorum(Keeper *keeper, bool replicationQuorum);
-static bool set_node_hostname(Keeper *keeper, const char *hostname);
 static bool set_formation_number_sync_standbys(Monitor *monitor,
 											   char *formation,
 											   int groupId,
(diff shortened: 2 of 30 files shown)


src/monitor/node_active_protocol.c
```diff
@@ -329,30 +342,17 @@ node_active(PG_FUNCTION_ARGS)
...
 	if (pgAutoFailoverNode == NULL)
 	{
-		ereport(ERROR, (errmsg("node %s:%d is not registered",
-							   nodeHost, nodePort)));
-	}
-	else if (strcmp(pgAutoFailoverNode->formationId, formationId) != 0)
-	{
-		ereport(ERROR, (errmsg("node %s:%d does not belong to formation %s",
-							   nodeHost, nodePort, formationId)));
-	}
```
JelteF, line 348: This check looks like it should still be done

src/bin/pg_autoctl/monitor.c
```diff
@@ -1518,65 +1529,60 @@ printCurrentState(void *ctx, PGresult *result)
...
+	for (int i = 0; i < maxNameSize; i++)
 	{
-		if (i < maxHostNameSize)
-		{
-			nameSeparatorHeader[i] = '-';
-		}
-		else
-		{
-			nameSeparatorHeader[i] = '\0';
-		}
```
JelteF, line 1530: What was this case used for and why is it not needed anymore now?

src/bin/pg_autoctl/monitor.c
```diff
@@ -1518,65 +1529,60 @@ printCurrentState(void *ctx, PGresult *result)
...
-		char *groupId = PQgetvalue(result, currentTupleIndex, 2);
-		char *nodeId = PQgetvalue(result, currentTupleIndex, 3);
-		char *currentState = PQgetvalue(result, currentTupleIndex, 4);
-		char *goalState = PQgetvalue(result, currentTupleIndex, 5);
-		char *candidatePriority = PQgetvalue(result, currentTupleIndex, 6);
-		char *replicationQuorum = PQgetvalue(result, currentTupleIndex, 7);
+		char *name = PQgetvalue(result, currentTupleIndex, 0);
+
+		/* char *hostname = PQgetvalue(result, currentTupleIndex, 1); */
+		/* char *nodeport = PQgetvalue(result, currentTupleIndex, 2); */
```
JelteF, line 1566: I think it's still very useful to print these in the state table. It's quite common to want to know where the name points to. Especially in cases where the default names are used, since they are not very useful.

src/bin/pg_autoctl/cli_create_drop_node.c
```diff
@@ -242,7 +244,7 @@ cli_create_postgres_getopts(int argc, char **argv)
 
 	int optind =
 		cli_create_node_getopts(argc, argv, long_options,
-								"C:D:H:p:l:U:A:Sd:n:f:m:MRVvqhP:r:xsN",
+								"C:D:H:p:l:U:A:Sd:a:n:f:m:MRVvqhP:r:xsN",
```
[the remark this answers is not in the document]
DimCitus: My understanding is that without this, `getopt_long` would fail to parse the long option too. If that's not the case then we could consider using an empty string here all the time...

src/monitor/pgautofailover.sql
```diff
@@ -113,6 +114,9 @@ CREATE TABLE pgautofailover.node
     candidatepriority	 int not null default 100,
     replicationquorum	 bool not null default true,
 
+    -- node names must be unique in a given formation
+    UNIQUE (formationid, nodename),
```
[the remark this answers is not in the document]
DimCitus: You could have a single monitor that manages your production and staging environment, two formation that are the same...

src/bin/pg_autoctl/monitor.c
DimCitus: We do a `bzero` before the loop now.

src/bin/pg_autoctl/monitor.c
DimCitus: I'm trying to stay within the 80 columns width if that's possible. The JSON output has all the information. Do you think it's important enough to have host:port on the `pg_autoctl show state` output to have very long lines?

src/bin/pg_autoctl/cli_get_set_properties.c
```diff
@@ -106,19 +105,22 @@ static CommandLine set_node_candidate_priority_command =
...
+				 "  --pgdata      path to data directory\n"
+				 "  --name        pg_auto_failover node name\n"
+				 "  --hostname    hostname used to connect from other nodes\n"
+				 "  --pgport      PostgreSQL's port number\n",
+				 cli_node_metadata_getopts,
+				 cli_set_node_metadata);
 
 
 static CommandLine *set_node_subcommands[] = {
+	&set_node_metadata_command,
```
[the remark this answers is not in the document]
DimCitus: I think we make all the get/set commands available in the multiple standby PR anyway, which is planed on the next release. That said, I don't see how things change in this PR in terms of what's exposed, which is decided in `cli_root.c` with the following array members for `root_subcommands_with_debug` and `root_subcommands`.

``` c
	&get_commands,
	&set_commands,
```

src/monitor/node_active_protocol.c
DimCitus: Good point, re-adding it now!

src/monitor/pgautofailover.sql
```diff
@@ -158,51 +163,46 @@ CREATE TABLE pgautofailover.event
...
+   OUT node_id          bigint,
+   OUT node_name        text,
+   OUT node_host        text,
+   OUT node_port        int
  )
 RETURNS record LANGUAGE SQL STRICT SECURITY DEFINER
 AS $$
       update pgautofailover.node
          set sysidentifier = node_sysidentifier
-       where nodeid = node_id
```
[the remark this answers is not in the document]
DimCitus: One is the column name, the other the argument name. Postgres allows to use `functionName.paramName` to disambiguate and I think that in this very case it's a clear win to use that feature. I don't suppose there was a bug before as the names are different.

src/monitor/pgautofailover.sql
```diff
@@ -219,17 +219,15 @@ RETURNS record LANGUAGE C STRICT SECURITY DEFINER
...
+      pgautofailover.register_node(text,text,int,name,text,bigint,int,pgautofailover.replication_state,text, int, bool)
    to autoctl_node;
 
 
 CREATE FUNCTION pgautofailover.node_active
  (
     IN formation_id           		text,
-    IN node_name              		text,
-    IN node_port              		int,
-    IN current_node_id        		int default -1,
```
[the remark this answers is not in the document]
DimCitus: This comes from the first early prototype 2 years ago where there was no separate registration API, so you got your nodeid on the first call to node_active. It's been one of the first things I changed.

src/monitor/node_metadata.c
```diff
@@ -693,11 +752,25 @@ AddAutoFailoverNode(char *formationId,
...
+	 * expression in the INSERT query.
+	 */
 	const char *insertQuery =
+		"WITH seq(nodeid) AS "
+		"(SELECT nextval('pgautofailover.node_nodeid_seq'::regclass)) "
 		"INSERT INTO " AUTO_FAILOVER_NODE_TABLE
-		" (formationid, groupid, nodehost, nodeport, sysidentifier, "
-		"  goalstate, reportedstate, candidatepriority, replicationquorum)"
-		" VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9) "
+		" (formationid, nodeid, groupid, nodename, nodehost, nodeport, "
```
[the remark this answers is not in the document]
DimCitus: Well a node exists only within a formation, that was the idea. I'm not sure doing one or the other bears any weight here really...

src/monitor/node_active_protocol.c
DimCitus: Done.

src/monitor/pgautofailover.sql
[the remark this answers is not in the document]
JelteF: I guess I missed the underscore because it was EOD on friday

src/bin/pg_autoctl/monitor.c
JelteF: The hostname for sure. I think it's very common to want to connect to a node that is in the wrong state. The port I can see not being super useful for most production users, but I don't think that matters much in linesize.

src/bin/pg_autoctl/cli_create_drop_node.c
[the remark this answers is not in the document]
JelteF: I believe it's not necessary, but I think doing this change everywhere fits better in another PR. A few we would probably want to keep though.

src/bin/pg_autoctl/cli_get_set_properties.c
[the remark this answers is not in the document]
JelteF: So in the current state you can only change the name when using `PG_AUTOCTL_DEBUG=true`. If that's changed by the multiple standby PR I think that's fine.

src/monitor/node_metadata.c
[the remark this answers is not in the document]
JelteF: I just checked the table definition, the `formationid` is also first there. So this is good to keep as as, let's keep it consistent with that.

JelteF:
Before merging I think it would be good to have few more things:
1. If you rerun `pg_autoctl create` with a different `--name`/`--hostname`/`--port` it should change them. Otherwise the `.cfg` file and the monitor get out of sync.
2. It needs a test to see that the `name` on the monitor is as expected, and that it can be changed using `pg_autoctl set node ...` (and also using `pg_autoctl create` like described above).
3. The hostname in the default `show state` output (like discussed in the other comment)

DimCitus:
> Before merging I think it would be good to have few more things:

Indeed, that PR was shy of a brick's load. I think we're good now.

> 1. If you rerun `pg_autoctl create` with a different `--name`/`--hostname`/`--port` it should change them. Otherwise the `.cfg` file and the monitor get out of sync.

That was done in the previous version of the PR and somehow I lost it in this one. Added again now.

> 2. It needs a test to see that the `name` on the monitor is as expected, and that it can be changed using `pg_autoctl set node ...` (and also using `pg_autoctl create` like described above).

Done.

> 3. The hostname in the default `show state` output (like discussed in the other comment)

Here's what we have now, which I think is too long, but well, there's a balance to keep and we certainly need to include the minimum necessary information.

```
$ pg_autoctl show state
 Name |      Host:Port |   Node |     Current State |    Assigned State | Priority | Quorum
------+----------------+--------+-------------------+-------------------+----------+-------
    a | localhost:5001 |    0/1 |           primary |           primary |      100 |      t
    b | 127.0.0.1:5002 |   0/52 |         secondary |         secondary |      100 |      t
```

tests/test_basic_operation.py
```diff
@@ -42,7 +45,10 @@ def test_004_init_secondary():
     global node2
     node2 = cluster.create_datanode("/tmp/basic/node2")
     node2.create()
```
JelteF, line 47: Something like this would be good to test too, to see that the `--name` argument of create works:

```suggestion
    node2.create()
    node2.create(name="create_b")
    eq_(node2.get_nodename(), "create_b")
```

src/bin/pg_autoctl/monitor.c
```diff
@@ -1510,79 +1514,108 @@ printCurrentState(void *ctx, PGresult *result)
...
+		sformat(composedId, sizeof(hostport), "%s/%s", groupId, nodeId);
+
 		if (pg_autoctl_debug)
 		{
-			fformat(stdout, "%*s | %6s | %5s | %17s | %17s | %8s | %6s\n",
-					maxNameSize, name, groupId, nodeId,
+			fformat(stdout, "%*s | %*s | %6s | %17s | %17s | %8s | %6s\n",
+					maxNameSize, name,
+					maxHostSize, hostport,
+					composedId,
```
JelteF, line 1610: I believe `composedId` should also have a `maxComposedIdSize`.

Right now the new state table is not lined up correctly in all cases.

With a small nodeid:
```
  Name |  Host:Port |  Node |     Current State |    Assigned State
-------+------------+--------+-------------------+------------------
node_1 |     l:5501 |    0/1 |            single |            single
```
And with a big nodeid:
```
       Name |          Host:Port |  Node |     Current State |    Assigned State
------------+--------------------+--------+-------------------+------------------
     node_1 |             l:5501 |    0/1 |           primary |           primary
node_100001 | 172.18.199.43:5502 | 0/100001 |         secondary |         secondary

``

tests/test_basic_operation.py
DimCitus: Right, added now.

src/bin/pg_autoctl/monitor.c
DimCitus: Fair enough, even though I think it's pushing it to edge cases, I don't see a good reason why we wouldn't expand our efforts in that direction too. Done now.

PR #345, opened 2020-07-30
Add a regression test for multiple standbys

onderkalaci:
Which ensures that pg_auto_failover can stay operational even if the primary
and the mostAdvancedLSN secondary becomes inaccessible

I opened a PR for a single test for now (a) for ease of review (b) get some feedback before adding other tests

### DIFF
--- a/tests/test_multi_standbys.py
+++ b/tests/test_multi_standbys.py
@@ -296,3 +296,71 @@ def test_020_start_node3_again():
     assert node1.has_needed_replication_slots()
     assert node2.has_needed_replication_slots()
     assert node3.has_needed_replication_slots()
+
+# test_021_XXX, test_022_XXX, test_023_XXX, test_024_XXX and test_025_XXX
+# are meant to test the scenario when the most advanced secondary
+# becomes inaccessible at the same time when the primary is inaccessible
+def test_021_prepare_candidate_priorities():
+    # we are aiming to promote node2
+    assert node2.set_candidate_priority(100)  # the next primary
+
+    # other nodes are already candidates for primary, but with less
+    # priority
+    assert node1.get_candidate_priority() == 90
+    assert node3.get_candidate_priority() == 90
+
+def test_022_prepare_replication_quorums():
+    # for the purpose of this test, we need one node
+    # async, to allow that we should decrement the sync stanbys
+    node1.set_number_sync_standbys(0)
+
+    # to emulate one node is behind, it is easier to make it async
+    # we want node2 to be behind others
+    assert node2.set_replication_quorum("false")
+
+    # others should be sync
+    assert node1.get_replication_quorum()
+    assert node3.get_replication_quorum()
+
+def test_023_secondary_gets_behind_primary():
+
+    # make sure that node2 gets behind of the primary
+    node2.ifdown()
+
+    # primary ingests some data
+    node1.run_sql_query("INSERT INTO t1 VALUES (5), (6)")
+    node1.run_sql_query("CHECKPOINT")
+
+    # ensure that the healthy secondary gets the change
+    results = node3.run_sql_query("SELECT count(*) FROM t1")
+    assert results == [(100006,)]
+
+def test_024_secondary_reports_lsn():
+    # make the primary and mostAdvanced secondary inaccessible
+    # and the candidate for failover as accessible
+    # which means that node2 will not be able to fetch wal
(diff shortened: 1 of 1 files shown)


tests/test_multi_standbys.py
```diff
@@ -296,3 +296,67 @@ def test_020_start_node3_again():
     assert node1.has_needed_replication_slots()
     assert node2.has_needed_replication_slots()
     assert node3.has_needed_replication_slots()
+
+def test_021_most_advanced_standby_inaccessible():
+
+    global node1
+    global node2
+    global node3
```
DimCitus, line 304: Not needed here, you don't assign anything to those variables. Python is weird sometimes.

tests/test_multi_standbys.py
```diff
@@ -296,3 +296,67 @@ def test_020_start_node3_again():
...
+
+    node1.set_candidate_priority(50)   # the current primary
+    node2.set_candidate_priority(100)  # the next primary
+    node3.set_candidate_priority(0)    # never candidate for a failover
```
DimCitus, line 308: Consider using `assert` or `eq_` for the priorities that are already set the way you need them to be, so that we don't have to wait for a full round-trip with the monitor and the primary 3 times here.

tests/test_multi_standbys.py
```diff
@@ -296,3 +296,67 @@ def test_020_start_node3_again():
...
+
+    # to emulate one node is behind, it is easier to make it async
+    # we want node2 to be behind others
+    node1.set_number_sync_standbys(0)
```
DimCitus, line 312: The comment does not apply to this line. Please comment this line specifically, and move the comment to the next block of three lines.

tests/test_multi_standbys.py
DimCitus: Also please set the candidate priorities for nodes in the expected order (node1, node2, node3), making it easier to follow along.

onderkalaci: > Also please set the candidate priorities for nodes in the expected order (node1, node2, node3), making it easier to follow along.

Aren't they  already in that order? Or do you mean the highest should be the first line?

onderkalaci:
> Also note that this test_multi_standbys is getting pretty long. We could (should?) split it in different test suites, I think, but at the moment travis seems to have a problem with having too many tests to run so I'm not sure how to make Travis happy again.

I have no experience with Travis, so not sure what we can do about this. For now, I just added the functions to the same file. 

The problem(?) is that I have plans to add  5-6 more complex tests like this and it is likely to deepen this problem if  we  don't do anything about it.

In citus, we have many parallel jobs in CircleCI, which helped a lot for decreasing the overall times. I guess we could consider following similar approach, though I have no clue on how to move things to CircleCI and have parallel jobs :)

Maybe we should ask Utku/Metin to assign someone  who has experience with this.

tests/test_multi_standbys.py
```diff
@@ -296,3 +296,71 @@ def test_020_start_node3_again():
...
+    assert results == [(100006,)]
+
+def test_024_secondary_reports_lsn():
+    # make the primary and mostAdvanced secondary inaccessible
+    # and the candidate for failover as accessible
+    # which means that node2 will not be able to fetch wal
+    # and blocked until the other secondary is up
+    node1.ifdown()
+    node3.ifdown()
+    node2.ifup()
```
DimCitus, line 345: Please add same line comments with the expected current role of the nodes:

``` Python
    node1.ifdown()   # primary
    node3.ifdown()   # most advanced standby
    node2.ifup()     # failover candidate
```

tests/test_multi_standbys.py
```diff
@@ -296,3 +296,71 @@ def test_020_start_node3_again():
...
+    monitor.failover()
+
+    # node2 reports its LSN while others are inaccessible
+    assert node2.wait_until_state(target_state="report_lsn")
+
+def test_025_finalize_failover_after_most_advanced_secondary_gets_back():
+    # when they are accessible again, both should become
+    # secondaries
+    node3.ifup()
+    node1.ifup()
```
DimCitus, line 358: Same as before, would be easier to follow with same-line comments with the current role of those nodes.

onderkalaci:
Would merging break your regular work flow? You need to fetch this commit from remote after merging the PR, so @DimCitus please merge this whenever works best for you.

Issue #353, opened 2020-08-07
pg_autoctl does not know how to reach state "report_lsn" from "primary" after node removal


marcocitus:
I tried setting up a primary with two standby's, then took down the primary, and after the failover completed I removed the old primary.

The node that was promoted to primary then transitioned to report_lsn, which failed:
```
00:03:12 15309 INFO  Monitor assigned new state "primary"
00:03:12 15309 INFO  FSM transition from "wait_primary" to "primary": A healthy secondary appeared
00:03:12 15309 INFO  Setting synchronous_standby_names to 'ANY 1 (pgautofailover_standby_4, pgautofailover_standby_2)'
00:03:12 15309 INFO  Waiting until standby node has caught-up to LSN 0/5000368
00:03:12 15309 INFO  Standby reached LSN 0/5000410, thus advanced past LSN 0/5000368
00:03:12 15309 INFO  Transition complete: current state is now "primary"
00:03:34 15309 ERROR Failed to fetch current replication properties from standby node: no standby connected in pg_stat_replication.
00:03:34 15309 WARN  HINT: check pg_autoctl and Postgres logs on standby nodes
00:03:34 15309 WARN  Failed to update the keeper's state from the local PostgreSQL instance.
00:03:34 15309 INFO  Fetched current list of 1 other nodes from the monitor to update HBA rules, including 1 changes.
00:03:34 15309 INFO  Ensuring HBA rules for node 4 (172.28.217.37:6003)
00:03:34 15309 INFO  Monitor assigned new state "report_lsn"
00:03:34 15309 INFO  Dropping replication slot "pgautofailover_standby_2"
00:03:34 15309 FATAL pg_autoctl does not know how to reach state "report_lsn" from "primary"
00:03:34 15309 ERROR Failed to transition to state "report_lsn", retrying...
00:03:35 15309 ERROR Failed to fetch current replication properties from standby node: no standby connected in pg_stat_replication.
00:03:35 15309 WARN  HINT: check pg_autoctl and Postgres logs on standby nodes
00:03:35 15309 WARN  Failed to update the keeper's state from the local PostgreSQL instance.
00:03:35 15309 INFO  Monitor assigned new state "report_lsn"
00:03:35 15309 FATAL pg_autoctl does not know how to reach state "report_lsn" from "primary"
00:03:35 15309 ERROR Failed to transition to state "report_lsn", retrying...
[...]

PR #357, opened 2020-08-10
In some cases, we still need to call pg_controldata.

DimCitus:
As we renew the keeper->postgres.pgSetup cache on each round, we need to
manually make sure to re-install the current values we have in cache when
Postgres is not running. That said, we might also reach that code path at
pg_autoctl start-up when Postgres is not yet running.

### DIFF
--- a/src/bin/pg_autoctl/keeper.c
+++ b/src/bin/pg_autoctl/keeper.c
@@ -475,6 +475,7 @@ keeper_update_pg_state(Keeper *keeper)
 			return false;
 		}
 
+		/* update the state from the metadata we just obtained */
 		keeperState->pg_control_version = pgSetup->control.pg_control_version;
 		keeperState->catalog_version_no = pgSetup->control.catalog_version_no;
 		keeperState->system_identifier = pgSetup->control.system_identifier;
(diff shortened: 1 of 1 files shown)


JelteF:
Can you add a test for this and/or explain how to reproduce the problem locally?

DimCitus:
> Can you add a test for this and/or explain how to reproduce the problem locally?

I got this error while reproducing https://github.com/citusdata/pg_auto_failover/issues/353 and have now added test coverage for the situation in https://github.com/citusdata/pg_auto_failover/pull/358 ; so I think we should be good .

src/bin/pg_autoctl/keeper.c
```diff
@@ -483,6 +484,25 @@ keeper_update_pg_state(Keeper *keeper)
 	{
 		/* Postgres is not running. */
 		postgres->pgIsRunning = false;
+
+		/* keep the current values we have for the Postgres characteristics */
+		if (pgSetup->control.pg_control_version != 0)
```
JelteF, line 489: Shouldn't this be `keeperState->pg_control_version`?

DimCitus: I believe you're right. Cache invalidation still is in the top-3 of difficult things, right?

PR #358, opened 2020-08-10
Fix dropping the old-primary node.

DimCitus:
When we remove the old primary, its reported state is still that of a
primary. We might have moved on to a new primary already though, and in that
case we want to process the old-primary the same way we would a secondary.

To fix, consider the goalState rather than the reportedState, and filter out
the DRAINING, DEMOTE, and DEMOTE_TIMEOUT as meaning that we have a primary,
by testing with CanTakeWritesInState() rather than StateBelongsToPrimary().

Fixes https://github.com/citusdata/pg_auto_failover/issues/353.

### DIFF
--- a/src/monitor/node_active_protocol.c
+++ b/src/monitor/node_active_protocol.c
@@ -1030,7 +1030,7 @@ RemoveNode(AutoFailoverNode *currentNode)
 	LockFormation(currentNode->formationId, ExclusiveLock);
 
 	/* when removing the primary, initiate a failover */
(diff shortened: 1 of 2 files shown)


JelteF:
Can you add a test for this and/or explain how to reproduce the problem locally?

DimCitus:
> Can you add a test for this and/or explain how to reproduce the problem locally?

Added a test, and to reproduce we can follow the bug report at https://github.com/citusdata/pg_auto_failover/issues/353 ; that I forgot to link. Done now.

PR #359, opened 2020-08-10
Implement a retry policy for deadlocks and other transient failures.

DimCitus:
After seeing the following happen in Travis, we could process some error
conditions as transient and retry. It would be best for test stability, and
also for users who script around some pg_autoctl commands.

    ERROR Monitor ERROR:  deadlock detected
    ERROR Monitor DETAIL:  Process 21381 waits for ShareLock on transaction 532; blocked by process 21382.
    ERROR Monitor Process 21382 waits for ExclusiveLock on advisory lock [16385,822708183,0,11]; blocked by process 21381.
    ERROR Monitor HINT:  See server log for query details.
    ERROR Monitor CONTEXT:  while updating tuple (0,11) in relation "node"
    ERROR Monitor SQL statement "UPDATE pgautofailover.node SET goalstate = $1, statechangetime = now() WHERE nodeid = $2"
    ERROR SQL query: SELECT pgautofailover.stop_maintenance($1)
    ERROR SQL params: '2'
    ERROR Failed to stop_maintenance of node 2 from the monitor
    FATAL Failed to stop maintenance from the monitor, see above for details

### DIFF
--- a/src/bin/pg_autoctl/cli_enable_disable.c
+++ b/src/bin/pg_autoctl/cli_enable_disable.c
@@ -437,6 +437,13 @@ cli_enable_maintenance(int argc, char **argv)
 
 	char *channels[] = { "state", NULL };
 
+	ConnectionRetryPolicy retryPolicy = {
+		POSTGRES_PING_RETRY_TIMEOUT,
+		-1, /* unbounded number of attempts */
+		5 * 1000,               /* sleep up to 5s between attempts */
+		1 * 1000                /* first retry happens after 1 second */
+	};
+
 	keeper.config = keeperOptions;
 
 	(void) exit_unless_role_is_keeper(&(keeper.config));
@@ -483,8 +490,14 @@ cli_enable_maintenance(int argc, char **argv)
 		}
 	}
 
+	/*
+	 * Set a retry policy for cases when we have a transient error on the
+	 * monitor.
+	 */
+
 	if (!monitor_start_maintenance(&(keeper.monitor),
-								   keeper.state.current_node_id))
+								   keeper.state.current_node_id,
+								   &retryPolicy))
 	{
 		log_fatal("Failed to start maintenance from the monitor, "
 				  "see above for details");
@@ -525,6 +538,13 @@ cli_disable_maintenance(int argc, char **argv)
 
 	char *channels[] = { "state", NULL };
 
+	ConnectionRetryPolicy retryPolicy = {
+		POSTGRES_PING_RETRY_TIMEOUT,
+		-1, /* unbounded number of attempts */
+		5 * 1000,               /* sleep up to 5s between attempts */
+		1 * 1000                /* first retry happens after 1 second */
+	};
+
 	keeper.config = keeperOptions;
 
 	(void) exit_unless_role_is_keeper(&(keeper.config));
@@ -558,7 +578,8 @@ cli_disable_maintenance(int argc, char **argv)
 	}
 
 	if (!monitor_stop_maintenance(&(keeper.monitor),
-								  keeper.state.current_node_id))
+								  keeper.state.current_node_id,
+								  &retryPolicy))
 	{
 		log_fatal("Failed to stop maintenance from the monitor, "
 				  "see above for details");
--- a/src/bin/pg_autoctl/cli_enable_disable.c
+++ b/src/bin/pg_autoctl/cli_enable_disable.c
@@ -437,14 +437,11 @@ cli_enable_maintenance(int argc, char **argv)
 
 	char *channels[] = { "state", NULL };
 
-	ConnectionRetryPolicy retryPolicy = {
-		POSTGRES_PING_RETRY_TIMEOUT,
-		-1, /* unbounded number of attempts */
-		5 * 1000,               /* sleep up to 5s between attempts */
-		1 * 1000                /* first retry happens after 1 second */
-	};
+	ConnectionRetryPolicy retryPolicy = { 0 };
 
 	keeper.config = keeperOptions;
+
+	(void) pgsql_set_monitor_interactive_retry_policy(&retryPolicy);
 
 	(void) exit_unless_role_is_keeper(&(keeper.config));
 
@@ -538,14 +535,11 @@ cli_disable_maintenance(int argc, char **argv)
 
 	char *channels[] = { "state", NULL };
 
-	ConnectionRetryPolicy retryPolicy = {
-		POSTGRES_PING_RETRY_TIMEOUT,
-		-1, /* unbounded number of attempts */
-		5 * 1000,               /* sleep up to 5s between attempts */
-		1 * 1000                /* first retry happens after 1 second */
-	};
+	ConnectionRetryPolicy retryPolicy = { 0 };
 
 	keeper.config = keeperOptions;
+
+	(void) pgsql_set_monitor_interactive_retry_policy(&retryPolicy);
 
 	(void) exit_unless_role_is_keeper(&(keeper.config));
 
--- a/src/bin/pg_autoctl/cli_enable_disable.c
+++ b/src/bin/pg_autoctl/cli_enable_disable.c
@@ -441,8 +441,6 @@ cli_enable_maintenance(int argc, char **argv)
 
 	keeper.config = keeperOptions;
 
-	(void) pgsql_set_monitor_interactive_retry_policy(&retryPolicy);
-
 	(void) exit_unless_role_is_keeper(&(keeper.config));
 
 	if (!keeper_config_read_file(&(keeper.config),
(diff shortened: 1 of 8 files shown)


src/bin/pg_autoctl/monitor.c
```diff
@@ -2554,8 +2600,19 @@ monitor_start_maintenance(Monitor *monitor, int nodeId)
 								   paramCount, paramTypes, paramValues,
 								   &context, &parseSingleValueResult))
 	{
+		if (monitor_retryable_error(context.sqlstate))
+		{
+			log_warn("Failed to start_maintenance of node %d on the monitor, "
+					 "retrying in %ds.",
+					 nodeId, PG_AUTOCTL_KEEPER_SLEEP_TIME);
+
+			sleep(PG_AUTOCTL_KEEPER_SLEEP_TIME);
```
JelteF, line 2609: Can we use our exponential random backoff mechanism to implement this sleep?

DimCitus: Not easily. Our dynamic retry policy depends on a state that is attached to the PGSQL object, and used privately in the context of re-connecting. The policy we set on a monitor PGSQL object is meant to defer retries to another part of the code, and so we enforce `pgsql_set_main_loop_retry_policy` on the monitor's connection, which means no retry by default.

We could change the level of abstraction where our function `pgsql_compute_connection_retry_sleep_time` works and then handle that context in the monitor's function where we implement some retry, or maybe on their callers. I can have a try at that.

DimCitus: Done now, including with the registration monitor API where we already had a crude retry mechanism.

src/bin/pg_autoctl/cli_enable_disable.c
```diff
@@ -437,6 +437,13 @@ cli_enable_maintenance(int argc, char **argv)
 
 	char *channels[] = { "state", NULL };
 
+	ConnectionRetryPolicy retryPolicy = {
+		POSTGRES_PING_RETRY_TIMEOUT,
+		-1, /* unbounded number of attempts */
+		5 * 1000,               /* sleep up to 5s between attempts */
+		1 * 1000                /* first retry happens after 1 second */
+	};
```
JelteF, line 445: We use this same retry policy in 3 places. Can we give it a name that makes sense and make it a constant/function?

src/bin/pg_autoctl/pgsql.c
```diff
@@ -470,28 +504,11 @@ pgsql_retry_open_connection(PGSQL *pgsql, uint64_t startTime)
...
 	{
 		int sleep = 0;
-		uint64_t now = time(NULL);
 
-		/* Any signal is reason enough to break out from this retry loop. */
-		if (asked_to_quit ||
-			asked_to_stop ||
-			asked_to_stop_fast ||
-			asked_to_reload)
+		if (pgsql_retry_policy_expired(&(pgsql->retryPolicy)))
```
JelteF, line 508: Afaict this call does something completely different than the code that was here before.
1. It doesn't break when it reaches a signal, instead it enters this if block.
2. It does the opposite of what it previously did whit `true`/`false` for the `maxR`+`maxT`.

src/bin/pg_autoctl/monitor.c
```diff
@@ -2554,8 +2609,24 @@ monitor_start_maintenance(Monitor *monitor, int nodeId)
 								   paramCount, paramTypes, paramValues,
 								   &context, &parseSingleValueResult))
 	{
+		if (monitor_retryable_error(context.sqlstate) &&
+			!pgsql_retry_policy_expired(retryPolicy))
+		{
+			int sleep = pgsql_compute_connection_retry_sleep_time(retryPolicy);
```
JelteF, line 2615: Let's do this everywhere. People expect `sleep` to be a function, even github syntax highlighting does that it seems.

```suggestion
			int sleepTime = pgsql_compute_connection_retry_sleep_time(retryPolicy);
```

src/bin/pg_autoctl/monitor.c
```diff
@@ -2591,6 +2663,21 @@ monitor_stop_maintenance(Monitor *monitor, int nodeId)
...
+
+			log_warn("Failed to start_maintenance of node %d on the monitor, "
```
JelteF, line 2671:
```suggestion
			log_warn("Failed to stop_maintenance of node %d on the monitor, "
```

src/bin/pg_autoctl/monitor.c
```diff
@@ -2591,6 +2663,21 @@ monitor_stop_maintenance(Monitor *monitor, int nodeId)
...
+					 "retrying in %d ms.",
+					 nodeId, sleep);
+
+			/* we have milliseconds, pg_usleep() wants microseconds */
+			(void) pg_usleep(sleep * 1000);
+
+			return monitor_start_maintenance(monitor, nodeId, retryPolicy);
```
JelteF, line 2678:
```suggestion
			return monitor_stop_maintenance(monitor, nodeId, retryPolicy);
```

src/bin/pg_autoctl/pgsql.c
```diff
@@ -265,16 +264,49 @@ pgsql_compute_connection_retry_sleep_time(PGSQL *pgsql)
...
+	 * of our attempts count budget, whichever comes first.
+	 *
+	 * maxR = 0 (zero) means no retry at all, checked before the loop
+	 * maxR < 0 (zero) means unlimited number of retries
+	 */
+	if ((now - retryPolicy->startTime) >= retryPolicy->maxT ||
+		(retryPolicy->maxR > 0 &&
+		 retryPolicy->attempts >= retryPolicy->maxR))
+	{
+		return false;
```
JelteF, line 306: As also described in my other comment. I believe this should be `return true` and below `return false`.

JelteF:
There's quite a lot of serious errors in this last changeset. Can you try to test it a bit better after fixing these? It's definitely possible that I missed some edgecases.

src/bin/pg_autoctl/cli_enable_disable.c
DimCitus: Good point. The abstraction to set retry policies was a level too high to easily do that, so I changed them to take a ConnectionRetryPolicy instead of a PGSQL argument, and added a new one.

src/bin/pg_autoctl/pgsql.c
DimCitus: I think your later comment is right, the return values have been inverted in that function.

src/bin/pg_autoctl/monitor.c
DimCitus: Done now.

src/bin/pg_autoctl/monitor.c
DimCitus: Oops. Good catch. Fixed now.

src/bin/pg_autoctl/monitor.c
DimCitus: Another good catch, thanks. Copy/paste and all.

src/bin/pg_autoctl/pgsql.c
DimCitus: Done that way, you're right.

DimCitus:
> There's quite a lot of serious errors in this last changeset. Can you try to test it a bit better after fixing these? It's definitely possible that I missed some edgecases.

I just did a first pass over what you reported, will schedule more review time with QA and testing, and maybe a better approach to make sure this does the right thing. Turning to draft meanwhile.

DimCitus:
See also #392 and #395, where we can't `C-c` out of the `pg_autoctl create` command when there's a loop over `monitor_register_node`. This PR should fix it with a proper retry mechanism that allows for signal handling.

DimCitus:
I have now reviewed and finished this PR, it's again ready for review. The main thing is that we changed the registration process with proper transaction handling and this broke our “poor's man” retry implementation here. Good news, the new implementation is easier to review, makes more sense, and should be more robust in the wild.

DimCitus:
Extra note: sometimes the Travis nightly build for the master branch fails, and that's because the `pg_autoctl enable maintenance` or `pg_autoctl disable maintenance` functions are hitting a deadlock on the monitor. This PR with the retry policy for those interactive commands should fix the flakiness of the nightly builds.

Of course sometimes we get this problem in branch/PR builds too, and then I usually restart the individual unlucky build.

src/bin/pg_autoctl/pgsql.c
JelteF: The first item still stands. With the new code it will log the connection error on any signal, with the old code it wouldn't. I'm not sure if we care about that though.

src/bin/pg_autoctl/keeper.c
```diff
@@ -1239,49 +1243,95 @@ keeper_register_and_init(Keeper *keeper, NodeState initialState)
...
-	 *
-	 * We register to the monitor in a SQL transaction that we only COMMIT
-	 * after we have updated our local state file. If we fail to do so, we
-	 * ROLLBACK the transaction, and thus we are not registered to the monitor
-	 * and may try again. If we are disconnected halfway through the
-	 * registration (process killed, crash, etc), then the server issues a
-	 * ROLLBACK for us upon disconnection.
+	 * We implement a specific retry policy for cases where we have a transient
+	 * error on the monitor, such as OBJECT_IN_USE which indicates that another
+	 * standby is concurfrently being added to the same group.
```
JelteF, line 1250:
```suggestion
	 * standby is concurently being added to the same group.
```

src/bin/pg_autoctl/monitor.c
```diff
@@ -635,21 +681,11 @@ monitor_register_node(Monitor *monitor, char *formation,
 	{
 		if (strcmp(parseContext.sqlstate, STR_ERRCODE_OBJECT_IN_USE) == 0)
```
JelteF, line 682: Shouldn't this use `monitor_retryable_error` too?
```suggestion
		if (monitor_retryable_error(parseContext.sqlstate, STR_ERRCODE_OBJECT_IN_USE))
```

JelteF (changes requested):

JelteF:
I confirmed that this indeed solves the Ctrl+C issue that was reported in #392

src/bin/pg_autoctl/pgsql.c
DimCitus: Yeah I think it's better to log about “cancelling” the connection attempt, it might be what the user is trying to achieve with interrupting (C-c) the flow at this point.

src/bin/pg_autoctl/keeper.c
DimCitus: Fixed, thanks!

src/bin/pg_autoctl/monitor.c
DimCitus: Yeah good point, we can check also for the generic monitor_retryable_error, and add the specific error that we can also retry during registration.

JelteF (approved):
Blocked until this is merged https://github.com/citusdata/pg_auto_failover/pull/402, but otherwise good to merge.

PR #360, opened 2020-08-11
Fix pg_autoctl status when pg_autoctl is not running.

### LINKED ISSUE #310: Fix pg_autoctl status when it's not running
At the moment `pg_autoctl status` expects our services to be running, and the output is not useful when it's not the case:

```
$ pg_autoctl status
15:52:42 93791 INFO  pg_autoctl pid file "/tmp/pg_autoctl/tmp/plop/b/pg_autoctl.pid" does not exists
15:52:42 93791 INFO  pg_autoctl is running with pid 0
15:52:42 93791 ERROR Failed get postmaster pid, file "/private/tmp/plop/b/postmaster.pid" does not exists
15:52:42 93791 ERROR Failed to get Postgres pid, see above for details
```

DimCitus:
And when it's running and --json is used, match the same format as in the
command pg_autoctl show file --pid --contents --json, adding the status of
the pids (running, or stale).

Fixes #310

### DIFF
--- a/src/bin/pg_autoctl/cli_service.c
+++ b/src/bin/pg_autoctl/cli_service.c
@@ -426,18 +426,22 @@ cli_service_status(int argc, char **argv)
 	{
 		log_debug("pg_autoctl pid file \"%s\" does not exists", pathnames->pid);
 
+		/*
+		 * pg_autoctl should be the parent process of Postgres. That said, when
+		 * in maintenance, operators could stop pg_autoctl and then start/stop
+		 * Postgres to make some configuration changes, and then use pg_autoctl
+		 * again.
+		 *
+		 * So check if Postgres is running, and complain about it when it's the
+		 * case and pg_autoctl is not running, as it will get in the way when
+		 * starting pg_autoctl again.
+		 */
 		if (pg_setup_is_running(pgSetup))
 		{
 			log_fatal("Postgres is running at \"%s\" with pid %d",
 					  pgSetup->pgdata, pgSetup->pidFile.pid);
 		}
 
-		/*
-		 * That's not supposed to happen, because pg_autoctl should be the
-		 * parent process of Postgres. That said, when in maintenance,
-		 * operators could stop pg_autoctl and then start/stop Postgres to make
-		 * some configuration changes, and then use pg_autoctl again.
-		 */
 		log_info("pg_autoctl is not running at \"%s\"", pgSetup->pgdata);
 		exit(PG_CTL_STATUS_NOT_RUNNING);
 	}
--- a/src/bin/pg_autoctl/cli_service.c
+++ b/src/bin/pg_autoctl/cli_service.c
@@ -424,28 +424,33 @@ cli_service_status(int argc, char **argv)
 
 	if (!file_exists(pathnames->pid))
 	{
-		log_info("pg_autoctl pid file \"%s\" does not exists", pathnames->pid);
+		log_debug("pg_autoctl pid file \"%s\" does not exists", pathnames->pid);
 
 		if (pg_setup_is_running(pgSetup))
 		{
 			log_fatal("Postgres is running at \"%s\" with pid %d",
 					  pgSetup->pgdata, pgSetup->pidFile.pid);
-			exit(EXIT_CODE_INTERNAL_ERROR);
 		}
+
+		log_info("pg_autoctl is not running at \"%s\"", pgSetup->pgdata);
+		exit(PG_CTL_STATUS_NOT_RUNNING);
 	}
 
 	/* ok now we have a pidfile for pg_autoctl */
-	if (read_pidfile(pathnames->pid, &pid))
+	if (read_pidfile(pathnames->pid, &pid) && pid > 0)
 	{
 		if (kill(pid, 0) != 0)
 		{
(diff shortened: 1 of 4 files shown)


src/bin/pg_autoctl/cli_service.c
```diff
@@ -460,29 +465,8 @@ cli_service_status(int argc, char **argv)
...
-			/* can't happen */
-			exit(EXIT_CODE_INTERNAL_ERROR);
-		}
-
-		/* concatenate JSON objects into a container object */
-		json_object_set_value(root, "postgres", jsPostgres);
-		json_object_set_value(root, "pg_autoctl", jsPGAutoCtl);
-
-		(void) cli_pprint_json(js);
+		(void) fprint_pidfile_as_json(pathnames->pid, includeStatus);
```
JelteF, line 470: The new json format is missing all the information about the actual postgres process, all you see is info about the postgres service.
new:
```json
{
  "pid": 5599,
  "status": "running",
  "pgdata": "monitor",
  "version": "1.3",
  "semId": 655360,
  "services": [
    {
      "name": "postgres",
      "pid": 5603,
      "status": "running",
      "version": "1.3",
      "pgautofailover": "1.3"
    },
    {
      "name": "listener",
      "pid": 5604,
      "status": "running",
      "version": "1.3",
      "pgautofailover": "1.3"
    }
  ]
}
```

old
```json
{
  "postgres": {
    "pgdata": "/home/jelte/work/pg_auto_failover/monitor",
    "pg_ctl": "/home/jelte/.pgenv/pgsql/bin/pg_ctl",
    "version": "12.3",
    "host": "/tmp",
    "port": 5500,
    "proxyport": 0,
    "pid": 5608,
    "in_recovery": false,
    "control": {
      "version": 0,
      "catalog_version": 0,
      "system_identifier": "0"
    },
    "postmaster": {
      "status": "ready"
    }
  },
  "pg_autoctl": {
    "pid": 5599
  }
}
```

An easy way to fix this would be to make the JSON instead be like this:
```js
{
  "postgres": {
     // all postgres process values from the old one
  },
  "pg_autoctl": {
    // the pidfile data
  }
}
```

This would actually make the new JSON format compatible with the old format, since it only adds fields and doesn't remove them.

src/bin/pg_autoctl/cli_service.c
```diff
@@ -424,28 +424,33 @@ cli_service_status(int argc, char **argv)
...
+
+		log_info("pg_autoctl is not running at \"%s\"", pgSetup->pgdata);
+		exit(PG_CTL_STATUS_NOT_RUNNING);
 	}
 
 	/* ok now we have a pidfile for pg_autoctl */
-	if (read_pidfile(pathnames->pid, &pid))
+	if (read_pidfile(pathnames->pid, &pid) && pid > 0)
 	{
 		if (kill(pid, 0) != 0)
```
JelteF, line 442: This check is already done in `read_pidfile`, so it can be removed. When removing this I think none of the `pid > 0` added in this PR are needed anymore.

src/bin/pg_autoctl/cli_service.c
```diff
@@ -424,28 +424,33 @@ cli_service_status(int argc, char **argv)
 
 	if (!file_exists(pathnames->pid))
 	{
-		log_info("pg_autoctl pid file \"%s\" does not exists", pathnames->pid);
+		log_debug("pg_autoctl pid file \"%s\" does not exists", pathnames->pid);
 
 		if (pg_setup_is_running(pgSetup))
 		{
 			log_fatal("Postgres is running at \"%s\" with pid %d",
```
JelteF, line 431: I don't really understand this log message now. It seems like it would be quite confusing to get both this one and the one below.

src/bin/pg_autoctl/cli_service.c
DimCitus: Thanks, good thinking! Done now.

src/bin/pg_autoctl/cli_service.c
DimCitus: I cleaned the pidfile code a little, in the spirit of https://github.com/citusdata/pg_auto_failover/pull/364 (which we can cancel now), and cleaned the caller code too.

Fixes #364.

src/bin/pg_autoctl/cli_service.c
DimCitus: I have added comments about the situation. In maintenance mode, operators can do what they want. Including stop pg_autoctl, restart Postgres, etc.

src/bin/pg_autoctl/cli_service.c
```diff
@@ -432,25 +432,24 @@ cli_service_status(int argc, char **argv)
 					  pgSetup->pgdata, pgSetup->pidFile.pid);
 		}
 
+		/*
+		 * That's not supposed to happen, because pg_autoctl should be the
+		 * parent process of Postgres. That said, when in maintenance,
+		 * operators could stop pg_autoctl and then start/stop Postgres to make
+		 * some configuration changes, and then use pg_autoctl again.
```
JelteF, line 439: I guess this should be for the log above?

DimCitus: Ah yeah would be clearer that way, done now.

PR #361, opened 2020-08-11
Fix/fatal auth pgautofailover monitor

DimCitus:
Create the pgautofailover_monitor user on the Postgres nodes to avoid FATAL errors in the logs.

As we want to avoid using a password that we could leak when not using SSL connections, we hardcode the password (to the same string as the username). We don't actually authenticate at all, let alone use the password send, so we also hard code the authentication method for the health checks to trust.

Fixes https://github.com/citusdata/pg_auto_failover/issues/277

### DIFF
--- a/src/monitor/health_check_worker.c
+++ b/src/monitor/health_check_worker.c
@@ -47,7 +47,7 @@
 
 #define CONN_INFO_TEMPLATE \
 	"host=%s port=%u user=pgautofailover_monitor " \
-	"password=user=pgautofailover_monitor dbname=postgres " \
+	"password=pgautofailover_monitor dbname=postgres " \
 	"connect_timeout=%u"
 #define MAX_CONN_INFO_SIZE 1024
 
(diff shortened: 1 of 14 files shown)


src/monitor/health_check_worker.c
```diff
@@ -45,16 +45,9 @@
 #include "utils/memutils.h"
 #include "tcop/utility.h"
 
-/*
- * The healthcheck only checks if it gets a response back from postgres. So
- * both user and password are actually useless, because we do not need to
- * authenticate. They are provided though to override any settings set through
- * PGPASSWORD environment variable or .pgpass file. This way it does not matter
- * that TLS is not necessarily used, because no secret information is sent.
```
JelteF, line 53: Let's keep part of this comment

DimCitus: Oops, good point. I guess the git revert killed that comment. Re-installed now!

PR #371, opened 2020-08-17
Fix pg_autoctl get formation settings.

DimCitus:
To make it work when there's more than one group in a formation, implement
the feature server-side on the monitor, and have the client-sise command
line just call and format that properly to stdout.

WIP: still need to add the JSON formatted output.


DimCitus:
> The `--formation doesnotexist` argument together with the `--json` argument still outputs everything for the default formation. Without `--json` it correctly outputs nothing.

Oops. Was a typo where we would use `'default'` hard-coded in the SQL call rather than `$1`.

PR #372, opened 2020-08-17
Improve testing for pre-existing Postgres directory in monitor init.

DimCitus:
Fixes #323.


onderkalaci (approved):
This all makes sense, and verified that it works as expected.

Could we  add a test  where we add  one line to any of the existing `create_monitor`, something like below?
```
+ mkdir  /tmp/multi_async/monitor
 monitor = cluster.create_monitor("/tmp/multi_async/monitor")
```

DimCitus:
> Could we add a test where we add one line to any of the existing `create_monitor`, something like below?

Good point, I added a test!

PR #373, opened 2020-08-17
Reduce chatter when postmaster.pid is read empty.

DimCitus:
That's not a very interesting race condition, really.


onderkalaci (approved):
I can see how we might unnecessarily warn users, and after the fix we'll already be giving errors if  `get_pgpid ` fails due to actual problems

PR #374, opened 2020-08-18
Feature/default node names for citus formations

DimCitus:
In Citus formation, use "coordinator" or "worker" as a prefix for nodes registered without `--name` option.

In passing, fix several problems in the extension regression tests, and also fix the system identifier exclusion constraint to allow different groups with the same groupId in different formations to have different system identifiers.

### DIFF
--- a/src/monitor/Makefile
+++ b/src/monitor/Makefile
@@ -14,7 +14,7 @@ MODULE_big = $(EXTENSION)
 OBJS = $(patsubst ${SRC_DIR}%.c,%.o,$(wildcard ${SRC_DIR}*.c))
 PG_CPPFLAGS = -std=c99 -Wall -Werror -Wno-unused-parameter -Iinclude -I$(libpq_srcdir) -g
 SHLIB_LINK = $(libpq)
-REGRESS = create_extension monitor dummy_update drop_extension upgrade
+REGRESS = create_extension monitor workers dummy_update drop_extension upgrade
 
 PG_CONFIG ?= pg_config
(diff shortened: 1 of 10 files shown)


onderkalaci (commented):
The second and third commits are clear, but I cannot follow the first commit. Some parts of the commit is required as `register_node ` returns one more column, but the rest seems not  related?

It'd be good if you can explain the reason (and it'd be great if you can split the PR unless they are relevant to here?)

DimCitus:
> The second and third commits are clear, but I cannot follow the first commit. Some parts of the commit is required as `register_node ` returns one more column, but the rest seems not related?
> 
> It'd be good if you can explain the reason (and it'd be great if you can split the PR unless they are relevant to here?)

I guess you're mentioning the commit https://github.com/citusdata/pg_auto_failover/pull/374/commits/5f31a7b6c7756eb6b6ec80d14b7b484e6ec1162c (Fix regression tests for the monitor extension). It contains extra stuff because the whole test suite was found broken and I fixed it in this PR. In an ideal world we could separate much of the fixes in their own PR, but then we need to find a way to have Travis run `make installcheck` for us for that to make sense. And I didn't find a way to do that because `pg_regress` only works with unlimited core size settings in `ulimit` and we don't have that in our docker environment.

onderkalaci (approved):
> In an ideal world we could separate much of the fixes in their own PR, but then we need to find a way to have Travis run make installcheck for us for that to make sense

Ok, I was not aware of this

PR #376, opened 2020-08-20
Implement pg_autoctl perform promotion.

DimCitus:
This command allows to promote a given node (by name, or the current local
one) to being the primary on a group. When the node is already a primary, no
action is taken and the command returns true immediately.

### DIFF
--- a/src/bin/pg_autoctl/cli_perform.c
+++ b/src/bin/pg_autoctl/cli_perform.c
@@ -21,6 +21,7 @@
 
 static int cli_perform_failover_getopts(int argc, char **argv);
 static void cli_perform_failover(int argc, char **argv);
+static void cli_perform_promotion(int argc, char **argv);
 
 CommandLine perform_failover_command =
 	make_command("failover",
@@ -42,9 +43,20 @@ CommandLine perform_switchover_command =
 				 cli_perform_failover_getopts,
 				 cli_perform_failover);
 
+CommandLine perform_promotion_command =
+	make_command("promotion",
+				 "Perform a failover that promotes a target node",
+				 " [ --pgdata --formation --group ] ",
+				 "  --pgdata      path to data directory	 \n" \
+				 "  --formation   formation to target, defaults to 'default' \n" \
+				 "  --name        node name to target, defaults to current node \n",
+				 cli_get_name_getopts,
+				 cli_perform_promotion);
+
 CommandLine *perform_subcommands[] = {
 	&perform_failover_command,
 	&perform_switchover_command,
+	&perform_promotion_command,
 	NULL,
 };
 
@@ -302,5 +314,70 @@ cli_perform_failover(int argc, char **argv)
 	{
 		log_error("Failed to wait until a new primary has been notified");
 		exit(EXIT_CODE_INTERNAL_ERROR);
+	}
+}
+
+
+/*
+ * cli_perform_promotion calls the function pgautofailover.perform_promotion()
+ * on the monitor.
+ */
+static void
+cli_perform_promotion(int argc, char **argv)
+{
+	Keeper keeper = { 0 };
+	Monitor *monitor = &(keeper.monitor);
+	KeeperConfig *config = &(keeper.config);
+
+	int groupId = 0;
+
+	PgInstanceKind nodeKind = NODE_KIND_UNKNOWN;
+
+	char *channels[] = { "state", NULL };
+
+	keeper.config = keeperOptions;
+
+	if (!monitor_init_from_pgsetup(monitor, &keeper.config.pgSetup))
+	{
+		/* errors have already been logged */
+		exit(EXIT_CODE_BAD_ARGS);
+	}
+
+	/* grab --name from either the command options or the configuration file */
+	(void) cli_ensure_node_name(&keeper);
+
+	if (!monitor_get_groupId_from_name(monitor,
+									   config->formation,
+									   config->name,
+									   &groupId))
+	{
+		/* errors have already been logged */
+		exit(EXIT_CODE_BAD_ARGS);
+	}
+
+	/* start listening to the state changes before we call perform_promotion */
+	if (!pgsql_listen(&(monitor->pgsql), channels))
+	{
+		log_error("Failed to listen to state changes from the monitor");
+		exit(EXIT_CODE_MONITOR);
+	}
+
+	/*
+	 * pgautofailover.perform_promotion returns true when a promotion has been
+	 * triggered, and false when none was necessary. When an error occurs, it
+	 * reports an error condition, which is logged about already.
+	 */
+	if (monitor_perform_promotion(monitor, config->formation, config->name))
+	{
+		/* process state changes notification until we have a new primary */
+		if (!monitor_wait_until_some_node_reported_state(monitor,
+														 config->formation,
+														 groupId,
+														 nodeKind,
+														 PRIMARY_STATE))
+		{
+			log_error("Failed to wait until a new primary has been notified");
+			exit(EXIT_CODE_INTERNAL_ERROR);
+		}
 	}
 }
--- a/src/monitor/node_active_protocol.c
+++ b/src/monitor/node_active_protocol.c
@@ -1320,6 +1320,11 @@ perform_promotion(PG_FUNCTION_ARGS)
 		IsCurrentState(currentNode, REPLICATION_STATE_PRIMARY))
 	{
 		/* return false: no promotion is happening */
+		ereport(NOTICE,
+				(errcode(ERRCODE_OBJECT_NOT_IN_PREREQUISITE_STATE),
+				 errmsg("cannot perform promotion: node %s in formation %s "
+						"is already a primary.",
+						nodeName, formationId)));
 		PG_RETURN_BOOL(false);
 	}
 
@@ -1363,7 +1368,7 @@ perform_promotion(PG_FUNCTION_ARGS)
 							nodeName, formationId)));
 		}
 
-		currentNode->candidatePriority += 100;
+		currentNode->candidatePriority += MAX_USER_DEFINED_CANDIDATE_PRIORITY;
 
 		ReportAutoFailoverNodeReplicationSetting(
 			currentNode->nodeId,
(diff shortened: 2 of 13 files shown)


src/monitor/node_active_protocol.c
```diff
@@ -1259,6 +1260,120 @@ perform_failover(PG_FUNCTION_ARGS)
...
+	char *nodeName = text_to_cstring(nodeNameText);
+
+	List *groupNodesList = NULL;
+	int totalNodesCount = 0;
+
+	AutoFailoverNode *currentNode = NULL;
+
+	checkPgAutoFailoverVersion();
+
+	currentNode = GetAutoFailoverNodeByName(formationId, nodeName);
```
onderkalaci, line 1282: If the node is  in maintenance (or unhealthy?), it is not/may not  get promoted.

In MS case, we could promote another  node, in single standby we error out as there is no primary but might not  be very clear to the users `Monitor ERROR:  couldn't find the primary node in formation "default", group 0`


So, maybe we give a proper error here if the node unhealthy AND in maintenance mode?

src/monitor/node_active_protocol.c
```diff
@@ -1259,6 +1260,120 @@ perform_failover(PG_FUNCTION_ARGS)
...
+
+	LockFormation(formationId, ShareLock);
+	LockNodeGroup(formationId, currentNode->groupId, ExclusiveLock);
+
+	/*
+	 * If the current node is the primary, that's done.
+	 */
+	if (IsCurrentState(currentNode, REPLICATION_STATE_SINGLE) ||
+		IsCurrentState(currentNode, REPLICATION_STATE_PRIMARY))
+	{
```
onderkalaci, line 1299: minor: might be good to have  `ereport(NOTICE, .. node is already primary ..)` kind of information to the user

src/monitor/node_active_protocol.c
```diff
@@ -1259,6 +1260,120 @@ perform_failover(PG_FUNCTION_ARGS)
...
+		if (currentNode->candidatePriority == 0)
+		{
+			ereport(ERROR,
+					(errcode(ERRCODE_OBJECT_NOT_IN_PREREQUISITE_STATE),
+					 errmsg("cannot perform promotion: node %s in formation %s "
+							"has a candidate priority of 0 (zero).",
+							nodeName, formationId)));
+		}
+
+		currentNode->candidatePriority += 100;
```
onderkalaci, line 1344: minor: Maybe use a macro instead of `100` for _readability,_ such as `MAX_CANDIDATE_PRIORITY`. But, if you  decide to do that, we  should also go over the uses of `100`  in the code  and use  the macro for completeness.

src/monitor/node_active_protocol.c
```diff
@@ -1259,6 +1260,120 @@ perform_failover(PG_FUNCTION_ARGS)
...
+			"Updating candidate priority to %d for node %d \"%s\" (%s:%d)",
+			currentNode->candidatePriority,
+			currentNode->nodeId,
+			currentNode->nodeName,
+			currentNode->nodeHost,
+			currentNode->nodePort);
+
+		NotifyStateChange(currentNode, message);
+
+		DirectFunctionCall2(perform_failover,
```
onderkalaci, line 1364: I think it might be good to have a single line comment that if `perform_failover` fails,  we rollback the +100 on the metadata anyway. It might not be  very clear for everyone

src/bin/pg_autoctl/cli_perform.c
```diff
@@ -304,3 +316,68 @@ cli_perform_failover(int argc, char **argv)
...
+
+	/*
+	 * pgautofailover.perform_promotion returns true when a promotion has been
+	 * triggered, and false when none was necessary. When an error occurs, it
+	 * reports an error condition, which is logged about already.
+	 */
+	if (monitor_perform_promotion(monitor, config->formation, config->name))
+	{
+		/* process state changes notification until we have a new primary */
+		if (!monitor_wait_until_some_node_reported_state(monitor,
```
onderkalaci, line 373: As far as I can see, `monitor_wait_until_some_node_reported_state ` has  a 60 seconds timeout, after that we see 
```
12:43:31 65553 ERROR Failed to receive monitor's notifications
12:43:31 65553 ERROR Failed to wait until a new primary has been notified
```

We could likely to exceed the timeout on some real-world scenarios with async replicas. 

If my above observation is accurate, I can see two options: (a) wait indefinitely -- which sounds scary :)   (b) somehow check that LSN is increasing (such  as `standby_fetch_missing_wal` )and reset the timeout as long as the LSN increases

onderkalaci (commented):
With this approach, we are relaxing the invariant `100 >=  candidatePriority >= 0` and this seems to add an additional complexity while thinking about edge cases. I wonder if there is any opportunity to make this change more explicitly and not changing the invariant?

One option could be add something like `bool userInitiatedPerformPromotion` to  `AutoFailoverNode` struct/metadata, and on `SelectFailoverCandidateNode()` function have a short-circuit to return that node. I can see that my suggestion would end with a little more code/changes, but feels easier to maintain ( at least for me).

My above notes are optional, feel free to ignore if it doesn't make sense to you.

DimCitus:
> With this approach, we are relaxing the invariant `100 >= candidatePriority >= 0` and this seems to add an additional complexity while thinking about edge cases. I wonder if there is any opportunity to make this change more explicitly and not changing the invariant?

In fact I didn't think it could be possible to implement `perform promotion` within our current design so easily, until I though about organising a “rigged election”, and then we already have all the pieces we need.

> One option could be add something like `bool userInitiatedPerformPromotion` to `AutoFailoverNode` struct/metadata, and on `SelectFailoverCandidateNode()` function have a short-circuit to return that node. I can see that my suggestion would end with a little more code/changes, but feels easier to maintain ( at least for me).

The problem with that approach is that the selection of a candidate note to promote does not happen within the same transaction in the monitor. It doesn't even happen in the same connection, and will often happen in a connection from another node. There's no way to easily share information like what you're proposing in those different contexts, this would need either shared memory or table level storage. The problem with shared memory is that the monitor could well be restarted between the metadata changes and the actual candidate election.

The way the Group FSM works has managed to avoid extra information being needed, we rely mostly on the state themselves to decide what's next. So your proposal would mean introducing a whole parallel election mechanism with its own set of states and transitions so that we can carry the information forward during the failover orchestration.

I don't think this feature needs this much complexity, and I rather prefer the “rigged election” mechanism. Even when in cases, as you're mentioning, we might end up promoting another node if the manually selected one is not up to the task.

src/monitor/node_active_protocol.c
DimCitus: I think it's fine to re-use error message from the perform failover code path. The intention is to show in the docs that the `pg_autoctl perform promotion` is a special case of the `pg_autoctl perform switchover` command.

src/monitor/node_active_protocol.c
DimCitus: Added.

src/monitor/node_active_protocol.c
DimCitus: Good point, now added with a comment that also covers the user range being 0..100 while the internal range is different (0..200).

src/monitor/node_active_protocol.c
DimCitus: Good point, added now.

src/bin/pg_autoctl/cli_perform.c
DimCitus: If there's no activity (downtime and maintenance windows) then the LSN will not increase either. I think it's fine to have a timeout here, and then users can resort to `watch -n 1 pg_autoctl show state` or their favorite way to follow along what's happening. It might also be a good time to have a look at the logs to understand what's going on, in the case when it's not just taking more time...

onderkalaci: >  I think it's fine to have a timeout here

With the current code, as far as I can see, the promotion would fail even if everything is OK, only if just it takes some time to fetch the wal (in case of async replication) (or some checkpointing?). Am I missing anything on this?

I think we should definitely have the timeout, but it should not break a potentially viable promotion.

> If there's no activity (downtime and maintenance windows) then the LSN will not increase either.

Yeah, in that case we should be fine to fail the promotion anyway?

I think in the vast majority of the cases we would not hit this timeout, but still I can see that some users might hit this. The workaround is to set the priority to highest value, and execute `perform switchover`. 

If we are going with the current code -- which I'm fine with --, it could be nice to document it somewhere (either a github issue or the documentation).

src/monitor/node_active_protocol.c
onderkalaci: I agree that error messages should come from failover code. But, the following behavior is confusing:
- node 1: primary, node2: secondary node3: maintenance
- perform promotion on node 3
- node2 becomes the primary as failover code would ignore node3

This seems like a new case that failover code is not aware of?

DimCitus: Oh I see what you mean, we could refuse the target node for `perform_promotion` when it's not currently in the SECONDARY state. That makes sense, going to add this test.

onderkalaci:
> I don't think this feature needs this much complexity

I'm not suggesting such a big change. My suggestion is to have have a boolean column on the `pgautofaolover.node` table and corresponding field on `AutoFailoverNode ` struct, so pass the information as we do for `candidatePriority`, nothing new. The main difference is that make it explicit enough to prevent any future confusion/maintainability issues.

As I noted, I'm OK with the current approach as well with a minor concern.

src/monitor/node_active_protocol.c
DimCitus: Added now!

DimCitus:
> I'm not suggesting such a big change. My suggestion is to have have a boolean column on the `pgautofaolover.node` table and corresponding field on `AutoFailoverNode ` struct, so pass the information as we do for `candidatePriority`, nothing new. The main difference is that make it explicit enough to prevent any future confusion/maintainability issues.
> 
> As I noted, I'm OK with the current approach as well with a minor concern.

OK. I believe it would make the code more complex though, because we need to ensure the boolean is true for only one node at a time, if there's a failure somewhere we need to clean-up the boolean always and expose a new way for the users to clean it up, and then we will have an even more complex election mechanism.

The trick with the “rigged election” as implemented now is that it is very easy to understand and fix, should something go wrong. We don't even have a new parameter to the already complex failover mechanism. I think I like that a lot, actually.

src/bin/pg_autoctl/cli_perform.c
DimCitus: I'm confused now. The timeout is applied to the client part where we wait for notifications. The only impact when we reach the timeout is that we stop displaying the state changes that are happening on the monitor. The state changes are still happening. It might even be a network timeout problem, after all. The promotion/failover is happening, we just are not receiving notifications about it. Or it's taking a long time, and it is continuing in the background.

Reaching this timeout has no impact on the failover mechanics.

src/monitor/node_active_protocol.c
```diff
@@ -1281,6 +1282,157 @@ perform_failover(PG_FUNCTION_ARGS)
...
+
+		/* if we reach this point, then a failover is in progress */
+		PG_RETURN_BOOL(true);
+	}
+	else
+	{
+		char message[BUFSIZE] = { 0 };
+
+		/*
+		 * In the general case, we perform a little trick:
```
onderkalaci, line 1378: minor: maybe move this comment after the `if (currentNode->candidatePriority == 0)` check is done

onderkalaci (approved):

PR #378, opened 2020-08-26
List pg_stat_statements in .control as dependency

soumyadeep2007:
This ensures that pg_autoctl tries to create the extension if it doesn't
exist, like btree_gist. This ensures fast failure in case
pg_stat_statements isn't installed (possible if we have a dev build of
the server):
```
16:43:35 100236 WARN  NOTICE:  installing required extension "pg_stat_statements"
16:43:35 100236 ERROR Failed to create extension
"pgautofailover"[58P01]: ERROR:  could not open extension control file "/usr/local/pgsql/share/extension/pg_stat_statements.control": No such file or directory
```
Otherwise, we fail much later, while "pg_autoctl create postgres .."
does initdb.
Primary's startup.log:
```
2020-08-25 15:52:07.680 PDT [47143] FATAL:  could not access file "pg_stat_statements": No such file or directory
2020-08-25 15:52:07.680 PDT [47143] LOG:  database system is shut down
```

bf278df introduced pg_stat_statements but did not include it as a
dependency in the control file.


DimCitus:
Hi @soumyadeep2007 ; thanks for your contribution to pg_auto_failover!

I think we still have a misunderstanding to clear out here, and we need a different approach to solve the issue you're reporting, I believe. Here's my thinking:

  - the pgautofailover extension is only used on the monitor, and we neither use or require pg_stat_statements on the monitor ;
  - the pg_stat_statements extension is now required on the Postgres nodes, initialised with the command `pg_autoctl create postgres` ; and we do not install the pgautofailover extension on those nodes.

I believe that in your dev environment you're using the same Postgres installation for the monitor and the Postgres nodes, and that would explain why you're seeing the error message earlier in your case. I don't think this translate to actually improving the situation for our end-users, though.

To help diagnose the problem we could use the equivalent of `ls -l $(pg_config --sharedir)/extension/pg_stat_statements.control` in the code path for `pg_autoctl create postgres` before installing our configuration file. That could improve the user experience when the extension is not available. As you're saying, though, this only happens in development versions of Postgres nowadays, as the contribs come installed within the main packages.

soumyadeep2007:
> * the pgautofailover extension is only used on the monitor, and we neither use or require pg_stat_statements on the monitor ;
> * the pg_stat_statements extension is now required on the Postgres nodes, initialised with the command `pg_autoctl create postgres` ; and we do not install the pgautofailover extension on those nodes.
> 
+1
> I believe that in your dev environment you're using the same Postgres installation for the monitor and the Postgres nodes, and that would explain why you're seeing the error message earlier in your case. I don't think this translate to actually improving the situation for our end-users, though.

Yep. I was using the same server binary for all the nodes and yes, this change will not affect end users.
> 
> To help diagnose the problem we could use the equivalent of `ls -l $(pg_config --sharedir)/extension/pg_stat_statements.control` in the code path for `pg_autoctl create postgres` before installing our configuration file. That could improve the user experience when the extension is not available. As you're saying, though, this only happens in development versions of Postgres nowadays, as the contribs come installed within the main packages.

I agree. For anyone developing on pg_auto_failover with a Postgres server dev build, it would be helpful to have that early error feedback. Otherwise, it was non-trivial for me at least, to discover the root cause amid the many ERROR messages pg_auto_failover reports in that scenario.

DimCitus:
> > To help diagnose the problem we could use the equivalent of `ls -l $(pg_config --sharedir)/extension/pg_stat_statements.control` in the code path for `pg_autoctl create postgres` before installing our configuration file. That could improve the user experience when the extension is not available. As you're saying, though, this only happens in development versions of Postgres nowadays, as the contribs come installed within the main packages.
> 
> I agree. For anyone developing on pg_auto_failover with a Postgres server dev build, it would be helpful to have that early error feedback. Otherwise, it was non-trivial for me at least, to discover the root cause amid the many ERROR messages pg_auto_failover reports in that scenario.

Do you want to change your current PR to this approach? You can find calls to other Postgres binaries in https://github.com/citusdata/pg_auto_failover/blob/master/src/bin/pg_autoctl/pgctl.c#L138 and other places. This call to `pg_config --sharedir` should be easy to integrate, if you want to have a look.

soumyadeep2007:
Hey @DimCitus I took a stab. I didn't add any tests. Please let me know if you want me to.

src/bin/pg_autoctl/pgctl.c
```diff
@@ -252,6 +252,66 @@ config_find_pg_ctl(PostgresSetup *pgSetup)
...
+			log_error("Extension %s is not installed", extName);
+			free_program(&prog);
+			return false;
+		}
+	}
+	else
+	{
+		(void) log_program_output(prog, LOG_INFO, LOG_ERROR);
+		log_error("Failed to run \"%s\" on \"%s\", see above for details",
+				  pg_config_path, pgSetup->pgdata);
```
DimCitus, line 306: We're not using PGDATA at all in this function, so let's not confuse the users by using it in the error message now...

DimCitus (changes requested):
Thanks for pushing this through! I think we need small adjustments before we can merge, and I've left detailed review for all I could see.

soumyadeep2007:
Thanks for the detailed feedback! I addressed your comments in my latest push

src/bin/pg_autoctl/keeper_pg_init.c
```diff
@@ -709,6 +709,15 @@ create_database_and_extension(Keeper *keeper)
 	/* publish our new pgSetup to the caller postgres state too */
 	postgres->postgresSetup.ssl = initPostgres.postgresSetup.ssl;
 
+	/* Ensure pg_stat_statements is installed in the server extension dir used
+	 * to create the Postgres instance
+	 */
```
DimCitus, line 714: Minor style issue: the first line of the comments is always empty in the code base, for reasons...

src/bin/pg_autoctl/pgctl.c
```diff
@@ -252,6 +252,67 @@ config_find_pg_ctl(PostgresSetup *pgSetup)
...
+find_extension_control_file(PostgresSetup *pgSetup, char *extName)
+{
+	char pg_config_path[MAXPGPATH] = { 0 };
+	char extension_path[MAXPGPATH] = { 0 };
+	char *share_dir;
+	char extension_control_file_name[MAXPGPATH] = { 0 };
+	char *lines[1];
+	Program prog;
+
+	log_info("Checking if the %s extension is installed", extName);
```
DimCitus, line 270: I think we can make it a `log_debug`, we don't expect this to either fail or take a long time...

src/bin/pg_autoctl/pgctl.c
```diff
@@ -252,6 +252,67 @@ config_find_pg_ctl(PostgresSetup *pgSetup)
 }
 
 
+/*
+ * find_extension_control_file ensures that the extension is present in the
+ * given Postgres installation. This does the equivalent of:
+ * ls -l $(pg_config --sharedir)/extension/pg_stat_statements.control
+ */
+bool
+find_extension_control_file(PostgresSetup *pgSetup, char *extName)
```
DimCitus, line 261: I would still like it better if it took a `const char *pg_ctl` argument rather than the whole PostgresSetup, singling the read-only aspect of the operation in terms of in-memory data structures...

src/bin/pg_autoctl/pgctl.c
```diff
@@ -252,6 +252,67 @@ config_find_pg_ctl(PostgresSetup *pgSetup)
...
+			return false;
+		}
+		share_dir = lines[0];
+
+		join_path_components(extension_path, share_dir, "extension");
+		sformat(extension_control_file_name, MAXPGPATH, "%s.control", extName);
+		join_path_components(extension_path, extension_path, extension_control_file_name);
+		if (!file_exists(extension_path))
+		{
+			log_error("Failed to find extension control file %s", extension_path);
```
DimCitus, line 298: Please quote the filename in the error message, so that we are safe even with paths containing spaces, and most importantly, make it visually obvious which part of the error message is the file path.

``` c
			log_error("Failed to find extension control file \"%s\"", extension_path);
```

DimCitus (changes requested):
Looks good to me!

There are still a couple of very minor things to fix and then we can merge. Thanks for your efforts polishing the code here.

src/bin/pg_autoctl/keeper_pg_init.c
soumyadeep2007: Whoops, sorry I missed that

src/bin/pg_autoctl/pgctl.c
soumyadeep2007: Sorry! I totally missed your earlier comment

src/bin/pg_autoctl/keeper_pg_init.c
```diff
@@ -709,6 +709,17 @@ create_database_and_extension(Keeper *keeper)
 	/* publish our new pgSetup to the caller postgres state too */
 	postgres->postgresSetup.ssl = initPostgres.postgresSetup.ssl;
 
+	/*
+	 * Ensure pg_stat_statements is installed in the server extension dir used
+	 * to create the Postgres instance
+	 */
+	config_find_pg_ctl(&config->pgSetup);
```
DimCitus, line 716: Can you explain why you needed to add a call to that function here? This is supposed to be done when parsing the command line arguments, or when reading our configuration file when it exists already. It's quite surprising to see that call here, smells like “modularity violation” kind of thing.

src/bin/pg_autoctl/pgctl.c
DimCitus: bump.

DimCitus (changes requested):

src/bin/pg_autoctl/keeper_pg_init.c
soumyadeep2007: Ah, sorry that was a remnant from my previous implementation. In my previous implementation's call site, `pgSetup` didn't have the `pg_ctl` field set.
Thanks for spotting this!

DimCitus (approved):

PR #380, opened 2020-08-26
Fix unhealthy node detection when reported by pg_autoctl.

### LINKED ISSUE #379: Primary is not running, but failover is not triggered
Steps to repro:

Basically, have 1 monitor, 3 nodes. Add a syntax error to the Primary's postgresql.conf, restart primary. Primary does not start again, but failover is not triggered either.

```
-- setup monitor
pg_autoctl create monitor --pgport 10000 --pgdata /Users/onderkalaci/Documents/pg_auto_failover/monitor  --auth trust --ssl-self-signed --run

-- node 1, primary
pg_autoctl create postgres --pgport 10001 --pgdata /Users/onderkalaci/Documents/pg_auto_failover/worker1a --auth trust --monitor postgres://autoctl_node@192.168.2.4:10000/pg_auto_failover?sslmode=require --pgctl /Users/onderkalaci/Documents/citus_code/pgenv/pgsql/bin//pg_ctl  --ssl-self-signed --run

-- node 2
pg_autoctl create postgres --pgport 10002 --pgdata /Users/onderkalaci/Documents/pg_auto_failover/worker1b --auth trust --monitor postgres://autoctl_node@192.168.2.4:10000/pg_auto_failover?sslmode=require --pgctl /Users/onderkalaci/Documents/citus_code/pgenv/pgsql/bin//pg_ctl  --ssl-self-signed --run

-- node 3
pg_autoctl create postgres --pgport 10003 --pgdata /Users/onderkalaci/Documents/pg_auto_failover/worker1c --auth trust --monitor postgres://autoctl_node@192.168.2.4:10000/pg_auto_failover?sslmode=require --pgctl /Users/onderkalaci/Documents/citus_code/pgenv/pgsql/bin//pg_ctl  --ssl-self-signed --run

```

Now, add an error to primary nodes' Postgres config and reload
```
echo "shared_preload_libraries='pgautofailover, wrong_extension'" >> /Users/onderkalaci/Documents/pg_auto_failover/worker1a/postgresql.auto.conf

ctrl+c to the keeper process of node 1 and start again
```

Later, the keeper starts to fail:
```
11:39:13 70520 ERROR Failed get postmaster pid, file "/Users/onderkalaci/Documents/pg_auto_failover/worker1a/postmaster.pid" does not exists
11:39:13 70520 ERROR Failed to get Postgres pid, see above for details
11:39:13 70520 WARN  Postgres logs from "/Users/onderkalaci/Documents/pg_auto_failover/worker1a/startup.log":
11:39:13 70520 ERROR 2020-08-26 11:39:03.225 CEST [70835] FATAL:  could not access file "wrong_extension": No such file or directory
11:39:13 70520 ERROR 2020-08-26 11:39:03.225 CEST [70835] LOG:  database system is shut down
11:39:13 70520 WARN  Failed to start Postgres instance at "/Users/onderkalaci/Documents/pg_auto_failover/worker1a"
11:39:13 70941 INFO   /Users/onderkalaci/Documents/citus_code/pgenv/pgsql/bin/postgres -D /Users/onderkalaci/Documents/pg_auto_failover/worker1a -p 10001 -h *
11:39:16 70521 ERROR Failed get postmaster pid, file "/Users/onderkalaci/Documents/pg_auto_failover/worker1a/postmaster.pid" does not exists
11:39:16 70521 ERROR Failed to get Postgres pid, see above for details
11:39:16 70521 ERROR Failed to ensure that Postgres is running in "/Users/onderkalaci/Documents/pg_auto_failover/worker1a"
11:39:16 70521 ERROR Failed to restart PostgreSQL, see PostgreSQL logs for instance at "/Users/onderkalaci/Documents/pg_auto_failover/worker1a".
11:39:16 70521 WARN  pg_autoctl failed to ensure current state "primary": PostgreSQL is not running
11:39:17 70521 ERROR Failed to fetch current replication properties from standby node: no standby connected in pg_stat_replication.
11:39:17 70521 WARN  HINT: check pg_autoctl and Postgres logs on standby nodes
11:39:17 70521 WARN  Failed to update the keeper's state from the local PostgreSQL instance.
11:39:17 70521 ERROR Failed to restart PostgreSQL 4 times in the last 34s, reporting PostgreSQL not running to the pg_auto_failover monitor.

```

The state shows `Reachable = no`:
```
pg_autoctl show state --pgdata /Users/onderkalaci/Documents/pg_auto_failover/monitor/
  Name |  Node |         Host:Port |       LSN | Reachable |       Current State |      Assigned State
-------+-------+-------------------+-----------+-----------+---------------------+--------------------
node_1 |     1 | 192.168.2.4:10001 | 0/5000C58 |       yes |           secondary |           secondary
node_2 |     2 | 192.168.2.4:10002 | 0/5000BE0 |        no |             primary |             primary
node_3 |     3 | 192.168.2.4:10003 | 0/5000C58 |       yes |           secondary |           secondary
```

Connect to monitor, and see `health = 0`  : 
```SQL
pg_auto_failover=# select * from  pgautofailover.node where reportedstate = 'primary';
-[ RECORD 1 ]-------+------------------------------
formationid         | default
nodeid              | 2
groupid             | 0
nodename            | node_2
nodehost            | 192.168.2.4
nodeport            | 10002
sysidentifier       | 6865218689184959116
goalstate           | primary
reportedstate       | primary
reportedpgisrunning | f
reportedrepstate    | unknown
reporttime          | 2020-08-26 11:29:14.196976+02
reportedlsn         | 0/5000BE0
walreporttime       | 2020-08-26 11:20:12.957677+02
health              | 0
healthchecktime     | 2020-08-26 11:29:05.431315+02
statechangetime     | 2020-08-26 11:29:14.196976+02
candidatepriority   | 50
replicationquorum   | t
```

Failover never happens, and we cannot connect to the primary:
```
 psql -p 10001 postgres
psql: error: could not connect to server: could not connect to server: No such file or directory
	Is the server running locally and accepting
	connections on Unix domain socket "/tmp/.s.PGSQL.10001"?
```

DimCitus:
Fixes #379

### DIFF
--- a/src/monitor/group_state_machine.c
+++ b/src/monitor/group_state_machine.c
@@ -1574,11 +1574,6 @@ static bool
 IsUnhealthy(AutoFailoverNode *pgAutoFailoverNode)
 {
 	TimestampTz now = GetCurrentTimestamp();
-	List *pgIsNotRunningStateList =
-		list_make4_int(REPLICATION_STATE_DRAINING,
-					   REPLICATION_STATE_DEMOTED,
-					   REPLICATION_STATE_DEMOTE_TIMEOUT,
-					   REPLICATION_STATE_PREPARE_MAINTENANCE);
 
 	if (pgAutoFailoverNode == NULL)
 	{
@@ -1608,8 +1603,7 @@ IsUnhealthy(AutoFailoverNode *pgAutoFailoverNode)
 	 * If the keeper reports that PostgreSQL is not running, then the node
 	 * isn't Healthy.
 	 */
-	if (!pgAutoFailoverNode->pgIsRunning &&
-		IsStateIn(pgAutoFailoverNode->goalState, pgIsNotRunningStateList))
+	if (!pgAutoFailoverNode->pgIsRunning)
 	{
 		return true;
 	}
--- a/tests/pgautofailover_utils.py
+++ b/tests/pgautofailover_utils.py
@@ -373,20 +373,12 @@ class PGNode:
 
     def restart_postgres(self):
         """
-        Restart the postgres configuration by running:
-          pg_ctl -D ${self.datadir} restart
+        Restart Postgres with pg_autoctl do service restart postgres
         """
-        restart_command = [shutil.which('pg_ctl'), '-D', self.datadir, 'restart']
-        with self.vnode.run(restart_command) as restart_proc:
-            out, err = self.cluster.communicate(restart_proc, COMMAND_TIMEOUT)
-            if restart_proc.returncode > 0:
-                print("restarting postgres for '%s' failed, out: %s\n, err: %s"
-                      % (self.vnode.address, out, err))
-                return False
-            elif restart_proc.returncode is None:
-                print("restarting postgres for '%s' timed out")
-                return False
-            return True
+        command = PGAutoCtl(self)
+
+        command.execute("service restart postgres",
+                        'do', 'service', 'restart', 'postgres')
 
     def pg_is_running(self, timeout=COMMAND_TIMEOUT):
         """
--- a/tests/test_ensure.py
+++ b/tests/test_ensure.py
@@ -2,6 +2,7 @@ import pgautofailover_utils as pgautofailover
 from nose.tools import *
 
 import time
+import os.path
 
 cluster = None
 node1 = None
@@ -65,3 +66,25 @@ def test_004_demoted():
     # postgres again before calling the transition function
     print("re-starting pg_autoctl on node1")
     assert node1.wait_until_state(target_state="secondary")
+
+def test_005_inject_error_in_node2():
+    assert node2.wait_until_state(target_state="primary")
+
+    # break Postgres setup on the primary, and restart Postgres: then
+    # Postgres keeps failing to start, and pg_autoctl still communicates
+    # with the monitor, which should still orchestrate a failover.
+    pgconf = os.path.join(node2.datadir, "postgresql.conf")
+
+    with open(pgconf, "a+") as f:
+        f.write("\n")
+        f.write("shared_preload_libraries='wrong_extension'\n")
+
+    node2.restart_postgres()
+
+    # the first step is the promotion of the other node as the new primary:
+    assert node1.wait_until_state("wait_primary")
+
+    # then when the failover happens, the new primary postgresql.conf gets
+    # copied over, and we get the nodes back to primary/secondary
+    assert node2.wait_until_state("secondary")
+    assert node1.wait_until_state("primary")


src/monitor/group_state_machine.c
```diff
@@ -1608,8 +1603,7 @@ IsUnhealthy(AutoFailoverNode *pgAutoFailoverNode)
 	 * If the keeper reports that PostgreSQL is not running, then the node
 	 * isn't Healthy.
 	 */
-	if (!pgAutoFailoverNode->pgIsRunning &&
-		IsStateIn(pgAutoFailoverNode->goalState, pgIsNotRunningStateList))
+	if (!pgAutoFailoverNode->pgIsRunning)
```
onderkalaci, line 1606: I remember that we added these states in MS [PR](https://github.com/citusdata/pg_auto_failover/pull/326/commits/f91be34d0300455b41aa1ab54a02e8cd231b4e71) to prevent some of the edge cases. Given that the PR is pretty large, I cannot easily remember what the reason  was.

I'm mostly skeptical about `REPLICATION_STATE_PREPARE_MAINTENANCE `.  Do you remember any edge cases around it?

DimCitus: Well the idea with having a list of states in the `IsUnhealthy` function would be to implement exceptions where we could have the information that `pgIsRunning` is false and yet would not consider the node as UnHealthy.

I think that in the current version of the Group FSM code, we don't depend on the value of pgIsRunning or the result of the IsUnhealthy function call result when we are in a state allowing Postgres to be down. This will impact a couple code path in the multiple standby (s)election mechanism, where we skip a node when it is unhealthy. That said, we don't want to target a node that is in one of the listed states anyway, so we're good.

JelteF: Some extra context on this. It seems that real bug is that the previous logic was simply incorrect, i.e. it's missing a `!` before `IsStateIn`. Instead of reporting unhealthy when the node is not running and it's in a state where it should be running, it reports unhealthy when it's in a state where it should not be running. 

This already seems the case in the commit linked by @onderkalaci. So I don't see how this logic ever worked as expected. This makes me think that it's fine to completely remove it for now.

JelteF (changes requested):
I think the fix is fine as explained in my inline comment. I do think that adding a regression test for this seems required before merging this. This is pretty much core HA behaviour (failing over when primary is down) that was broken without existing tests catching it.

DimCitus:
> I think the fix is fine as explained in my inline comment. I do think that adding a regression test for this seems required before merging this. This is pretty much core HA behaviour (failing over when primary is down) that was broken without existing tests catching it.

We had almost the right test for the situation, but in the test we have we stop both Postgres and `pg_autoctl`, we were missing a test where `pg_autoctl` is running fine and contacting the monitor, but Postgres won't start. I have now added such a test.

JelteF (approved):

PR #383, opened 2020-09-01
Default password_encryption based on auth param to pg_autoctl create

soumyadeep2007:
This automatically sets the password_encryption GUC while performing a
create {postgres|monitor} if the supplied --auth method is password
based (one of ["md5", "password", "scram-sha-256"].

This saves the cluster operator from having to set this GUC and reload
the server configuration on each node in case the --auth option is
either {password, scram-sha-256}. (The operator didn't have to set this
GUC for --auth md5 as the default value of password_encryption is md5)


src/bin/pg_autoctl/pgctl.c
```diff
@@ -506,6 +506,22 @@ prepare_guc_settings_from_pgsetup(const char *configFilePath,
...
 		}
+		else if (streq(setting->name, "password_encryption"))
+		{
+			Assert(strlen(pgSetup->authMethod) > 0);
+
+			/*
+			 * Set password_encryption if the authMethod is password based.
+			 */
+			if (streq(pgSetup->authMethod, "md5") ||
+				streq(pgSetup->authMethod, "password") ||
```
JelteF, line 517: This code doesn't work for the `password` case. In that case `password_encryption` should still be set to `md5` or `scram-sha-256`. 

When running `pg_autoctl create monitor --pgdata monitor --ssl-self-signed --auth password --run`, postgres doesn't start and the logs show these errors:
```
11:47:11 21215 ERROR 2020-09-01 09:47:01.724 GMT [21220] LOG:  invalid value for parameter "password_encryption": "password"
11:47:11 21215 ERROR 2020-09-01 09:47:01.724 GMT [21220] HINT:  Available values: md5, scram-sha-256.
11:47:11 21215 ERROR 2020-09-01 11:47:01.724 CEST [21220] FATAL:  configuration file "/home/jelte/work/pg_auto_failover/monitor/postgresql-auto-failover.conf" contains errors
```

According to the docs `password` supports `scram-sha-256` (which makes sense because you can hash the plaintext password to whatever you want): https://www.postgresql.org/docs/12/auth-password.html
So let's use that when specifying `password`.

src/bin/pg_autoctl/monitor_pg_init.c
```diff
@@ -47,6 +47,7 @@ GUC monitor_default_settings[] = {
 	{ "log_disconnections", "off" },
 	{ "log_lock_waits", "on" },
 	{ "log_statement", "ddl" },
+	{ "password_encryption", NULL }, /* will be set later */
```
JelteF, line 50:
```suggestion
	{ "password_encryption", NULL }, /* will be set later based on --auth */
```

src/bin/pg_autoctl/primary_standby.c
```diff
@@ -53,6 +53,7 @@ static void local_postgres_update_pg_failures_tracking(LocalPostgresServer *post
 	{ "log_connections", "off" }, \
 	{ "log_disconnections", "off" }, \
 	{ "log_lock_waits", "on" }, \
+	{ "password_encryption", NULL }, /* will be set later */ \
```
JelteF, line 56:
```suggestion
	{ "password_encryption", NULL }, /* will be set later based on --auth */ \
```

JelteF (changes requested):
Nice improvement! I left some comments.

src/bin/pg_autoctl/monitor_pg_init.c
DimCitus: I'm not so comfortable with having a NULL in there, what about using `"md5"` (or something else) as a placeholder value, or even `"placeholder"` to make it explicit? What could go wrong when we use NULL?

JelteF: because we explicitely handle it and don't write anything in case of other auth methods it shouldn't matter. I had the same thought though. So maybe indeed to be on the safe side for future refactors we should put md5 there.

src/bin/pg_autoctl/pgctl.c
soumyadeep2007: > This code doesn't work for the `password` case. In that case `password_encryption` should still be set to `md5` or `scram-sha-256`.

I see. I thought there would be an option to store the password as plain-text, but evidently not as you pointed out and from the doc: _"Previous PostgreSQL releases supported storing the password on the server in plain text. This is no longer possible."_

> According to the docs password supports scram-sha-256 (which makes sense because you can hash the plaintext password to whatever you want): https://www.postgresql.org/docs/12/auth-password.html
So let's use that when specifying password

+1. Yup a much better default than md5.

Ill make these changes.

src/bin/pg_autoctl/monitor_pg_init.c
soumyadeep2007: Yeah I was kind of enforcing it with [here](https://github.com/citusdata/pg_auto_failover/pull/383/commits/706c82f8c8023d1ea26c6f3264e623a9e5122b94#diff-4b0fb807d0a2ae6005c10e98bb5a675dR511) with the Assert.

I initially thought of specifying `md5` as a default, but I had decided against it as this default may become inconsistent with PG if ever PG changes the default value to something else. But I don't feel too strongly about that.

src/bin/pg_autoctl/primary_standby.c
soumyadeep2007: Opted to remove the comment as I set the default as `md5` now

JelteF (approved):
Looks good now

PR #384, opened 2020-09-01
Move the keeper reload_configuration into a hooks array.

DimCitus:
This allows more flexibility for extending the behavior at reload.

### DIFF
--- a/src/bin/pg_autoctl/keeper.c
+++ b/src/bin/pg_autoctl/keeper.c
@@ -22,6 +22,7 @@
 #include "pghba.h"
 #include "pgsetup.h"
 #include "primary_standby.h"
+#include "signals.h"
 #include "state.h"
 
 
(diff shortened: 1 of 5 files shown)


src/bin/pg_autoctl/primary_standby.c
```diff
@@ -80,14 +80,14 @@ GUC postgres_default_settings_13[] = {
 
 GUC citus_default_settings_pre_13[] = {
 	DEFAULT_GUC_SETTINGS_FOR_PG_AUTO_FAILOVER_PRE_13,
-	{ "shared_preload_libraries", "'citus','pg_stat_statements'" },
+	{ "shared_preload_libraries", "'citus,pg_stat_statements'" },
```
onderkalaci, line 83: When I  do
```
alter system set shared_preload_libraries TO 'citus,pg_stat_statements';
```
Postgres cannot start, complaining that the syntax is wrong.

It can  start with the following:
```
alter system set shared_preload_libraries TO citus,pg_stat_statements;
```

So, do we really need the single  quotes?

DimCitus: I guess that ALTER SYSTEM command parsing does the list splitting in a different way than the configuration file parser. This fix comes from trying this code in a Citus environment.

src/bin/pg_autoctl/keeper.c
```diff
@@ -1965,3 +1966,93 @@ keeper_config_accept_new(Keeper *keeper, KeeperConfig *newConfig)
...
 										&(newConfig->pgSetup));
 }
+
+
+/*
+ * reload_configuration reads the supposedly new configuration file and
+ * integrates accepted new values into the current setup.
+ */
+bool
+keeper_reload_configuration(Keeper *keeper, bool firstLoop)
```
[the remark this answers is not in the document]
DimCitus: I also would like to be able to retry the SIGHUP processing, but it's pretty hard to invent a solid approach that would not surprise anybody. As you're saying we can't really retry the SIGHUP action infinitely. I think the current implementation is easy to understand and makes sense.

Also we have many other function that return a bool and that are used in context where we can't really act on their return value. I prefer to leave it that way, so that if we come up with a better error handling strategy later, we have the information we need already.

src/bin/pg_autoctl/primary_standby.c
onderkalaci: Nothing this as reference, you explanation sounds right. Because even if  I do:
```
alter system set shared_preload_libraries TO citus,pg_stat_statements;
```

On `postgresql.auto.conf`, it shows up:
```
shared_preload_libraries = 'citus, pg_stat_statements'
```

onderkalaci (approved):

Issue #385, opened 2020-09-02
Bulk feedback on master branch pre 1.4 release


trecloux:
Hi All,

This issue is following a [twitter thread](https://twitter.com/tapoueh/status/1301025400810766338) with @DimCitus.

As an introduction, we (https://github.com/auchanretailfrance) are evaluating pg_auto_failover in a Kubernetes context and started to build an Helm chart to support this context. If we succeed, we will opensource the Helm chart.

After testing a custom build of master branch ( commit 0e4d2c972f3265789c5bbfb60a26ef9d76e6bc57 ), here is our feedback.

- pg_autofailover is generating PostgreSQL tuning parameters.
  With PostgreSQL >= 12, these parameters are duplicated in `postgresql-auto-failover.conf` and `postgresql-auto-failover-standby.conf`
  With PostgreSQL < 12, these parameters are duplicated in `postgresql-auto-failover.conf` and `recovery.conf`.
  When PostgreSQL 11 is parsing `recovery.conf` file, it raises an error and PostgreSQL can't start.
  This is a blocker for PostgreSQL 11.

- in a container environment, it looks like the postgresql tuning is based on the host resources, not on the container/cgroup limits. Ex `7537 MB` for a container with 1GB memory limit.

- `--nodename` parameter was replaced by `--hostname`, this is breaking existing scripts and CLI error message does not mention the new parameter name. It might be interesting to outline this change in the release notes, and/or in the CLI error message.

trecloux:
Here are the details to reproduce the recovery.conf parsing error

- build docker image from Dockerfile at root of repository
```
sudo docker build -t pg_auto_failover .
Sending build context to Docker daemon  7.312MB
Step 1/15 : FROM debian:buster-slim
 ---> c7346dd7f20e
....
Step 15/15 : ENV PATH $PATH:/usr/lib/postgresql/11/bin
 ---> Running in abc9e556baeb
Removing intermediate container abc9e556baeb
 ---> ebe48efec9b2
Successfully built ebe48efec9b2
Successfully tagged pg_auto_failover:latest

```

- create the container, create the monitor, run the monitor 

```
sudo docker run -ti --user postgres --name pg_autofailover  pg_auto_failover
[...]

PR #388, opened 2020-09-03
Fix/tuning and recovery setup

DimCitus:
See #385

### DIFF
--- a/src/bin/pg_autoctl/pgctl.c
+++ b/src/bin/pg_autoctl/pgctl.c
@@ -45,11 +45,13 @@ static bool pg_include_config(const char *configFilePath,
 							  const char *configIncludeComment);
 static bool ensure_default_settings_file_exists(const char *configFilePath,
 												GUC *settings,
-												PostgresSetup *pgSetup);
+												PostgresSetup *pgSetup,
+												bool includeTuning);
 static bool prepare_guc_settings_from_pgsetup(const char *configFilePath,
 											  PQExpBuffer config,
 											  GUC *settings,
-											  PostgresSetup *pgSetup);
+											  PostgresSetup *pgSetup,
+											  bool includeTuning);
 static void log_program_output(Program prog, int outLogLevel, int errorLogLevel);
 
 
@@ -322,6 +324,7 @@ pg_add_auto_failover_default_settings(PostgresSetup *pgSetup,
 									  char *configFilePath,
 									  GUC *settings)
 {
+	bool includeTuning = true;
 	char pgAutoFailoverDefaultsConfigPath[MAXPGPATH];
 
 	/*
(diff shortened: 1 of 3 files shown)


src/bin/pg_autoctl/pgctl.c
```diff
@@ -671,18 +680,8 @@ prepare_guc_settings_from_pgsetup(const char *configFilePath,
...
 	}
 
-	/*
-	 * Only bother to apply settings on Linux, macOS and Windows are considered
-	 * development environment and we decide to refrain from tuning a dedicated
-	 * Postgres system on those targets.
-	 *
-	 * Also disable Postgres tuning when running the unit test suite.
-	 */
-#if defined(__linux__)
```
JelteF, line 681: I don't see this linux check anywhere now. Is that on purpose?

JelteF (commented):
Mayby (in a seperate PR) we can add a `--no-tuning` flag. To allow people to disable this feature in case it causes bad settings for them.

src/bin/pg_autoctl/pgtuning.c
```diff
@@ -262,44 +271,90 @@ pgtuning_edit_guc_settings(GUC *settings, DynamicTuning *tuning,
...
+
+				(void) pretty_print_bytes(pretty, sizeof(pretty),
+										  tuning->shared_buffers);
+
+				appendPQExpBuffer(contents, "%s = '%s'\n",
+								  setting->name, pretty);
+			}
+			else
+			{
+				appendPQExpBuffer(contents, "%s = %s\n",
```
JelteF, line 301: Here we now take the settings from our default tuning array:
```
GUC postgres_tuning[] = {
	{ "track_functions", "pl" },
	{ "shared_buffers", "'128 MB'" },
	{ "work_mem", "'4 MB'" },
	{ "maintenance_work_mem", "'64MB'" },
	{ "effective_cache_size", "'4 GB'" },
	{ "autovacuum_max_workers", "3" },
	{ "autovacuum_vacuum_scale_factor", "0.08" },
	{ "autovacuum_analyze_scale_factor", "0.02" },
	{ NULL, NULL }
};

```

We didn't do this before. Do these numbers make sense if we cannot get tuning info? Or would it be better not to tune these settings at all in this case, i.e. removing the `else` case here.

src/bin/pg_autoctl/pgctl.c
DimCitus: Good point. Unfortunately there's no way to comment code that's removed, and I forgot to add a note about that in the PR itself. Yes it's on purpose. You reported that your laptop was fine with running more than 4 Postgres nodes with 1/4th of the RAM in shared_buffers for each, for one thing, and for another thing, I don't think we can decide that only Linux is production, come to think of it.

Also makes it easier to test what's happening for others, because it's happening the same on all platforms now.

src/bin/pg_autoctl/pgtuning.c
DimCitus: Before those changes we would only call the function when we wanted to install locally tuned values. Now we call `pgtuning_prepare_guc_settings` no matter what, and when we see that we're not going to adjust to the host specifics, we use our own default values.

I believe our defaults are either the same as Postgres, or better. So yeah I think it makes sense.

JelteF (approved):

Issue #392, opened 2020-09-08
Adding a standby while another is in maintenance is not possible and cannot be killed by ctrl+C


JelteF:
Setup:
```bash
pg_autoctl create postgres --pgdata node1 --run
pg_autoctl create postgres --pgdata node2 --run
pg_autoctl enable maintenance --pgdata node2 
```

Then this command fails continously and cannot be killed by ctrl+c anymore

```
pg_autoctl create postgres --pgdata node3
```
Output:
```
^C15:08:02 500 INFO  pg_autoctl received signal SIGINT, terminating
15:08:02 500 ERROR Failed to send signal Interrupt to service postgres with pid 505
15:08:02 506 ERROR Monitor ERROR:  primary node 172.28.149.103:5501 is already in state wait_primary
15:08:02 506 ERROR Monitor HINT:  Retry registering in a moment
15:08:02 506 ERROR SQL query: SELECT * FROM pgautofailover.register_node($1, $2, $3, $4, $5, $6, $7, $8::pgautofailover.replication_state, $9, $10, $11)
15:08:02 506 ERROR SQL params: 'default', '172.28.149.103', '5503', 'postgres', '', '6868238261832629440', '-1', 'init', 'standalone', '50', 'true'
15:08:02 506 WARN  Failed to register node 172.28.149.103:5503 in group -1 of formation "default" with initial state "init" because the monitor is already registering another standby, retrying in 1s
```

Current cluster state
```

  Name |  Node |           Host:Port |       LSN | Reachable |       Current State |      Assigned State
-------+-------+---------------------+-----------+-----------+---------------------+--------------------
node_1 |     1 | 172.28.149.103:5501 | 0/3000728 |       yes |        wait_primary |        wait_primary
node_2 |     2 | 172.28.149.103:5502 | 0/3000728 |       yes |         maintenance |         maintenance
```

JelteF:
It would of course be best if it would allow a node to be added while another is in maintenance. I understand this might be problematic, because it could cause weird edge cases. The command should at least be able to be killed though.

DimCitus:
We should allow the registration of a new standby in that state, I can't see what's getting in the way there (other than the current implementation that's a little too restrictive for its own good). 

The general case is there to make sure that the primary has had enough time to prepare the HBA entries, and really, we could lift the limitation about only one standby at a time now that we edit the HBA eagerly, but we still have the race condition that the primary reports having added all the standbys and then it missed one.

[...]

PR #396, opened 2020-09-09
Implement proper signal handling in select calls.

DimCitus:
See #392.

### DIFF
--- a/src/bin/pg_autoctl/monitor.c
+++ b/src/bin/pg_autoctl/monitor.c
@@ -3162,6 +3162,15 @@ monitor_wait_until_primary_applied_settings(Monitor *monitor,
 	while (!applySettingsTransitionDone)
 	{
 		/* Sleep until something happens on the connection. */
+		sigset_t sig_mask;
+		sigset_t sig_mask_orig;
+
+		/* we have milliseconds, we want seconds and nanoseconds separately */
+		int seconds = PG_AUTOCTL_LISTEN_NOTIFICATIONS_TIMEOUT;
+		int nanosecs = 0;
+		struct timespec timeout = { .tv_sec = seconds, .tv_nsec = nanosecs };
+
+		int ret;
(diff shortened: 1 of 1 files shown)


JelteF (approved):
I feel like it might be possible to move this `pselect`+(`un`)`block_signals` code to a shared function, since both places look so similar to the other one. If that's not the case for whatever reason feel free to merge this. It looks like a good change.

However, like discussed in chat though, this doesn't fix the issue described in https://github.com/citusdata/pg_auto_failover/issues/392. So another fix for that one is still needed.

DimCitus:
> I feel like it might be possible to move this `pselect`+(`un`)`block_signals` code to a shared function, since both places look so similar to the other one. If that's not the case for whatever reason feel free to merge this. It looks like a good change.

I have now refactored the code in a way that allows better code share / re-use. This needs some more infrastructure, as in context structures and function pointers, to make it for the lack of closures and dynamic dispatch in C.

src/bin/pg_autoctl/monitor.c
```diff
@@ -2982,25 +3031,28 @@ monitor_stop_maintenance(Monitor *monitor, int nodeId)
...
  * it's expected that the caller keeps polling the results to drain the queue
  * of notifications received from the previous calls loop.
  */
-bool
-monitor_get_notifications(Monitor *monitor, int timeoutMs)
+static bool
+monitor_process_notifications(Monitor *monitor,
+							  int timeoutMs,
+							  char *channels[],
+							  void *NotificationContext,
```
JelteF, line 3050: Looks like a type otherwise:
```suggestion
							  void *notificationContext,
```

JelteF (approved):

src/bin/pg_autoctl/monitor.c
DimCitus: Good point, done now.

PR #398, opened 2020-09-10
Add some targets to quickly spawn a cluster

JelteF:
With these changes you can now run `make cluster` to spawn a 2 node
cluster and `make cluster3` to spawn a 3 node cluster.

It basically runs the commands that I would normally run to test out a
new feature manually. So this saves me quite some time.

There's ofcourse various improvements possible to these commands. For
instance making them more configurable, with base directories or port
prefixes. However, I don't think these are necessary for the first PR.
We can add these improvements later on just as easy.

Killing it is as easy as `Ctrl+B & Enter`

Somewhat related, to make tmux easier to work with. This is my minimal config that you can put in `~/.tmux.conf`:
```tmux
set -g default-terminal "screen-256color"

# https://stackoverflow.com/a/40902312/2570866
# Version-specific commands [grumble, grumble]
# See: https://github.com/tmux/tmux/blob/master/CHANGES
run-shell "tmux setenv -g TMUX_VERSION $(tmux -V | cut -c 6-)"

if-shell -b '[ $(printf "%s\n" "$TMUX_VERSION" "2.1" | sort -V | tail -n1) = 2.1 ]' \
  "set -g mouse-select-pane on; set -g mode-mouse on; \
    set -g mouse-resize-pane on; set -g mouse-select-window on"

# In version 2.1 "mouse" replaced the previous 4 mouse options
if-shell -b '[ $(printf "%s\n" "$TMUX_VERSION" "2.1" | sort -V | head -n1) = 2.1 ]' \
  "set -g mouse on"

# UTF8 is autodetected in 2.2 onwards, but errors if explicitly set
if-shell -b '[ $(printf "%s\n" "$TMUX_VERSION" "2.2" | sort -V | tail -n1) = 2.2 ]' \
  "set -g utf8 on; set -g status-utf8 on; set -g mouse-utf8 on"

# Intuitive splitting
bind | split-window -h
bind _ split-window -v
# Easy swithing pane
bind-key C-a last-pane

set -g status-bg black
set -g status-fg white
set -g status-left '#S | '
setw -g window-status-format "#I #W |"
setw -g window-status-current-format "#[fg=green] #I #W #[fg=white]|"

set -g history-limit 3000

# Turn on vim awesomeness
set-window-option -g mode-keys vi

bind h select-layout even-vertical

# Make ctrl left/right work
set-window-option -g xterm-keys on

set -ga terminal-overrides ',*:sitm@,ritm@'
```


DimCitus (commented):
I have cleaned the approach to make it easier to maintain, and added the command `pg_autoctl do tux script` to produce a tmux script for a number of nodes.

src/bin/pg_autoctl/cli_do_tmux.c
```diff
@@ -0,0 +1,441 @@
...
+	int pgport = options.firstPort;
+
+	if (script == NULL)
+	{
+		log_error("Failed to allocate memory");
+		exit(EXIT_CODE_INTERNAL_ERROR);
+	}
+
+	tmux_add_command(script, "set-option -g default-shell /bin/bash");
+	tmux_add_command(script, "new -s pgautofailover");
```
JelteF, line 349: This session name should include the root I think, right now you cannot start two clusters at the same time.

src/bin/pg_autoctl/cli_do_root.c
```diff
@@ -230,6 +230,27 @@ CommandLine do_pgsetup_commands =
 					 "Manage a local Postgres setup", NULL, NULL,
 					 NULL, do_pgsetup);
 
+CommandLine do_tmux_script =
+	make_command("script",
+				 "Produce a tmux script for a demo or a test case",
+				 "[option ...]",
+				 "  --root          path where to create a cluster\n" \
+				 "  --first-port    first Postgres port to use (5500)\n" \
```
JelteF, line 238: There's no makefile variable that allows you to set this. Thus disallowing you to spawn two different clusters at the same time.

Makefile
```diff
@@ -91,7 +98,25 @@ $(PDF):
...
+tmux-script: $(TMUX_SCRIPT) ;
+
+tmux-clean:
+	pkill pg_autoctl || true
+	rm -rf $(TMUX_TOP_DIR)
+
+cluster: install tmux-clean $(TMUX_SCRIPT)
+	mkdir -p $(TMUX_TOP_DIR)/run
+	tmux start-server \; source-file $(TMUX_SCRIPT)
+	pkill pg_autoctl || true
```
JelteF, line 119: I found this trick on stackoverflow to send Ctrl+C to all panes on exit:
```
tmux set-hook before-kill-window 'run-shell "
   tmux list-panes -t \"#{window_id}\" -F \"##{pane_id}\" | xargs -I {} tmux send-keys -t {} C-c
"'
```

Source: https://unix.stackexchange.com/a/568928/138818

This would have as the advantage that running `make cluster` doesn't kill pg_autoctl processes from other clusters.

Makefile
```diff
@@ -91,7 +98,25 @@ $(PDF):
...
+         --nodes $(NODES)        \
+         --layout $(TMUX_LAYOUT) > $@
+
+tmux-script: $(TMUX_SCRIPT) ;
+
+tmux-clean:
+	pkill pg_autoctl || true
+	rm -rf $(TMUX_TOP_DIR)
+
+cluster: install tmux-clean $(TMUX_SCRIPT)
```
JelteF, line 116: Because `tmux-clean` and `$(TMUX_SCRIPT)` are targets that don't depend on eachother they are executed in parallel when you do:
```
make cluster -j9
```

This triggers errors like:
```
tmux start-server \; source-file ./tmux/script.tmux
./tmux/script.tmux: No such file or directory
```

I think it's best to simply call `make tmux-clean` and `make tmux-script` in the `cluster` target, instead of making it make dependencies. That way the order in which they are executed is clear.

JelteF (commented):
Left some comments, most of them are related to spawning two clusters at the same time not working. I guess you can postpone those for a future improvement PR, but the one about `make cluster -j9` causing issues seems common enough that it should be fixed.

src/bin/pg_autoctl/cli_do_tmux.c
DimCitus: The root could be any path on-disk. I've used the first pgport instead, which is known to be a number with not too many digits.

src/bin/pg_autoctl/cli_do_root.c
DimCitus: Fixed now, use FIRST_PGPORT from the Makefile.

Makefile
DimCitus: I see. Fixed now!

Makefile
DimCitus: When I try to integrate that thing in the tmux script we output, then I get errors that I don't have around anymore. I guess manually killing processes is a better option at the moment.

```
$ cat -n tmux/script-5500.tmux
     2	new-session -s pgautofailover-5500
     3	set-hook before-kill-window 'run-shell "tmux list-panes -t \"#{window_id}\" -F \"##{pane_id}\" | xargs -I {} tmux send-keys -t {} C-c
     4	"'
```

DimCitus (approved):

Issue #382, opened 2020-08-28
Setting replication.password doesn't set the password in the primary's database instance


soumyadeep2007:
## Description
Following the steps in the documentation [here](https://github.com/citusdata/pg_auto_failover/blob/7ac18c5d0a07fb23738032bd74e5b8896a48cd81/docs/security.rst#authentication-with-passwords) to setup password based authentication is not enough without an added:
`psql -p ${PRIMARY_PORT} postgres -c "alter user pgautofailover_replicator password 'h4ckm3m0r3';`

This `ALTER` command needs to be run before the secondary is created and started:
```
$ export PGPASSWORD=h4ckm3m0r3
$ pg_autoctl create postgres       \
       --auth scram-sha-256        \
       ...                         \
       --monitor postgres://autoctl_node:h4ckm3@monitor.host/pg_auto_failover

$ pg_autoctl config set replication.password h4ckm3m0r3
```

Log excerpt when we try to create the secondary (`pg_basebackup` fails in auth):
```
16:14:46 538293 INFO   /usr/local/pgsql/bin/pg_basebackup -w -d application_name=pgautofailover_standby_5 host=127.0.0.1 port=8800 user=pgautofailover_replicator sslmode=prefer --pgdata /home/pivotal/failover/backup/node_5 -U pgautofailover_replicator --verbose --progress -
-max-rate 100M --wal-method=stream --slot pgautofailover_standby_5                                                                                                                                                                                                                
16:14:46 538293 INFO  pg_basebackup: error: could not connect to server: FATAL:  password authentication failed for user "pgautofailover_replicator"
16:14:46 538293 ERROR Failed to run pg_basebackup: exit code 1
16:14:46 538293 ERROR Failed initialize standby server, see above for details
16:14:46 538293 ERROR Failed to transition from state "wait_standby" to state "catchingup", see above.

```

## RCA:

If we look at `pg_shadow` after `pg_autoctl config set replication.password h4ckm3m0r3`:

```
$ psql postgres -p ${PRIMARY_PORT} -c "select usename, passwd from pg_shadow where usename='pgautofailover_replicator';"
          usename          | passwd 
---------------------------+--------
 pgautofailover_replicator | 
(1 row)
```
We see that `pgautofailover_replicator` is created w/ a blank password.
In the code `keeper_cli_create_replication_user` creates the `pgautofailover_replicator` user with the password that it reads from the config during the course of a `create postgres`. At that time `replication.password` is undefined in the config.

[...]

PR #399, opened 2020-09-11
React correctly to change in replication.password

soumyadeep2007:
Addresses #382 .

Followed the steps outlined in https://github.com/citusdata/pg_auto_failover/issues/382#issuecomment-683686687

### DIFF
--- a/src/bin/pg_autoctl/cli_config.c
+++ b/src/bin/pg_autoctl/cli_config.c
@@ -36,9 +36,12 @@ static void cli_monitor_config_get(int argc, char **argv);
 
 static void cli_config_set(int argc, char **argv);
 static void cli_keeper_config_set(int argc, char **argv);
-static bool cli_keeper_config_validate_replication(KeeperConfig *config, PGSQL *pgsql);
-static bool cli_keeper_config_validate_and_commit(KeeperConfig *config, const char *key,
+static bool cli_keeper_config_validate_and_commit(KeeperConfig *config,
+												  const char *key,
 												  char *value);
+static bool cli_keeper_config_validate_and_commit_replication_primary(Keeper *keeper);
+static bool cli_keeper_config_validate_replication_secondary(Keeper *keeper,
+															 NodeAddress *primary);
 static void cli_monitor_config_set(int argc, char **argv);
 
 static CommandLine config_check =
@@ -636,94 +639,99 @@ cli_config_set(int argc, char **argv)
 
 
 /*
- * cli_keeper_config_validate_replication validates the act of setting the
- * value of "replication.password" on a given primary or secondary.
+ * cli_keeper_config_validate_and_commit_replication_primary checks that ALTER
+ * USER PASSWORD succeeds on a primary with the current replication
+ * configuration, then writes the configuration to disk. If the configuration
+ * write fails, the ALTER USER will be rolled back. If the primary is not
+ * running, validation is skipped (and this function returns true).
  */
 static bool
-cli_keeper_config_validate_replication(KeeperConfig *config, PGSQL *pgsql)
+cli_keeper_config_validate_and_commit_replication_primary(Keeper *keeper)
 {
-	Keeper keeper = { 0 };
-	NodeAddress primary = { 0 };
-	bool primaryUnreachable;
+	KeeperConfig *config = &keeper->config;
+	PGSQL *pgsql = &keeper->postgres.sqlClient;
+	bool success = false;
 
-	if (!keeper_init(&keeper, config))
-	{
-		/* already logged */
-		return false;
-	}
 
-	/* get the primary info to populate the replication source */
-	if (!monitor_get_primary(&keeper.monitor,
-							 config->formation,
-							 keeper.state.current_group,
-							 &primary))
+	/* Skip validation if the primary isn't running. */
+	if (!pg_is_running(config->pgSetup.pg_ctl, config->pgSetup.pgdata))
 	{
-		log_warn("Failed to contact monitor to get primary conninfo, skipping "
-				 "replication.password validation");
+		log_warn("Failed to connect to the primary, skipping password "
+				 "validation");
+
+		if (!keeper_config_write_file(config))
+		{
+			log_error("Failed to write pg_autoctl configuration file \"%s\"",
+					  config->pathnames.config);
+			return false;
+		}
+
 		return true;
 	}
 
 	/*
-	 * We can check if we are validating against a primary by comparing our
-	 * local nodeId with the nodeId from the primary info obtained from the
-	 * monitor.
+	 * Validate the replication password in a transaction block by setting
+	 * the password using ALTER USER PASSWORD. The transaction block will
+	 * encapsulate the subsequent change to the config file as well and we
+	 * ROLLBACK/COMMIT depending on whether both the ALTER and the config
+	 * file update are successful.
 	 */
-	if (primary.nodeId == keeper.state.current_node_id)
+	if (!pgsql_execute(pgsql, "BEGIN"))
 	{
-		/* we are the primary */
-		PGSQL *pgsql_primary = &keeper.postgres.sqlClient;
-
-		/* Skip validation if the primary isn't running. */
-		if (!pg_is_running(config->pgSetup.pg_ctl, config->pgSetup.pgdata))
-		{
-			log_warn("Failed to connect to the primary, skipping password "
-					 "validation");
-			return true;
-		}
-
-		/*
-		 * Validate the replication password in a transaction block by setting
-		 * the password using ALTER USER PASSWORD. The transaction block will
-		 * encapsulate the subsequent change to the config file as well and we
-		 * ROLLBACK/COMMIT depending on whether both the ALTER and the config
-		 * file update are successful.
-		 */
-		if (!pgsql_execute(pgsql_primary, "BEGIN"))
-		{
-			log_error("Failed to open a SQL transaction to update the "
-					  "replication password");
-			pgsql_finish(pgsql_primary);
-			return false;
-		}
-
-		if (!pgsql_set_password(pgsql_primary,
-								PG_AUTOCTL_REPLICA_USERNAME,
-								config->replication_password))
-		{
-			/*
-			 * We can safely disconnect without an explicit ROLLBACK in the
-			 * caller.
-			 */
-			log_error("Failed to set replication password, see above for details");
-			pgsql_finish(pgsql_primary);
-			return false;
-		}
-
-		/*
-		 * Copy the SQL connection state to the caller, for later
-		 * COMMIT/ROLLBACK.
-		 */
-		*pgsql = keeper.postgres.sqlClient;
-
-		return true;
+		log_error("Failed to open a SQL transaction to update the "
+				  "replication password");
+		goto finish;
 	}
 
+	if (!pgsql_set_password(pgsql,
+							PG_AUTOCTL_REPLICA_USERNAME,
+							config->replication_password))
+	{
+		log_error("Failed to set replication password, see above for details");
+		goto finish;
+	}
+
+	/* write the new configuration settings to file */
+	if (!keeper_config_write_file(config))
+	{
+		log_error("Failed to write pg_autoctl configuration file \"%s\"",
+				  config->pathnames.config);
+		goto finish;
+	}
+
+	if (!pgsql_execute(pgsql, "COMMIT"))
+	{
+		log_error("Failed to COMMIT replication password update transaction "
+				  "on the primary, see above for details.");
+		goto finish;
+	}
+
+	success = true;
+
+finish:
+	pgsql_finish(pgsql); /* will implicitly ROLLBACK in failure cases */
+	return success;
+}
+
+
+/*
+ * cli_keeper_config_validate_replication_secondary checks that the current
+ * replication configuration can be used to connect to the primary. If the
+ * primary cannot be contacted, validation is skipped (and this function
+ * returns true).
+ */
+static bool
+cli_keeper_config_validate_replication_secondary(Keeper *keeper, NodeAddress *primary)
+{
+	KeeperConfig *config = &keeper->config;
+	bool primaryUnreachable;
+
 	(void) postgres_sprintf_replicationSlotName(
-		primary.nodeId,
+		primary->nodeId,
 		config->replication_slot_name,
 		sizeof(config->replication_slot_name));
-	if (!standby_init_replication_source(&keeper.postgres,
-										 &primary,
+	if (!standby_init_replication_source(&keeper->postgres,
+										 primary,
 										 PG_AUTOCTL_REPLICA_USERNAME,
 										 config->replication_password,
 										 config->replication_slot_name,
@@ -731,13 +739,13 @@ cli_keeper_config_validate_replication(KeeperConfig *config, PGSQL *pgsql)
 										 config->backupDirectory,
 										 NULL, /* no targetLSN */
 										 config->pgSetup.ssl,
-										 keeper.state.current_node_id))
+										 keeper->state.current_node_id))
 	{
 		/* can't happen at the moment */
 		return false;
 	}
 
-	if (!pgctl_identify_system(&keeper.postgres.replicationSource, &primaryUnreachable))
+	if (!pgctl_identify_system(&keeper->postgres.replicationSource, &primaryUnreachable))
 	{
 		if (primaryUnreachable)
 		{
@@ -756,59 +764,61 @@ cli_keeper_config_validate_replication(KeeperConfig *config, PGSQL *pgsql)
 
 
 /*
- * cli_keeper_config_validate_and_commit sets the config value in memory,
- * runs any required validation, writes the updated config value to the config
- * file and then follows up with a COMMIT/ROLLBACK if the aforementioned steps
- * were performed inside a transaction boundary.
+ * cli_keeper_config_validate_and_commit runs any required validation and writes
+ * the updated configuration to disk.
  */
 static bool
 cli_keeper_config_validate_and_commit(KeeperConfig *config, const char *key, char *value)
 {
-	PGSQL pgsql = { 0 };
-	bool success = true;
+	Keeper keeper = { 0 };
+	NodeAddress primary = { 0 };
 
-	if (!keeper_config_set_setting(config, key, value))
+	if (!keeper_init(&keeper, config))
 	{
-		/* we already logged about it */
+		/* already logged */
 		return false;
 	}
 
-	if (keeper_config_setting_requires_validation(key) &&
-		!cli_keeper_config_validate_replication(config, &pgsql))
+	/* get the primary info from the monitor */
+	if (!monitor_get_primary(&keeper.monitor,
+							 config->formation,
+							 keeper.state.current_group,
+							 &primary))
 	{
-		/* we already logged about it */
+		log_warn("Failed to contact monitor to get primary conninfo, skipping "
+				 "replication.password validation");
+		goto write;
+	}
+
+	/*
+	 * We can check if we are validating against a primary by comparing our
+	 * local nodeId with the nodeId from the primary info obtained from the
+	 * monitor.
+	 */
+	if (keeper.state.current_node_id == primary.nodeId)
+	{
+		/*
+		 * The primary case handles the configuration write itself. Return
+		 * directly.
+		 */
+		return cli_keeper_config_validate_and_commit_replication_primary(&keeper);
+	}
+
+	if (!cli_keeper_config_validate_replication_secondary(&keeper, &primary))
+	{
+		/* already logged */
 		return false;
 	}
 
-	/* first write the new configuration settings to file */
+write:
 	if (!keeper_config_write_file(config))
 	{
 		log_error("Failed to write pg_autoctl configuration file \"%s\"",
 				  config->pathnames.config);
-		success = false;
+		return false;
 	}
 
-	/*
-	 * COMMIT/ROLLBACK any temporary state changes to the database. A
-	 * COMMIT/ROLLBACK would have been necessary if the validation step
-	 * initiated a transaction. In this case, we use the same client to perform
-	 * the command as the one that started the transaction.
-	 */
-	if (pgsql.connection != NULL)
-	{
-		char *cmd = success ? "COMMIT" : "ROLLBACK";
-
-		if (!pgsql_execute(&pgsql, cmd))
-		{
-			log_error("Failed to %s replication password update transaction "
-					  "on the primary, see above for details.", cmd);
-			success = false;
-		}
-
-		pgsql_finish(&pgsql);
-	}
-
-	return success;
+	return true;
 }
 
 
@@ -830,9 +840,30 @@ cli_keeper_config_set(int argc, char **argv)
 		/* we print out the value that we parsed, as a double-check */
 		char value[BUFSIZE];
 
-		if (!cli_keeper_config_validate_and_commit(&config, argv[0], argv[1]))
+		if (!keeper_config_set_setting(&config,
+									   argv[0],
+									   argv[1]))
 		{
 			/* we already logged about it */
+			exit(EXIT_CODE_BAD_CONFIG);
+		}
+
+		/* first write the new configuration settings to file */
+		if (keeper_config_setting_requires_validation(argv[0]))
+		{
+			if (!cli_keeper_config_validate_and_commit(&config,
+													   argv[0],
+													   argv[1]))
+			{
+				/* we already logged about it */
+				exit(EXIT_CODE_BAD_CONFIG);
+			}
+		}
+		else if (!keeper_config_write_file(&config))
+		{
+			log_fatal("Failed to write pg_autoctl configuration file \"%s\", "
+					  "see above for details",
+					  config.pathnames.config);
 			exit(EXIT_CODE_BAD_CONFIG);
 		}
 
--- a/src/bin/pg_autoctl/cli_config.c
+++ b/src/bin/pg_autoctl/cli_config.c
@@ -36,6 +36,9 @@ static void cli_monitor_config_get(int argc, char **argv);
 
 static void cli_config_set(int argc, char **argv);
 static void cli_keeper_config_set(int argc, char **argv);
+static bool cli_keeper_config_validate_replication(KeeperConfig *config, PGSQL *pgsql);
+static bool cli_keeper_config_validate_and_commit(KeeperConfig *config, const char *key,
+												  char *value);
 static void cli_monitor_config_set(int argc, char **argv);
 
 static CommandLine config_check =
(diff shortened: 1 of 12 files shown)


src/bin/pg_autoctl/state.h
```diff
@@ -63,6 +63,10 @@ typedef enum
  */
 #define state_matches(x, y) (x == ANY_STATE || y == ANY_STATE || x == y)
 
+#define IS_STANDBY(state) (state_matches((state), CATCHINGUP_STATE) || \
+						   state_matches((state), SECONDARY_STATE) || \
+						   state_matches((state), MAINTENANCE_STATE))
+
```
DimCitus, line 69: As mentioned before, I don't think we need that. If we do though, we have more states that belong to a standby...

src/bin/pg_autoctl/keeper_config.c
```diff
@@ -579,6 +579,14 @@ keeper_config_set_setting(KeeperConfig *config,
 						  pg_is_not_running_is_ok))
 		{
 			config->pgSetup = pgSetup;
+			if (streq(path, "replication.password") &&
+				!keeper_config_validate_replication_password(config, value))
```
DimCitus, line 583: The `keeper_config_set_setting` is only called from `cli_keeper_config_set` which is invoked when using the command line `pg_autoctl config set section.option value`. It is not used when loading values from the configuration file if it's been edited “at rest”, and it is not used at reload of the configuration file in the main server.

src/bin/pg_autoctl/keeper_config.c
```diff
@@ -814,3 +822,87 @@ keeper_config_update_with_absolute_pgdata(KeeperConfig *config)
...
+
+/*
+ * keeper_config_validate_replication_password gates the update of the replication
+ * password in the config file if the server is a standby and it is running. If
+ * the server is not running, validation will be performed at server restart. If
+ * the server is a primary, the ensuing reload request will lead to the node
+ * active process updating the password in the database.
+ */
+bool
+keeper_config_validate_replication_password(KeeperConfig *config, const char *password)
```
DimCitus, line 835: This looks to me like a modularity violation. The `keeper_config.c` file is supposed to deal with the configuration file and data structure, not with the actual Postgres instance that we're managing. That code belongs to `primary_standby.c` on a first reading of it, with maybe a wrapper to call it from `keeper.c`.

DimCitus (changes requested):
Thanks for this work! I think it's a good start, and more work is needed in terms of code architecture and integration. Which is fine! I have left some comments to help drive you towards completion hope it helps! Questions welcome of course.

soumyadeep2007:
Thank you for the detailed feedback! I will address it.

soumyadeep2007:
I rebased the branch against latest master and then made a separate commit (to be squashed later) to address the feedback so far.

I had a few questions and comments about the changes you requested and my latest commit to address those:

* `keeper_ensure_configuration` already has logic to validate the conn string w/ the new password via `IDENTIFY_SYSTEM`, updating `primary_conninfo` and restarting the standby. So my commit just leverages that w/o any additional validation. I simply ensure that the `replicationSource` supplied to `pg_setup_standby_mode` has the right primary node info.

* I have gotten rid of the changes to `keeper_config_set_setting`. Now the flow is that it will set the password in the config unhindered by validation (equivalent to someone setting it in the file at rest) -> reload signal -> password change will be acted upon on reload of the keeper config.
By moving the handling of the change in replication password to `keeper_ensure_configuration`, we are accepting the change in replication password in the config file and in the keeper's in-memory config data structure (even if the password is "bad" in the case of a standby) - **something I thought we originally wanted to guard with validation. Is this not a requirement?**
However,  we do ensure that the bad password never makes it to the replication settings in the conf files.
Maybe this isn't a bad pattern?

* When we are in `keeper_ensure_configuration`, we have no guarantee that the server is running. If it is not, we retry the config reload later. Retries of reload is introduced in this config by looking at the return value of a reload hook and by returning false when we can't act upon a change in replication password or any other failure. Maybe this change is too invasive? 
**Alternative approach:**
Instead, maybe we can have the retry logic by using a `replication_password_changed` global, set that in `keeper_config_accept_new` and use it to call `keeper_ensure_configuration` inside `keeper_node_active`? We can clear the global on a successful catalog update (for a primary) and after a  connect->update_primary_conninfo->restart (for a standby) to break the retry cycle?

* What should be the behavior when we supply a bad replication password to the standby? Should we infinitely retry or shutdown everything after a certain number of retries?

I somehow don't feel good about the changes. I probably have several gaps in understanding of the requirements.

src/bin/pg_autoctl/keeper.c
```diff
@@ -947,6 +946,28 @@ keeper_ensure_configuration(Keeper *keeper, bool postgresNotRunningIsOk)
 			}
 		}
 	}
+	else
```
DimCitus, line 949: We have many states where we don't want to ALTER ROLE PASSSWORD, such as MAINTENANCE and its related states. I think this should not be an else, but rather its own `if`. I think the list of states in which we want to act on the password change are SINGLE, WAIT_PRIMARY, JOIN_PRIMARY, PRIMARY, APPLY_SETTINGS.

src/bin/pg_autoctl/keeper.c
```diff
@@ -895,7 +894,7 @@ keeper_ensure_configuration(Keeper *keeper, bool postgresNotRunningIsOk)
 
 		/* prepare a replicationSource from the primary and our SSL setup */
 		if (!standby_init_replication_source(postgres,
-											 NULL, /* primaryNode is done */
+											 &(upstream->primaryNode), /* primaryNode is done */
```
DimCitus, line 897: Keeping the comment here makes this very confusing. Can we have a new comment?

src/bin/pg_autoctl/keeper.c
```diff
@@ -2100,12 +2122,18 @@ keeper_reload_configuration(Keeper *keeper, bool firstLoop)
...
+		}
 	}
 
-	/* we're done reloading now. */
-	asked_to_reload = 0;
+	/*
+	 * If there are no errors reloading the configuration, we are done.
+	 * Otherwise, we will have to retry the reload.
+	 */
+	asked_to_reload = (errors > 0);
```
DimCitus, line 2138: I'm not fond of eternal reload retries, maybe we need a counter and a MaxR/MaxT strategy as in other places. That said, for a first PR, we may be able to accept a reload-forever policy here. I'm worried about picking a password that we would always fail to SET. Is that possible?

src/bin/pg_autoctl/keeper_config.c
```diff
@@ -578,7 +578,6 @@ keeper_config_set_setting(KeeperConfig *config,
 						  missing_pgdata_is_ok,
 						  pg_is_not_running_is_ok))
 		{
-			config->pgSetup = pgSetup;
```
DimCitus, line 581: Can you explain why this line is removed in the PR?

src/bin/pg_autoctl/pgsql.c
```diff
@@ -1601,6 +1601,30 @@ pgsql_set_default_transaction_mode_read_write(PGSQL *pgsql)
...
+/*
+ * pgsql_set_password sets the password for the given username.
+ */
+bool
+pgsql_set_password(PGSQL *pgsql, const char *username, const char *password)
+{
+	char command[BUFSIZE];
+
+	sformat(command, sizeof(command), "ALTER USER %s PASSWORD '%s'", username,
+			IS_EMPTY_STRING_BUFFER(password) ? "''" : password);
```
DimCitus, line 1613: I think we need to either use `PQescapeLiteral` as documented at https://www.postgresql.org/docs/current/libpq-exec.html#LIBPQ-EXEC-ESCAPE-STRING ; or even better use an actual SQL parameter to the command here. Because `ALTER USER` is a DDL, we might have to do like we do in `pgsql_create_user` to escape the identifiers and the literals. Please copy what we do there.

DimCitus:
Hi @soumyadeep2007 ; thanks for this new version of the PR. There's still more work to do. Let's also try to review the high level approach here:

> * `keeper_ensure_configuration` already has logic to validate the conn string w/ the new password via `IDENTIFY_SYSTEM`, updating `primary_conninfo` and restarting the standby. So my commit just leverages that w/o any additional validation. I simply ensure that the `replicationSource` supplied to `pg_setup_standby_mode` has the right primary node info.

Sounds good, thanks!

> * I have gotten rid of the changes to `keeper_config_set_setting`. Now the flow is that it will set the password in the config unhindered by validation (equivalent to someone setting it in the file at rest) -> reload signal -> password change will be acted upon on reload of the keeper config.

Thanks.

>   By moving the handling of the change in replication password to `keeper_ensure_configuration`, we are accepting the change in replication password in the config file and in the keeper's in-memory config data structure (even if the password is "bad" in the case of a standby) - **something I thought we originally wanted to guard with validation. Is this not a requirement?**

The validation we need is that, on the one hand the primary is going to accept the password, and the only way to ensure that is by running the ALTER ROLE SET PASSWORD command; and on the other hand that the secondary can connect to the primary when using the new password, and that's done in `pg_setup_standby_mode` as you note later, using IDENTIFY_SYSTEM.

The way it's now done, we have to read the new password from the configuration file anyway, before to impact the system. That's too late to invalidate the change. So I don't see the point in the guard anymore. 

The guard would be useful before we set the password in the configuration, which could be done on the primary where we can just connect to the database from the `pg_autoctl config set replication.password` command. Then when the command is successful, we could update the configuration file. If we could be smart about using a Postgres transaction to sync the file edits with the password validation, that would be good then. See `keeper_register_and_init` for an example when we do that already.

I think that would be a better approach on the primary to implement the password change in the interactive process rather than signal the background process for a reload.

On a secondary server though, we have to ask the background process to orchestrate the change of the replication parameters for Postgres and restart Postgres. In Postgres 13 we might be able to just reload when the primary_conninfo is changed, that said, and we now have official support for Postgres 13, by the way.

Again, the validation on the standby could be done before we enter the `reload` signal processing on the background process. In the interactive process, we can try to connect using the password and the IDENTIFY_SYSTEM command and only change the configuration file and reload when that succeeds. I'm not completely sure that we want to impose an ordering in the password change here though.


> * When we are in `keeper_ensure_configuration`, we have no guarantee that the server is running. If it is not running or there is any sort of error - such as a bad password, we retry the config reload later. Retries of reload is introduced in this config by looking at the return value of a reload hook and by returning false when we can't act upon a change in replication password or any other failure (from `keeper_reload_configuration`). Maybe this 
change is too invasive?

Yeah I'm not sure I like the infinite-eternal-retry-reload thing.

>   **Alternative approach:**
>   Instead, maybe we can have the retry logic by using a `replication_password_changed` global, set that in `keeper_config_accept_new` and use it to call `keeper_ensure_configuration` inside `keeper_node_active`? We can clear the global on a successful catalog update (for a primary) and after a  connect->update_primary_conninfo->restart (for a standby) to break the retry cycle?

I don't like that approach either that said. The usual pattern to guard is: what if we restart `pg_autoctl`, then our internal global variable is not helping anymore. In the case of the reload, at least we broke from the hell loop.

> * What should be the behavior when we supply a bad replication password to the standby? Should we infinitely retry or shutdown everything after a certain number of retries?

Ideally we would reject it because the interactive command failed to connect to the primary. We would use a retry policy that continues retrying once in a while for a long time, so as to avoid requiring perfect timing of the operation on every node --- that's almost impossible to achieve unless you serialise, but then you need to figure out which node currently is the primary and cross your fingers hard that it doesn't change in the middle of the process.

> I somehow don't feel good about the changes. I probably have several gaps in understanding of the requirements.

I am not certain how much this comment is going to help. Let's discuss more and maybe organise a meeting if you feel like it's necessary.

src/bin/pg_autoctl/keeper.c
jchampio: > I'm worried about picking a password that we would always fail to SET. Is that possible?

Yes, if for example the [`passwordcheck` contrib module](https://www.postgresql.org/docs/13/passwordcheck.html) is installed and doesn't like the password.

soumyadeep2007:
@DimCitus 

w/ @jchampio 
Addressed much of the feedback you had for us.

We haven't yet done anything to change the retry logic on the client side `pg_autoctl set` process (the default interactive retry policy kicks in) and the retry logic for the keeper config reload (now there are infinite retries)

src/bin/pg_autoctl/cli_config.c
```diff
@@ -650,23 +801,12 @@ cli_keeper_config_set(int argc, char **argv)
 		/* we print out the value that we parsed, as a double-check */
 		char value[BUFSIZE];
 
-		if (!keeper_config_set_setting(&config,
-									   argv[0],
-									   argv[1]))
+		if (!cli_keeper_config_validate_and_commit(&config, argv[0], argv[1]))
```
DimCitus, line 804: Are we running the new validation code for any setting that can be set through `pg_autoctl config set` now, rather than only the replication password?

src/bin/pg_autoctl/cli_config.c
```diff
@@ -632,6 +632,157 @@ cli_config_set(int argc, char **argv)
 }
 
 
+static bool
+cli_keeper_config_validate_replication(KeeperConfig *config, PGSQL *pgsql)
```
DimCitus, line 636: Missing declaration, missing function-level comments.

src/bin/pg_autoctl/cli_config.c
```diff
@@ -632,6 +632,157 @@ cli_config_set(int argc, char **argv)
...
+	}
+
+	/* get the primary info to populate the replication source */
+	if (!monitor_get_primary(&keeper.monitor,
+							 config->formation,
+							 keeper.state.current_group,
+							 &primary))
+	{
+		log_warn("Monitor could not be contacted for primary conninfo, skipping "
+				 "password validation");
```
DimCitus, line 655: Usually our error messages start with "Failed to..." to make it explicit from the first words that it is an unexpected failure.

src/bin/pg_autoctl/cli_config.c
```diff
@@ -632,6 +632,157 @@ cli_config_set(int argc, char **argv)
...
+		return true;
+	}
+
+	if (primary.nodeId == keeper.state.current_node_id)
```
DimCitus, line 659: It's a little strange to decide if we're a primary or not with that check, but given that we need the primary information for the connection string later, why not. I guess I'm saying that a comment about this choice would make sense.

src/bin/pg_autoctl/cli_config.c
```diff
@@ -632,6 +632,157 @@ cli_config_set(int argc, char **argv)
...
+	{
+		/* we are the primary */
+		PGSQL *pgsql = &keeper.postgres.sqlClient;
+
+		/* Skip validation if the primary isn't running. */
+		if (!pg_is_running(config->pgSetup.pg_ctl, config->pgSetup.pgdata))
+		{
+			log_warn("Primary could not be contacted, skipping password "
+					 "validation");
```
DimCitus, line 668: Are we skipping the password validation, or the while `ALTER ROLE` operation altogether, and skipping the only actual change that we wanted to implement?

src/bin/pg_autoctl/cli_config.c
```diff
@@ -632,6 +632,157 @@ cli_config_set(int argc, char **argv)
...
+			pgsql_finish(pgsql);
+			return false;
+		}
+
+		if (!pgsql_set_password(pgsql,
+								PG_AUTOCTL_REPLICA_USERNAME,
+								config->replication_password))
+		{
+			log_error("Failed to set replication password, see above for details");
+			pgsql_finish(pgsql);
```
DimCitus, line 692: Disconnecting without issuing ROLLBACK first might be a good approach, though it needs at least some comments about it.

src/bin/pg_autoctl/cli_config.c
```diff
@@ -632,6 +632,157 @@ cli_config_set(int argc, char **argv)
...
+			return false;
+		}
+
+		/*
+		 * Copy the SQL connection state to the caller, for later
+		 * COMMIT/ROLLBACK.
+		 */
+		*pgsql = keeper.postgres.sqlClient;
```
DimCitus, line 700: I think this is a big no: having code that handles an SQL transaction from multiple functions, each doing early `return false`, without proper documentation about it, and when it's very unclear that this level of complexity is needed at all... please let's try another approach.

I think this whole code be way easier to read/write/maintain when done the following way:
  - check what setting is being set from the command line, and when `replication.password` invoke a specific function,
  - the `pg_autoctl set replication.password` specific function either forwards the change to Postgres, on the primary, or change the `primary_conninfo` connection string, on a secondary ; again, I think this first function should only recognise the current situation and then delegate to a specialised function,
  - now we have a function that has to ALTER ROLE on the primary and that's it,
  - and another function that has to use `pgctl_identify_system` to validate the new connection string, and when successful (maybe after some retries), then restart the secondary with the new replication settings.

Your implementation seems to be more complex than that in order to enable a smart handling of keeping the success of the transaction (ALTER ROLE on the primary, establishing a replication connection on the secondary) and the actual editing of the configuration file. What's complex is to be able to ROLLBACK the ALTER ROLE command when the configuration file edit failed. For that, I think it's fair to say that the main CLI entry point should delegate the configuration file handling to the `pg_autoctl set replication.password` specific function(s), and that the BEGIN/COMMIT/ROLLBACK and file update is only necessary in the sub-function that takes care of the primary. There's no Postgres transaction in the control flow of the secondary side of things.

src/bin/pg_autoctl/cli_config.c
```diff
@@ -632,6 +632,157 @@ cli_config_set(int argc, char **argv)
...
+	}
+
+	if (!pgctl_identify_system(&keeper.postgres.replicationSource, &primaryUnreachable))
+	{
+		if (primaryUnreachable)
+		{
+			log_warn("Primary could not be contacted, skipping password "
+					 "validation");
+			return true;
+		}
```
DimCitus, line 731: It seems a little surprising that `pgctl_identify_system` would also PQPing the primary server when it fails to establish a replication connection, so I think this responsibility should move up to the caller.

src/bin/pg_autoctl/cli_config.c
```diff
@@ -632,6 +632,157 @@ cli_config_set(int argc, char **argv)
...
+				  "password. See above for details.");
+		return false;
+	}
+
+	return true;
+}
+
+
+static bool
+cli_keeper_config_validate_and_commit(KeeperConfig *config, const char *key, char *value)
```
DimCitus, line 743: Missing function signature (up in the file) and function-level comments.

src/bin/pg_autoctl/cli_config.c
```diff
@@ -632,6 +632,157 @@ cli_config_set(int argc, char **argv)
...
+	if (!keeper_config_set_setting(config, key, value))
+	{
+		/* we already logged about it */
+		return false;
+	}
+
+	if (!cli_keeper_config_validate_replication(config, &pgsql))
+	{
+		return false;
+	}
```
DimCitus, line 757: This only needs to happen when setting `replication.password`, is that right?

src/bin/pg_autoctl/cli_config.c
```diff
@@ -650,23 +801,12 @@ cli_keeper_config_set(int argc, char **argv)
...
 
-		/* first write the new configuration settings to file */
-		if (!keeper_config_write_file(&config))
-		{
-			log_fatal("Failed to write pg_autoctl configuration file \"%s\", "
-					  "see above for details",
-					  config.pathnames.config);
-			exit(EXIT_CODE_BAD_CONFIG);
-		}
-
```
DimCitus, line 669: Again, that change seems too invasive. The special processing of the `replication.password` setting should happen entirely elsewhere, and this code left unchanged for the other settings.

src/bin/pg_autoctl/keeper.c
```diff
@@ -947,6 +946,32 @@ keeper_ensure_configuration(Keeper *keeper, bool postgresNotRunningIsOk)
...
+			 state->current_role == WAIT_PRIMARY_STATE ||
+			 state->current_role == JOIN_PRIMARY_STATE ||
+			 state->current_role == PRIMARY_STATE ||
+			 state->current_role == APPLY_SETTINGS_STATE)
+	{
+		/*
+		 * On the primary, the keeper's replication password may have been set
+		 * in the keeper config while the server was stopped. So, set the
+		 * password in the server catalog.
+		 */
```
DimCitus, line 959: It would be better to only call `ALTER ROLE ... PASSWORD` in the three following situations:

  - when we start Postgres for the first time in `pg_autoctl` (because it might have been changed at rest),
  - or the first time after an interactive password change,
  - or any time we check that Postgres is running if we failed to update the password on the previous occasion.

To enable that,  we could have a cache of the password in-memory. The cache should probably live in the `LocalPostgresServer` data structure, invalidated at reload time (first "reload" to set the value, when the cache is unset, and we know that thanks to IS_EMPTY_STRING_BUFFER, then only when different from the existing value).

It looks like the `pgsql_set_password` call could then happen in `ensure_postgres_service_is_running` in the case when the function had to actually restart Postgres, and that an update is expected.  Because `ensure_postgres_service_is_running` only has access to the `LocalPostgresServer` rather than the whole `Keeper`, we might need to also maintain a boolean in `LocalPostgresServer ` to drive the decision to call to `pgsql_set_password`. The boolean would be maintained in the `keeper_ensure_configuration` when we also update our password cache.

Note that we would still keep retrying to set the password an infinite amount of time with that scheme, though it's now easy to solve that problem, adding a counter to the password change tracking boolean. We could decide that after N password change attempts we bail out, but I'm not sure what we should do in that case. The user certainly needs to find another password that Postgres will like better. At least the `pgsql_set_password` retry mechanism would be separate from the reload signal handling mechanism and would not make it so that we keep reloading the whole setup again and again. When someone is in the middle of editing their pg_auto_failover setup, it's best to not have rewrites of the configuration file happening in the background.

src/bin/pg_autoctl/keeper.c
DimCitus: See another comment in this PR for an alternative approach that allows removing the reload-again-and-again hack, which I think is a terrible idea now that I realise that the first thing we do is `keeper_config_update` which calls `keeper_config_write_file`. That would prevent users from doing any other edits to the configuration file, or they would need to time it perfectly with the reload signal.

src/bin/pg_autoctl/pgctl.c
```diff
@@ -1989,7 +1989,7 @@ prepare_conninfo_sslmode(PQExpBuffer buffer, SSLOptions sslOptions)
  * the IDENTIFY_SYSTEM command to check that HBA is ready.
  */
 bool
-pgctl_identify_system(ReplicationSource *replicationSource)
+pgctl_identify_system(ReplicationSource *replicationSource, bool *unreachable)
```
DimCitus, line 1992: As noted in another comment I don't think that's the best way to implement that check. Let's move the ping calls to the caller.

src/bin/pg_autoctl/pgsql.c
```diff
@@ -1608,6 +1608,85 @@ pgsql_set_default_transaction_mode_read_write(PGSQL *pgsql)
...
+		log_error("Failed to set password for username \"%s\": %s", userName,
+				  PQerrorMessage(connection));
+
+		success = false;
+		goto cleanup;
+	}
+
+	command = createPQExpBuffer();
+	appendPQExpBuffer(command, "ALTER USER %s", escapedUsername);
+	appendPQExpBuffer(command, " PASSWORD %s", escapedPassword);
```
DimCitus, line 1652: We would either use a PQExpBuffer and re-use the same escaped string memory area for both the username and the password, or use a string buffer of BUFSIZE characters and then use `sformat`. Given the code here, I think we should use the second approach.

src/bin/pg_autoctl/pgsql.c
```diff
@@ -2562,17 +2641,35 @@ typedef struct IdentifySystem
...
+		{
+			/*
+			 * Perform a PQping to determine why the connection open attempt was
+			 * unsuccessful.
+			 */
+			pingResult = PQping(pgsql->connectionString);
+			if (pingResult == PQPING_NO_RESPONSE)
+			{
+				*unreachable = true;
+			}
```
DimCitus, line 2670: Let's move the PQPing responsibility to the caller. And then think about unstable networks. See also the `PGConnStatus status` field in `PGSQL`, which might already solve your problem in a way that the caller can distinguish (as in https://github.com/citusdata/pg_auto_failover/blob/master/src/bin/pg_autoctl/service_keeper.c#L605 for instance):

``` c
/*
 * Allow higher level code to distinguish between failure to connect to the
 * target Postgres service and failure to run a query or obtain the expected
 * result. To that end we expose PQstatus() of the connection.
 *
 * We don't use the same enum values as in libpq because we want to have the
 * unknown value when we didn't try to connect yet.
 */
typedef enum
{
	PG_CONNECTION_UNKNOWN = 0,
	PG_CONNECTION_OK,
	PG_CONNECTION_BAD
} PGConnStatus;
```

DimCitus (changes requested):
Hi folks,

We're making good progress here! I didn't like several of the approaches in the current implementation of the PR, and left very detailed comments about that. It looks like the strategy is beginning to make sense, it might all be about tactical editing now!

src/bin/pg_autoctl/cli_config.c
soumyadeep2007: Apologies. We forgot to guard the replication validation check like this:

```diff
-	if (!cli_keeper_config_validate_replication(config, &pgsql))
+	if (streq(key, "replication.password") && !cli_keeper_config_validate_replication(config, &pgsql))
	{
```
We will correct this.

jchampio: I remember why we did this -- the other configuration settings seemed to be validated as a whole, and so any co-dependencies between them (say, hostname and port) would be handled. We were worried that setting some _other_ configuration could effectively invalidate the password (especially given the reliance on HBA), and that we wouldn't end up discovering the failure until it was already committed. Is the cost of running the validation too high?

DimCitus: Good point, we should re-compute the primary_conninfo connection string for every config setting that would impact it. In practice it's mostly SSL and password. 

I'm not so sure about when a node's hostname and port are edited, we could also re-check although I suppose it comes with network updates that might have required a system / pg_autoctl restart already, and the change needs to be done on the monitor and then all other nodes in cascade, which we already take into account. Making sure that we test the change locally at the right time (once the HBA have been edited everywhere) is going to be rather complex, if not impossible. I would not sweat over it.

Doing the primary conninfo change for all parameters is too much in my view, yes.

src/bin/pg_autoctl/cli_config.c
soumyadeep2007: w/ @jchampio
If the `ALTER ROLE` isn't issued (which means it was not possible to perform validation on the primary), we still write the config value to the file..so that the password change can be applied to the primary on config reload / restart in `keeper_ensure_configuration()`.

src/bin/pg_autoctl/pgsql.c
jchampio: `PG_CONNECTION_BAD` is returned for both the "we couldn't make a network connection at all" and the "we connected but then your password was rejected" cases, unfortunately. `PQping()` differentiates between the two cases with `PQPING_NO_RESPONSE` and `PQPING_OK`.

jchampio: >think about unstable networks

In what way? The underlying connection code is already retrying several times. Or is there another angle you're referring to?

jchampio: >Let's move the PQPing responsibility to the caller

I think this forces us to duplicate most of the dance for formulating a replication conninfo string, but we may be able to get a good API out of it that can be used in both places. We'll give it a shot.

DimCitus: Thanks for the explanation about needing PQPing to distinguish between the two cases. Something to the same effect should be added as a comment in the code. Unstable networks: failed to connect, then successful ping, because it's all up again. Maybe if we can ping we should retry connecting? I don't know.

About the API I guess we have the connection string in the caller, but I only say that now because I see that the change happens in pgsql.c and I don't remember that that's where we build connection strings.

src/bin/pg_autoctl/cli_config.c
soumyadeep2007: w/ @jchampio

> I think this whole code be way easier to read/write/maintain when done the following way:

We did try that approach _(We can try the approach again and present what it looks like)._ However we had a few pain points:

* There was a lot of code duplication (a lot of the keeper setup code and `keeper_config_write_file`) between the function to handle the primary case and the function to handle the secondary case.

* In addition to the code duplication, if we want to validate the entire config holistically (one config value change might affect other config values adversely as @jchampio mentions [here](https://github.com/citusdata/pg_auto_failover/pull/399#discussion_r513629674)), then we can't have a specialized function.

> and another function that has to use pgctl_identify_system to validate the new connection string, and when successful (maybe after some retries), then restart the secondary with the new replication settings.

The subsequent config reload (`cli_pg_autoctl_reload()` call) would lead to `keeper_ensure_configuration()` restarting the standby anyway. Do we want the replication settings update in the recovery conf file and an explicit restart from the client? Is this to do with ensuring that the effect of a `pg_autoctl config set` command is immediately enforced?

src/bin/pg_autoctl/pgsql.c
jchampio: >Unstable networks: failed to connect, then successful ping, because it's all up again.

Yeah, this is IMO a deficiency in the libpq API. I want to know why a specific connection failed without having to issue a second connection, which might fail (or succeed) in a different way. But as far as I can tell, the current API does not give us that option.

soumyadeep2007: > About the API I guess we have the connection string in the caller, but I only say that now because I see that the change happens in pgsql.c and I don't remember that that's where we build connection strings.

The connection string is built inside `pgctl_identify_system` and is hidden from its callers. So unfortunately, we don't have access to the connection string in `cli_keeper_config_validate_replication` where we have to distinguish between the two cases of connection failure.

DimCitus: I see now. Okay I think we should extend what we're doing in `pgsql_retry_open_connection` in order to keep the information we gather when using PGPing there. This retry mechanism is already being used for a connection of type PGSQL_CONN_UPSTREAM as used in `pgctl_identify_system`. 

So what about extending the `PGConnStatus` enum with some of the `PQPING_` diagnostic information and keeping track of what we know there? Or maybe we need a new enum for `PGConnPingStatus` and a new member in `struct PGSQL` to keep track of that?

That way we could differentiate between `PQPING_NO_RESPONSE` and `PQPING_OK` I suppose, without having to implement the PQPing retry mechanism again.

src/bin/pg_autoctl/cli_config.c
DimCitus: I think some code duplication is much less of a problem than complex transaction management in different independent functions with early exists for error handling. Let's make it right, and if we can get rid of code duplication in a simple way, we can do it later. Now it's impossible for me to follow all the possible cases where we might exit with a transaction left open.

About configuration settings that have an effect on the primary_conninfo, please list them. I still don't think it would be wise to recompute and test our primary_conninfo when the user changes our `network_partition_timeout` for instance.

Then about restarting Postgres, you're right, that sounds like a job for the background process upon processing a RELOAD signal.

src/bin/pg_autoctl/keeper.c
DimCitus: In that instance I specifically meant `ensure_postgres_service_is_running` as written later.

jchampio:
Added a squash commit to the end of the patchset. This fixes an issue with our transaction handling in the primary case (the connection was previously being closed before a `COMMIT` could be correctly issued), adjusts `test_auth` with the new behavioral guarantees, and implements the duplicated logic for the primary and standby code paths.

Several action items mentioned above have changed considerably, so let's regroup. Here's where I think we are in the roadmap:
- Primary/standby logic is no longer combined into a single code path.
- Ping logic has been pulled inside connection establishment. It's not possible to remove the `*unreachable` output parameter from `pgctl_identify_system()` without a bigger API change, unfortunately, as the caller isn't required to have a `PGSQL` object ready by the time that function is called.
- A password cache system has been set up, which I believe (but am not entirely sure) meets the three requirements set out [here](https://github.com/citusdata/pg_auto_failover/pull/399#discussion_r512157463).
- Testing is sparse. We're having trouble implementing race-free tests due to all of the moving pieces here (we have to get the tests to wait for state changes out of, and back to, the `secondary` state, and there aren't currently any test primitives for that).

Unfortunately I'm not able to mark conversations as resolved. 😁 Are there other outstanding items I've missed above that are still needed?

PR #402, opened 2020-09-15
When secondary, don't ensure current state before a transition.

DimCitus:
Getting out of the secondary state means we should not ensure the secondary
state anymore, it's already too late. We had cases of Postgres being
restarted because of that, and if that happens when we don't have a recovery
setup in place, then Postgres restarts as a primary. Then it can't be a
secondary again anymore, WAL are "corrupted".

### DIFF
--- a/src/bin/pg_autoctl/keeper.c
+++ b/src/bin/pg_autoctl/keeper.c
@@ -171,6 +171,17 @@ keeper_should_ensure_current_state_before_transition(Keeper *keeper)
 		return false;
 	}
 
+	if (keeperState->current_role == SECONDARY_STATE &&
+		keeperState->assigned_role != SECONDARY_STATE)
+	{
+		/*
+		 * We might have a different primary server to reconnect to, or be
+		 * asked to report lsn, etc. Ensuring the secondary state does not
+		 * sound productive there.
+		 */
+		return false;
+	}
+
 	/* in all other cases, yes please ensure the current state */
 	return true;
 }
--- a/src/bin/pg_autoctl/primary_standby.c
+++ b/src/bin/pg_autoctl/primary_standby.c
@@ -1132,6 +1132,13 @@ standby_follow_new_primary(LocalPostgresServer *postgres)
 
 	log_info("Follow new primary %s:%d", primaryNode->host, primaryNode->port);
 
+	if (!pgctl_identify_system(replicationSource))
+	{
+		log_error("Failed to establish a replication connection "
+				  "to the new primary, see above for details");
+		return false;
+	}
+
 	/* cleanup our existing standby setup, including postgresql.auto.conf */
 	if (!pg_cleanup_standby_mode(pgSetup->control.pg_control_version,
 								 pgSetup->pg_ctl,
@@ -1284,6 +1291,13 @@ standby_restart_with_current_replication_source(LocalPostgresServer *postgres)
 	PGSQL *pgsql = &(postgres->sqlClient);
 	PostgresSetup *pgSetup = &(postgres->postgresSetup);
 	ReplicationSource *replicationSource = &(postgres->replicationSource);
+
+	if (!pgctl_identify_system(replicationSource))
+	{
+		log_error("Failed to establish a replication connection "
+				  "to the primary node, see above for details");
+		return false;
+	}
 
 	/* cleanup our existing standby setup, including postgresql.auto.conf */
 	if (!pg_cleanup_standby_mode(pgSetup->control.pg_control_version,


src/bin/pg_autoctl/keeper.c
```diff
@@ -171,6 +171,17 @@ keeper_should_ensure_current_state_before_transition(Keeper *keeper)
 		return false;
 	}
 
+	if (keeperState->current_role == SECONDARY_STATE &&
```
onderkalaci, line 174: After this change, we won't be getting into `keeper_ensure_current_state()` and hence not calling `keeper_maintain_replication_slots()`, that might be OK because we call it on `catching up-> secondary` transition anyway. 

If so, do we still need the `case SECONDARY_STATE:` in `keeper_ensure_current_state`?

Also, on `keeper_fsm_step` we call `keeper_ensure_current_state()` without `keeper_should_ensure_current_state_before_transition()` check. Is that intended?

onderkalaci (commented):
I've been  thinking on this PR for a while now and still cannot fully understand if what it solves is the root cause or its effect.

Ok, we don't  strictly need to check if the secondary is running (or restarting) before it transitions to another one, but should not be a problem either.

>  We had cases of Postgres being
restarted because of that, and if that happens when we don't have a recovery
setup in place, then Postgres restarts as a primary

How wouldn't we have the recovery setup while still `keeperState->current_role == SECONDARY_STATE`? Do we know the set of transitions that would lead to this?

src/bin/pg_autoctl/keeper.c
DimCitus: We still ensure the current state when the current state and assign state are the same, no change there. The change is about ensure current state before implementing an FSM transition only: when the state is asked to change.

DimCitus:
> > We had cases of Postgres being
> > restarted because of that, and if that happens when we don't have a recovery
> > setup in place, then Postgres restarts as a primary
> 
> How wouldn't we have the recovery setup while still `keeperState->current_role == SECONDARY_STATE`? Do we know the set of transitions that would lead to this?

So from the logs I have, this happens at secondary ➜ catchingup.
[test.pg12.debug2.log](https://github.com/citusdata/pg_auto_failover/files/5231881/test.pg12.debug2.log)

What I believe happens is that we lose the primary node first in our tests (`node1.ifdown()`) and so the monitor assigns CATCHING-UP to the standby nodes. Because we can't connect to the primary anymore, we can't implement this transition (the primary_conninfo testing fails). 

Also when doing secondary to catching-up, we must be ready for the primary to have changed in the meantime. So to take that into account, we redo the recovery config and compute a new one from the current primary. First, we cleanup the current recovery setup, then we fetch the current primary, check that we can connect to it, and then install the new recovery setup.

Because we can't connect to the current primary, we fail to re-install any recovery setup, and later when we are assigned secondary ➜ report_lsn (instead of catching-up) we would first make sure that Postgres is running. But we don't have a recovery setup. So we start as a primary there. Boom, standby is “corrupted”, or at least, can't be a standby anymore.


To follow what's happening one need to follow the following clues: when Postgres is starting the first LOG entries are telling us if it was a primary or a standby on the previous run. Here's what we have when it was a standby in the previous run:

```
LOG:  database system was shut down in recovery at ...
```

And here's what we see when it was *not* a standby in the previous run:

```
LOG:  database system was shut down at ...
```

src/bin/pg_autoctl/keeper.c
DimCitus: We call `keeper_ensure_current_state` in two places in the main keeper loop. One place is when there is a state transition to implement, and then sometimes we want to ensure the current state before implementing the transition (that's mostly when everything is under control). The other place is when the current and assigned state are the same, which is most of the times really.

This PR doesn't change `keeper_ensure_current_state`, it only makes it so that when we are asked to implement a transition to another state, we sometimes don't ensure the current state first. On the primary that's already what we did with the exceptions about DRAINING, DEMOTED, and DEMOTE_TIMEOUT states. It turns out we need the same kind of provisions for some cases on the secondary node(s).

onderkalaci (approved):
>  First, we cleanup the current recovery setup, then we fetch the current primary, check that we can connect to it, and then install the new recovery setup.

Thanks for the detailed explanation.

For the above, if we change the order to:
-   fetch the current primary,
 -  check we can connect to new primary
-  install  new  recovery

Could we avoid this situation?  If we  can't connect, the transition would fail, and the secondary would point to the old primary, which seems fine. We'll be electing a new primary and then it'll connect to that (unless it becomes the primary)

Ok, I'm curious about your answer to the above question. Overall, it feels easier to think about all the cases if we are sure that `SECONDARY` has the recovery, as opposed to it may not have sometimes. My concern is that we  could have other edge cases or could add in the future.

That said, what we are doing in this PR does seem OK by itself. Any state change that requires restart does it when needed anyway. So, I don't think this should block the release just because of this

DimCitus:
> Could we avoid this situation? If we can't connect, the transition would fail, and the secondary would point to the old primary, which seems fine. We'll be electing a new primary and then it'll connect to that (unless it becomes the primary)

That's a good point. I have added the code to enable that now, on-top of the previous changes. Avoiding the “ensure” step also makes the transitions slower, so I still think we should keep the changes to `keeper_should_ensure_current_state_before_transition()`.

PR #403, opened 2020-09-15
Fix/test primary conninfo

DimCitus:
Stop using `pg_receivewal` for testing primary_conninfo. We can actually use libpq and the replication protocol command IDENTIFY_SYSTEM instead.


src/bin/pg_autoctl/pgctl.c
```diff
@@ -2046,6 +1917,69 @@ prepare_conninfo_sslmode(PQExpBuffer buffer, SSLOptions sslOptions)
...
+		/* errors have already been logged */
+		return false;
+	}
+
+	if (!pgsql_identify_system(&replicationClient))
+	{
+		/* errors have already been logged */
+		return false;
+	}
+
```
onderkalaci, line 1978: I guess it'd be  nice to use `pgsql_finish ` as we don't need the connection  anymore.

(We already fo  it in `pgsql_identify_system `, but still could be nicer to follow the code as we're injecting `replication=1`  and the readers could likely to expect  it  closed here)

src/bin/pg_autoctl/cli_do_misc.c
```diff
@@ -491,33 +491,28 @@ keeper_cli_promote_standby(int argc, char **argv)
 
 
 /*
- * keeper_cli_receiwal allows to run pg_receivewal targetting the current
- * primary conninfo of a standby node, either to check that the connection
- * makes sense, or to actually start a sub-process receiving WAL.
+ * keeper_cli_identify_system connects to a Postgres server using the
```
onderkalaci, line 494: Minor: it might be good to add a sentence about what is `IDENTFY_SYSTEM`

src/bin/pg_autoctl/pgctl.c
```diff
@@ -2046,6 +1917,69 @@ prepare_conninfo_sslmode(PQExpBuffer buffer, SSLOptions sslOptions)
...
+{
+	NodeAddress *primaryNode = &(replicationSource->primaryNode);
+
+	char primaryConnInfo[MAXCONNINFO] = { 0 };
+	char primaryConnInfoReplication[MAXCONNINFO] = { 0 };
+	PGSQL replicationClient = { 0 };
+	int len = 0;
+
+	/* sometimes we don't have a primary, just pretend everything is fine */
+	if (IS_EMPTY_STRING_BUFFER(primaryNode->host))
```
onderkalaci, line 1935: I'd  rather  move this check to where `pgctl_identify_system` is called as  we used to have:
```C
!IS_EMPTY_STRING_BUFFER(primaryNode->host) &&	
			!pg_receivewal(pgdata, pg_ctl, replicationSource, "0/1", LOG_DEBUG)
```

Otherwise, this check/comment seems hard to follow and unexpected

src/bin/pg_autoctl/pgctl.c
```diff
@@ -2046,6 +1917,69 @@ prepare_conninfo_sslmode(PQExpBuffer buffer, SSLOptions sslOptions)
...
+								  replicationSource->password,
+								  replicationSource->applicationName,
+								  replicationSource->sslOptions,
+								  false)) /* no need for escaping */
+	{
+		/* errors have already been logged. */
+		return false;
+	}
+
+	len = sformat(primaryConnInfoReplication, MAXCONNINFO,
```
onderkalaci, line 1955: Minor: I guess this information is not likely that  every reader of the code is aware of. Maybe add a comment from the docs: `A Boolean value of true (or on, yes, 1) tells the backend to go into physical replication walsender mode.`

src/bin/pg_autoctl/pgsql.c
```diff
@@ -2204,7 +2205,7 @@ pgsql_get_postgres_metadata(PGSQL *pgsql,
 		" coalesce(rep.sync_state, '') as sync_state,"
 		" case when pg_is_in_recovery()"
 		" then coalesce(pg_last_wal_receive_lsn(), pg_last_wal_replay_lsn())"
-		" else pg_current_wal_lsn()"
+		" else pg_current_wal_flush_lsn()"
```
onderkalaci, line 2208: Could we document (maybe at least on the PR) for the reason of this change?

onderkalaci (approved):
Posting this docs for reference as it contains some relevant information: https://www.postgresql.org/docs/12/protocol-replication.html

The  changes looks good with few minor comments, mostly around the code documentation of the change.

onderkalaci:
(also tested locally, all works as expected)

src/bin/pg_autoctl/pgctl.c
DimCitus: The connection is both opened and closed within `pgsql_identify_system` so it don't belongs to the upper layer in the pgctl module, the way I see it... which means I'm confused by your comment...

onderkalaci:
Lastly, marking #350 as related PR where we added the logic that is replaced with the PR

src/bin/pg_autoctl/cli_do_misc.c
DimCitus: Added!

src/bin/pg_autoctl/pgctl.c
DimCitus: Good point, it's easier to read that way. Done now.

src/bin/pg_autoctl/pgctl.c
DimCitus: I added an extract from the Postgres docs to explain what we're doing here.

src/bin/pg_autoctl/pgsql.c
DimCitus: That's because the `pg_current_wal_lsn` gets an LSN that is not known to have been flushed on-disk yet, whereas `pg_current_wal_flush_lsn` embeds that guarantee. I guess in an ideal world that change would have happened on its own PR, but that's how it is now.

PR #386, opened 2020-09-02
Release/1.4.0

DimCitus:
Prepare the upgrade scripts and changelog for version 1.4.0.

PR #405, opened 2020-09-16
Adjust formation->number_sync_standbys when dropping a node.

### LINKED ISSUE #404: set_formation_number_sync_standbys seems too restrictive
I wonder whether some of these errors are necessary, especially when the value is 0:
```sql
pg_auto_failover=# SELECT set_formation_number_sync_standbys('default', 0);
ERROR:  Couldn't find the primary node in formation "default", group 0

pg_auto_failover=# SELECT pgautofailover.set_formation_number_sync_standbys('default', 0);
ERROR:  cannot set number_sync_standbys when current goal state for primary node 4 (172.23.156.186:6001) is "single", and current reported state is "single"
DETAIL:  The primary node so must be in state "primary" to be able to apply configuration changes to its synchronous_standby_names setting

pg_auto_failover=# SELECT pgautofailover.set_formation_number_sync_standbys('default', 0);
ERROR:  cannot set number_sync_standbys when current goal state for primary node 4 (172.23.156.186:6001) is "wait_primary", and current reported state is "wait_primary"
DETAIL:  The primary node so must be in state "primary" to be able to apply configuration changes to its synchronous_standby_names setting
```

I wanted to use it after reducing a multiple standby cluster to single, since I ended up with a single-node cluster that had number_sync_standbys 1.

DimCitus:
Fixes #404.

### DIFF
--- a/src/monitor/node_active_protocol.c
+++ b/src/monitor/node_active_protocol.c
@@ -993,9 +993,12 @@ RemoveNode(AutoFailoverNode *currentNode)
 {
 	bool currentNodeIsPrimary = false;
 
+	AutoFailoverFormation *formation = NULL;
 	List *otherNodesGroupList = NIL;
 	ListCell *nodeCell = NULL;
 	AutoFailoverNode *firstStandbyNode = NULL;
+
+	int countSyncStandbys = 0;
 
 	char message[BUFSIZE] = { 0 };
 
@@ -1005,6 +1008,8 @@ RemoveNode(AutoFailoverNode *currentNode)
 	}
 
 	LockFormation(currentNode->formationId, ExclusiveLock);
+
+	formation = GetFormation(currentNode->formationId);
 
 	/* when removing the primary, initiate a failover */
 	currentNodeIsPrimary = CanTakeWritesInState(currentNode->goalState);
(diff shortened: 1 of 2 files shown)


src/monitor/node_active_protocol.c
```diff
@@ -1060,6 +1065,47 @@ RemoveNode(AutoFailoverNode *currentNode)
...
+	 * otherNodesGroupList is the list of all the remaining nodes, and that
+	 * includes the current primary, which might be setup with replication
+	 * quorum set to true (and probably is).
+	 */
+	countSyncStandbys = CountSyncStandbys(otherNodesGroupList) - 1;
+
+	LogAndNotifyMessage(message, BUFSIZE,
+						"CountSyncStandbys: %d", countSyncStandbys);
+
+	if (countSyncStandbys < (formation->number_sync_standbys + 1))
```
onderkalaci, line 1080: Is otherwise ever possible? I mean `countSyncStandbys  >= formation->number_sync_standbys + 1`?

We normally prevent it, still being defensive here is good, just want to ask in case I oversee some cases.

onderkalaci (approved):
This all makes sense, I've tried different scenarios with drop node all works as expected.

src/monitor/pgautofailover.sql
```diff
@@ -741,49 +741,6 @@ grant execute on function
...
-  end if;
-
-  return old;
-end
-$$;
-
-comment on function pgautofailover.adjust_number_sync_standbys()
-        is 'adjust formation number_sync_standbys when removing a node, if needed';
-
-CREATE TRIGGER adjust_number_sync_standbys
```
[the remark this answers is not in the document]
DimCitus: Yes. I will update https://github.com/citusdata/pg_auto_failover/pull/386 as soon as this one is merged in master's branch.

src/monitor/node_active_protocol.c
DimCitus: That's the code that runs when we just dropped a standby. So it might be that we had two standby nodes and number_sync_standbys set to 1, all good. One standby removed. One standby left, number_sync_standbys now must be set to zero. That's what this test is all about.

Now, it could be that we had three standby nodes, and still had number_sync_standbys set to 1. Now a standby is removed. Two standby left, number_sync_standbys = 1, it's all good. That's your “otherwise” example, right?

PR #406, opened 2020-09-16
Fix network partition detections.

DimCitus:
We might fail to connect to the monitor early in keeper_node_active()
function and that was not a code path taken into account. Oops.


JelteF:
This should have a regression test, since it's quite a serious bug that postgres is not killed.

JelteF (approved):

PR #407, opened 2020-09-16
Fix some problems in make cluster

JelteF:
My main problems were with the fact that nodes were spawned at the same
time and the node with name `node_2` was not the node with directory
`node2`. This is fixed by doing three things:
1. Adding `--name` to the create commands
2. Make `show state` order by name instead of nodeid, so `node2` would
   actually be above `node3` in the table
3. Fix the `do wait` commands to use `PG_AUTOCTL_DEBUG`.

My other issue was that I could not use `--pgdata node2` in the
interactive shell. This had two reasons:
1. It did not actually `cd` to the `TMUX_ROOT` directory
2. The `XDG` paths were relative, so once I `cd`ed there the XDG paths
   were suddenly wrong.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,8 +28,8 @@ FIRST_PGPORT ?= 5500
 
 TMUX_EXTRA_COMMANDS ?= ""
 TMUX_LAYOUT ?= even-vertical	# could be "tiled"
-TMUX_TOP_DIR = ./tmux
-TMUX_SCRIPT = ./tmux/script-$(FIRST_PGPORT).tmux
+TMUX_TOP_DIR = $(PWD)/tmux
+TMUX_SCRIPT = $(TMUX_TOP_DIR)/script-$(FIRST_PGPORT).tmux
 
 all: monitor bin ;
 
--- a/src/bin/pg_autoctl/monitor.c
+++ b/src/bin/pg_autoctl/monitor.c
@@ -1641,7 +1641,7 @@ monitor_print_state(Monitor *monitor, char *formation, int group)
 		{
 			sql =
 				"SELECT * FROM pgautofailover.current_state($1) "
-				"ORDER BY node_id";
+				"ORDER BY nodename";
 
 			paramCount = 1;
 			paramTypes[0] = TEXTOID;
@@ -1654,7 +1654,7 @@ monitor_print_state(Monitor *monitor, char *formation, int group)
 		{
 			sql =
 				"SELECT * FROM pgautofailover.current_state($1,$2) "
-				"ORDER BY node_id";
+				"ORDER BY nodename";
 
 			groupStr = intToString(group);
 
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -28,8 +28,8 @@ FIRST_PGPORT ?= 5500
 
 TMUX_EXTRA_COMMANDS ?= ""
 TMUX_LAYOUT ?= even-vertical	# could be "tiled"
-TMUX_TOP_DIR = ./tmux
-TMUX_SCRIPT = ./tmux/script-$(FIRST_PGPORT).tmux
+TMUX_TOP_DIR = $(PWD)/tmux
```
DimCitus, line 31: I think instead of $(PWD) we should use the classic GNU Make trick for the top-level directory:

``` make
TOP := $(dir $(abspath $(lastword $(MAKEFILE_LIST))))
```

Put that in the first line of the Makefile.

src/bin/pg_autoctl/monitor.c
```diff
@@ -1641,7 +1641,7 @@ monitor_print_state(Monitor *monitor, char *formation, int group)
 		{
 			sql =
 				"SELECT * FROM pgautofailover.current_state($1) "
-				"ORDER BY node_id";
+				"ORDER BY nodename";
```
DimCitus, line 1644: I don't think that's a good idea, the node_id is stable, the nodename can change at any time.

src/bin/pg_autoctl/monitor.c
```diff
@@ -1654,7 +1654,7 @@ monitor_print_state(Monitor *monitor, char *formation, int group)
 		{
 			sql =
 				"SELECT * FROM pgautofailover.current_state($1,$2) "
-				"ORDER BY node_id";
+				"ORDER BY nodename";
```
DimCitus, line 1657: Same. Can we avoid doing that?

DimCitus (changes requested):

DimCitus:
> 1. Adding `--name` to the create commands

Good point!

> 2. Make `show state` order by name instead of nodeid, so `node2` would
>    actually be above `node3` in the table

Let's not do that?

> 3. Fix the `do wait` commands to use `PG_AUTOCTL_DEBUG`.

Oh good catch! I guess with that fix alone the waiting makes it all good with the pgport and nodded matching together again?

> 1. It did not actually `cd` to the `TMUX_ROOT` directory

Good point.

> 2. The `XDG` paths were relative, so once I `cd`ed there the XDG paths
>    were suddenly wrong.

I'm missing the fix for that, can't see that change in the PR?

src/bin/pg_autoctl/monitor.c
JelteF: I guess we could make node3 wait on node2 being up. The reason I did this was that it's quite confusing to have node2 be the third node in the list.

Makefile
JelteF: > > The XDG paths were relative, so once I cded there the XDG paths
> > were suddenly wrong.
>
> I'm missing the fix for that, can't see that change in the PR?

This makes the XDG paths absolute for me. I now see that it's better to do it in the C code, so it also works if you provide your own TMUX_TOP_DIR. So I think I will undo this change and do it in the better way.

JelteF:
Closing this in favor of #409

PR #409, opened 2020-09-17
Feature/tmux session

DimCitus:
Implement `pg_autoctl do tmux session`, a command that starts the tmux session and then kills
the pg_autoctl process created within at exit, and the tmux session too.
It allows to finish the test session with `C-b d`.

See #407, which it kind of replaces, fixing the same list of problems in passing.

### DIFF
--- a/src/bin/pg_autoctl/cli_do_root.c
+++ b/src/bin/pg_autoctl/cli_do_root.c
@@ -234,14 +234,39 @@ CommandLine do_tmux_script =
 	make_command("script",
 				 "Produce a tmux script for a demo or a test case",
 				 "[option ...]",
-				 "  --root          path where to create a cluster\n" \
-				 "  --first-port    first Postgres port to use (5500)\n" \
-				 "  --nodes         number of Postgres nodes to create (2)",
+				 "  --root          path where to create a cluster\n"
+				 "  --first-pgport  first Postgres port to use (5500)\n"
+				 "  --nodes         number of Postgres nodes to create (2)"
+				 "  --layout        tmux layout to use (even-vertical)",
 				 cli_do_tmux_script_getopts,
 				 cli_do_tmux_script);
 
+CommandLine do_tmux_session =
+	make_command("session",
+				 "Run a a tmux session for a demo or a test case",
+				 "[option ...]",
+				 "  --root          path where to create a cluster\n"
+				 "  --first-pgport  first Postgres port to use (5500)\n"
+				 "  --nodes         number of Postgres nodes to create (2)"
+				 "  --layout        tmux layout to use (even-vertical)",
+				 cli_do_tmux_script_getopts,
+				 cli_do_tmux_session);
+
+CommandLine do_tmux_stop =
+	make_command("stop",
+				 "Stop pg_autoctl processed that belong to a tmux session ",
+				 "[option ...]",
+				 "  --root          path where to create a cluster\n"
+				 "  --first-pgport  first Postgres port to use (5500)\n"
+				 "  --nodes         number of Postgres nodes to create (2)"
+				 "  --layout        tmux layout to use (even-vertical)",
+				 cli_do_tmux_script_getopts,
+				 cli_do_tmux_stop);
+
 CommandLine *do_tmux[] = {
 	&do_tmux_script,
+	&do_tmux_session,
+	&do_tmux_stop,
 	NULL
 };
 
--- a/src/bin/pg_autoctl/cli_do_root.c
+++ b/src/bin/pg_autoctl/cli_do_root.c
@@ -255,13 +255,23 @@ CommandLine do_tmux_session =
 
 CommandLine do_tmux_stop =
 	make_command("stop",
-				 "Stop pg_autoctl processed that belong to a tmux session ",
+				 "Stop pg_autoctl processes that belong to a tmux session ",
 				 "[option ...]",
 				 "  --root          path where to create a cluster\n"
 				 "  --first-pgport  first Postgres port to use (5500)\n"
 				 "  --nodes         number of Postgres nodes to create (2)",
 				 cli_do_tmux_script_getopts,
 				 cli_do_tmux_stop);
+
+CommandLine do_tmux_clean =
+	make_command("clean",
+				 "Clean-up a tmux session processes and root dir",
+				 "[option ...]",
+				 "  --root          path where to create a cluster\n"
+				 "  --first-pgport  first Postgres port to use (5500)\n"
+				 "  --nodes         number of Postgres nodes to create (2)",
+				 cli_do_tmux_script_getopts,
+				 cli_do_tmux_clean);
 
 CommandLine do_tmux_wait =
 	make_command("wait",
@@ -278,6 +288,7 @@ CommandLine *do_tmux[] = {
 	&do_tmux_session,
 	&do_tmux_stop,
 	&do_tmux_wait,
+	&do_tmux_clean,
 	NULL
 };
 
--- a/src/bin/pg_autoctl/cli_do_root.c
+++ b/src/bin/pg_autoctl/cli_do_root.c
@@ -252,21 +252,32 @@ CommandLine do_tmux_session =
 				 cli_do_tmux_script_getopts,
 				 cli_do_tmux_session);
 
+
 CommandLine do_tmux_stop =
 	make_command("stop",
 				 "Stop pg_autoctl processed that belong to a tmux session ",
 				 "[option ...]",
 				 "  --root          path where to create a cluster\n"
 				 "  --first-pgport  first Postgres port to use (5500)\n"
-				 "  --nodes         number of Postgres nodes to create (2)"
-				 "  --layout        tmux layout to use (even-vertical)",
+				 "  --nodes         number of Postgres nodes to create (2)",
 				 cli_do_tmux_script_getopts,
 				 cli_do_tmux_stop);
+
+CommandLine do_tmux_wait =
+	make_command("wait",
+				 "Wait until a given node has been registered on the monitor",
+				 "[option ...] nodename",
+				 "  --root          path where to create a cluster\n"
+				 "  --first-pgport  first Postgres port to use (5500)\n"
+				 "  --nodes         number of Postgres nodes to create (2)",
+				 cli_do_tmux_script_getopts,
+				 cli_do_tmux_wait);
 
 CommandLine *do_tmux[] = {
 	&do_tmux_script,
 	&do_tmux_session,
 	&do_tmux_stop,
+	&do_tmux_wait,
 	NULL
 };
 
--- a/src/bin/pg_autoctl/cli_do_root.c
+++ b/src/bin/pg_autoctl/cli_do_root.c
@@ -232,11 +232,11 @@ CommandLine do_pgsetup_commands =
 
 CommandLine do_tmux_script =
 	make_command("script",
-				 "Produce a tmux script for a demo or a test case",
+				 "Produce a tmux script for a demo or a test case (debug only)",
 				 "[option ...]",
 				 "  --root          path where to create a cluster\n"
 				 "  --first-pgport  first Postgres port to use (5500)\n"
-				 "  --nodes         number of Postgres nodes to create (2)"
+				 "  --nodes         number of Postgres nodes to create (2)\n"
 				 "  --layout        tmux layout to use (even-vertical)",
 				 cli_do_tmux_script_getopts,
 				 cli_do_tmux_script);
@@ -247,7 +247,7 @@ CommandLine do_tmux_session =
 				 "[option ...]",
 				 "  --root          path where to create a cluster\n"
 				 "  --first-pgport  first Postgres port to use (5500)\n"
-				 "  --nodes         number of Postgres nodes to create (2)"
+				 "  --nodes         number of Postgres nodes to create (2)\n"
 				 "  --layout        tmux layout to use (even-vertical)",
 				 cli_do_tmux_script_getopts,
 				 cli_do_tmux_session);
--- a/src/bin/pg_autoctl/cli_do_tmux.c
+++ b/src/bin/pg_autoctl/cli_do_tmux.c
@@ -73,6 +73,7 @@ static void tmux_pg_autoctl_create_postgres(PQExpBuffer script,
 static bool tmux_start_server(const char *root, const char *scriptName);
 static bool pg_autoctl_stop(const char *root, const char *name);
 static bool tmux_stop_pg_autoctl(TmuxOptions *options);
+static void tmux_process_options(TmuxOptions *options);
 
 
 /*
@@ -276,6 +277,9 @@ tmux_add_send_keys_command(PQExpBuffer script, const char *fmt, ...)
 static bool
 tmux_prepare_XDG_environment(const char *root)
 {
+	log_info("Preparing XDG setting for self-contained session in \"%s\"",
+			 root);
+
 	for (int i = 0; xdg[i][0] != NULL; i++)
 	{
 		char *var = xdg[i][0];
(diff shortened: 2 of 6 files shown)


src/bin/pg_autoctl/cli_do_tmux.c
```diff
@@ -439,7 +514,333 @@ cli_do_tmux_script(int argc, char **argv)
...
+{
+	TmuxOptions options = tmuxOptions;
+
+	/* prepare the XDG environment */
+	if (!tmux_prepare_XDG_environment(options.root))
+	{
+		exit(EXIT_CODE_INTERNAL_ERROR);
+	}
+
+	if (!tmux_stop_pg_autoctl(&options))
```
JelteF, line 842: Shouldn't this try to stop the tmux session as well?

src/bin/pg_autoctl/cli_do_root.c
```diff
@@ -234,14 +234,39 @@ CommandLine do_tmux_script =
...
+				 cli_do_tmux_session);
+
+CommandLine do_tmux_stop =
+	make_command("stop",
+				 "Stop pg_autoctl processed that belong to a tmux session ",
+				 "[option ...]",
+				 "  --root          path where to create a cluster\n"
+				 "  --first-pgport  first Postgres port to use (5500)\n"
+				 "  --nodes         number of Postgres nodes to create (2)"
+				 "  --layout        tmux layout to use (even-vertical)",
```
JelteF, line 262: Layout doesn't seem very useful for the `stop` command.

src/bin/pg_autoctl/cli_do_tmux.c
```diff
@@ -439,7 +514,333 @@ cli_do_tmux_script(int argc, char **argv)
...
+		success = false;
+		log_fatal("Failed to start the tmux session, see above for details");
+	}
+
+	/*
+	 * Stop our pg_autoctl processes and kill the tmux session.
+	 */
+	log_info("tmux session ended: kill pg_autoct processes");
+	success = success && tmux_stop_pg_autoctl(&options);
+	success = success && tmux_kill_session(&options);
```
JelteF, line 818: Maybe it makes more sense to kill the session first. Otherwise it might still spawn a `pg_autoctl` process in between this process trying to kill it, but before the `tmux` session is killed.

JelteF (changes requested):
I found some issues:
# Issue 1: wait command errors

I get this output in the `wait` command for `node2` and `node3` (which waits on `node1`):
```
jelte@MININT-53UGE2E ~/work/pg_auto_failover (feature/tmux-session %) $ ./src/bin/pg_autoctl/pg_autoctl do pgsetup wait --pgdata /home/jelte/work/pg_auto_failover/tmux/node1
09:40:28 19733 ERROR Given --pgport 5432 doesn't match PostgreSQL port 5501 from "/home/jelte/work/pg_auto_failover/tmux/node1/postmaster.pid"
09:40:28 19733 FATAL Failed to discover PostgreSQL setup, please fix previous errors.
09:40:28 19733 ERROR pg_setup_wait_until_is_ready: pg_setup_init is false
09:40:28 19733 INFO  Postgres status is: "unknown"
```

This seems like something is not going as intended there.

# Issue 2: Assertion in logging.c

I triggered an assertion at some point. This is the backtrace:
```
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
#1  0x00007f709b5688b1 in __GI_abort () at abort.c:79
#2  0x00007f709b55842a in __assert_fail_base (fmt=0x7f709b6dfa38 "%s%s%s:%u: %s%sAssertion `%s' failed.\n%n", assertion=assertion@entry=0x5591f8e6ab9e "progname", 
    file=file@entry=0x5591f8e6ab94 "logging.c", line=line@entry=193, function=function@entry=0x5591f8e6ac00 <__PRETTY_FUNCTION__.6149> "pg_log_generic_v") at assert.c:92
#3  0x00007f709b5584a2 in __GI___assert_fail (assertion=assertion@entry=0x5591f8e6ab9e "progname", file=file@entry=0x5591f8e6ab94 "logging.c", line=line@entry=193, 
    function=function@entry=0x5591f8e6ac00 <__PRETTY_FUNCTION__.6149> "pg_log_generic_v") at assert.c:101
#4  0x00005591f8e4aca7 in pg_log_generic_v (level=PG_LOG_WARNING, fmt=0x5591f8e6ab30 "could not remove file or directory \"%s\": %m", ap=ap@entry=0x7ffc2565eed0)
    at logging.c:193
#5  0x00005591f8e4af34 in pg_log_generic (level=level@entry=PG_LOG_WARNING, fmt=fmt@entry=0x5591f8e6ab30 "could not remove file or directory \"%s\": %m") at logging.c:179
#6  0x00005591f8e4a80d in rmtree (path=path@entry=0x7ffc2565f8e0 "./tmux", rmtopdir=rmtopdir@entry=true) at rmtree.c:113
#7  0x00005591f8e0ed9a in ensure_empty_dir (dirname=dirname@entry=0x7ffc2565f8e0 "./tmux", mode=mode@entry=448) at file_utils.c:96
#8  0x00005591f8e3c6e2 in cli_do_tmux_session (argc=<optimized out>, argv=<optimized out>) at cli_do_tmux.c:756
#9  0x00005591f8e46283 in commandline_run (command=0x5591f907c580 <do_tmux_session>, argc=0, argc@entry=9, argv=0x7ffc25660c08, argv@entry=0x7ffc25660bc0)
    at /home/jelte/work/pg_auto_failover/src/bin/pg_autoctl/../lib/subcommands.c/commandline.c:71
#10 0x00005591f8e46369 in commandline_run (command=0x5591f907ca40 <do_tmux_commands>, argc=9, argc@entry=10, argv=0x7ffc25660bc0, argv@entry=0x7ffc25660bb8)
    at /home/jelte/work/pg_auto_failover/src/bin/pg_autoctl/../lib/subcommands.c/commandline.c:94
#11 0x00005591f8e46369 in commandline_run (command=0x5591f907ca00 <do_commands>, argc=10, argc@entry=11, argv=0x7ffc25660bb8, argv@entry=0x7ffc25660bb0)
    at /home/jelte/work/pg_auto_failover/src/bin/pg_autoctl/../lib/subcommands.c/commandline.c:94
#12 0x00005591f8e46369 in commandline_run (command=0x7ffc25660660, argc=11, argv=0x7ffc25660bb0)
    at /home/jelte/work/pg_auto_failover/src/bin/pg_autoctl/../lib/subcommands.c/commandline.c:94
#13 0x00005591f8e32881 in main (argc=12, argv=0x7ffc25660ba8) at main.c:120
```

# Issue 3: Killing of processes stops after first failure

The code suggests that even when it fails in stopping a `pg_autoctl` process it will continue to stop the others as well. However, this is not what happens. I guess somewher deeper in the code we call `exit(EXIT_CODE_INTERNAL_ERROR)`.
```
10:13:34 20368 INFO  tmux session ended: kill pg_autoct processes
10:13:34 20368 INFO   ./src/bin/pg_autoctl/pg_autoctl stop --pgdata /home/jelte/work/pg_auto_failover/tmux/node1
10:13:34 20400 FATAL Expected configuration file does not exists: "/home/jelte/work/pg_auto_failover/tmux/config/pg_autoctl/home/jelte/work/pg_auto_failover/tmux/node1/pg_autoctl.cfg"
10:13:34 20400 WARN  HINT: Check your PGDATA setting: "/home/jelte/work/pg_auto_failover/tmux/node1"
10:13:34 20368 WARN  Failed to stop pg_autoctl for "/home/jelte/work/pg_auto_failover/tmux/node1"
Makefile:120: recipe for target 'cluster' failed
make: *** [cluster] Error 12
```

To reproduce, run `make cluster` and `C-b d` right away before everything is ready.

src/bin/pg_autoctl/cli_do_tmux.c
```diff
@@ -439,7 +514,333 @@ cli_do_tmux_script(int argc, char **argv)
...
+	return success;
+}
+
+
+/*
+ * keeper_cli_tmux_script generates a tmux script to run a test case or a demo
+ * for pg_auto_failover easily.
+ */
+void
+cli_do_tmux_script(int argc, char **argv)
```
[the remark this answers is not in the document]
DimCitus: I couldn't find a good reason to ditch this command entirely. Have added root directory normalisation and things.

src/bin/pg_autoctl/cli_do_tmux.c
```diff
@@ -382,18 +464,21 @@ cli_do_tmux_script(int argc, char **argv)
 			tmux_add_send_keys_command(script,
 									   "%s do pgsetup wait --pgdata %s/node1",
```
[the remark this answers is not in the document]
DimCitus: Good point. Thing is, we don't need to wait until Postgres is running, which could take quite a long time on a secondary because we then are waiting for `pg_basebackup` to complete first. So I have added a new command `pg_autoctl do tmux wait nodename`. When given a monitor, it waits until Postgres is running. When given a Postgres node, it waits until the state file exists on disk: we create the file once registered to the monitor. Seems to work well in my testing.

src/bin/pg_autoctl/cli_do_tmux.c
DimCitus: I have used it as a debugging tool mostly when something went wrong after `C-b d`, I'm not sure we have another use case than “things went wrong, help me”. When things went wrong, being able to `tmux at` and take it from there has been quite helpful to me.

src/bin/pg_autoctl/cli_do_root.c
DimCitus: Good point. Removed from the `--help` output.

src/bin/pg_autoctl/cli_do_tmux.c
DimCitus: I'm still not sure how that works, the kill session thing, with respect to processes running in the shell in the panes. That's why I want to first stop our processes and then kill the session. Does it make sense?

src/bin/pg_autoctl/cli_do_tmux.c
```diff
@@ -365,35 +467,30 @@ cli_do_tmux_script(int argc, char **argv)
...
-									   "%s do pgsetup wait --pgdata %s/node1",
-									   pg_autoctl_argv0,
-									   options.root);
-		}
+		/*
+		 * Force node ordering to easy debugging of interactive sessions: each
+		 * node waits until the previous one has been started or registered.
+		 */
+		tmux_add_send_keys_command(script,
+								   "%s do tmux wait --root %s %s -vv",
```
JelteF, line 475: Do we really want `-vv` for this command?

src/bin/pg_autoctl/cli_do_tmux.c
```diff
@@ -245,6 +270,68 @@ tmux_add_send_keys_command(PQExpBuffer script, const char *fmt, ...)
 }
 
 
+/*
+ * tmux_prepare_XDG_environment set XDG environment variables in the current
+ * process tree.
+ */
+static bool
+tmux_prepare_XDG_environment(const char *root)
```
JelteF, line 278: Something weird is going on with this. If I run two clusters at the same time, the second cluster will use the XDG variables from the first somehow. To reproduce:
```bash
#shell 1
make -j12 cluster
# shell 2
make TMUX_TOP_DIR=tmux2 FIRST_PGPORT=6500 cluster -j9
```

The second cluster will then use the XDG paths of the first. I guess the reason for this is that the tmux daemon is already running and thus using `setenv` doesn't have any effect for the second cluster. So I think it's probably better to set them always using `sendkeys` again.

JelteF (changes requested):

src/bin/pg_autoctl/cli_do_tmux.c
DimCitus: I think I was totally confused with the tmux process model. If the tmux server is started at the right time then it all makes sense with the previous coding, but when the tmux server has already been running then it keeps the first XDG values. I had strange test cases locally, and I think that's what happened. Should be fixed now.

DimCitus:
Issue with killing pg_autoctl processes stopping after first failure is now fixed too.

src/bin/pg_autoctl/cli_do_tmux.c
```diff
@@ -365,49 +463,446 @@ cli_do_tmux_script(int argc, char **argv)
...
+				log_info("No pidfile for pg_autoctl for node \"%s\"", name);
+				--countRunning;
+				continue;
+			}
+
+			if (kill(pid, 0) == -1 && errno == ESRCH)
+			{
+				log_info("Pid %d for node \"%s\" is not running anymore",
+						 pid, name);
+				--countRunning;
```
JelteF, line 648: Getting this warning:
```
cli_do_tmux.c:648:5: warning: decrement of a boolean expression [-Wbool-operation]
     --countRunning;
     ^~
```

src/bin/pg_autoctl/cli_do_tmux.c
```diff
@@ -365,49 +463,446 @@ cli_do_tmux_script(int argc, char **argv)
...
+				sformat(name, sizeof(name), "node%d", i + 1);
+			}
+
+			sformat(pgdata, sizeof(pgdata), "%s/%s", options->root, name);
+
+			if (!pg_autoctl_getpid(pgdata, &pid))
+			{
+				/* we don't have a pid */
+				log_info("No pidfile for pg_autoctl for node \"%s\"", name);
+				--countRunning;
```
JelteF, line 640: Getting this warning:
```
cli_do_tmux.c:640:5: warning: decrement of a boolean expression [-Wbool-operation]
     --countRunning;
     ^~
```

JelteF (changes requested):

src/bin/pg_autoctl/cli_do_tmux.c
```diff
@@ -246,117 +276,185 @@ tmux_add_send_keys_command(PQExpBuffer script, const char *fmt, ...)
...
 	 */
-	for (int i = 0; i < 3; i++)
+	for (int i = 0; xdg[i][0] != NULL; i++)
 	{
 		char *var = xdg[i][0];
-		char *dir = xdg[i][1];
 
-		tmux_add_send_keys_command(script,
-								   "export %s=\"%s/%s\"", var, root, dir);
+		tmux_add_command(script, "set-option update-environment %s", var);
```
JelteF, line 367: This still doesn't fix my issue where a second cluster has the first clusters XDG env vars. After a little googling I think we should change this to (non pseudo code version of):
```suggestion
		tmux_add_command(script, "setenv %s %s", var, normalize_path(xdg[i][1]));
```

JelteF (changes requested):

src/bin/pg_autoctl/cli_do_tmux.c
```diff
@@ -649,13 +665,74 @@ tmux_stop_pg_autoctl(TmuxOptions *options)
...
+	Program program;
+	int returnCode;
+
+	char tmux[MAXPGPATH] = { 0 };
+	char command[BUFSIZE] = { 0 };
+
+	if (!search_path_first("tmux", tmux))
+	{
+		log_fatal("Failed to find program tmux in PATH");
+		return false;
```
JelteF, line 691: I don't think this will ever happen, but we won't handle this as an error when calling `tmux_has_session`. 
The reason is that the return value of this function is used as a way to signal if the session exists, not if an error happens. Is this on purpose or an oversight?

src/bin/pg_autoctl/cli_do_tmux.c
DimCitus: There's now `pg_autoctl do tmux clean` that cleans processes, directories and kills the tmux session.

src/bin/pg_autoctl/cli_do_tmux.c
DimCitus: So the `set-option update-environment` does not work as I though it would. I didn't try the `setenv` command, I have added back the sending of the `export var=value` commands as we used to have before. I could then use `pg_autoctl do tmux session` on 3 different environments in parallel.

src/bin/pg_autoctl/cli_do_tmux.c
DimCitus: Fixed, that was an oversight indeed.

JelteF (approved):

PR #414, opened 2020-09-21
Fix taking control of an already running  Postgres instance.

### LINKED ISSUE #412: Rogue postgres cannot be stopped anymore by pg_autoctl
To reproduce:

```bash
# shell 1
pg_autoctl create monitor --pgdata monitor --ssl-self-signed --auth trust --pgport 5500 --run

# shell 2
pg_autoctl create postgres --pgdata node1 --monitor "$(pg_autoctl show uri --pgdata monitor --monitor)" --pgport 5501 --ssl-self-signed --auth trust --run
# <Ctrl+C>
pg_ctl start -D node1
pg_autoctl create postgres --pgdata node1 --monitor "$(pg_autoctl show uri --pgdata monitor --monitor)" --pgport 5501 --ssl-self-signed --auth trust --run

# shell 3
pg_autoctl create postgres --pgdata node2 --monitor "$(pg_autoctl show uri --pgdata monitor --monitor)" --pgport 5502 --ssl-self-signed --auth trust --run

# Shell 4
pg_autoctl perform failover --pgdata node1
```

This situation mostly occurs when upgrading from before 1.3 and when initially starting to use pg_autofailover, when adding HA to an already running postgres. Both of these are a bit of an edge case that is not commonly triggered by devs during normal use. So I think it would be good to add a test case for starting from a running postgres specifically, since afaik we still want to support this use case.

DimCitus:
When we optimised away calls to `pg_controldata`, we missed that parts of the code depend on it having been run previously. This is another case of cache invalidation games, and we force getting the control data again when we don't have it in `pg_setup_pgdata_exists`. Given that fix, then we unblock a loop at https://github.com/citusdata/pg_auto_failover/blob/master/src/bin/pg_autoctl/service_postgres_ctl.c#L273 which otherwise is an infinite loop.

Fix #412.

### DIFF
--- a/src/bin/pg_autoctl/pgsetup.c
+++ b/src/bin/pg_autoctl/pgsetup.c
@@ -830,8 +830,9 @@ pg_setup_get_local_connection_string(PostgresSetup *pgSetup,
 
 
 /*
- * pg_setup_pgdata_exists returns true when the pg_controldata probe was
- * susccessful.
+ * pg_setup_pgdata_exists returns true when PGDATA exists, hosts a
+ * global/pg_control file (so that it looks like a Postgres cluster) and when
+ * the pg_controldata probe was successful.
  */
 bool
 pg_setup_pgdata_exists(PostgresSetup *pgSetup)
@@ -860,7 +861,10 @@ pg_setup_pgdata_exists(PostgresSetup *pgSetup)
 	{
 		bool missingPgdataIsOk = false;
 
-		return pg_controldata(pgSetup, missingPgdataIsOk);
+		/* errors are logged from within pg_controldata */
+		(void) pg_controldata(pgSetup, missingPgdataIsOk);
+
+		return pgSetup->control.system_identifier != 0;
 	}
 
 	return true;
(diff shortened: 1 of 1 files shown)


src/bin/pg_autoctl/pgsetup.c
```diff
@@ -852,7 +852,18 @@ pg_setup_pgdata_exists(PostgresSetup *pgSetup)
...
-	return pgSetup->control.system_identifier != 0;
+	/*
+	 * Now that we know that PGDATA exists, let's grab the system identifier if
+	 * we don't have it already.
+	 */
+	if (pgSetup->control.system_identifier == 0)
+	{
+		bool missingPgdataIsOk = false;
+
+		return pg_controldata(pgSetup, missingPgdataIsOk);
```
JelteF, line 863: Is it on purpuse that we use the error boolean from `pg_controldata` as the return value from this function? The name of the function suggests that it returns true when pgdata exists. The comment of the function says that it returns true when `pg_controldata` was successful. This is a bit confusing.

JelteF (approved):
This indeed fixes the issue and the explanation is good. I think this is good to merge, except  for maybe updating the function name as suggested in my other comment.

src/bin/pg_autoctl/pgsetup.c
DimCitus: Good point, I tried to clarify the comment for the function, that was not kept up-to-date when we optimised away calls to pg_controldata, apparently.

JelteF (approved):

PR #425, opened 2020-09-23
Update tutorial for release 1.4.0.

DimCitus:


docs/tutorial.rst
```diff
@@ -374,12 +374,6 @@ and query the data, this time from node B.
    ssh -l ha-admin -t `vm_ip monitor` \
      pg_autoctl perform switchover --pgdata monitor
 
-   # watch the progress
-   ssh -t -l ha-admin `vm_ip monitor` -- \
-     watch -n 1 -d pg_autoctl show state --pgdata monitor
-
-   # press CTRL-C to stop watching
```
JelteF, line 381: Why remove this section?

docs/tutorial.rst
```diff
@@ -394,8 +388,7 @@ It shows
 
 .. code-block:: bash
 
-  .
```
JelteF, line 397: I thought this `.` was needed for good indentation or something in the final HTML: https://github.com/citusdata/pg_auto_failover/commit/6efcbf2c207b42c049d772e4697d410e84859d9b

The same for the other dots that were removed.

docs/tutorial.rst
```diff
@@ -263,7 +263,7 @@ Next connect to node B and do the same process. We'll do both steps at once:
        --ssl-self-signed \
        --username ha-admin \
        --dbname appdb \
-       --nodename ha-demo-b.internal.cloudapp.net \
+       --hostname ha-demo-b.internal.cloudapp.net \
        --pgctl /usr/lib/postgresql/11/bin/pg_ctl \
        --monitor 'postgres://autoctl_node@ha-demo-monitor.internal.cloudapp.net/pg_auto_failover?sslmode=require'
```
JelteF, line 268: should we add a `--name` argument to these nodes as well?

JelteF (approved):
Looks fine overal, added some small comments.

docs/tutorial.rst
DimCitus: Because it's not needed anymore, now that the `pg_autoctl perform failover` shows what's happening thanks to using the notifications from the monitor.

docs/tutorial.rst
DimCitus: I have added non-breaking white space instead. They display nice in Emacs, and don't disturb the readers of the HTML, where we still have the intended markup.

docs/tutorial.rst
DimCitus: I think it's good to show the defaults. In a later version of the tutorial, when we use multiple zones, I will given them names like east-a and west-c and things.

PR #439, opened 2020-09-30
Fix when to ProceedWithMSFailover after a candidate is selected.

DimCitus:
We might reach intermediate states on the newly elected primary such as
wait_primary/primary and in that situation we can already move forward and
have the secondary nodes re-configured to follow the new primary.

### DIFF
--- a/src/monitor/node_metadata.c
+++ b/src/monitor/node_metadata.c
@@ -1537,6 +1537,23 @@ IsBeingPromoted(AutoFailoverNode *node)
 
 
 /*
+ * CandidateNodeIsReadyToStreamWAL returns whether a newly selected candidate
+ * node, possibly still being promoted, is ready for the other standby nodes is
+ * REPORT_LSN to already use the new primary as an upstream node.
+ */
+bool
+CandidateNodeIsReadyToStreamWAL(AutoFailoverNode *node)
+{
+	return node != NULL &&
+		   (node->reportedState == REPLICATION_STATE_STOP_REPLICATION ||
+			node->goalState == REPLICATION_STATE_STOP_REPLICATION
+
+			|| node->reportedState == REPLICATION_STATE_WAIT_PRIMARY ||
+			node->goalState == REPLICATION_STATE_WAIT_PRIMARY);
+}
+
+
+/*
  * IsParticipatingInPromotion returns whether a node is currently participating
  * in a promotion, either as a candidate that IsBeingPromoted, or as a
  * "support" node that is reporting its LSN or re-joining as a secondary.
--- a/src/monitor/node_metadata.c
+++ b/src/monitor/node_metadata.c
@@ -1549,7 +1549,10 @@ CandidateNodeIsReadyToStreamWAL(AutoFailoverNode *node)
 			node->goalState == REPLICATION_STATE_STOP_REPLICATION
 
 			|| node->reportedState == REPLICATION_STATE_WAIT_PRIMARY ||
-			node->goalState == REPLICATION_STATE_WAIT_PRIMARY);
+			node->goalState == REPLICATION_STATE_WAIT_PRIMARY
+
+			|| node->reportedState == REPLICATION_STATE_PRIMARY ||
+			node->goalState == REPLICATION_STATE_PRIMARY);
 }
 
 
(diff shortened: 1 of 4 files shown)


DimCitus:
When performing the demo on Monday (2020-09-28) I reached a situation that would stall the group FSM on the monitor:

```
ha-admin@ha-demo-dim-monitor:~$ pg_autoctl show state
       Name |  Node |      Host:Port |        LSN | Reachable |       Current State |      Assigned State
------------+-------+----------------+------------+-----------+---------------------+--------------------
    paris-a |     4 | 10.11.1.6:5432 | 0/20000330 |       yes |           secondary |           secondary
    paris-b |     5 | 10.11.1.8:5432 | 0/20000178 |       yes |          report_lsn |          report_lsn
     node_6 |     6 | 10.11.1.5:5432 | 0/20000178 |       yes |          report_lsn |          report_lsn
amsterdam-a |     7 | 10.51.1.5:5432 | 0/20000178 |       yes |          report_lsn |          report_lsn
amsterdam-b |     8 | 10.51.1.6:5432 | 0/20000330 |       yes |        wait_primary |             primary
amsterdam-c |    11 | 10.51.1.4:5432 | 0/20000108 |        no |             demoted |             demoted
```

What happens is that node `paris-a` reaches secondary first and has `replication-quorum = false`, so the primary stays in wait_primary/primary states. Because of the current code, other secondary nodes then are not assigned `join_secondary` and remain in the `report_lsn` state, which means that no secondary participating in the replication quorum may make progress at this point.

The solution coded here is to accept to move forward any `report_lsn` node as soon as the selected new primary is in a position to stream WAL to them rather than only when it has reached a stable state (report state == assigned state is stable) in the Group FSM.

It would be best to be able to add a test case for the situation, but I anticipate that to require lots of work in the best case, so I wanted to first share the solution and see what Travis makes of it with our test suite.

JelteF (commented):
Looking at `IsBeingPromoted` it also returns true in case of `REPLICATION_STATE_FAST_FORWARD`. Is it possible for the node that supposed to be fast forwarded from to enter `REPLICATION_STATE_JOIN_SECONDARY` before this fast forward has completed?

Finally, I would be hapier if this had a test.

DimCitus:
> Looking at `IsBeingPromoted` it also returns true in case of `REPLICATION_STATE_FAST_FORWARD`. Is it possible for the node that supposed to be fast forwarded from to enter `REPLICATION_STATE_JOIN_SECONDARY` before this fast forward has completed?

That's a good analysis, thanks, and I think the previous code (first commit of the series in the PR) would be open to that problem. I have fixed it now by securing our move to later in the FSM progression.

> Finally, I would be hapier if this had a test.

I would too. I just don't have enough round tuits to make it happen this week, and I kind of anticipate writing a test case to require some level of efforts. We might be able to reproduce with just primary/sync standby/async standby and make sure that the sync standby stay stuck in REPORT_LSN longer than the other one, with a call to `node2.ifdown()` in the test suite or something like that, but we need to have the timing right.

DimCitus:
> Finally, I would be hapier if this had a test.

Now with a test case!

tests/test_multi_async.py
```diff
@@ -106,3 +107,74 @@ def test_006_async_failover():
...
+
+    # make sure we reached primary on node1 before next tests
+    assert node2.wait_until_state(target_state="primary")
+
+def test_010_promote_node1():
+    print()
+    print("Calling pgautofailover.perform_promotion(node1) on the monitor")
+
+    # we don't use node1.perform_promotion() here because using the
+    # pg_autoctl client means we would lsiten to notification and get back
```
JelteF, line 158:
```suggestion
    # pg_autoctl client means we would listen to notification and get back
```

src/monitor/node_metadata.c
```diff
@@ -1536,6 +1536,39 @@ IsBeingPromoted(AutoFailoverNode *node)
...
+			(node->reportedState == REPLICATION_STATE_STOP_REPLICATION &&
+			 (node->goalState == REPLICATION_STATE_STOP_REPLICATION ||
+			  node->goalState == REPLICATION_STATE_WAIT_PRIMARY)) ||
+
+			(node->reportedState == REPLICATION_STATE_WAIT_PRIMARY &&
+			 (node->goalState == REPLICATION_STATE_WAIT_PRIMARY ||
+			  node->goalState == REPLICATION_STATE_PRIMARY)) ||
+
+			(node->reportedState == REPLICATION_STATE_PRIMARY &&
+			 node->goalState == REPLICATION_STATE_PRIMARY));
```
JelteF, line 1568: In a second PR let's do this approach, with checking both reported and goal state, for `IsBeingPromoted` too.

JelteF (approved):

src/monitor/node_metadata.c
DimCitus: Yeah that's a good point, I'll work on that next.

JelteF (approved):

Issue #434, opened 2020-09-26
Monitor has to be restarted twice to trigger upgrade from 1.3 to 1.4


s4ke:
In our tests for upgrading automatically from 1.3 to 1.4 we found that we had to restart the monitor service twice after upgrading the libraries before it triggered the upgrade from 1.3 to 1.4.

This is the procedure we are using:

https://github.com/neuroforgede/pg_auto_failover_ansible/blob/master/roles/postgres-cluster-pg-auto-failover-upgrade-pre-1.4/tasks/upgrade.yml

If we leave out lines 73-77, the monitor does not trigger an upgrade. I was able to reproduce this multiple times. Once those lines are removed, I have to trigger the failover manually so that the step in lines 82-110 is completed.

The tests in question can be found here and require Vagrant:

https://github.com/neuroforgede/pg_auto_failover_ansible/tree/master/test/upgrade_1.3_to_1.4

EDIT:

one more thing: The docs state that we should watch for "ALTER EXTENSION pgautofailover UPDATE TO" in the logs, but at least for me when upgrading from 1.3 to 1.4 this did not show up.

JelteF:
So the flow that should be possible to use to upgrade pg_auto_failover from 1.3 to 1.4 is documented here:
https://pg-auto-failover.readthedocs.io/en/latest/operations.html#upgrading-from-previous-pg-auto-failover-versions

These steps assume a systemd managed `pg_autoctl` process. If it's not managed by systemd on the monitor you should also stop the postgres process, not just `pg_autoctl`. 

As you can see [in the section above](https://pg-auto-failover.readthedocs.io/en/latest/operations.html#upgrading-pg-auto-failover-from-versions-1-4-onwward) the upgrade flow for future versions should be much more straightforward.

s4ke:
We are provisioning the monitor node with a sytemd service like this:

```
- name: "create systemd config for cluster"
  shell: >
    su -c 'PATH="$PATH:{{ postgresql_cluster_bin_path }}" pg_autoctl -q show systemd --pgdata "{{ postgresql_cluster_data_dir }}"' postgres \
    | tee /etc/systemd/system/{{ postgresql_cluster_daemon }}.service
  when: (not pgdata_dir_version.stat.exists) and (postgresql_cluster_is_monitor | default('False') | bool)
```

This generates something similar to this (postgresql_main_cluster.service):

```
[Unit]
Description = pg_auto_failover

[Service]
[...]

PR #441, opened 2020-09-30
clarify that monitor might have to be restarted twice when upgrading from version < 1.4

s4ke:
#434


s4ke:
@DimCitus In light of #434 being fixed now, should we close this PR then?

DimCitus:
I believe so yes. Thanks for tracking it down, I might have skipped that one otherwise!

PR #435, opened 2020-09-29
Tutorial Part II.

DimCitus:
Add an interesting story where we migrate from one region to another thanks
to a short list of pg_auto_failover commands. This time this happens in
Europe.

PR #442, opened 2020-10-01
pg_autoctl do azure create region

DimCitus:
Work in progress, see #435


DimCitus:
I believe this PR is ready for review now. Here what I'm trying to achieve:

  1. make is very easy for contributors to run a full QA environment/setup on azure, that's two commands now
  2. offer an easier way to play around with Azure VMs from the tutorial
  3. prepare for the multi-region tutorial where we will need more automation because some Azure objects (gateways) can take up to 45 mins to be created. best would be to then have a fully automated tool which implements the waiting and finish the deployment as soon as the regions are connected together

JelteF:
I think one file is missing, probably `azure.c`

DimCitus:
> I think one file is missing, probably `azure.c`

```
$ git log -n1 -- src/bin/pg_autoctl/azure.c 
commit bc7925497a07a325af72750397dbaa6d6b05eb73 (HEAD -> feature/azure-scripting, origin/feature/azure-scripting)
Author: Dimitri Fontaine <dimitri@citusdata.com>
Date:   Mon Oct 5 18:18:10 2020 +0200
```

Similar results are shown here for `azure.h` and `cli_do_azure.c` and I don't think I have added other files. What could it be then? Can you share more about what's wrong on your side?

JelteF:
Ugh, it was collapsed on github and I missed it.

jonels-msft (commented):
I did want to bring up an alternate design and get your opinion. What about putting the azure commands into shell scripts that you bundle as "extras" in this repo rather than baking in pg_autoctl subcommands?

Here are possible **disadvantages of azure subcommands**:

1. It seems more verbose and error prone to, e.g. build argument arrays in azure.c than to run commands in a shell script.
1. It's harder for a user to preview what the command will do (or modify it) than with separate scripts. (Granted there's a `--script` option to obtain them)
1. You have to change and recompile pg_autoctl when the Azure command line interface changes.
1. As a potentially vendor-neutral tool, it's odd to bake in Azure knowledge. Might be easier in the long run to separate concerns.

Possible **advantages of azure subcommands**:

1. Simpler for packaging, it's a single self-contained binary. Doesn't need helper scripts installed on the machine.
2. Cooperates in a standard logging process.
3. Makes the functionality more discoverable since the program usage screen shows it. Provides a unified way to get `--help` for subcommands too.

Finally, what about a **hybrid approach**, where you can extend pg_autoctl with a subcommand `foo` by adding a script of the name `pg_autoctl-foo` to the path? That way the core subcommands can be written in the tool in C, and extra helpers can be distributed with the program as scripts. That's how git works, isn't it?

DimCitus:
Thanks for diving into the pros/cons of `pg_autoctl do azure...` subcommands, @jonels-msft. I agree with your analysis, and that's why I added the `--script` option. Also, I think we should still have the full Azure CLI copy/paste option in the docs, but not as the only option there. 

When we add the gateways for Site-to-Site Vnet VPN, this is ~45 mins of provisioning time, so I want to have a way to script around it. There's an azure command that waits until a resource is done being provisioned, so we might use that, but still, the copy/paste strategy might not be our best approach then.

I don't like the `git` approach where we would add sub-commands as external commands (script shell or something else), because it means installing more stuff into your PATH somehow. If in the end we're going to ship scripts, I'd rather develop said scripts in C and benefit from the other pieces we have put together for `pg_autoctl`. 

Also it helps with QA where a developer can now provision a whole cluster on Azure in two commands and like ten minutes, *from sources*, and play around with it. Getting down to a single command is not done yet because I'm yet to pick a spelling for the command.

```
$ pg_autoctl do azure tutorial single-region --prefix ha-demo --location useast
$ pg_autoctl do azure tutorial multi-regions --prefix ha-demo --location eastus --location westus --location francecentral
$ pg_autoctl do azure tutorial swap-primary-region --prefix ha-demo --location westus
```

Maybe it should be `pg_autoctl do azure scenario` rather than `tutorial`?

JelteF:
Did you consider instead of creating a shell script creating an ARM template? This should significantly speed up the provisioning time, since things will be provisioned in parallel: https://docs.microsoft.com/en-us/azure/azure-resource-manager/templates/overview#why-choose-arm-templates

DimCitus:
> Did you consider instead of creating a shell script creating an ARM template? This should significantly speed up the provisioning time, since things will be provisioned in parallel: https://docs.microsoft.com/en-us/azure/azure-resource-manager/templates/overview#why-choose-arm-templates

I'm discovering ARM templates now. There's a command to export an existing group as a template, so I have just done that for the current resource group that I have created with 5 VMs (app, monitor, 3 Postgres nodes) and the usual networking bits. That's a 1166 lines JSON file with variables and other ARM template specifics. I'm not sure how anyone would actually be able to review that before applying it to their own resource group. Also it contains my ssh public keys so I don't think it would work with anyone else.

I'm not convinced that Azure ARM templates is the way to go for us and I will continue diving into this area. I don't want to depend on things such as Terraform and the like, but re-inventing those would be even worse. I still think the sweet spot is a single script to automate that, and that including such a script in `pg_autoctl` itself makes it easier for us to maintain and ship the script. That it happens to be a script written in C does not make me happy, but at least we benefit from all the utility code in the project.

jonels-msft:
I reviewed the docs and will add my comments tomorrow.

docs/tutorial-script-single.rst
```diff
@@ -0,0 +1,280 @@
+.. _script_single:
```
jonels-msft, line 1: Since this file is not included in the table of contents, it should be marked

```rst
:orphan:
```

on the first line. That'll fix a sphinx warning.

docs/tutorial-script-single.rst
```diff
@@ -0,0 +1,280 @@
...
+platform. The techniques here are relevant to any cloud provider or
+on-premise network. We use four virtual machines: a primary database, a
+secondary database, a monitor, and an "application." The monitor watches the
+other nodes’ health, manages global state, and assigns nodes their roles.
+
+This document shows you how to create the Azure environment with the virtual
+machines needed for the rest of the "single standby" tutorial. You have two
+options:
+
+  1. use a pg_autoctl command that automates all the steps so that you can
```
jonels-msft, line 21: Although this list explains that the `pg_autoctl do azure …` and `az … create` methods are two alternatives, the structure of the article makes it look like the reader should run both of them, one after another.

What if we split the two methods into their own pages, and linked to them in this list?

docs/tutorial-script-single.rst
```diff
@@ -0,0 +1,280 @@
...
+    ha-demo-nsg                      Microsoft.Network/networkSecurityGroups                eastus
+    ha-demo-a-ip                     Microsoft.Network/publicIPAddresses                    eastus
+    ha-demo-app-ip                   Microsoft.Network/publicIPAddresses                    eastus
+    ha-demo-b-ip                     Microsoft.Network/publicIPAddresses                    eastus
+    ha-demo-monitor-ip               Microsoft.Network/publicIPAddresses                    eastus
+    ha-demo-net                      Microsoft.Network/virtualNetworks                      eastus
+
+.. _script_single_automated:
+
+Using the pg_autoctl Azure integration commands
```
jonels-msft, line 147: This is where an unwary person might continue copy and pasting code blocks, and try to recreate a region unnecessarily.

docs/tutorial-script-single.rst
```diff
@@ -0,0 +1,280 @@
...
+      --nodes 2
+
+Finaly, the option ``--from-source`` allows to provision the VM nodes from
+your current local pg_auto_failover source tree, using ``rsync`` to upload
+your local sources to the VMs:
+
+.. code-block:: bash
+
+   PG_AUTOCTL_DEBUG=1 pg_autoctl do azure create region \
+      --prefix ha-demo-dim \
```
jonels-msft, line 190: In the previous section where we ran commands by hand, we used `ha-demo`.

Why does this one change to `ha-demo-dim`?

docs/tutorial-script-single.rst
```diff
@@ -0,0 +1,280 @@
...
+Using the pg_autoctl Azure integration commands
+-----------------------------------------------
+
+The following command creates a new Azure resource group and then deploy
+some network settings (vnet, nsg, nsg rules, subnet), and then creates the
+target Virtual Machines and provision them:
+
+.. code-block:: bash
+
+   PG_AUTOCTL_DEBUG=1 pg_autoctl do azure create region \
```
jonels-msft, line 156: I love the logging output from this command, it's clean and informative. :)

docs/tutorial-script-single.rst
```diff
@@ -0,0 +1,280 @@
...
+   ha-demo-dim-paris-b        51.103.34.39      10.11.11.6
+   ha-demo-dim-paris-monitor  51.11.244.130     10.11.11.4
+
+
+To connect to the monitor and run the ``pg_autoctl show state`` command you
+can use the following command:
+
+.. code-block:: bash
+
+   $ PG_AUTOCTL_DEBUG=1 pg_autoctl do azure show state --prefix ha-demo-dim --region paris
```
jonels-msft, line 265: Also saw only node B created when I ran this.

jonels-msft (changes requested):
Added some questions, and a suggestion about the flow.

I also noticed things to fix in tutorial.rst, but they are unrelated to the changes here, and I'll deal with them elsewhere.

DimCitus:
Replaced by https://github.com/citusdata/pg_auto_failover/pull/544 that is intended at helping us do QA, without impact on docs.

Issue #443, opened 2020-10-01
How do we run tests locally? 


rheaton:
Hello,

I am a new contributor, and I am having a difficult time running the python tests and getting my machine set up. Is there documentation to help contributors write and run tests? 

Thanks,
Rachel

-------

Note:
I'm are on a mac Catalina 10.15

Steps tried:
1. Installed psycopg2 (On my computer, v 2.7.5 wouldn't install, so used v 2.8.6 and updated Pipfile) `pip install psycopg2`
1. `PIPENV_PIPFILE=test/Pipfile  pipenv install --python=$(which python)`
1. `make test`

```
sudo -E env "PATH=/Users/rheaton/.cargo/bin:/usr/local/opt/mysql@5.7/bin:/usr/local/opt/python/libexec/bin:/usr/local/opt/ruby/bin:/Users/rheaton/go/bin:/usr/local/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/Applications/Wireshark.app/Contents/MacOS:/Users/rheaton/workspace/core-services-wiki/scripts:/Users/rheaton/.fzf/bin:/usr/local/sbin:/usr/local/Cellar/postgresql/12.4/bin:/Users/rheaton/Library/Python/3.8/bin" USER=rheaton \
		/usr/local/bin/nosetests			\
		--verbose				\
		--nocapture				\
		--stop					\
		--where=tests
ERROR

======================================================================
ERROR: test suite for <module 'test_auth' from '/Users/rheaton/workspace/pg_auto_failover/tests/test_auth.py'>
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/site-packages/nose-1.3.7-py3.8.egg/nose/suite.py", line 210, in run
    self.setUp()
  File "/usr/local/lib/python3.8/site-packages/nose-1.3.7-py3.8.egg/nose/suite.py", line 293, in setUp
    self.setupContext(ancestor)
  File "/usr/local/lib/python3.8/site-packages/nose-1.3.7-py3.8.egg/nose/suite.py", line 316, in setupContext
    try_run(context, names)
  File "/usr/local/lib/python3.8/site-packages/nose-1.3.7-py3.8.egg/nose/util.py", line 471, in try_run
    return func()
  File "/Users/rheaton/workspace/pg_auto_failover/tests/test_auth.py", line 12, in setup_module
    cluster = pgautofailover.Cluster()
[...]

PR #444, opened 2020-10-02
Add documentation about how to run tests to CONTRIBUTING.md

rheaton:
Resolves: https://github.com/citusdata/pg_auto_failover/issues/443

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -56,6 +56,25 @@ __EOF__
 chmod +x .git/hooks/pre-commit
 ```
 
+### Running tests
+
+The integration tests are written using Python and the
+[nose](https://nose.readthedocs.io/en/latest/index.html) testing framework.
+They are run in a docker container, so you need
+[docker](https://docs.docker.com/get-docker/) installed locally.
+
+```bash
+make run-test
+```
+
+You can filter the tests you are running with the `TEST` environment variable.
+
+```bash
+make TEST=multi run-test       # runs tests matching tests/test_multi*
+make TEST=single run-test      # runs tests _not_ matching tests/test_multi*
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -50,3 +50,6 @@ docs/tikz/*.png
 
 # Exclude our demo/test tmux directory
 tmux/
+
+# Exclude CLion directory
+.idea/
```
DimCitus, line 55: I think personal workflow and tooling exclude list should not make it to the shared repository. It's good that we all use our favorite tool, the shared repository should only deal with what our build process produces. Please consider moving that to `.git/info/exclude` instead.

CONTRIBUTING.md
```diff
@@ -56,6 +56,25 @@ __EOF__
...
+```bash
+make run-test
+```
+
+You can filter the tests you are running with the `TEST` environment variable.
+
+```bash
+make TEST=multi run-test       # runs tests matching tests/test_multi*
+make TEST=single run-test      # runs tests _not_ matching tests/test_multi*
+make TEST=test_sample run-test # runs tests/test_sample.py
```
DimCitus, line 75: Can we use an example that you can run? this one gives me `ERROR: Failure: ImportError (No module named 'test_sample')` because we don't have a test_sample file. A short one that runs in a short amount of time (about 30s on my laptop) is `make TEST=test_auth run-test`.

DimCitus (changes requested):
Thanks for this PR! It's looking good, and I think that after a couple more edits we will be able to merge it!

rheaton:
@DimCitus Updated to match your comments! :)

DimCitus (approved):

PR #446, opened 2020-10-05
After removing a standby node, transition primary to apply settings

### LINKED ISSUE #445: Multi-standby: Removing a standby doesn't update synchronous_standby_names on primary
When there are at least 2 standby nodes, the following scenario can be created:

## Version
Master: db44bcb2feed71ceda746f5be348d62cbd93939c

## Summary
When we have a multi-standby setup, and we drop a standby from the cluster with `pg_autoctl drop node` does not update the` synchronous_standby_names` GUC on the primary to exclude the dropped standby. This can lead to situations where transactions can get stuck on the primary.

## Repro
1. Create a cluster with one primary and 3 standbys (default `replication_quorum` and `candidate_priority` )
1. `pg_autoctl set formation number-sync-standbys 2 --pgdata ~/failover/monitor/` 
1. On the primary:
```
postgres=# show synchronous_standby_names ;
               synchronous_standby_names                
--------------------------------------------------------------------------------------
 ANY 2 (pgautofailover_standby_2, pgautofailover_standby_3, pgautofailover_standby_4)
(1 row)
```
1. Let’s drop *two* of the standbys using `pg_autoctl drop node --destroy`
1. On the primary we can see that synchronous_standby_names has not been updated:
```
postgres=# show synchronous_standby_names ;
               synchronous_standby_names                
--------------------------------------------------------------------------------------
 ANY 2 (pgautofailover_standby_2, pgautofailover_standby_3, pgautofailover_standby_4)
(1 row)
```
 6. As a consequence transactions hang on the primary.

Branch with a failing test, illustrating the scenario: https://github.com/citusdata/pg_auto_failover/compare/master...rheaton:synchronous-standby-names-primary?expand=1

rheaton:
The primary node is assigned to APPLY_SETTINGS_STATE so that the synchronous_standby_names are correctly updated on the primary.

Fixes #445

with @soumyadeep2007 and @jchampio

### DIFF
--- a/src/monitor/node_active_protocol.c
+++ b/src/monitor/node_active_protocol.c
@@ -1129,6 +1129,14 @@ RemoveNode(AutoFailoverNode *currentNode)
 		if (primaryNode)
 		{
 			(void) ProceedGroupState(primaryNode);
+			LogAndNotifyMessage(
+				message, BUFSIZE,
+				"Setting goal state of %s:%d to apply_settings "
+				"after removing standby node %s:%d from formation %s.",
+				primaryNode->nodeHost, primaryNode->nodePort,
+				currentNode->nodeHost, currentNode->nodePort,
+				formation->formationId);
+			SetNodeGoalState(primaryNode, REPLICATION_STATE_APPLY_SETTINGS, message);
 		}
 	}
 
--- a/src/monitor/node_active_protocol.c
+++ b/src/monitor/node_active_protocol.c
@@ -1108,22 +1108,6 @@ RemoveNode(AutoFailoverNode *currentNode)
 			formation->number_sync_standbys,
 			formation->formationId,
 			countSyncStandbys);
-
-		AutoFailoverNode *primaryNode =
-			GetPrimaryNodeInGroup(currentNode->formationId,
-								  currentNode->groupId);
-
-		if (primaryNode)
-		{
-			LogAndNotifyMessage(
-				message, BUFSIZE,
-				"Setting goal state of %s:%d to apply_settings "
-				"after removing standby node %s:%d from formation %s.",
-				primaryNode->nodeHost, primaryNode->nodePort,
-				currentNode->nodeHost, currentNode->nodePort,
-				formation->formationId);
-			SetNodeGoalState(primaryNode, REPLICATION_STATE_APPLY_SETTINGS, message);
-		}
 	}
 
 	/* now proceed with the failover, starting with the first standby */
@@ -1145,6 +1129,12 @@ RemoveNode(AutoFailoverNode *currentNode)
 		if (primaryNode)
 		{
 			(void) ProceedGroupState(primaryNode);
+			if (primaryNode->goalState == primaryNode->reportedState)
+			{
+				const char *message = "whatever";
+				SetNodeGoalState(primaryNode,
+								 REPLICATION_STATE_APPLY_SETTINGS, message);
+			}
 		}
 	}
 
--- a/src/monitor/node_active_protocol.c
+++ b/src/monitor/node_active_protocol.c
@@ -1108,6 +1108,22 @@ RemoveNode(AutoFailoverNode *currentNode)
 			formation->number_sync_standbys,
 			formation->formationId,
 			countSyncStandbys);
+
+		AutoFailoverNode *primaryNode =
+			GetPrimaryNodeInGroup(currentNode->formationId,
+								  currentNode->groupId);
+
+		if (primaryNode)
+		{
+			LogAndNotifyMessage(
+				message, BUFSIZE,
+				"Setting goal state of %s:%d to apply_settings "
+				"after removing standby node %s:%d from formation %s.",
+				primaryNode->nodeHost, primaryNode->nodePort,
+				currentNode->nodeHost, currentNode->nodePort,
+				formation->formationId);
+			SetNodeGoalState(primaryNode, REPLICATION_STATE_APPLY_SETTINGS, message);
+		}
 	}
 
 	/* now proceed with the failover, starting with the first standby */
@@ -1129,14 +1145,6 @@ RemoveNode(AutoFailoverNode *currentNode)
 		if (primaryNode)
 		{
 			(void) ProceedGroupState(primaryNode);
-			LogAndNotifyMessage(
-				message, BUFSIZE,
-				"Setting goal state of %s:%d to apply_settings "
-				"after removing standby node %s:%d from formation %s.",
-				primaryNode->nodeHost, primaryNode->nodePort,
-				currentNode->nodeHost, currentNode->nodePort,
-				formation->formationId);
-			SetNodeGoalState(primaryNode, REPLICATION_STATE_APPLY_SETTINGS, message);
 		}
 	}
 
--- a/tests/pgautofailover_utils.py
+++ b/tests/pgautofailover_utils.py
@@ -1062,7 +1062,7 @@ SELECT reportedstate
 
     def get_synchronous_standby_names(self):
         """
-            Gets number sync standbys  via pg_autoctl
+            Gets synchronous standby names via pg_autoctl
         """
         command = PGAutoCtl(self)
         out, err, ret = command.execute("get synchronous_standby_names",
@@ -1072,7 +1072,7 @@ SELECT reportedstate
 
     def get_synchronous_standby_names_local(self):
         """
-            Gets number sync standbys  via pg_autoctl
+            Gets synchronous standby names via sql query on data node
         """
         query = "select current_setting('synchronous_standby_names')"
 
--- a/tests/pgautofailover_utils.py
+++ b/tests/pgautofailover_utils.py
@@ -1070,6 +1070,22 @@ SELECT reportedstate
 
         return out.strip()
 
+    def get_synchronous_standby_names_local(self):
+        """
+            Gets number sync standbys  via pg_autoctl
+        """
+        query = "select current_setting('synchronous_standby_names')"
+
+        result = self.run_sql_query(query)
+        return result[0][0]
+
+    def print_synchronous_standby_names(self):
+        print("synchronous_standby_names       = '%s'\nsynchronous_standby_names_local = '%s'" %
+              (self.get_synchronous_standby_names(),
+               self.get_synchronous_standby_names_local())
+              )
+        return
+
     def list_replication_slot_names(self):
         """
             Returns a list of the replication slot names on the local Postgres.
--- a/tests/pgautofailover_utils.py
+++ b/tests/pgautofailover_utils.py
@@ -964,6 +964,26 @@ SELECT reportedstate
         command.execute("drop node", 'drop', 'node')
         return True
 
+    def do_fsm_assign(self, target_state):
+        """
+        Runs `pg_autoctl do fsm assign` on a node
+
+        :return:
+        """
+        command = PGAutoCtl(self)
+        command.execute("do fsm assign", 'do', 'fsm', 'assign', target_state)
+        return True
+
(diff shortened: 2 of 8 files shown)


tests/pgautofailover_utils.py
```diff
@@ -1070,6 +1070,22 @@ def get_synchronous_standby_names(self):
 
         return out.strip()
 
+    def get_synchronous_standby_names_local(self):
+        """
+            Gets number sync standbys  via pg_autoctl
```
jchampio, line 1075: Just noticed that this, as well as the `get_synchronous_standby_names` docstring, are copies of the `get_number_sync_standbys` docstring.

src/monitor/node_active_protocol.c
```diff
@@ -1129,6 +1129,14 @@ RemoveNode(AutoFailoverNode *currentNode)
...
 		{
 			(void) ProceedGroupState(primaryNode);
+			LogAndNotifyMessage(
+				message, BUFSIZE,
+				"Setting goal state of %s:%d to apply_settings "
+				"after removing standby node %s:%d from formation %s.",
+				primaryNode->nodeHost, primaryNode->nodePort,
+				currentNode->nodeHost, currentNode->nodePort,
+				formation->formationId);
+			SetNodeGoalState(primaryNode, REPLICATION_STATE_APPLY_SETTINGS, message);
```
DimCitus, line 1139: I don't think we want to do that when `ProceedGroupState` sets the goal state for the primary node to anything other than PRIMARY, such as WAIT_PRIMARY or JOIN_PRIMARY. In those cases, we should also verify that the `synchronous_standby_names` is fetched again from the monitor, but I think it is, at least for WAIT_PRIMARY.

tests/test_multi_standbys.py
```diff
@@ -111,20 +111,21 @@ def test_005_number_sync_standbys():
     print("set number_sync_standbys = 2")
     assert node1.set_number_sync_standbys(2)
     assert node1.get_number_sync_standbys() == 2
-    print("synchronous_standby_names = '%s'" %
-          node1.get_synchronous_standby_names())
+    node1.print_synchronous_standby_names()
+    assert node1.get_synchronous_standby_names() == node1.get_synchronous_standby_names_local()
```
DimCitus, line 115: It's best to use `eq_` from the nosetest package, like we do in other places. In case of failures it gives more details about what's wrong, making it easier to debug.

tests/test_multi_standbys.py
```diff
@@ -111,20 +111,21 @@ def test_005_number_sync_standbys():
...
 
     print("set number_sync_standbys = 0")
     assert node1.set_number_sync_standbys(0)
     assert node1.get_number_sync_standbys() == 0
-    print("synchronous_standby_names = '%s'" %
-          node1.get_synchronous_standby_names())
+    node1.print_synchronous_standby_names()
+    assert node1.get_synchronous_standby_names() == node1.get_synchronous_standby_names_local()
```
DimCitus, line 121: Same, let's use `eq_`.

tests/test_multi_standbys.py
```diff
@@ -111,20 +111,21 @@ def test_005_number_sync_standbys():
...
 
     print("set number_sync_standbys = 1")
     assert node1.set_number_sync_standbys(1)
     assert node1.get_number_sync_standbys() == 1
-    print("synchronous_standby_names = '%s'" %
-          node1.get_synchronous_standby_names())
+    node1.print_synchronous_standby_names()
+    assert node1.get_synchronous_standby_names() == node1.get_synchronous_standby_names_local()
```
DimCitus, line 127: Let's use `eq_`.

tests/test_multi_standbys.py
```diff
@@ -133,6 +134,8 @@ def test_006_number_sync_standbys_trigger():
     node4.drop()
     assert node1.get_number_sync_standbys() == 1
     assert node1.wait_until_state(target_state="primary")
+    node1.print_synchronous_standby_names()
+    assert node1.get_synchronous_standby_names() == node1.get_synchronous_standby_names_local()
```
DimCitus, line 138: `eq_`

DimCitus (changes requested):
Overriding the decision made by `ProceedGroupState` in every possible case seems wrong to me, or at least very surprising. Should we implement the fix that way, at least we should certainly stop calling into `ProceedGroupState` entirely, no?

src/monitor/node_active_protocol.c
```diff
@@ -1108,6 +1108,22 @@ RemoveNode(AutoFailoverNode *currentNode)
...
+		if (primaryNode)
+		{
+			LogAndNotifyMessage(
+				message, BUFSIZE,
+				"Setting goal state of %s:%d to apply_settings "
+				"after removing standby node %s:%d from formation %s.",
+				primaryNode->nodeHost, primaryNode->nodePort,
+				currentNode->nodeHost, currentNode->nodePort,
+				formation->formationId);
+			SetNodeGoalState(primaryNode, REPLICATION_STATE_APPLY_SETTINGS, message);
```
jchampio, line 1125: Can this have the opposite problem, where the primary node's goal state is overwritten by `ProceedGroupState` without anything actually happening? @soumyadeep2007 and I are wondering if perhaps `ProceedGroupState` needs to _know_ that a node has been dropped, somehow.

rheaton: I was wondering this too -- I did replace the `ProceedGroupState` call with this one, and all the tests passed, but I went with Dimitri's advice for the push up. I'm not sure about the test coverage.

DimCitus: So, `synchronous_standby_names` is computed on the monitor in a function of the same name at https://github.com/citusdata/pg_auto_failover/blob/master/src/monitor/node_active_protocol.c#L2127. This function is called from the client side implementation of the “node-active protocol” from the function `fsm_apply_settings` at https://github.com/citusdata/pg_auto_failover/blob/master/src/bin/pg_autoctl/fsm_transition.c#L626. Without surprise it's called in the transition from PRIMARY to APPLY_SETTINGS already (see https://github.com/citusdata/pg_auto_failover/blob/master/src/bin/pg_autoctl/fsm.c#L320):

``` c
	{ PRIMARY_STATE, APPLY_SETTINGS_STATE, COMMENT_PRIMARY_TO_APPLY_SETTINGS, &fsm_apply_settings },
```

The `fsm_apply_settings` is also called from https://github.com/citusdata/pg_auto_failover/blob/master/src/bin/pg_autoctl/fsm_transition.c#L555 in the function `fsm_enable_sync_rep` which is used in more places in the FSM (https://github.com/citusdata/pg_auto_failover/blob/master/src/bin/pg_autoctl/fsm.c#L268):

``` c
	{ WAIT_PRIMARY_STATE, PRIMARY_STATE, COMMENT_WAIT_PRIMARY_TO_PRIMARY, &fsm_enable_sync_rep },
	{ JOIN_PRIMARY_STATE, PRIMARY_STATE, COMMENT_JOIN_PRIMARY_TO_PRIMARY, &fsm_enable_sync_rep },
```

Now, when removing a node from the group that was not the primary node, the primary node might be in any of SINGLE, PRIMARY, WAIT_PRIMARY, JOIN_PRIMARY, or APPLY_SETTINGS state. So a standby node might be removed during its WAIT_STANDBY to CATCHINGUP transition, say, allowing the primary to go from JOIN_PRIMARY to either WAIT_PRIMARY or PRIMARY. In that case, some transition will happen on the client side. 

I think we should only assign APPLY_SETTINGS when the primary node was already in the PRIMARY before the node removal, and still is in the PRIMARY state after the node removal. From there, we should probably extend that to “the state on the primary is the same after calling `ProceedGroupState`. If `ProceedGroupState` has assigned a state, that's the new state we should reach, that's it.

Now, to make the `synchronous_standby_names` change that is needed, we should extend the existing FSM transitions in the client side to call `fsm_apply_settings` where it's currently missing.

rheaton:
@DimCitus We are going to pick this up next week, when I have more time for OSS work :)

jchampio:
@DimCitus While implementing the `APPLY_SETTINGS` suggestion, we ran into a race that probably warrants a closer look.

>I think we should only assign APPLY_SETTINGS when the primary node was already in the PRIMARY before the node removal, and still is in the PRIMARY state after the node removal. From there, we should probably extend that to “the state on the primary is the same after calling ProceedGroupState”. If ProceedGroupState has assigned a state, that's the new state we should reach, that's it.

The issue with this is that the primary could already be in a state transition before we call `ProceedGroupState`. In our failing race, we see this when we drop a node at the same time that the primary is going through a `join_primary -> primary` transition. So `RemoveNode` sees that `join_primary -> primary` is already happening and does not ask for `apply_settings`, but the primary node may have grabbed a stale `synchronous_standby_names` from before the drop operation.

Here's a full sequence to illustrate:
```
+-------+            +---------+                                    +---------+       
| user  |            | monitor |                                    | primary |       
+-------+            +---------+                                    +---------+       
    |                     |                                              |            
    |                     | ------------------------                     | ------------------------
    |                     |-| standbys = (1, 2, 3) |                     |-| standbys = (1, 2, 3) |
    | add secondary 4     | ------------------------                     | ------------------------
    |-------------------->|                                              |            
    |                     | --------------------------------             |            
    |                     | | add node                     |             |
    |                     |-| standbys = (1, 2, 3, 4)      |             |
    |                     | | start new primary transition |             |
    |                     | -------------------------------|             |
    |                     |                                              |            
    |                     | request primary -> join_primary              |            
    |                     |--------------------------------------------> | transition 
    |                     |                                              |----------- 
    |                     |                                              |          | 
    |                     |                                         done |<---------- 
    |                     | <--------------------------------------------|            
    |                     |                                              |            
    |                     | request join_primary -> primary              |            
    |                     |--------------------------------------------->| transition 
    |                     |                                              |-----------
    |                     |                get synchronous_standby_names |          |
    |                     |<---------------------------------------------|          |
    | drop secondary 2    |--------------------------------------------->| ---------------------------
    |-------------------->|                                              |-| standbys = (1, 2, 3, 4) |
    |                     | ---------------------------------            | ---------------------------
    |                     | | drop node                     |            |          |
    |                     |-| standbys = (1, 3, 4)          |            |          |
    |                     | | primary already in transition |            |          |
    |                     | ---------------------------------            |          |
    |                     |                                         done |<---------- 
    |                     |<---------------------------------------------|            
    |                     |                                              |    
    |                     | ------------------------                     | ---------------------------
    |                     |-| standbys = (1, 3, 4) |                     |-| standbys = (1, 2, 3, 4) |
    |                     | ------------------------                     | ---------------------------
```

If `apply_settings` really needs to be a state, then it seems like we have to have some concept that tells the primary, "move to `apply_settings` as soon as you're done with your current transition." But that doesn't feel particularly robust to me.

Combined with the fact that we're having to introduce new `apply_settings_*` states so that we make sure that we don't accidentally move from e.g. `wait_primary -> apply_settings -> primary`, I think the `apply_settings` state is starting to show some cracks.

DimCitus:
Thanks @jchampio for the very detailed analysis here. Thanks to this summary, I think we should see about removing the APPLY_SETTINGS state entirely, and instead work in `keeper_node_active` directly, same as we do when we call [keeper_refresh_other_nodes](https://github.com/citusdata/pg_auto_failover/blob/master/src/bin/pg_autoctl/service_keeper.c#L705) and then add entries to the HBA when that's needed.

DimCitus:
Also this reminds me I wanted to explore using https://texample.net/tikz/examples/sequence-diagram/ in our documentation diagrams about the FSM in https://pg-auto-failover.readthedocs.io/en/latest/failover-state-machine.html ; it looks like quite some work but having a couple well chosen examples in there might help understand our design much better.

PR #447, opened 2020-10-07
Docs/tikz color review

DimCitus:
Review the colours and fonts and text size and make the diagrams easier to read and understand.


docs/tikz/arch-single-standby.tex
```diff
@@ -1,32 +1,49 @@
...
+\definecolor{mtxt}{HTML}{2F2F2F} % off-black
+
+\definecolor{apbox}{HTML}{0078D4} % MS blue
+\definecolor{aptxt}{HTML}{2F2F2F} % off-black
+
+%% sans-serif fonts, large by default, and bold too
+\sffamily
+\sbweight
+\bfseries
+\large
```
JelteF, line 28: You can use `\input{some-shared-file.tex}` to include another `.tex` file. I think it's good to have some shared file(s) where we define these colors and import packages and define hese colors etc.

docs/tikz/arch-multi-standby.tex
```diff
@@ -1,78 +1,103 @@
...
 
-  \node  (app) at (-6,18)  [app]        {Application};
-  \node  (m)   at (6,18)   [monitor]    {Monitor};
+  \node  (app) at (-8,18)  [app]        {\textbf{\Large Application}};
+  \node  (m)   at (8,18)   [monitor]    {\textbf{\Large Monitor}};
 
   \tikzstyle{sql}=[->,color=blue!60,text=black!90,line width=0.15em]
   \tikzstyle{sqlf}=[->,color=blue!35,text=black!90,line width=0.15em,loosely dashed]
   \tikzstyle{sr}=[>->,color=black!70,text=black!90,line width=0.15em]
-  \tikzstyle{hc}=[<->,color=green!50,text=black!90,line width=0.15em,dotted]
```
JelteF, line 64: The arrows still use the old colors. I guess these should also be changed, right?

JelteF (changes requested):
Overall, awesome improvements!

Both 3 standby images now look a bit messed up (nodes not next to eachother, text overlapping arrows):
afbeelding

And I left two comments.

Sidenote: the rich svg diff tool that github has is awesome.

docs/tikz/arch-single-standby.tex
DimCitus: Good point, done that way now!

docs/tikz/arch-multi-standby.tex
DimCitus: Also good point, fixed to use the same colour as the boxes now.

DimCitus:
> Overall, awesome improvements!

Thanks to Claire who had lots of good input about the diagrams, from herself directly and also from colleagues specialised in diagrams. So that's the result with a pro eye, and I like it too!

> Both 3 standby images now look a bit messed up (nodes not next to eachother, text overlapping arrows):

Fixed now.

> Sidenote: the rich svg diff tool that github has is awesome.

Mmm, I didn't see that one...

docs/tikz/arch-multi-standby.tex
```diff
@@ -30,24 +18,7 @@
...
-	  minimum height=2.5cm,minimum width=4cm,inner sep=0.5cm]
-
-  \tikzstyle{primary}=[mpnode,text=ptxt,draw=white,
-  	  rectangle split part fill={pbox,pbox,white}]
-
-  \tikzstyle{standby}=[mpnode,text=stxt,draw=white,
-	  rectangle split part fill={sbox,sbox,white}]
-
-  \tikzstyle{monitor}=[node,text=mtxt,draw=mbox,fill=mbox]
+  \input{common.tex}
```
JelteF, line 21: This now has these empty blocks, because the definition of primary and standby changed because of the shared code.

afbeelding

JelteF (changes requested):

docs/tikz/arch-multi-standby.tex
DimCitus: Ah yeah, that's why the arrow is shorter. Fixed now!

JelteF (approved):

PR #455, opened 2020-10-13
Ensure that the monitor owns the Postgres (sub-)process.

### LINKED ISSUE #434: Monitor has to be restarted twice to trigger upgrade from 1.3 to 1.4

DimCitus:
When we don't own the Postgres subprocess, we can't ensure that a pending
monitor upgrade is being done properly. This is mostly a problem when
upgrading from 1.3.1 to 1.4(.0, fix going to appeart in .1).

Fixes #434

### DIFF
--- a/src/bin/pg_autoctl/defaults.h
+++ b/src/bin/pg_autoctl/defaults.h
@@ -73,6 +73,7 @@
 #define PG_AUTOCTL_KEEPER_SLEEP_TIME 1      /* seconds */
 #define PG_AUTOCTL_KEEPER_RETRY_TIME_MS 350 /* milliseconds */
 #define PG_AUTOCTL_MONITOR_SLEEP_TIME 10 /* seconds */
+#define PG_AUTOCTL_MONITOR_RETRY_TIME 1  /* seconds */
 
 #define PG_AUTOCTL_LISTEN_NOTIFICATIONS_TIMEOUT 60
 
--- a/src/bin/pg_autoctl/pgsql.c
+++ b/src/bin/pg_autoctl/pgsql.c
@@ -869,6 +869,13 @@ pgsql_execute_with_params(PGSQL *pgsql, const char *sql, int paramCount,
 			strlcpy(ctx->sqlstate, sqlstate, SQLSTATE_LENGTH);
 		}
 
+		/* if we get a connection exception, track that */
+		if (sqlstate &&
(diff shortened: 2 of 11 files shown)


JelteF:
This doesn't seem to work for me:
```bash
pg_autoctl create monitor --pgdata monitor --auth trust --ssl-self-signed
pg_ctl start -D monitor
pg_autoctl run --pgdata monitor
```

Looking at `htop` the postgres process is not a child of the monitor after running these commands.

DimCitus:
> This doesn't seem to work for me:

Yes I realise this morning that when I fixed the test suite I basically de-activated the fix.

``` c
			if (pg_setup_is_ready(pgSetup, pgIsNotRunningIsOk) &&
				!read_pidfile(pgServicePidFile, &pgServicePid))
```

The Postgres controller is started first, so there is no point in time where the Postgres service pid file would not already exist when the listener is started, short of race conditions. Today I will try to solve this differently, maybe the Postgres controller code is where this should happen, but then conditionally, we might not want to restart Postgres on a keeper, at least not at `pg_autoctl create postgres` time.

DimCitus:
It should be all good now. I have fixed the approach to avoid the “whack-a-mole” game in between the two processes (listener and postgres controller) and instead implemented the new `PG_EXPECTED_STATUS_RUNNING_AS_SUBPROCESS` enum value for `ExpectedPostgresStatus`.

JelteF (approved):

Issue #452, opened 2020-10-09
Second node creation fails because of pg_hba issue


kevinelliott:
Setting up the monitor on the monitor host has no problems:

```
postgres@db1-monitor:~$ pg_autoctl create monitor --ssl-self-signed --hostname db1-monitor.int --auth trust --run  --pgctl /usr/lib/postgresql/12/bin/pg_ctl 
18:07:16 5828 INFO  Using default --ssl-mode "require"
18:07:16 5828 INFO  Using --ssl-self-signed: pg_autoctl will create self-signed certificates, allowing for encrypted network traffic
18:07:16 5828 WARN  Self-signed certificates provide protection against eavesdropping; this setup does NOT protect against Man-In-The-Middle attacks nor Impersonation attacks.
18:07:16 5828 WARN  See https://www.postgresql.org/docs/current/libpq-ssl.html for details
18:07:16 5828 INFO  Initialising a PostgreSQL cluster at "./monitor"
18:07:16 5828 INFO  /usr/lib/postgresql/12/bin/pg_ctl initdb -s -D ./monitor --option '--auth=trust'
18:07:17 5828 INFO   /usr/bin/openssl req -new -x509 -days 365 -nodes -text -out /var/lib/postgresql/monitor/server.crt -keyout /var/lib/postgresql/monitor/server.key -subj "/CN=db1-monitor.int"
18:07:17 5828 INFO  Started pg_autoctl postgres service with pid 5848
18:07:17 5828 INFO  Started pg_autoctl listener service with pid 5849
18:07:17 5848 INFO   /usr/bin/pg_autoctl do service postgres --pgdata ./monitor -v
18:07:17 5854 INFO   /usr/lib/postgresql/12/bin/postgres -D /var/lib/postgresql/monitor -p 5000 -h *
18:07:17 5848 INFO  Postgres is now serving PGDATA "/var/lib/postgresql/monitor" on port 5000 with pid 5854
18:07:17 5849 WARN  NOTICE:  installing required extension "btree_gist"
18:07:17 5849 INFO  Granting connection privileges on 10.116.0.0/20
18:07:17 5849 INFO  Your pg_auto_failover monitor instance is now ready on port 5000.
18:07:17 5849 INFO  Monitor has been successfully initialized.
18:07:17 5849 INFO   /usr/bin/pg_autoctl do service listener --pgdata ./monitor -v
[...]

PR #458, opened 2020-10-15
Implement HBA hostname DNS checks, and warn users when needed.

DimCitus:
Postgres HBA implementation with hostname is pretty picky, in order to be
secure enough. The reverse DNS lookup requirements are often missed, and
sometimes hard or impossible to achieve (Cloud VM environments).

As more users get confused with their hostname based setup, we should
provide more guidance when that's possible.

See #452.

### DIFF
--- a/src/bin/pg_autoctl/ipaddr.c
+++ b/src/bin/pg_autoctl/ipaddr.c
@@ -594,3 +594,95 @@ findHostnameFromLocalIpAddress(char *localIpAddress, char *hostname, int size)
 
 	return true;
 }
+
+
+/*
+ * resolveHostnameForwardAndReverse returns true when we could do a forward DNS
+ * lookup for the hostname, and then a reverse DNS lookup, and the reverse DNS
+ * lookup returns a single IP address that is the same as the forward lookup
+ * returned.
+ *
+ * When Postgres runs the DNS checks in the HBA implementation, the client IP
+ * address is looked-up in a reverse DNS query, and that name is compared to
+ * the hostname in the HBA file. Then, a forward DNS query is performed on the
+ * hostname, and one of the IP addresses returned must match with the client IP
+ * address.
+ *
+ *  client ip -- reverse dns lookup --> hostname
+ *   hostname -- forward dns lookup --> { ... client ip ... }
+ *
+ * At this point we don't have a client IP address. That said, the Postgres
+ * check will always fail if we fail to get our hostname back from at least one
+ * of the IP addresses that our hostname forward-DNS query returns.
+ */
+bool
+resolveHostnameForwardAndReverse(const char *hostname, char *ipaddr, int size)
+{
+	int error;
+	struct addrinfo *lookup, *ai;
+
+	bool foundHostnameFromAddress = false;
+
+	error = getaddrinfo(hostname, NULL, 0, &lookup);
+	if (error != 0)
+	{
+		log_warn("Failed to resolve DNS name \"%s\": %s",
+				 hostname, gai_strerror(error));
+		return false;
+	}
+
+	/* loop over the forward DNS results for hostname */
+	for (ai = lookup; ai; ai = ai->ai_next)
+	{
+		int ret;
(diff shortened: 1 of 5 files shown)


src/bin/pg_autoctl/pghba.c
```diff
@@ -624,3 +648,60 @@ pghba_enable_lan_cidr(PGSQL *pgsql,
...
+	if (strchr(hostname, '/') || ip_address_type(hostname) != IPTYPE_NONE)
+	{
+		return true;
+	}
+
+	if (!resolveHostnameForwardAndReverse(hostname, ipaddr, sizeof(ipaddr)))
+	{
+		/* warn users about possible DNS misconfiguration */
+		log_warn("Failed to resolve hostname \"%s\" to an IP address that "
+				 "resolves batck to the hostname on a reverse DNS lookup.",
```
JelteF, line 694:
```suggestion
				 "resolves back to the hostname on a reverse DNS lookup.",
```

src/bin/pg_autoctl/pghba.c
```diff
@@ -624,3 +648,60 @@ pghba_enable_lan_cidr(PGSQL *pgsql,
...
+				 hostname);
+
+		log_warn("Postgres might deny connection attempts from \"%s\", "
+				 "even with the new HBA rules.",
```
JelteF, line 698: I think it's good to add some more actionable info in a log message on how to fix this issue, e.g.
> You probably don't want to use hostnames in your HBA config, but instead use IPs, since IPs are much easier to set up correctly. If you do want hostnames, you should set up the DNS records correctly as described in the postgres docs <link to postgres docs here>.

src/bin/pg_autoctl/ipaddr.c
```diff
@@ -594,3 +594,95 @@ findHostnameFromLocalIpAddress(char *localIpAddress, char *hostname, int size)
...
+		{
+			log_debug("Failed to resolve hostname from address \"%s\": %s",
+					  ipaddr, gai_strerror(ret));
+			continue;
+		}
+
+		/* compare reverse-DNS lookup result with our hostname */
+		if (strcmp(hbuf, hostname) == 0)
+		{
+			foundHostnameFromAddress = true;
```
JelteF, line 681: maybe just `freeaddrinfo` and `return true`

src/bin/pg_autoctl/ipaddr.c
```diff
@@ -594,3 +594,95 @@ findHostnameFromLocalIpAddress(char *localIpAddress, char *hostname, int size)
 
 	return true;
 }
+
+
+/*
+ * resolveHostnameForwardAndReverse returns true when we could do a forward DNS
+ * lookup for the hostname, and then a reverse DNS lookup, and the reverse DNS
+ * lookup returns a single IP address that is the same as the forward lookup
+ * returned.
```
JelteF, line 603: > reverse DNS lookup returns a single IP 

a reverse dns lookup returns a hostname right? Can you double check this comment being correct?

JelteF (changes requested):

src/bin/pg_autoctl/pghba.c
DimCitus: Fixed, thanks!

src/bin/pg_autoctl/pghba.c
DimCitus: Added an hint in that vein, though not this exact phrase.

src/bin/pg_autoctl/ipaddr.c
DimCitus: I like that the variable name describes the situation, but if you feel that a return would be less surprising to read I should just do that then?

src/bin/pg_autoctl/ipaddr.c
DimCitus: That was more like a draft, thanks for catching it. Reworded now.

JelteF (approved):

Issue #448, opened 2020-10-08
Can not auto discover keeper hostname


RekGRpth:
Can not auto discover keeper hostname
because of, function `inet_addr` in file `ipaddr.c` requires

> Internet host address IPv4 in numbers-and-dots notation

as argument, but not hostname! (as it passed in function `discover_hostname` in file `cli_create_drop_node.c`)

RekGRpth:
how about such solution?
```diff
diff --git a/src/bin/pg_autoctl/ipaddr.c b/src/bin/pg_autoctl/ipaddr.c
index d1c34d1..d9b8a3d 100644
--- a/src/bin/pg_autoctl/ipaddr.c
+++ b/src/bin/pg_autoctl/ipaddr.c
@@ -49,6 +49,7 @@ fetchLocalIPAddress(char *localIpAddress, int size,
        struct sockaddr_in name;
        socklen_t namelen = sizeof(name);
        int err = -1;
+       struct hostent *he;
 
        struct sockaddr_in serv;
 
@@ -63,7 +64,14 @@ fetchLocalIPAddress(char *localIpAddress, int size,
 
        memset(&serv, 0, sizeof(serv));
        serv.sin_family = AF_INET;
-       serv.sin_addr.s_addr = inet_addr(serviceName);
+       if((he = gethostbyname2(serviceName, AF_INET)))
+       {
+               serv.sin_addr.s_addr = ((struct in_addr *)he->h_addr)->s_addr;
+       }
+       else
+       {
+               serv.sin_addr.s_addr = inet_addr(serviceName);
+       }
        serv.sin_port = htons(servicePort);
 
        err = connect(sock, (const struct sockaddr *) &serv, sizeof(serv));
```

DimCitus:
Hi @RekGRpth ; it is very hard for me to understand what you're trying to solve here. Can you please include enough information in your ticket? The basics for bug reports are always the same:

  1. tell us what you did (e.g. copy/paste a `pg_autoctl` command line)
  2. tell us what you obtained (e.g. copy/paste the output of the command line, and then point to the exact part that is a problem for you)
  3. tell us how what you obtained is different from what you expected

Without that I can't help you. And this issue is missing all the 3 points, jumping to a reaction to something only you saw and know about. Help me help you?

RekGRpth:
1) I just try register new keeper without specifying --hostname option

DimCitus:
[...]

PR #467, opened 2020-10-17
448

handle2001:
This PR modifies the fetchLocalIPAddress() function to use getaddrinfo() to determine the first available IP address on the host and save it to localIpAddress. This PR addresses https://github.com/citusdata/pg_auto_failover/issues/448


handle2001:
Fixed up changes with citus_indent

DimCitus:
Thanks for your work @handle2001 ! I think it's a little confused with respect to what I had in mind and shared in #448, so I have prepared #479 to better show what I think the PR should look like.

DimCitus:
Fixed in #479. Thanks for your contribution!

PR #471, opened 2020-10-20
Remove distractions from SSH output

jonels-msft:
At the cost of complicating the commands a little, I've removed warnings and MOTD output from the SSH output.

Also registered the server ssh fingerprints at the start rather than having it prompt the user. In one code block we sshed to node B twice, and if a user copied and pasted both commands it would fail because the fingerprint prompt would consume the pasted input.


jonels-msft:
These are the "unrelated changes" I mentioned in my review of https://github.com/citusdata/pg_auto_failover/pull/442

DimCitus (approved):

PR #472, opened 2020-10-22
Fix semmle issues

SaitTalhaNisanci:
This PR fixes some of the issues that semmle analysis tool found as part of the security review.

We also have some warning in file_utils that I am not sure if/how we should fix:

image

The errors are false-positives after this PR.

### DIFF
--- a/src/monitor/conninfo.c
+++ b/src/monitor/conninfo.c
@@ -51,7 +51,8 @@ ReadPrimaryHostAddress(char **primaryName, char **primaryPort)
 	options = PQconninfoParse(connInfo, &errorMessage);
 	if (options == NULL)
 	{
-		free(connInfo);
+		pfree(connInfo);
+		return -1;
 	}
 
 	for (currentOption = options; currentOption->keyword != NULL; currentOption++)
--- a/src/monitor/formation_metadata.c
+++ b/src/monitor/formation_metadata.c
@@ -656,6 +656,13 @@ FormationNumSyncStandbyIsValid(AutoFailoverFormation *formation,
 	List *standbyNodesGroupList = NIL;
 	int count = 0;
 
+	if (formation == NULL)
+	{
+		ereport(ERROR,
+				(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+				 errmsg("the given formation must not be NULL")));
+	}
+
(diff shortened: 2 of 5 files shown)


src/monitor/formation_metadata.c
```diff
@@ -745,6 +749,10 @@ AutoFailoverFormationGetDatum(FunctionCallInfo fcinfo, AutoFailoverFormation *fo
 	Datum values[5];
 	bool isNulls[5];
 
+	if (formation == NULL) {
+		return PG_RETURN_DATUM(resultDatum);
```
JelteF, line 753: I think it makes more sense to error out in this case.

src/monitor/conninfo.c
```diff
@@ -52,6 +52,7 @@ ReadPrimaryHostAddress(char **primaryName, char **primaryPort)
 	if (options == NULL)
 	{
 		free(connInfo);
```
JelteF, line 54: Should this instead be `pfree`?

SaitTalhaNisanci: I think so

SaitTalhaNisanci:
Also a note on the function that gives warnings `init_ps_buffer`:

```c
 * Inspired / stolen from Postgres code src/backend/utils/misc/ps_status.c with
 * most of the portability bits removed. At the moment we prefer simple code
 * that works on few targets to highly portable code.
```

so it seems that we took this function from postgres and looking at the function it seems not a critical code as it just sets the process title, so I guess these 2 warnings are okay, what do you think @JelteF?

SaitTalhaNisanci:
image

it seems like I dont' have the restart button for the failed job

src/monitor/formation_metadata.c
```diff
@@ -656,6 +656,11 @@ FormationNumSyncStandbyIsValid(AutoFailoverFormation *formation,
 	List *standbyNodesGroupList = NIL;
 	int count = 0;
 
+	if (formation == NULL)
+	{
+		return false;
```
JelteF, line 661: I think we should error out here as well

SaitTalhaNisanci: I think we can also add error code for these two errors, probably `errcode(ERRCODE_INVALID_PARAMETER_VALUE)`

DimCitus (approved):

PR #474, opened 2020-10-26
Register components manually

SaitTalhaNisanci:
Component Governance detector cannot detect git components currently, so
we should register them manually for SDL.

Example result: https://msdata.visualstudio.com/Database%20Systems/_componentGovernance/citus-ha?_a=components&typeId=4985361&alerts-view-option=active

We also have `pg` and `subcommands`, as far as I remember now we kinda own `subcommands` and for `pg` we seem to take only two files, it seemed not necessary to add whole postgres repo for this.

### DIFF
--- a/cgmanifest.json
+++ b/cgmanifest.json
@@ -1,0 +1,67 @@
+{
+    "Registrations": [ 
+        {
+            "Component": { 
+                "Type": "git", 
+                "git": {
+                    "RepositoryUrl": "https://github.com/rxi/log.c",
+                    "CommitHash": "f9ea34994bd58ed342d2245cd4110bb5c6790153"
+                }
+            },
+            "DevelopmentDependency": false
+        },
+        {
+            "Component": { 
+                "Type": "git", 
+                "git": {
+                    "RepositoryUrl": "https://github.com/kgabis/parson",
+                    "CommitHash": "8beeb5ea4da5eedff8d3221307ef04855804a920"
+                }
+            },
+            "DevelopmentDependency": false
+        },
+        {
+            "Component": { 
+                "Type": "git", 
+                "git": {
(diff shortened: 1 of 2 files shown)


cgmanifest.json
```diff
@@ -0,0 +1,24 @@
...
+    "Registrations": [ 
+        {
+            "Component": { 
+                "Type": "git", 
+                "git": {
+                    "RepositoryUrl": "https://github.com/rxi/log.c",
+                    "CommitHash": "f9ea34994bd58ed342d2245cd4110bb5c6790153"
+                }
+            },
+            "DevelopmentDependency": true
```
SaitTalhaNisanci, line 11: This should probably be false

JelteF: yes, I agree

cgmanifest.json
```diff
@@ -0,0 +1,57 @@
...
+            "Component": { 
+                "Type": "git", 
+                "git": {
+                    "RepositoryUrl": "https://github.com/postgres/postgres",
+                    "CommitHash": "9213462c539e6412fe0498a7f8e20b662e15c4ec"
+                }
+            },
+            
+            "license": "PostgreSQL",
+            "licenseDetail": [
```
SaitTalhaNisanci, line 33: For some reason, NOTICE generation doesn't automatically add this definition. But we can manually append this to the NOTICE file. It is probably okay for it to be not automated as we will generate the notice file only once.

JelteF (changes requested):
Missing this dependency: https://github.com/mattiasgustavsson/libs

Which we use for parsing .ini files.

JelteF (approved):

DimCitus:
Thanks team!

PR #476, opened 2020-10-26
Get rid of the APPLY_SETTINGS state and transitions.

DimCitus:
In case an upgrade should happen while the primary is in APPLY_SETTINGS, we
still keep some level of compatibility with the state: we still know how to
move the FSM forward in that case.

That said, we don't assign APPLY_SETTINGS anymore and apply cache
invalidation to the synchronous_standby_names settings in the same way we
have done already for the otherNodes list and the HBA editing.

Finally, when using pg_autoctl set node|formation ... we return directly to
the user, we don't ensure the setting made its way to the primary node
already when the command is finished.

See #446.


jchampio:
> Finally, when using pg_autoctl set node|formation ... we return directly to
the user, we don't ensure the setting made its way to the primary node
already when the command is finished.

This aspect is interesting -- over in #446 we added a test for `set formation number-sync-standbys` that now fails after this change, since we don't see the primary immediately updated. What's a good strategy for tests (or end users, for that matter) to wait for the primary to receive and apply recently-made settings before continuing?

DimCitus:
> > Finally, when using pg_autoctl set node|formation ... we return directly to
> > the user, we don't ensure the setting made its way to the primary node
> > already when the command is finished.
> 
> This aspect is interesting -- over in #446 we added a test for `set formation number-sync-standbys` that now fails after this change, since we don't see the primary immediately updated. What's a good strategy for tests (or end users, for that matter) to wait for the primary to receive and apply recently-made settings before continuing?

Yeah that's the main point of having a specific state for apply settings, we get notified when it's been applied, and we get stuck if we fail to. The current testing strategy is not very sophisticated: add a call to `time.sleep(2)` and usually that's okay. Note that we still have places where tests are flaky because of this kind of issue, such as in https://github.com/citusdata/pg_auto_failover/pull/426 that still lacks a proper fix.

In this specific case the client could maybe loop over querying the primary to check if the setting has been applied, when used with a new option `--wait` or something? I'm not sure I want to introduce direct queries between nodes as an expected mode of operation, but I can see that we will need something for at least the tests.

DimCitus:
> > This aspect is interesting -- over in #446 we added a test for `set formation number-sync-standbys` that now fails after this change, since we don't see the primary immediately updated. What's a good strategy for tests (or end users, for that matter) to wait for the primary to receive and apply recently-made settings before continuing?

Well actually... after chatting about this approach with @JelteF we came in agreement that getting rid of APPLY_SETTINGS is not the best way forward, as the monitor needs to know that the primary is in sync with the monitor, and the FSM assignments is how we implement that.

I have taken the liberty to continue from your work in #446 and filling in the gaps I could see, you will see the result in #480 that also includes your additional test cases for the situation.

jchampio:
@DimCitus I think the new `test_014_update_standby_names_when_adding_and_removing_a_standby` wasn't fully baked yet; we were trying to fix up the race conditions. There's commented code that needs to be rewritten or readded (not sure which yet) in order to prevent races. I'll file an issue.

PR #478, opened 2020-10-27
Review memory/malloc errors and potential leaks.

### LINKED ISSUE #473: Refactor: Memory allocation errors
When a memory allocation fails such as `malloc` we log an error about this however the error is not standardized. It would probably make sense to standardize this error with a variable and use it everywhere.

Example:

```c
		char *result = (char *) malloc(length * sizeof(char));
		if (result == NULL)
		{
			log_error("Failed to allocate memory, probably because it's all used");
			return NULL;
		}
```

suggestion:

```c
#define ALLOCATION_FAILED_ERROR <a nice message>
...
		char *result = (char *) malloc(length * sizeof(char));
		if (result == NULL)
		{
			log_error(ALLOCATION_FAILED_ERROR);
			return NULL;
		}
```

DimCitus:
We avoid using malloc in most of the code base, and still require it when
reading a file from disk. Some places in the code didn't get the memo that
read_file() and read_file_if_exists() leak memory, and the caller needs to
call free().

Also, because we only have a handful of malloc() failed error messages, it's
a good thing to always use the same error message format. And this format
should not require memory allocation of course. I believe %m is okay with
that.

Fixes #473

### DIFF
--- a/src/bin/pg_autoctl/keeper.c
+++ b/src/bin/pg_autoctl/keeper.c
@@ -851,6 +851,8 @@ keeper_ensure_configuration(Keeper *keeper, bool postgresNotRunningIsOk)
 
 		char upstreamConfPath[MAXPGPATH] = { 0 };
 
+		bool replicationSettingsHaveChanged = false;
+
 		char *currentConfContents = NULL;
 		long currentConfSize = 0L;
 
@@ -882,6 +884,7 @@ keeper_ensure_configuration(Keeper *keeper, bool postgresNotRunningIsOk)
 							 pgSetup->pgdata,
 							 relativeConfPathName);
 
+		/* to check if replicationSettingsHaveChanged, read current file */
 		if (file_exists(upstreamConfPath))
 		{
 			if (!read_file(upstreamConfPath,
@@ -927,8 +930,17 @@ keeper_ensure_configuration(Keeper *keeper, bool postgresNotRunningIsOk)
 			return false;
 		}
 
-		if (currentConfContents == NULL ||
-			strcmp(newConfContents, currentConfContents) != 0)
+		replicationSettingsHaveChanged =
+			currentConfContents == NULL ||
+			strcmp(newConfContents, currentConfContents) != 0;
+
+		if (currentConfContents != NULL)
+		{
+			free(currentConfContents);
+		}
+		free(newConfContents);
+
+		if (replicationSettingsHaveChanged)
 		{
 			log_info("Replication settings at \"%s\" have changed, "
 					 "restarting Postgres", upstreamConfPath);
--- a/src/bin/pg_autoctl/keeper.c
+++ b/src/bin/pg_autoctl/keeper.c
@@ -909,6 +909,7 @@ keeper_ensure_configuration(Keeper *keeper, bool postgresNotRunningIsOk)
 											 state->current_node_id))
 		{
 			/* can't happen at the moment */
+			free(currentConfContents);
 			return false;
 		}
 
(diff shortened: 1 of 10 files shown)


src/bin/pg_autoctl/lock_utils.c
```diff
@@ -194,14 +193,21 @@ semaphore_cleanup(const char *pidfile)
...
 		return false;
 	}
 
-	if (!stringToInt(semIdString, &(semaphore.semId)))
+	if (!stringToInt(fileLines[PIDFILE_LINE_SEM_ID], &(semaphore.semId)))
 	{
 		/* errors have already been logged */
+		free(fileContents);
 		return false;
 	}
```
JelteF, line 212: I think you missed one here in the happy path:
```suggestion
	}
	free(fileContents);
```

src/bin/pg_autoctl/keeper.c
```diff
@@ -927,8 +930,17 @@ keeper_ensure_configuration(Keeper *keeper, bool postgresNotRunningIsOk)
 			return false;
```
JelteF, line 930: this error and the two above can leak `currentConfContents`

src/bin/pg_autoctl/pidfile.c
```diff
@@ -332,11 +327,14 @@ read_service_pidfile_version_strings(const char *pidfile,
 		/* extension version string, comes later in the file */
 		if (pidLine == PIDFILE_LINE_EXTENSION_VERSION)
 		{
+			free(fileContents);
```
JelteF, line 330: This `free` should happen after the `strlcpy` in the line below.

JelteF (approved):
Looks good overall, left some small comments

src/bin/pg_autoctl/keeper.c
```diff
@@ -927,8 +930,17 @@ keeper_ensure_configuration(Keeper *keeper, bool postgresNotRunningIsOk)
...
 		}
 
-		if (currentConfContents == NULL ||
-			strcmp(newConfContents, currentConfContents) != 0)
+		replicationSettingsHaveChanged =
+			currentConfContents == NULL ||
+			strcmp(newConfContents, currentConfContents) != 0;
+
+		if (currentConfContents != NULL)
```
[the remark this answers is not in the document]
DimCitus: Thanks, fixed then!

src/bin/pg_autoctl/lock_utils.c
```diff
@@ -194,14 +193,21 @@ semaphore_cleanup(const char *pidfile)
...
+				  "it contains %d lines, semaphore id is expected in line %d",
+				  pidfile,
+				  lineCount,
+				  PIDFILE_LINE_SEM_ID);
+		free(fileContents);
 		return false;
 	}
 
-	if (!stringToInt(semIdString, &(semaphore.semId)))
+	if (!stringToInt(fileLines[PIDFILE_LINE_SEM_ID], &(semaphore.semId)))
```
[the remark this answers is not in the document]
DimCitus: Yeah, took me some time to figure out that more work needed to happen here than just the memory freeing...

src/bin/pg_autoctl/lock_utils.c
DimCitus: Done now, thanks.

src/bin/pg_autoctl/keeper.c
DimCitus: Good point, fixed.

src/bin/pg_autoctl/pidfile.c
DimCitus: Good point, done too!

PR #479, opened 2020-10-27
Fix/hostname discovery

DimCitus:
See #448.

### DIFF
--- a/src/bin/pg_autoctl/ipaddr.c
+++ b/src/bin/pg_autoctl/ipaddr.c
@@ -63,8 +63,16 @@ fetchLocalIPAddress(char *localIpAddress, int size,
 
 	memset(&serv, 0, sizeof(serv));
 	serv.sin_family = AF_INET;
-	serv.sin_addr.s_addr = inet_addr(serviceName);
 	serv.sin_port = htons(servicePort);
+	serv.sin_addr.s_addr = inet_addr(serviceName);
+
+	if (serv.sin_addr.s_addr == INADDR_NONE)
+	{
+		log_error("Failed to parse dotted IPv4 address \"%s\": %m", serviceName);
+		return false;
+	}
+
+	log_info("fetchLocalIPAddress: \"%s\"", serviceName);
 
 	err = connect(sock, (const struct sockaddr *) &serv, sizeof(serv));
 	if (err < 0)
--- a/src/bin/pg_autoctl/ipaddr.c
+++ b/src/bin/pg_autoctl/ipaddr.c
@@ -25,16 +25,20 @@
 
 #include "postgres_fe.h"
 
+#include "defaults.h"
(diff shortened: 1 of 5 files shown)


src/bin/pg_autoctl/ipaddr.c
```diff
@@ -25,16 +26,22 @@
...
 static bool ipv4eq(struct sockaddr_in *a, struct sockaddr_in *b);
 static bool ipv6eq(struct sockaddr_in6 *a, struct sockaddr_in6 *b);
 static bool fetchIPAddressFromInterfaceList(char *localIpAddress, int size);
+static bool ipaddr_sockaddr_to_string(struct addrinfo *ai,
+									  char *ipaddr, size_t size);
+static bool ipaddr_getsockname(int sock, char *ipaddr, size_t size);
+
 
 /*
  * Connect in UDP to a known DNS server on the external network, and grab our
```
JelteF, line 47: This comment is outdated now. We don't connect to a DNS server on a postgres node and we don't use UDP either.

src/bin/pg_autoctl/ipaddr.c
```diff
@@ -79,31 +128,23 @@ fetchLocalIPAddress(char *localIpAddress, int size,
 		}
 		else
 		{
-			log_warn("Failed to connect: %m");
+			log_warn("Failed to connect to any of the IP addresses for "
+					 "monitor hostname \"%s\" and port %d",
```
JelteF, line 132: 8.8.8.8 is not a monitor
```suggestion
					 "hostname \"%s\" and port %d",
```

JelteF, line 132: Maybe keep the `%m` at the end for debugging?

JelteF (approved):
Some small comments, but other than that looks good. I pushed one small change to the tmux commands that caused cluster creation to fail on master, but succeed on this branch. This is meant as a regression test of sorts. If you can think of a better way to test this, please do. But this seems better than nothing.

src/bin/pg_autoctl/ipaddr.c
DimCitus: Good point, fixed now!

src/bin/pg_autoctl/ipaddr.c
DimCitus: Good point again. Also when given a dotted IP address then `getaddrinfo` does not actually do a DNS query to return several IP addresses, it only does the parsing and prepares the needed binary structure. I have now added a specific message to the case of using `DEFAULT_INTERFACE_LOOKUP_SERVICE_NAME`.

DimCitus: It is now done at the connect call site. Also we do `getenv` in between so the errno has changed.

PR #480, opened 2020-10-29
Review APPLY_SETTINGS transitions in cases including node removal.

DimCitus:
When removing a node we might end up with a race condition if the primary in
parallel is already in a transition, or in any other case when
ProceedGroupState on the primary doesn't change the assigned state.

In that case we now force a new APPLY_SETTINGS state to be assigned, so that
the primary has a chance of fetching the (new) current synchronous standby
names settings.

Also, it might be that a node has been dropped while reaching APPLY_SETTINGS
itself, so in the transition from APPLY_SETTINGS to PRIMARY we also
implement fsm_apply_settings, to protect against race conditions.

Any transition to PRIMARY now also calls fsm_apply_settings.

Fixes #446, fixes #476.

### DIFF
--- a/src/monitor/node_active_protocol.c
+++ b/src/monitor/node_active_protocol.c
@@ -1121,7 +1121,32 @@ RemoveNode(AutoFailoverNode *currentNode)
 
 		if (primaryNode)
 		{
+			ReplicationState goalState = primaryNode->goalState;
+
 			(void) ProceedGroupState(primaryNode);
+
+			/*
+			 * When the removal of a secondary node has no impact on the
+			 * primary node state, we still need to change the replication
+			 * settings to adjust to the possibly new
+			 * synchronous_standby_names, so we force APPLY_SETTINGS in that
+			 * case.
+			 */
+			if (primaryNode->goalState == goalState &&
+				goalState != REPLICATION_STATE_APPLY_SETTINGS)
+			{
+				LogAndNotifyMessage(
+					message, BUFSIZE,
+					"Setting goal state of " NODE_FORMAT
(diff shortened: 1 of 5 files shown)


JelteF (approved):

src/monitor/node_active_protocol.c
```diff
@@ -1121,7 +1121,32 @@ RemoveNode(AutoFailoverNode *currentNode)
...
+					message, BUFSIZE,
+					"Setting goal state of " NODE_FORMAT
+					" to apply_settings after removing standby " NODE_FORMAT
+					" from formation %s.",
+					NODE_FORMAT_ARGS(primaryNode),
+					NODE_FORMAT_ARGS(currentNode),
+					formation->formationId);
+
+				SetNodeGoalState(primaryNode,
+								 REPLICATION_STATE_APPLY_SETTINGS, message);
```
jchampio, line 1148: Not all valid primary states have a transition defined to `APPLY_SETTINGS` -- this results in things like
```
16:10:57 199 FATAL fsm.c:514 pg_autoctl does not know how to reach state "apply_settings" from "join_primary"
16:10:57 199 ERROR service_keeper.c:461 Failed to transition to state "apply_settings", retrying...
```
We have a unit test waiting in the wings that exposes things like this; I'll clean it up and file a draft.

DimCitus: Oh good point. Yeah let's add the unit tests in a PR, and with a fix. The fix is that we don't need to go to apply_settings when we're join_primary, because the setting is going to be reset on the transition to primary anyway, as it should.

Issue #482, opened 2020-10-29
Dropping nodes during `join_primary` results in illegal state transition


jchampio:
After #480, dropping a node at the same time that the primary is in the `join_primary` state results in a hung transition.

Here's a sample state log:
```
2020-10-29 17:15:52.809786+00:00   node_5      wait_standby/wait_standby      unknown        0/0 node 5 "node_5" (172.27.1.7:5432) reported new state "wait_standby"
2020-10-29 17:15:52.827077+00:00   node_1           primary/join_primary         sync  0/6012978 Setting goal state of node 1 "node_1" (172.27.1.3:5432) to join_primary after node 5 "node_5" (172.27.1.7:5432) joined.
2020-10-29 17:15:52.879759+00:00   node_1      join_primary/join_primary         sync  0/6012978 node 1 "node_1" (172.27.1.3:5432) reported new state "join_primary"
2020-10-29 17:15:52.884601+00:00   node_5      wait_standby/catchingup        unknown        0/0 Setting goal state of node 5 "node_5" (172.27.1.7:5432) to catchingup after node 1 "node_1" (172.27.1.3:5432) converged to wait_primary.
2020-10-29 17:15:55.188883+00:00   node_5        catchingup/catchingup        unknown  0/8000000 node 5 "node_5" (172.27.1.7:5432) reported new state "catchingup"
2020-10-29 17:15:55.738690+00:00   node_1      join_primary/apply_settings       sync  0/8000000 Setting goal state of node 1 "node_1" (172.27.1.3:5432) to apply_settings after removing standby node 3 "node_3" (172.27.1.5:5432) from formation default.
```

And the failure message from the primary:
```
17:17:22 199 FATAL fsm.c:514 pg_autoctl does not know how to reach state "apply_settings" from "join_primary"
17:17:22 199 ERROR service_keeper.c:461 Failed to transition to state "apply_settings", retrying...
```

Draft PR with a failing unit test is forthcoming.

Akkowicz:
Hi,

Is there some way to force the primary in 'join_primary' state to transition back to 'primary'?
I've changed node priority while primary was still in 'join_primary' state and this is what I've ended up with:

```
postgres@database-pgautoctl-monitor:~$ pg_autoctl show state --pgdata monitor --formation cluster-REDACTED
         Name |  Node |                    Host:Port |          TLI: LSN |   Connection |       Current State |      Assigned State
[...]

Issue #464, opened 2020-10-15
Running postgres regression tests


rheaton:
Hello!

Jacob and I spent a bit of time trying to get the regression tests running locally, but it looks like there may be some prerequisite set-up required. 

Would it be possible to document this or make an easy way to run them (especially since we haven't found a way to run them in CI yet)?

Here are our notes from trying to set them up locally (we were planning to add them to CONTRIBUTING.md if we got it working) 😇

 -------

#### Postgres Regress tests

If your installed version of postgres hasn't packaged `pg_regress`:

Clone [postgres](https://github.com/postgres/postgres/tree/master)
Checkout the branch of the stable version that matches your local postgres install

e.g.
```bash
psql --version
> psql (PostgreSQL) 12.4
git clone git@github.com:postgres/postgres.git
cd postgres
git co REL_12_stable
```

Configure and make postgres
```bash
./configure
make -j8 // set to the number of processors you want to run against
```

Make the regress test runner (you may need to make and chown directories)
```bash
cd src/test/regress
make
```

Generates an executable `/usr/local/pgsql/lib/pgxs/src/test/regress/pg_regress`

Set up a monitor
```bash
pg_autoctl create monitor --pgport 7800 --auth trust --pgdata ~/failover/monitor -v --no-ssl --hostname 127.0.0.1 --run
```

Running a test
_Note_: it appears the tests may need to be run in an order, but there isn't
a schedule for running them. Additionally, other nodes may need to be set up.

`/usr/local/pgsql/lib/pgxs/src/test/regress/pg_regress --port 7800 --dbname pg_auto_failover --use-existing workers`

cc @jchampio

DimCitus:
Hi,

So the pgautofailover extension regression test can be run thanks to PGXS `make installcheck` target. This requires that a Postgres instance is running and setup correctly with the extension installed and loaded, and the easiest way to achieve that is of course `pg_autoctl create monitor`.

So here's my personal recipe to run the extension regression tests:

```
$ export PGDATA=/tmp/pgaf/m
$ export PGPORT=5000
$ ./src/bin/pg_autoctl/pg_autoctl create monitor --auth trust --no-ssl --hostname localhost --run
```

Then in another terminal

```
$ PGPORT=5000 make -C src/monitor/ installcheck
```

[...]

PR #483, opened 2020-10-29
test_multi_standbys: add failing unit test

jchampio:
Rewrite `test_014` to exercise a "simultaneous" add/drop situation, which leads to a hung FSM. This does not fix the problem, just expose it. See #482 for full details.

One additional issue: the `wait_until_state(join_primary)` call is itself racing against the primary, which might have already transitioned out of that state by the time we reach the call. I'm not sure how to fix that without a lot more plumbing.

Based on code originally authored with @rheaton.

### DIFF
--- a/tests/test_multi_standbys.py
+++ b/tests/test_multi_standbys.py
@@ -11,7 +11,6 @@ node2 = None
 node3 = None
 node4 = None
 node5 = None
-node6 = None
 
 def setup_module():
     global cluster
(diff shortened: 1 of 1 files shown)


jchampio:
We were also dabbling with testing other valid primary states, such as `wait_primary`. But all the inherent races involved with dropping a secondary at the same time that a primary is in the middle of a multi-step transition out of `single` led us to tackle this one first.

DimCitus:
I kind of liked the idea of driving the process manually with `fsm assign` and `fsm step`, I think that requires a setup without a monitor though nowadays. Do you want to investigate in that? You'd be writing the first coverage of the `--disable-monitor` feature, but it might be good if that allowed to construct difficult to reach racy situations and make sure we're good.

jchampio:
> Do you want to investigate in that?

Sure! No promises on a quick turnaround, though; a couple of other plates are being juggled right now. 😄

jchampio:
@DimCitus I picked this back up and attempted to rebase it on top of the `--disable-monitor` features in #494.

Unfortunately, the code that has the bug under test is inside the monitor (`RemoveNode()`), so disabling it doesn't help us catch the error case. What I think I'd like to see is the ability to run a monitor, so that its code and interactions can be tested, but prevent it from making state transitions until I say so, in the vein of `do fsm`.

DimCitus:
Hi @jchampio ; then maybe you should add an SQL regression test file to the monitor extension test suite over at https://github.com/citusdata/pg_auto_failover/blob/master/src/monitor/sql/monitor.sql. See also https://github.com/citusdata/pg_auto_failover/issues/464 for allowing `make -C src/monitor installcheck` to run on the Docker environment for the test suite. Even without that, we can make progress for this PR.

jchampio:
@DimCitus FYI those regression tests aren't currently passing; it looks like 3d6f4af introduced the failures.

PR #487, opened 2020-10-30
Add support for Citus worker specific transitions in IsBeingPromoted.

DimCitus:


JelteF:
should this also be added in `CandidateNodeIsReadyToStreamWAL`. I would guess yes.

DimCitus:
> should this also be added in `CandidateNodeIsReadyToStreamWAL`. I would guess yes.

Yeah, looks like you're right, thanks! I will note that in this context it's an optimisation rather than a bug: we would react a little later in the FSM. Better to react early though.

JelteF (approved):

JelteF:
Maybe add a comment to the code about this.

onderkalaci (approved):

DimCitus:
> Maybe add a comment to the code about this.

Not sure where/what to comment, so I've been lazy and merged thanks to two approvals ;-)

Issue #481, opened 2020-10-29
test_multi_standbys has intermittent failures


jchampio:
The `test_014_update_standby_names_when_adding_and_removing_a_standby` committed recently occasionally fails with

```
======================================================================
FAIL: test_multi_standbys.test_014_update_standby_names_when_adding_and_removing_a_standby
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/usr/src/pg_auto_failover/tests/test_multi_standbys.py", line 245, in test_014_update_standby_names_when_adding_and_removing_a_standby
    eq_(node1.get_synchronous_standby_names(),
AssertionError: '' != '*'
```

It's likely missing some synchronization code to ensure that the primary in question has actually transitioned to the final state.

The test has commented calls such as `assert node1.wait_until_state(...)`, which we were having trouble with while writing the test. For example, waiting until `primary` is probably correct, but waiting for `join_primary` earlier in the test causes a race where, if we don't call it fast enough, the primary will have already transitioned to `primary` state and the test will fail.

DimCitus:
The test was removed in https://github.com/citusdata/pg_auto_failover/pull/494 and should be added back later on, in a version that actually covers the problem area, and in a stable way. Meanwhile I'm closing this issue, I think #483 is where the work happens?

PR #488, opened 2020-10-30
Fix when we switch synchronous_standby_names to '*'.

DimCitus:
Because we assign both SECONDARY and PRIMARY in the same monitor
ProceedGroupState call, we might end up with the primary fetching its new
synchronous_standby_names from the monitor before when the secondary has
made it to its new state.

With the previous coding, the primary would then retrieve '', meaning sync
replication is disabled, when what's needed is '*'.

We add some needed test coverage from the situation and fix it in the
monitor by returning '*' as soon as the other node is assigned SECONDARY
rather than only after it has reported the state.

### DIFF
--- a/src/monitor/node_active_protocol.c
+++ b/src/monitor/node_active_protocol.c
@@ -2178,7 +2178,7 @@ synchronous_standby_names(PG_FUNCTION_ARGS)
 
 		if (secondaryNode != NULL &&
 			secondaryNode->replicationQuorum &&
-			IsCurrentState(secondaryNode, REPLICATION_STATE_SECONDARY))
+			secondaryNode->goalState == REPLICATION_STATE_SECONDARY)
 		{
 			/* enable synchronous replication */
 			PG_RETURN_TEXT_P(cstring_to_text("*"));
--- a/tests/test_auth.py
+++ b/tests/test_auth.py
@@ -55,7 +55,8 @@ def test_004_failover():
     print()
     print("Calling pgautofailover.failover() on the monitor")
     cluster.monitor.failover()
-    assert node2.wait_until_state(target_state="primary")
-    assert node1.wait_until_state(target_state="secondary")
 
+    assert node2.wait_until_state(target_state="primary")
     eq_(node2.get_synchronous_standby_names_local(), '*')
+
+    assert node1.wait_until_state(target_state="secondary")
--- a/tests/test_auth.py
+++ b/tests/test_auth.py
@@ -49,9 +49,13 @@ def test_003_init_secondary():
     assert node2.wait_until_state(target_state="secondary")
     assert node1.wait_until_state(target_state="primary")
 
+    eq_(node1.get_synchronous_standby_names_local(), '*')
+
 def test_004_failover():
     print()
     print("Calling pgautofailover.failover() on the monitor")
     cluster.monitor.failover()
     assert node2.wait_until_state(target_state="primary")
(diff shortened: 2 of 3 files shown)


tests/test_auth.py
```diff
@@ -49,9 +49,13 @@ def test_003_init_secondary():
     assert node2.wait_until_state(target_state="secondary")
     assert node1.wait_until_state(target_state="primary")
 
+    eq_(node1.get_synchronous_standby_names_local(), '*')
```
JelteF, line 52: To better check for the found issue, this seems like a better test:
```suggestion
    assert node1.wait_until_state(target_state="primary")
    eq_(node1.get_synchronous_standby_names_local(), '*')
    assert node2.wait_until_state(target_state="secondary")
```

Same for the other places where you added this new check after `wait_until_state` calls.

JelteF (changes requested):

tests/test_auth.py
DimCitus: Fair, done that way now.

jchampio:
With this PR branch, I still get intermittent failures as in #481:
```
FAIL: test_multi_standbys.test_014_update_standby_names_when_adding_and_removing_a_standby
  File "/usr/lib/python3/dist-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/usr/src/pg_auto_failover/tests/test_multi_standbys.py", line 245, in test_014_update_standby_names_when_adding_and_removing_a_standby
    eq_(node1.get_synchronous_standby_names(),
AssertionError: '' != '*'
```

I'm not sure that the work here is related, because `test_014` is missing calls `wait_until_state()`.

src/monitor/node_active_protocol.c
```diff
@@ -2178,7 +2178,7 @@ synchronous_standby_names(PG_FUNCTION_ARGS)
 
 		if (secondaryNode != NULL &&
 			secondaryNode->replicationQuorum &&
-			IsCurrentState(secondaryNode, REPLICATION_STATE_SECONDARY))
+			secondaryNode->goalState == REPLICATION_STATE_SECONDARY)
```
onderkalaci, line 2181: Would this lead to any issues if we have any failures during `catching up -> secondary`. Though, in the worst case, the writes might be blocked until the failure is detected, which sounds OK to me. Do you see any other edge cases on such failure?

onderkalaci (approved):
This looks good, and I verified that the race condition is gone. (And, wow it was almost always setting the sync standbys to `` before the patch)

> With this PR branch, I still get intermittent failures as in #481:

As we talked in a private chat, we realized that we might need to find the reason for #481 separately as that's multiple standbys are not going through this code-path

src/monitor/node_active_protocol.c
DimCitus: I agree with your analysis, and failed to think of other edge cases. We already react properly to network splits and other failures by re-assigning CATCHING-UP to the failed or split secondary, so we should be fine. That said, thinking more about it, we might want to also change the test at the following place:

https://github.com/citusdata/pg_auto_failover/blob/master/src/monitor/group_state_machine.c#L813

Maybe that's better in another PR though, I'll merge this one and prepare a new one.

JelteF (approved):
As discussed through chat, some test ordering should still be changed. Other than that this looks good to me.

src/monitor/node_active_protocol.c
DimCitus: See #493

PR #489, opened 2020-10-30
Fix/non default formation

DimCitus:
When we are supposed to read a configuration file, rather than creating one
for the first time, avoid setting a default formation in our command line
parsing structure. It's impossible to then distinguish between the user
saying --formation default to override what's in the configuration with the
case where we just want to use whatever is the default.

### DIFF
--- a/src/bin/pg_autoctl/cli_common.c
+++ b/src/bin/pg_autoctl/cli_common.c
@@ -1479,6 +1479,66 @@ cli_common_pgsetup_init(ConfigFilePaths *pathnames, PostgresSetup *pgSetup)
 
 
 /*
+ * cli_common_set_formation reads the formation name from the configuration
+ * file where it's not been given on the command line. When the local node is a
+ * monitor, the target formation should be found on the command line with the
+ * option --formation, otherwise we default to FORMATION_DEFAULT.
+ */
+bool
+cli_common_ensure_formation(KeeperConfig *options)
+{
+	/* if --formation has been used, we're good */
+	if (!IS_EMPTY_STRING_BUFFER(options->formation))
+	{
+		return true;
+	}
+
+	switch (ProbeConfigurationFileRole(options->pathnames.config))
+	{
+		case PG_AUTOCTL_ROLE_MONITOR:
+		{
+			/* on a monitor node, default to using the "default" formation */
+			strlcpy(options->formation,
+					FORMATION_DEFAULT,
+					sizeof(options->formation));
+			break;
+		}
+
+		case PG_AUTOCTL_ROLE_KEEPER:
+		{
+			KeeperConfig config = { 0 };
+			bool monitorDisabledIsOk = true;
+
+			/* copy the pathnames to our temporary config struct */
+			config.pathnames = options->pathnames;
+
+			if (!keeper_config_read_file_skip_pgsetup(&config,
+													  monitorDisabledIsOk))
+			{
+				/* errors have already been logged */
+				exit(EXIT_CODE_BAD_CONFIG);
+			}
+
+			strlcpy(options->formation,
+					config.formation,
+					sizeof(options->formation));
+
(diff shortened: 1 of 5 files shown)


src/bin/pg_autoctl/cli_perform.c
```diff
@@ -192,6 +192,20 @@ cli_perform_failover_getopts(int argc, char **argv)
 
 	cli_common_get_set_pgdata_or_exit(&(options.pgSetup));
 
+	if (!keeper_config_set_pathnames_from_pgdata(&(options.pathnames),
```
onderkalaci, line 195: would it make sense to move `keeper_config_set_pathnames_from_pgdata ` call into `cli_common_ensure_formation ` to make it robust for other cases?

src/bin/pg_autoctl/cli_common.c
```diff
@@ -1478,6 +1478,66 @@ cli_common_pgsetup_init(ConfigFilePaths *pathnames, PostgresSetup *pgSetup)
 }
 
 
+/*
+ * cli_common_set_formation reads the formation name from the configuration
```
onderkalaci, line 1482: function comment seems obsolete

onderkalaci (commented):
I think the changes make sense. I want to play a bit more before approving. I'll do some testing with more cli commands to see if we miss anything

src/bin/pg_autoctl/cli_perform.c
DimCitus: I though about this, but in some cases (see `cli_get_name_getopts`) it has been done already in a way that's not easy to refactor:

``` c
	/* now that we have the command line parameters, prepare the options */
	(void) prepare_keeper_options(&options);

	/* ensure --formation, or get it from the configuration file */
	if (!cli_common_ensure_formation(&options))
	{
		exit(EXIT_CODE_BAD_ARGS);
	}
```

The call to `prepare_keeper_options` is taking care of the pathnames, and is not something we want to do in every command either.

src/bin/pg_autoctl/cli_common.c
DimCitus: Fixed.

src/bin/pg_autoctl/cli_common.c
```diff
@@ -1478,6 +1478,66 @@ cli_common_pgsetup_init(ConfigFilePaths *pathnames, PostgresSetup *pgSetup)
...
+{
+	/* if --formation has been used, we're good */
+	if (!IS_EMPTY_STRING_BUFFER(options->formation))
+	{
+		return true;
+	}
+
+	switch (ProbeConfigurationFileRole(options->pathnames.config))
+	{
+		case PG_AUTOCTL_ROLE_MONITOR:
```
onderkalaci, line 1498: My understanding and testing shows that the monitor is always part of the default formation. So, we cannot have a monitor on a non-default formation, right?

onderkalaci (approved):
As @DimCitus noted in an internal chat, the steps to repro the issue was:
```
$ pg_autoctl create monitor ...


$ pg_autoctl create formation foo

$ pg_autoctl create postgres --formation foo ...

$ pg_autoctl run
```
And:
```
if you use --run it manages to remember the formation name, otherwise it fails, without the PR
```

I verified the above case is fixed. In  addition to that, I used the cli with plenty of other commands when the formation is non-default, and everything works as expected

src/bin/pg_autoctl/cli_common.c
DimCitus: I am used to thinking of it in a different way. In my view the monitor is not part of any formation, it hosts one or many formations, and comes with the special formation named “default” created by the extension script, as a way to simplify the first deployment.

That's why when doing `pg_autoctl perform failover` from the monitor, a target formation needs to be picked.

PR #494, opened 2020-11-03
Fix --disable-monitor to work with the parenting of Postgres model.

DimCitus:
In the current version the pg_autoctl process is the parent process of the
Postgres main process. Except when using --disable-monitor, which makes no
sense, so here we fix that.

It is now possible to control and run a Postgres process under a pg_autoctl
process without using a monitor, as in the following example:

  $ pg_autoctl create postgres --auth trust --no-ssl --disable-monitor --run
  $ PG_AUTOCTL_DEBUG=1 pg_autoctl do fsm assign single

The `pg_autoctl do fsm assign` command now edits the keeper state file, and
the FSM transition that is necessary is then implemented in the keeper loop
as usual.

### DIFF
--- a/src/bin/pg_autoctl/cli_do_fsm.c
+++ b/src/bin/pg_autoctl/cli_do_fsm.c
@@ -320,13 +320,6 @@ cli_do_fsm_assign(int argc, char **argv)
 	/* assign the new state */
 	keeper.state.assigned_role = goalState;
 
-	/* roll the state machine */
-	if (!keeper_fsm_reach_assigned_state(&keeper))
-	{
-		/* errors have already been logged */
-		exit(EXIT_CODE_BAD_STATE);
-	}
-
 	if (!keeper_store_state(&keeper))
 	{
 		/* errors have already been logged */
--- a/src/bin/pg_autoctl/cli_do_fsm.c
+++ b/src/bin/pg_autoctl/cli_do_fsm.c
@@ -458,15 +458,13 @@ cli_do_fsm_get_nodes(int argc, char **argv)
 		exit(EXIT_CODE_BAD_CONFIG);
 	}
 
-	if (!keeper_read_nodes_from_file(&keeper))
+	if (!keeper_read_nodes_from_file(&keeper, &(keeper.otherNodes)))
 	{
 		/* errors have already been logged */
 		exit(EXIT_CODE_INTERNAL_ERROR);
 	}
 
 	(void) printNodeArray(&(keeper.otherNodes));
-
-	exit(EXIT_CODE_INTERNAL_ERROR);
 }
 
 
@@ -521,8 +519,17 @@ cli_do_fsm_set_nodes(int argc, char **argv)
 		exit(EXIT_CODE_BAD_ARGS);
 	}
 
+	/* now read keeper's state */
+	if (!keeper_init(&keeper, config))
+	{
+		/* errors have already been logged */
+		exit(EXIT_CODE_BAD_CONFIG);
+	}
+
 	/* now parse the nodes JSON file */
-	if (!parseNodesArray(contents, &(keeper.otherNodes)))
+	if (!parseNodesArray(contents,
+						 &(keeper.otherNodes),
+						 keeper.state.current_node_id))
 	{
 		/* errors have already been logged */
 		exit(EXIT_CODE_INTERNAL_ERROR);
@@ -538,6 +545,4 @@ cli_do_fsm_set_nodes(int argc, char **argv)
 	}
 
 	(void) printNodeArray(&(keeper.otherNodes));
-
-	exit(EXIT_CODE_INTERNAL_ERROR);
 }
--- a/src/bin/pg_autoctl/cli_do_fsm.c
+++ b/src/bin/pg_autoctl/cli_do_fsm.c
@@ -20,6 +20,7 @@
 #include "fsm.h"
 #include "keeper_config.h"
 #include "keeper.h"
+#include "parsing.h"
 #include "pgctl.h"
 #include "state.h"
 #include "string_utils.h"
@@ -31,6 +32,9 @@ static void cli_do_fsm_list(int argc, char **argv);
 static void cli_do_fsm_gv(int argc, char **argv);
 static void cli_do_fsm_assign(int argc, char **argv);
 static void cli_do_fsm_step(int argc, char **argv);
+
+static void cli_do_fsm_get_nodes(int argc, char **argv);
+static void cli_do_fsm_set_nodes(int argc, char **argv);
 
 static CommandLine fsm_init =
 	make_command("init",
@@ -64,7 +68,7 @@ static CommandLine fsm_gv =
 static CommandLine fsm_assign =
 	make_command("assign",
 				 "Assign a new goal state to the keeper",
-				 CLI_PGDATA_USAGE "<goal state> [ <id> <host> <port> ]",
+				 CLI_PGDATA_USAGE "<goal state>",
 				 CLI_PGDATA_OPTION,
 				 cli_getopt_pgdata,
 				 cli_do_fsm_assign);
@@ -77,6 +81,34 @@ static CommandLine fsm_step =
 				 cli_getopt_pgdata,
 				 cli_do_fsm_step);
 
+static CommandLine fsm_nodes_get =
+	make_command("get",
+				 "Get the list of nodes from file (see --disable-monitor)",
+				 CLI_PGDATA_USAGE,
+				 CLI_PGDATA_OPTION,
+				 cli_getopt_pgdata,
+				 cli_do_fsm_get_nodes);
+
+static CommandLine fsm_nodes_set =
+	make_command("set",
+				 "Set the list of nodes to file (see --disable-monitor)",
+				 CLI_PGDATA_USAGE "</path/to/input/nodes.json>",
+				 CLI_PGDATA_OPTION,
+				 cli_getopt_pgdata,
+				 cli_do_fsm_set_nodes);
+
+
+static CommandLine *fsm_nodes_[] = {
+	&fsm_nodes_get,
+	&fsm_nodes_set,
+	NULL
+};
+
+CommandLine fsm_nodes =
+	make_command_set("nodes",
+					 "Manually manage the keeper's nodes list", NULL, NULL,
+					 NULL, fsm_nodes_);
+
 static CommandLine *fsm[] = {
 	&fsm_init,
 	&fsm_state,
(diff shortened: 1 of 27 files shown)


jchampio (commented):
Thanks, this helps quite a bit! Running the supervisor manually was a hurdle.

Note the following spurious `FATAL` message in the logs when running with this patch (the service starts happily despite it):
```
11:01:18 15679 FATAL Database directory "/some/path/pg_auto_node0" not found
11:01:18 15679 ERROR Failed to reload configuration, see above for details
11:01:18 15679 WARN  Failed to reload pg_autoctl configuration, see above for details
11:01:18 15679 INFO  pg_autoctl service is running, current state is "init"
```

Also, the pg_autoctl service rails a CPU at 100% while waiting for the first state assignment.

src/bin/pg_autoctl/cli_do_fsm.c
```diff
@@ -320,13 +320,6 @@ cli_do_fsm_assign(int argc, char **argv)
 	/* assign the new state */
 	keeper.state.assigned_role = goalState;
 
-	/* roll the state machine */
-	if (!keeper_fsm_reach_assigned_state(&keeper))
```
jchampio, line 324: With this removal, how does the CLI make use of `keeper.otherNodes` as assigned in the 4-argument version of `do fsm assign`?

src/bin/pg_autoctl/service_keeper.c
```diff
@@ -304,6 +294,12 @@ keeper_node_active_loop(Keeper *keeper, pid_t start_pid)
 				pgsql_finish(&(keeper->monitor.pgsql));
 			}
 		}
+		else if (doSleep && config->monitorDisabled)
+		{
+			int timeoutMs = PG_AUTOCTL_KEEPER_SLEEP_TIME * 1000 * 1000;
```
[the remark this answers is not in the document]
DimCitus: Good point, done that way.

src/bin/pg_autoctl/cli_do_fsm.c
DimCitus: I was not sure if we needed to expose the node list when the monitor is disabled, but if we want to manage HBA and more importantly the replication slots, then it's fair to say that we need to maintain the list of other nodes.

What I think should happen now is that we implement a set of commands to manage the list of other nodes, basically a text file containing a single array, as in the following example. The monitor replacement would have to use the new command to provide this JSON list in a file that the background process can read on every loop, and react to changes exactly like when processing the result of `pgautofailover.get_other_nodes` on the monitor.

Then I will also implement a new test file for the `--disable-monitor` option, and we will be able to add the case you found in there, either in this PR or another one.

```
$ PG_AUTOCTL_DEBUG=1 pg_autoctl do monitor get others --pgdata ./node1/ --json
[
    {
        "node_id": 2,
        "node_lsn": "0/4000110",
        "node_host": "localhost",
        "node_name": "node2",
        "node_port": 5502,
        "node_is_primary": false
    },
    {
        "node_id": 3,
        "node_lsn": "0/4000110",
        "node_host": "localhost",
        "node_name": "node3",
        "node_port": 5503,
        "node_is_primary": false
    },
    {
        "node_id": 4,
        "node_lsn": "0/4000110",
        "node_host": "localhost",
        "node_name": "node4",
        "node_port": 5504,
        "node_is_primary": false
    }
]

```

jchampio: Sounds reasonable.

DimCitus: Added support for the following new commands, and the main service keeper loop now reads the file at each loop too:

```
  pg_autoctl do fsm nodes
    get  Get the list of nodes from file (see --disable-monitor)
    set  Set the list of nodes to file (see --disable-monitor)
```

Next up, unit tests using `--disable-monitor`.

DimCitus: And unit testing of `--disable-monitor` is in, with many fixes to make it possible.

src/bin/pg_autoctl/keeper.c
```diff
@@ -2190,3 +2213,176 @@ keeper_call_reload_hooks(Keeper *keeper, bool firstLoop)
...
+
+		return true;
+	}
+	else
+	{
+		for (int i = 0; i < keeper->otherNodes.count; i++)
+		{
+			NodeAddress *node = &(keeper->otherNodes.nodes[i]);
+
+			if (node->isPrimary)
```
JelteF, line 2303: Should we also check for group here? Or is nodes.json only for a single group?

src/bin/pg_autoctl/keeper.c
```diff
@@ -1075,10 +1079,11 @@ keeper_maintain_replication_slots(Keeper *keeper)
 		return true;
 	}
 
-	if (!monitor_get_other_nodes(monitor, keeper->state.current_node_id,
-								 ANY_STATE, &(keeper->otherNodes)))
+	if (!keeper_refresh_other_nodes(keeper, forceCacheInvalidation))
```
JelteF, line 1082: `keeper_refresh_other_nodes` does a bit more than `monitor_get_other_nodes`. It also updates pg_hba. Is this change in behavior on purpose?

tests/test_monitor_disabled.py
```diff
@@ -0,0 +1,67 @@
...
+
+def test_008_secondary():
+    print()
+    node1.do_fsm_assign("primary")
+    node2.do_fsm_assign("secondary")
+
+    node1.wait_until_local_state("primary")
+    node2.wait_until_local_state("secondary")
+
+    eq_(node1.get_synchronous_standby_names_local(), '*')
```
JelteF, line 67: I think this test (or another one) should also add a third node, since a lot of this change introduces support for the monitor with multiple standbys. It's good to test that this indeed works.

If having multiple standbys work with disabled monitor is not a goal of this PR, please ignore this and expand this test when we actually implement this. In this case we should probably error out when trying to use multiple standbys with a disabled monitor.

src/bin/pg_autoctl/parsing.c
```diff
@@ -679,3 +681,172 @@ buildPostgresURIfromPieces(URIParams *uriParams, char *pguri)
...
+		}
+
+		++nodesArrayIndex;
+	}
+
+	/* now ensure the array is sorted by nodeId */
+	(void) pg_qsort(nodesArray->nodes,
+					nodesArray->count,
+					sizeof(NodeAddress),
+					nodeAddressCmpByNodeId);
```
JelteF, line 849: Probably good to validate that nodeIds are unique at this point, given they're sorted here that should be easy.

src/bin/pg_autoctl/cli_do_fsm.c
```diff
@@ -399,3 +425,124 @@ cli_do_fsm_step(int argc, char **argv)
...
+}
+
+
+/*
+ * cli_do_fsm_set_nodes parses the list of nodes parsed from the nodes file
+ * on-disk. A JSON array of nodes objects is expected. A nodes file is only
+ * used when running with --disable-monitor.
+ */
+static void
+cli_do_fsm_set_nodes(int argc, char **argv)
```
JelteF, line 477: Future improvement: I guess it would be nice if this could read from STDIN.

JelteF (changes requested):
Looks good overall. Left some questions, small comments and some suggestions for future improvements.

tests/pgautofailover_utils.py
```diff
@@ -814,17 +830,65 @@ def get_nodeid(self):
...
         """
         command = PGAutoCtl(self)
-        out, err, ret = command.execute("get node id", 'do', 'fsm', 'state')
+        out, err, ret = command.execute("get node id", '-vv', 'do', 'fsm', 'state')
 
         self.state = json.loads(out)
         return (self.state['state']['current_role'],
                 self.state['state']['assigned_role'])
 
+    def wait_until_local_state(self, target_state,
```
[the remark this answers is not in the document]
DimCitus: Good point. I even think this should be the default behaviour rather than an option, and I'm not sure about the need for a `--no-wait` option actually. Implemented that way now, with a timeout of 30s as usual.

src/bin/pg_autoctl/primary_standby.c
```diff
@@ -244,7 +244,7 @@ local_postgres_update(LocalPostgresServer *postgres, bool postgresNotRunningIsOk
 {
 	PostgresSetup *pgSetup = &(postgres->postgresSetup);
 	PostgresSetup newPgSetup = { 0 };
-	bool missingPgdataIsOk = false;
+	bool missingPgdataIsOk = true;
```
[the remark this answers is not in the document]
DimCitus: We now allow the background "node-active" to run even when no local Postgres has been initialised yet (either by `pg_ctl initdb` or `pg_basebackup`). So we need to be open to a missing PGDATA in more places.

src/bin/pg_autoctl/nodestate_utils.c
```diff
@@ -412,3 +412,87 @@ nodestate_log(CurrentNodeState *nodeState, int logLevel, int nodeId)
...
+}
+
+
+/*
+ * printNodeEntry pretty prints a node.
+ */
+void
+printNodeEntry(NodeAddress *node)
+{
+	fformat(stdout, "%3d | %s | %6d | %18s | %8s\n",
```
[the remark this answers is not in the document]
DimCitus: I reviewed that part of the code to re-use the rest of the infrastructure for the aligned output.

src/bin/pg_autoctl/keeper.c
DimCitus: Yeah the nodes.json file is intended for a single group.

src/bin/pg_autoctl/keeper.c
DimCitus: Yes I think it's a good idea to handle both concerns in one place. That's what we do with the list of other nodes after all. In most cases we don't expect changes in HBA to happen at this point, anyway.

tests/test_monitor_disabled.py
DimCitus: I have augmented the tests to add a 3rd node. I though we might want to do that later when adding the test referred to in https://github.com/citusdata/pg_auto_failover/pull/483 ; though adding that part now makes sense too of course.

src/bin/pg_autoctl/parsing.c
DimCitus: Done now, good point.

src/bin/pg_autoctl/cli_do_fsm.c
DimCitus: Yeah, that's a good point. That's for another day though, I think.

JelteF (approved):

PR #496, opened 2020-11-03
Fix bug in monitor hostname discovery

DimCitus:
Use our new function `ipaddrGetLocalHostname` directly on `config->hostname` so that we can continue with the hostname properly set.


onderkalaci (commented):
I really don't have much time to understand/review the changes,  but I can confirm that it fixes the issue. @JelteF would you mind looking into the changes given that you reviewed the relevant PR?

JelteF (approved):

PR #499, opened 2020-11-05
Fix race conditions in multiple-standby Group FSM for failover.

DimCitus:
When handling a group with multiple standby nodes but only one of them is a
candidate for failover, some adjustments are needed in order for the FSM to
coorectly move forward when the non-candidates are showing first.

In practice, we should:

  - refrain from assigning PRIMARY to a WAIT_PRIMARY node when a standby is
    reaching SECONDARY if that standby has a candidatePriority = 0,

  - start the re-join process of a demoted primary even when the new primary
    is still transitioning between wait_primary and primary, we don't need a
	stable state there,

  - and as a consequence of the first item, allow a transition from
    wait_primary to join_primary when adding a secondary to the new primary
    node, if the secondary being added has candidatePriority = 0.

See https://github.com/citusdata/pg_auto_failover/pull/439 which missed those cases.

### DIFF
--- a/src/monitor/group_state_machine.c
+++ b/src/monitor/group_state_machine.c
@@ -397,7 +397,9 @@ ProceedGroupState(AutoFailoverNode *activeNode)
 	 *  join_primary -> primary
 	 */
 	if (IsCurrentState(activeNode, REPLICATION_STATE_WAIT_MAINTENANCE) &&
-		IsCurrentState(primaryNode, REPLICATION_STATE_JOIN_PRIMARY))
+		primaryNode->reportedState == REPLICATION_STATE_JOIN_PRIMARY &&
+		(primaryNode->goalState == REPLICATION_STATE_JOIN_PRIMARY ||
+		 primaryNode->goalState == REPLICATION_STATE_PRIMARY))
 	{
 		char message[BUFSIZE];
 
@@ -681,8 +683,12 @@ ProceedGroupState(AutoFailoverNode *activeNode)
 		/* it's safe to rejoin as a secondary */
 		AssignGoalState(activeNode, REPLICATION_STATE_SECONDARY, message);
 
-		/* compute next step for the primary depending on node settings */
-		return ProceedGroupStateForPrimaryNode(primaryNode);
+		if (activeNode->candidatePriority > 0)
+		{
+			AssignGoalState(primaryNode, REPLICATION_STATE_PRIMARY, message);
+		}
+
+		return true;
 	}
 
 	/*
@@ -899,6 +905,7 @@ ProceedGroupStateForPrimaryNode(AutoFailoverNode *primaryNode)
 	 * there's no visible reason to not be a primary rather than either
 	 * wait_primary or join_primary
 	 *
+	 *    wait_primary ➜ primary
 	 *    join_primary ➜ primary
 	 */
 	if (IsCurrentState(primaryNode, REPLICATION_STATE_WAIT_PRIMARY) ||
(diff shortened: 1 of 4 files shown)


src/monitor/group_state_machine.c
```diff
@@ -678,7 +682,11 @@ ProceedGroupState(AutoFailoverNode *activeNode)
 
 		/* it's safe to rejoin as a secondary */
 		AssignGoalState(activeNode, REPLICATION_STATE_SECONDARY, message);
-		AssignGoalState(primaryNode, REPLICATION_STATE_PRIMARY, message);
+
+		if (activeNode->candidatePriority > 0)
```
onderkalaci, line 686: I cannot follow this logic. Why do we check the  newly  transitioned secondary's candidate priority for promoting the other node? Or, maybe conversely, why would we NOT promote the primary if `activeNode->candidatePriority = 0`?

DimCitus: The state PRIMARY embeds the assumption that it's possible to failover. A node with candidate Priority of zero can not be the target of a failover, so might reach SECONDARY without allowing to get out of WAIT_PRIMARY or JOIN_PRIMARY state.

At some point in the past I had a proposal to introduce a STREAMING state that would be like a SECONDARY state, but without being a target for promotion. Maybe that would be a good way to address the current situation...

onderkalaci: I really cannot understand this part well. I'm not saying that this is wrong or anything, but not intuitive for me and I really like to understand  better before approving.

1. This code-path is triggered only for single-standby case. Do we  ever care  about `candidatePriority ` in the single-standby case?
2.
>  A node with candidate Priority of zero can not be the target of a failover

Why do  we care about `activeNode `'s (e.g., the secondary's) priority when it is assigned to be secondary at the moment?  even if it is zero, why  don't we just allow `activeNode` to be  the `REPLICATION_STATE_SECONDARY` and primaryNode to be  in  `REPLICATION_STATE_PRIMARY`?

3. > so might reach SECONDARY without allowing to get out of WAIT_PRIMARY or JOIN_PRIMARY state.

The  `activeNode`  reaches `REPLICATION_STATE_SECONDARY ` and `primaryNode ` reaches to `REPLICATION_STATE_PRIMARY ` right here. When/why wouldn't it  allow to  get out of `WAIT_PRIMARY` or `JOIN_PRIMARY`?


4. Optionally, can you give an example flow where this change fixes the problem?

DimCitus: This transition happens when a former primary node has been demoted, and is now accepted back in the group as a secondary. 

  - When the group has only two nodes, the other node must have been promoted, and is now in the state WAIT_PRIMARY.

  - When the group has more than two nodes, an election is taking place while the former primary is demoted. One of the other nodes is going to be selected and assigned PREPARE_PROMOTION, and then later may reach PRIMARY when one of the still healthy secondaries follow the new primary.

The situation we are interested in with this PR is the situation when the new primary node can be assigned PRIMARY again, meaning the group is now ready for another failover if necessary. If the first standby that follows the new primary is not a candidate for promotion, we can't move to PRIMARY yet, we need to wait until a secondary that is a candidate for promotion shows up.

onderkalaci: > This transition happens when a former primary node has been demoted, and is now accepted back in the group as a secondary.

Alright, now  I got it  thanks. Can you maybe  add this to  the comment in the code?

onderkalaci (approved):

Issue #490, opened 2020-10-31
Hooks for Node Failover/Switchover - Virtual IP?


s4ke:
One of our application is based on Django and uses the standard libpq high availability capabilities so that we don't have to do anything in case of a failover.

Other applications in our stack (such as Superset which uses SQLAlchemy) do not support this at the application layer.

We try to go as far as we can go without requiring a load balancer/proxy to keep everything as simple as possible (exactly the reason we opted to go for pg_auto_failover in the first place :) ).

Now, I am thinking how to solve our issue with HA for the rest of our application stack: We could introduce a virtual ip that always points to the current master (or secondary if we only do analytics) of the replication set.

One way we could implement this is by having a daemon that checks all database nodes for their state and have this as a separate software in our cluster - or we could implement something that hooks into the lifecycle pg_auto_failover. What do you think? Is this something that pg_auto_failover could support?

DimCitus:
Hi @s4ke ; I'm surprised SQLAlchemy has no way to setup a multi-host Postgres connection string, but I must admit I never tried, so I'll take your word for it.

About v-ip we have had a similar discussion and approach with using pgbouncer. It is already possible to use the LISTEN/NOTIFY based notifications for any client program to know when something is happening on a formation on the monitor, and you can also use the SQL API directly with `pgautofailover.get_primary('default', 0)` and such.

We could see about how to contribute that piece of software directly in the pg_auto_failover tool set if we have a portable enough way to maintain a virtual IP address. The current thinking about this has been, I believe, related t pgbouncer, as in the following pre-existing issues:

  - https://github.com/citusdata/pg_auto_failover/issues/440
  - https://github.com/citusdata/pg_auto_failover/issues/427

s4ke:
Hey @DimCitus ,

I was confused by that as well, but I think SQLAlchemy wants you to implement HA via some kind of database router just like Django does normally as well. But since Django uses libpq internally all you have to do to get proper HA is to do something like this:

```
    DATABASES = {
        'default': {
            'ENGINE': 'django_db_geventpool.backends.postgresql_psycopg2',
            'NAME': '{{ dbname }}',
            'USER': '{{ db_user }}',
            'PASSWORD': '{{ db_passwd }}',
            'HOST': '{{ db_hosts }}',
            'CONN_MAX_AGE': 0,
            'PORT': '{{ skipper_db_ports }}',
            'OPTIONS': {
                'application_name': {{ app_name }},
[...]

Issue #440, opened 2020-09-30
Execute command on state change


RekGRpth:
How about to add some options into keeper config to execute some commands on every state change? 
(Instead listen channel `state`)

RekGRpth:
For example, it may update pgbouncer config

RekGRpth:
I solved it by using [pglisten](https://github.com/RekGRpth/pgsidekick)
```sh
pglisten --listen=state --conninfo="$(pg_autoctl show uri --monitor)" | jq --unbuffered --raw-output "select(.formation == \""$(pg_autoctl config get pg_autoctl.formation)"\" and .groupId == $(pg_autoctl config get pg_autoctl.group) and .health == \"good\" and .reportedState == \"primary\" and .goalState == \"primary\") | {host:.host, port:.port} | join(\" \")" | while read -r HOST PORT; do
    echo "* = host=$HOST port=$PORT" >"$HOME/databases.ini"
    sv reload pgbouncer
done
```
but it seems very bad, because it establish extra connection to monitor from each keeper (although each keeper connect to monitor already)

DimCitus:
When we implement support for pgbouncer in pg_autoctl directly, it will most certainly use its own connection to the monitor and the notifications like you did. The only bit you're missing now is the “cache invalidation” part. You should have a way of producing the pgbouncer setup from scratch when you start the `pglisten` program, and ideally, that reset step should happen between the LISTEN command and processing the first notification.

DimCitus:
Other than that I'm not sure about including a generic facility in `pg_autoctl` for hooks and extra command. I'm mostly worried about robustness and simplicity here. My experience with for example the `archive_command` setting in Postgres is that you can't have something that is that flexible and still is simple and robust. I'd rather include the features we need in the tool itself and make it simple on the users, and ensure we implement proper error handling.

RekGRpth:
You may see how it done in repmgr

DimCitus:
> You may see how it done in repmgr

  - do you have a link to the docs with a particular section of paragraph that I should review?
  - how do they prevent a failover from failing because the external script that is run in a hook is failing, or worse, taking a very long time, stuck somewhere?

RekGRpth:
> how do they prevent a failover from failing because the external script that is run in a hook is failing, or worse, taking a very long time, stuck somewhere?

It needs to investigate their code

Issue #427, opened 2020-09-24
is it possible to integrate pgbouncer to support HA?


thomasdba:
thanks !

danielgustafsson:
This has actually been discussed and I have started on a patch to add support for pgBouncer, hopefully I can show something soon.

DimCitus:
See https://github.com/citusdata/pg_auto_failover/pull/539

PR #508, opened 2020-11-13
[POC] Add rudimentary support of pgbouncer

gkokolatos:
A proof of concept (POC) PR around the question:
As a user, given that you have a pg_auto_failover controlled formation, how do
you integrate it with other necessary tools in your environment, for example,
pgbouncer?

It is obvious to the reader of the code that the task pg_auto_failover performs,
is concrete and by no means it intends to solve all the HA needs of any user.
And it shouldn't try to solve all those needs either.

Why send a patch? Because for a user, there should exist a simple and
straightforward way to use different tools. Which when used together, covers
most of the needs.

The design of pg_auto_failover, which leverages LISTEN/NOTIFY to control the
state of a cluster combined with the helper function for the monitor, does allow
for extensibility. Here is where the POC patch enters.

Admittedly right now the POC is a bit too tightly coupled to the code. Yet one
can imagine the ability to be able to extend pg_autoctl to run the external
service of the user's choice with minimal effort.

As it stands, the user can issue:
$ pg_autoctl create pgbouncer --config <path>/pgbouncer.ini --monitor <monitor url>

Then, the user gets a pgbouncer instance that runs as a child process and
under the supervision of pg_autoctl. It is the pg_autoctl's responsibility to
fill in the [databases] section of pgbouncer's configuration during start-up and
also to update it on formation changes.

What are your thoughts?

Cheers,
//Georgios

### DIFF
--- a/src/bin/pg_autoctl/cli_common.h
+++ b/src/bin/pg_autoctl/cli_common.h
@@ -78,6 +78,8 @@ extern int monitorDisabledNodeId;
 
 #define CLI_PGDATA_USAGE " [ --pgdata ] [ --json ] "
 
+/* cli_pgbouncer */
+extern CommandLine create_pgbouncer_command;
 
 /* cli_do.c */
 extern CommandLine do_commands;
--- a/src/bin/pg_autoctl/cli_pgbouncer.c
+++ b/src/bin/pg_autoctl/cli_pgbouncer.c
@@ -1,0 +1,475 @@
+/*
+ * cli_pgbouncer.c
+ *     Implementation of a CLI to manage a pgbouncer instance.
+ *
+ * TODO: Add Copyright note
+ *
+ */
+#include <getopt.h>
+#include <sys/types.h>
+#include <unistd.h>
+#include <wait.h>
+
+#include "postgres_fe.h"
+
+#include "commandline.h"
+#include "cli_root.h"
+#include "defaults.h"
+#include "ini_file.h"
+#include "log.h"
+#include "monitor.h"
+#include "pgsql.h"
+#include "string_utils.h"
+
+struct PgbouncerConfig {
+	int		listenPort;
+
+	char	adminUsers[NAMEDATALEN];
+	char	monitorPgUri[MAXCONNINFO];
+	char	configFile[MAXPGPATH];
+	char	pidFile[MAXPGPATH];
+} pgbouncerConfig;
+
+/* Forward declaration of commands */
+static int pgbouncer_create_getopts(int argc, char **argv);
+static void pgbouncer_create(int argc, char **argv);
+
+CommandLine create_pgbouncer_command =
+	make_command("pgbouncer",
+				 "Create a new pgbouncer instance to connect to primary",
+				 "[ --config --monitor --help ] ",
+				 "  --config     pgbouncer config file (required)\n"
+				 "  --monitor    pg_auto_failover Monitor Postgres URL (required)\n"
+				 "  --help       show this message \n",
+				 pgbouncer_create_getopts,
+				 pgbouncer_create);
+
+/*--------------------
+ * Get opts section
+ */
+
+static int
+pgbouncer_create_getopts(int argc, char **argv)
+{
+	int	c;
+	int	errors = 0;
+	int	option_index = 0;
+
+	static struct option long_options[] = {
+		{ "config",	required_argument, NULL, 'c' },
+		{ "help", no_argument, NULL, 'h' },
+		{ "monitor", required_argument, NULL, 'm' },
+		{ NULL, 0, NULL, 0 }
+	};
+
+	optind = 0;
+
+	while ((c = getopt_long(argc, argv, "c:hm:",
+							long_options, &option_index)) != -1)
+	{
+		switch (c)
+		{
+			case 'c':
+				if (strlcpy(pgbouncerConfig.configFile,
+							optarg,
+							sizeof(pgbouncerConfig.configFile)) >=
+					sizeof(pgbouncerConfig.configFile))
+				{
+					log_error("config file too long, greater than  %ld",
+								sizeof(pgbouncerConfig.configFile) -1);
+					exit(EXIT_CODE_BAD_ARGS);
+				}
+
+				log_trace("--config %s", pgbouncerConfig.configFile);
+				break;
+
+			case 'm':
+				if (!validate_connection_string(optarg))
+				{
+					log_fatal("Failed to parse --monitor connection string, "
+							  "see above for details.");
+					exit(EXIT_CODE_BAD_ARGS);
+				}
+				/* validate_connection_string has already checked size */
+				strlcpy(pgbouncerConfig.monitorPgUri, optarg, MAXCONNINFO);
+				log_trace("--monitor %s", pgbouncerConfig.monitorPgUri);
+				break;
+
+			case 'h':
+				commandline_help(stderr);
+				exit(EXIT_CODE_QUIT);
+				break;
+
+			default:
+				/* getopt_long already wrote an error message */
+				errors++;
+		}
+	}
+
+	if (errors > 0 ||
+		IS_EMPTY_STRING_BUFFER(pgbouncerConfig.configFile) ||
+		IS_EMPTY_STRING_BUFFER(pgbouncerConfig.monitorPgUri)
+		)
+	{
+		commandline_help(stderr);
+		exit(EXIT_CODE_BAD_ARGS);
+	}
+
+	return optind;
+}
+
+
+/*--------------------
+ * The actual commands section
+ * 		and their helpers
+ */
+
+/*
+ * Write a databases.ini file for pgbouncer to point to primary
+ *
+ * 		Which databases to follow?
+ *		Which users to set up?
+ */
+static bool
+write_pgbouncer_databases_ini_section(NodeAddress primary)
+{
+	FILE	   *fileStream = NULL;
+	const char *filePath = "/tmp/databases.ini"; /* XXX: which path to use */
+	IniOption  *databasesOption;
+	char		buf[BUFSIZE];
+	bool		success = false;
+
+	if (snprintf(buf, sizeof(buf),
+					"host=%s "
+					"port=%d "
+					"dbname=postgres" /* XXX: This needs expansion */,
+					primary.host, primary.port) >= sizeof(buf))
+	{
+		log_error("Failed to write database section %m");
+		exit(EXIT_CODE_INTERNAL_ERROR);
+	}
+
+	databasesOption = (IniOption[]){
+		make_strbuf_option("databases", "primary", "primary",
+							true, BUFSIZE, buf),
+		INI_OPTION_LAST
+	};
+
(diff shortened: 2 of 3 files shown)


DimCitus:
Hi @gkokolatos ; thanks for contributing to pg_auto_failover!

In the context of this PR, see also #427 and #440 and #490.

> Admittedly right now the POC is a bit too tightly coupled to the code. Yet one
> can imagine the ability to be able to extend pg_autoctl to run the external
> service of the user's choice with minimal effort.

I think that we are good with a very tightly coupled integration of pgbouncer in pg_auto_failover. If we get later integrations with other tooling such as haproxy and odyssey then we might consider an extension like integration. Until such time, let's implement it the easy way and learn things!

> As it stands, the user can issue:
> $ pg_autoctl create pgbouncer --config /pgbouncer.ini --monitor
> 
> Then, the user gets a pgbouncer instance that runs as a child process and
> under the supervision of pg_autoctl. It is the pg_autoctl's responsibility to
> fill in the [databases] section of pgbouncer's configuration during start-up and
> also to update it on formation changes.
> 
> What are your thoughts?

After having read through your current PR, I think there's an easier way to get you started. It seems to me that we have two main ways to integrate with pgbouncer:

  - as you're doing here, run a `pg_autoctl` dedicated “instance” (process tree) that knows how to update the `pgbouncer.ini` setup when the primary changes, and hot-reload the setup in the running pgbouncer process

        $ pg_autoctl create pgbouncer ...

  - also we could (and should) support the use case where within a `pg_autoctl run` we also have an extra pgbouncer process running alongside PostgreSQL itself, in such a way that the same process tree opens both the 5432 and 6432 ports (by default) and you can connect to 6432 to benefit from pgbouncer support. In that case we would want the 6432 proxying port to follow through whichever node is the primary at the moment

        $ pg_autoctl enable pgbouncer ...

The second integration style makes it more obvious how to re-use existing configuration and framework, I believe, because you can access everything that already exists in terms of setup. In the `enable` case, a node has already been created, you can grab the already existing monitor URI, you have a place where to store configuration files, etc.

Some specific remarks to help you in the next steps of discovering the internal facilities we already have in the project and making your life easier, and then ours (reviewers) too:

  - please use our supervision infrastructure in your `dance` function, see e.g. https://github.com/citusdata/pg_auto_failover/blob/master/src/bin/pg_autoctl/service_monitor.c#L42 and the call to `supervisor_start`

  - please use our INI file abstractions to generate the `pgbouncer.ini` file, see e.g. https://github.com/citusdata/pg_auto_failover/blob/master/src/bin/pg_autoctl/systemd_config.c for how we generate the systemd unit file

  - please add a `pgbouncer` structure member in `struct ConfigFilePaths` at https://github.com/citusdata/pg_auto_failover/blob/master/src/bin/pg_autoctl/config.h#L34 and use the XDG standard to decide where to maintain the file

    I would suggest using XDG_DATA_HOME for the generated `pgbouncer.ini` file, and maybe have a template copy in XDG_CONFIG_HOME that we would generate from the user given `pgbouncer.ini` file: the user can provide us with their exact setup for pgbouncer, we use exactly what they give us, except for the `[databases]` section that we replace. Then we manage both a *template* file as a configuration file, and a runtime file based on it.

  - please use our file handling facilities such as `read_file` and `write_file` found in https://github.com/citusdata/pg_auto_failover/blob/master/src/bin/pg_autoctl/file_utils.c

  - finally, when using LISTEN/NOTIFY to maintain a cache, remember that we need to implement the following simple algorithm:
      1. connect to the monitor and issue a LISTEN command
      2. cache reset: grab current primary and generate the run-time `pgbouncer.ini` file
      3. cache invalidation: process notifications in a loop, producing a new run-time `pgbouncer.ini` file when needed

Finally, a word about formation and groups. When running `pg_autoctl enable pgbouncer`, I think we should only generate a single entry in the `[databases]` section for the current formation and group of the existing node. The target is the current primary of the formation and group this node is a member of. When running `pg_autoctl create pgbouncer --monitor ...` though, we need to generate a `[databases]` entry for each formation. See `pg_autoctl show uri` that already knows how to fetch the list of connection strings for each formation.

gkokolatos:
On Monday, November 16, 2020 12:32 PM, Dimitri Fontaine <notifications@github.com> wrote:

> Hi [@gkokolatos](https://github.com/gkokolatos) ; thanks for contributing to pg_auto_failover!

Hey! Thank you for reading the code.

> In the context of this PR, see also [#427](https://github.com/citusdata/pg_auto_failover/issues/427) and [#440](https://github.com/citusdata/pg_auto_failover/issues/440) and [#490](https://github.com/citusdata/pg_auto_failover/issues/490).
>
>> Admittedly right now the POC is a bit too tightly coupled to the code. Yet one
>> can imagine the ability to be able to extend pg_autoctl to run the external
>> service of the user's choice with minimal effort.
>
> I think that we are good with a very tightly coupled integration of pgbouncer in pg_auto_failover. If we get later integrations with other tooling such as haproxy and odyssey then we might consider an extension like integration. Until such time, let's implement it the easy way and learn things!

I am glad to hear that response. I am pleasantly surprised by it.

Let me try a proper implementation then. I will follow your comments closely and avoid doing the 'dance' ;)

>> As it stands, the user can issue:
>> $ pg_autoctl create pgbouncer --config /pgbouncer.ini --monitor
>>
>> Then, the user gets a pgbouncer instance that runs as a child process and
>> under the supervision of pg_autoctl. It is the pg_autoctl's responsibility to
>> fill in the [databases] section of pgbouncer's configuration during start-up and
>> also to update it on formation changes.
>>
>> What are your thoughts?
>
> After having read through your current PR, I think there's an easier way to get you started. It seems to me that we have two main ways to integrate with pgbouncer:
>
> -
>
> as you're doing here, run a pg_autoctl dedicated “instance” (process tree) that knows how to update the pgbouncer.ini setup when the primary changes, and hot-reload the setup in the running pgbouncer process
>
> $ pg_autoctl create pgbouncer ...
>
> -
>
> also we could (and should) support the use case where within a pg_autoctl run we also have an extra pgbouncer process running alongside PostgreSQL itself, in such a way that the same process tree opens both the 5432 and 6432 ports (by default) and you can connect to 6432 to benefit from pgbouncer support. In that case we would want the 6432 proxying port to follow through whichever node is the primary at the moment
>
> $ pg_autoctl enable pgbouncer ...
>
> The second integration style makes it more obvious how to re-use existing configuration and framework, I believe, because you can access everything that already exists in terms of setup. In the enable case, a node has already been created, you can grab the already existing monitor URI, you have a place where to store configuration files, etc.
>
> Some specific remarks to help you in the next steps of discovering the internal facilities we already have in the project and making your life easier, and then ours (reviewers) too:
>
> -
>
> please use our supervision infrastructure in your dance function, see e.g. https://github.com/citusdata/pg_auto_failover/blob/master/src/bin/pg_autoctl/service_monitor.c#L42 and the call to supervisor_start
>
> -
>
> please use our INI file abstractions to generate the pgbouncer.ini file, see e.g. https://github.com/citusdata/pg_auto_failover/blob/master/src/bin/pg_autoctl/systemd_config.c for how we generate the systemd unit file
>
> -
>
> please add a pgbouncer structure member in struct ConfigFilePaths at https://github.com/citusdata/pg_auto_failover/blob/master/src/bin/pg_autoctl/config.h#L34 and use the XDG standard to decide where to maintain the file
>
> I would suggest using XDG_DATA_HOME for the generated pgbouncer.ini file, and maybe have a template copy in XDG_CONFIG_HOME that we would generate from the user given pgbouncer.ini file: the user can provide us with their exact setup for pgbouncer, we use exactly what they give us, except for the [databases] section that we replace. Then we manage both a template file as a configuration file, and a runtime file based on it.

Sure.

> -
>
> please use our file handling facilities such as read_file and write_file found in https://github.com/citusdata/pg_auto_failover/blob/master/src/bin/pg_autoctl/file_utils.c
>
> -
>
> finally, when using LISTEN/NOTIFY to maintain a cache, remember that we need to implement the following simple algorithm:
>
> - connect to the monitor and issue a LISTEN command
> - cache reset: grab current primary and generate the run-time pgbouncer.ini file
> - cache invalidation: process notifications in a loop, producing a new run-time pgbouncer.ini file when needed

Sure.

> Finally, a word about formation and groups. When running pg_autoctl enable pgbouncer, I think we should only generate a single entry in the [databases] section for the current formation and group of the existing node. The target is the current primary of the formation and group this node is a member of. When running pg_autoctl create pgbouncer --monitor ... though, we need to generate a [databases] entry for each formation. See pg_autoctl show uri that already knows how to fetch the list of connection strings for each formation.

I think that is a part that we will need to discuss a bit further. What you are proposing makes perfect sense, do not get me wrong, though I can see some scenarios that we might need to allow for slightly more connections to the primary. But let us not run ahead of ourselves at this point.

As a first step, let me come with a PR which is as close as I can to the proper implementation of `pg_autoctl enable` and `pg_autoctl create` to get the structure of the code right. Once that falls into place, we will have to discuss a bit the behaviour / options that pgbouncer can support.

DimCitus (changes requested):
Per discussion ; GitHub seems to insist that I follow the PR review workflow.

DimCitus:
> I think that is a part that we will need to discuss a bit further. What you are proposing makes perfect sense, do not get me wrong, though I can see some scenarios that we might need to allow for slightly more connections to the primary. But let us not run ahead of ourselves at this point. 

Yeah, let's do the simple and straightforward implementation first, and see where that leads us. We might want to improve that as early as in the first PR, or more probably improve it in a later PR.

> As a first step, let me come with a PR which is as close as I can to the proper implementation of `pg_autoctl enable` and `pg_autoctl create` to get the structure of the code right. Once that falls into place, we will have to discuss a bit the behaviour / options that pgbouncer can support.

My 2 cents here: it might be even easier if you would focus on `pg_autoctl enable pgbouncer` and `pg_autoctl create postgres --enable-pgbouncer ...` first, in its own PR. And then in another PR, building on the pieces from the first one, deliver `pg_autoctl create pgbouncer` to run a pgbouncer only pg_autoctl agent.

gkokolatos:
On Monday, November 16, 2020 4:29 PM, Dimitri Fontaine <notifications@github.com> wrote:

>> I think that is a part that we will need to discuss a bit further. What you are proposing makes perfect sense, do not get me wrong, though I can see some scenarios that we might need to allow for slightly more connections to the primary. But let us not run ahead of ourselves at this point.
>
> Yeah, let's do the simple and straightforward implementation first, and see where that leads us. We might want to improve that as early as in the first PR, or more probably improve it in a later PR.

Absolutely.

>> As a first step, let me come with a PR which is as close as I can to the proper implementation of pg_autoctl enable and pg_autoctl create to get the structure of the code right. Once that falls into place, we will have to discuss a bit the behaviour / options that pgbouncer can support.
>
> My 2 cents here: it might be even easier if you would focus on pg_autoctl enable pgbouncer and pg_autoctl create postgres --enable-pgbouncer ... first, in its own PR. And then in another PR, building on the pieces from the first one, deliver pg_autoctl create pgbouncer to run a pgbouncer only pg_autoctl agent.

That is exactly what I was planning on doing.

If helpful for maintenance reasons, feel free to close this PR and I can reference it from the new one.

PR #510, opened 2020-11-16
Use PG_CONFIG has a hint to choose one of multiple Postgres versions.

DimCitus:
When more than one Postgres version is found in $PATH a classic way to
target a specific one, for Postgres contributors and extension authors,
consists of using PG_CONFIG.

With this PR we make it so that pg_autoctl also follows PG_CONFIG and use
the pg_ctl PATH in the same directory as the pg_config setup that way in the
environment.

In passing, cleanup the pg_ctl_version() API to stop leaking memory out of
its local stack/heap.

### DIFF
--- a/src/bin/pg_autoctl/pgctl.c
+++ b/src/bin/pg_autoctl/pgctl.c
@@ -82,29 +82,35 @@ static bool pg_write_standby_signal(const char *pgdata,
 
 
 /*
- * Get pg_ctl --version output.
- *
- * The caller should free the return value if not NULL.
+ * Get pg_ctl --version output in pgSetup->pg_version.
  */
-char *
-pg_ctl_version(const char *pg_ctl_path)
+bool
+pg_ctl_version(PostgresSetup *pgSetup)
 {
 	char *version;
-	Program prog = run_program(pg_ctl_path, "--version", NULL);
+	Program prog = run_program(pgSetup->pg_ctl, "--version", NULL);
 
 	if (prog.returnCode != 0)
 	{
 		errno = prog.error;
 		log_error("Failed to run \"pg_ctl --version\" using program \"%s\": %m",
-				  pg_ctl_path);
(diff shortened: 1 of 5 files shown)


JelteF (changes requested):

src/bin/pg_autoctl/pgctl.c
```diff
@@ -185,6 +191,47 @@ pg_controldata(PostgresSetup *pgSetup, bool missing_ok)
...
+		return false;
+	}
+
+	if (!get_env_copy("PG_CONFIG", PG_CONFIG, sizeof(PG_CONFIG)))
+	{
+		/* errors have already been logged */
+		return false;
+	}
+
+	(void) path_in_same_directory(PG_CONFIG, "pg_ctl", pgSetup->pg_ctl);
```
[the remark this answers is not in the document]
DimCitus: Good point, I went too fast by taking for granted what Postgresapp.com does there. Fixed now by actually calling `pg_config --bindir`, thanks!

src/bin/pg_autoctl/pgctl.c
```diff
@@ -114,6 +114,54 @@ pg_ctl_version(PostgresSetup *pgSetup)
...
+		log_error("Failed to run \"pg_config --bindir\" using program \"%s\": %m",
+				  pg_config);
+		free_program(&prog);
+		return false;
+	}
+
+	if (splitLines(prog.stdOut, lines, 1) != 1)
+	{
+		log_error("Unable to parse output from pg_config --bindir");
+		free_program(&prog);
```
JelteF, line 144: It seems `free_program` in the later function exit points is missing (i.e. the `return true` and the other `return false`.

DimCitus: Thanks, fixed now.

JelteF (approved):

PR #515, opened 2020-11-18
Use pg_config from the PATH when we can't find a pg_ctl.

### LINKED ISSUE #514: Use pg_config to find pg_ctl if no pg_ctl is in PATH
On Ubuntu 18.04 `pg_ctl` is not in `PATH` by default, but `pg_config` is. It would be nice if we would use `pg_config` to find `pg_ctl` if it is not in `PATH`. This should be fairly easy, because of the changes from https://github.com/citusdata/pg_auto_failover/pull/510

DimCitus:
This improves default integration with debian/ubuntu packaging system for
Postgres where the postgresql-common package installs a pg_config version in
the main PATH and keeps major-dependent versions of pg_ctl in their specific
directories, that are not automatically added to PATH.

Fixes #514

### DIFF
--- a/src/bin/pg_autoctl/cli_common.c
+++ b/src/bin/pg_autoctl/cli_common.c
@@ -949,7 +949,7 @@ set_first_pgctl(PostgresSetup *pgSetup)
 	}
 
 	/* then, use PATH and fetch the first entry there for the monitor */
-	if (search_path_first("pg_ctl", pgSetup->pg_ctl))
+	if (search_path_first("pg_ctl", pgSetup->pg_ctl, LOG_WARN))
 	{
 		if (!pg_ctl_version(pgSetup))
 		{
--- a/src/bin/pg_autoctl/cli_common.c
+++ b/src/bin/pg_autoctl/cli_common.c
@@ -949,17 +949,25 @@ set_first_pgctl(PostgresSetup *pgSetup)
 	}
 
 	/* then, use PATH and fetch the first entry there for the monitor */
-	if (!search_path_first("pg_ctl", pgSetup->pg_ctl))
+	if (search_path_first("pg_ctl", pgSetup->pg_ctl))
 	{
-		/* errors have already been logged */
-		exit(EXIT_CODE_BAD_ARGS);
+		if (!pg_ctl_version(pgSetup))
+		{
+			/* errors have been logged in pg_ctl_version */
+			exit(EXIT_CODE_PGCTL);
+		}
+
+		return;
 	}
 
-	if (!pg_ctl_version(pgSetup))
+	/* then, use PATH and fetch pg_config --bindir from there */
+	if (set_pg_ctl_from_pg_config(pgSetup))
 	{
-		/* errors have been logged in pg_ctl_version */
-		exit(EXIT_CODE_PGCTL);
+		return;
 	}
+
+	/* at this point we don't have any other ways to find a pg_ctl */
+	exit(EXIT_CODE_PGCTL);
 }
 
 
--- a/src/bin/pg_autoctl/file_utils.c
+++ b/src/bin/pg_autoctl/file_utils.c
@@ -540,7 +540,7 @@ search_path_first(const char *filename, char *result)
 {
 	SearchPath paths = { 0 };
 
-	if (!search_path(filename, &paths))
+	if (!search_path(filename, &paths) || paths.found == 0)
 	{
 		log_error("Failed to find %s command in your PATH", filename);
 		return false;
--- a/src/bin/pg_autoctl/file_utils.c
+++ b/src/bin/pg_autoctl/file_utils.c
@@ -536,13 +536,13 @@ path_in_same_directory(const char *basePath, const char *fileName,
  * in PATH.
  */
 bool
-search_path_first(const char *filename, char *result)
+search_path_first(const char *filename, char *result, int logLevel)
 {
 	SearchPath paths = { 0 };
 
 	if (!search_path(filename, &paths) || paths.found == 0)
 	{
-		log_error("Failed to find %s command in your PATH", filename);
+		log_level(logLevel, "Failed to find %s command in your PATH", filename);
 		return false;
 	}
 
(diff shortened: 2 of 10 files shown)


src/bin/pg_autoctl/pgctl.c
```diff
@@ -334,7 +334,51 @@ config_find_pg_ctl(PostgresSetup *pgSetup)
...
+			log_debug("Found pg_ctl for PostgreSQL %s at %s from pg_config "
+					  "found in PATH at \"%s\"",
+					  pgSetup->pg_version, pgSetup->pg_ctl, pg_configs[0]);
+
+			/* set our return value to 1 in that case, indicating success */
+			n = pg_configs_count;
+		}
+		else
+		{
+			log_warn("Failed to find either pg_ctl or pg_config in PATH");
```
JelteF, line 378: Might be good to have another message in case there's more than 1 `pg_config` found. I don't expect this to happen in real setups though, i.e. no `pg_ctl` but multiple `pg_config`

JelteF (changes requested):
I'm still getting this error:
```
16:22:06 1968 ERROR file_utils.c:545: Failed to find pg_ctl command in your PATH
```

I guess it's because `set_first_pgctl` is still the same.

src/bin/pg_autoctl/pgctl.c
DimCitus: Yeah that's fair. This also triggered a more extensive review of the patch, and I went as far as cleaning up the memory allocation scheme of the `search_path` function, which landed early in the development and did not get the memo about our favorite way to allocate memory (struct on the stack).

Also I exposed the facility to be easily tested for both contexts of a monitor and a keeper:

```
$ PG_AUTOCTL_DEBUG=1 pg_autoctl do pgsetup pg_ctl --pgdata /tmp/m
```

The PGDATA directory does not need to exist, but the command requires one to be known still, because it's sharing the same getops command as many other commands in the `pg_autoctl do` family.

Finally, we currently have a different processing of `--pgctl` in the monitor and the keeper. That's because on the monitor we're not picky about the Postgres version to use, whereas on the monitor we really insist on there being absolutely no ambiguity. That said, I have to wonder if we shouldn't just apply the same thoroughness checks in both cases, actually?

DimCitus:
> I'm still getting this error:
> 
> ```
> 16:22:06 1968 ERROR file_utils.c:545: Failed to find pg_ctl command in your PATH
> ```
> 
> I guess it's because `set_first_pgctl` is still the same.

Ah yeah indeed. That's the monitor code path. I did fix only the keeper code path in the previous version of this PR. Both are addressed now.

src/bin/pg_autoctl/file_utils.h
```diff
@@ -24,6 +24,21 @@
...
+ * In order to avoid dynamic memory allocations and tracking when searching the
+ * PATH environment, we pre-allocate 1024 paths entries. That should be way
+ * more than enough for all situations, and only costs 1024*1024 = 1MB of
+ * memory.
+ */
+typedef struct SearchPath
+{
+	int total;
+	int found;
+	char entries[1024][MAXPGPATH];
```
JelteF, line 37: It seems only `matches` is actually used outside of the `search_path` function. So I think this can be removed from the struct (as well as the `total`).

src/bin/pg_autoctl/cli_common.c
```diff
@@ -949,17 +949,25 @@ set_first_pgctl(PostgresSetup *pgSetup)
 	}
 
 	/* then, use PATH and fetch the first entry there for the monitor */
-	if (!search_path_first("pg_ctl", pgSetup->pg_ctl))
+	if (search_path_first("pg_ctl", pgSetup->pg_ctl))
```
JelteF, line 952: If this fails, there's an error log like this coming from `search_path_first`, even though it's not a hard error:
```
10:26:42 1793 ERROR Failed to find pg_ctl command in your PATH
```

I think it makes sense to add a loglevel to `search_path_first`.

src/bin/pg_autoctl/file_utils.c
```diff
@@ -561,59 +562,28 @@ search_path_first(const char *filename, char *result)
  *
  * The caller should free the result by calling search_path_destroy_result().
```
JelteF, line 563: Function comment should be updated.

src/bin/pg_autoctl/file_utils.c
```diff
@@ -622,59 +592,20 @@ search_path(const char *filename, char ***result)
 			*sep = '\0';
 		}
 
-		join_path_components(candidate, path, filename);
-		canonicalize_path(candidate);
+		strlcpy(result->entries[result->total++], path, MAXPGPATH);
```
JelteF, line 595: This line can be removed afaict. (see other comment too)

src/bin/pg_autoctl/file_utils.c
```diff
@@ -768,18 +699,15 @@ set_program_absolute_path(char *program, int size)
...
 		if (pg_autoctl_argv0[0] == '/')
 		{
 			strlcpy(program, pg_autoctl_argv0, size);
 			return true;
 		}
 
-		n = search_path(pg_autoctl_argv0, &pathEntries);
-
-		if (n < 1)
+		if (!search_path(pg_autoctl_argv0, &paths))
```
JelteF, line 710: should check `paths.found` too.

JelteF (changes requested):

src/bin/pg_autoctl/file_utils.h
DimCitus: Fair point. I wanted it to be more generic that our needs, which is seldom a good idea. Fixed now.

src/bin/pg_autoctl/cli_common.c
DimCitus: Good point, done now.

src/bin/pg_autoctl/file_utils.c
DimCitus: Function comment updated now.

src/bin/pg_autoctl/file_utils.c
DimCitus: Removed.

src/bin/pg_autoctl/file_utils.c
DimCitus: You did that, thanks!

JelteF: Not in this one, I did it in `search_path_first` only to fix the monitor creation for me. And then I noticed the same issue here during review.

DimCitus: Ah. Indeed. Sorry about that, done now.

JelteF (approved):

PR #519, opened 2020-11-19
Fix build for *BSD

eradman:
Only define `union semun` on Linux. Tested building pg_auto_failover with this change on

- Fedora 33
- Alpine 3.12.0
- Mac OS 10.14.6
- OpenBSD 6.8


DimCitus (approved):

DimCitus:
Thanks for your contribution @eradman !

PR #521, opened 2020-11-25
Fix unhealthy nodes processing when wait_primary.

DimCitus:
The FSM state wait_primary embeds the information that there is no solution
to failover, and as such there is no point in changing the state of the node
to draining or otherwise, as no failover process is going to kick-in.

### DIFF
--- a/src/monitor/group_state_machine.c
+++ b/src/monitor/group_state_machine.c
@@ -1085,15 +1085,46 @@ ProceedGroupStateForMSFailover(AutoFailoverNode *activeNode,
 	 * The currently selected node might not be marked healthy at this time
 	 * because in REPORT_LSN we shut Postgres down. We still should proceed
 	 * with the previously selected node in that case.
+	 *
+	 * Also pay attention that Postgres might be running again, as reported by
+	 * the pg_autoctl node-active protocol, while our health checks have not
+	 * picked-up on that fact yet. The IsHealthy() check requires both checks
+	 * to be all-green, finding the candidate is okay with only one of those
+	 * being good: we really need to avoid having two candidates at the same
+	 * time.
 	 */
-	if (nodeBeingPromoted != NULL &&
-		(nodeBeingPromoted->reportedState == REPLICATION_STATE_REPORT_LSN ||
-		 IsHealthy(nodeBeingPromoted)))
+	if (nodeBeingPromoted != NULL)
 	{
-		elog(LOG, "Found candidate " NODE_FORMAT,
-			 NODE_FORMAT_ARGS(nodeBeingPromoted));
+		char message[BUFSIZE] = { 0 };
 
-		return ProceedWithMSFailover(activeNode, nodeBeingPromoted);
+		List *knownUnreachableStates =
+			list_make2_int(REPLICATION_STATE_REPORT_LSN,
+						   REPLICATION_STATE_PREPARE_PROMOTION);
+
+		LogAndNotifyMessage(
+			message, BUFSIZE,
+			"Active " NODE_FORMAT
+			" found failover candidate " NODE_FORMAT
+			" being promoted, it reported state \"%s\"",
+			NODE_FORMAT_ARGS(activeNode),
+			NODE_FORMAT_ARGS(nodeBeingPromoted),
+			ReplicationStateGetName(nodeBeingPromoted->reportedState));
+
+		/*
+		 * In state REPORT_LSN we stop Postgres. In the transition from
+		 * REPORT_LSN to PREPARE_PROMOTION we do not start Postgres yet, we
+		 * only restart Postgres when reaching STOP_REPLICATION, and we set
+		 * default_transaction_read_only to on.
+		 */
+		if (IsStateIn(nodeBeingPromoted->reportedState, knownUnreachableStates) ||
+			nodeBeingPromoted->pgIsRunning ||
+			IsHealthy(nodeBeingPromoted))
+		{
+			elog(LOG, "Found candidate " NODE_FORMAT,
+				 NODE_FORMAT_ARGS(nodeBeingPromoted));
+
+			return ProceedWithMSFailover(activeNode, nodeBeingPromoted);
+		}
 	}
 
 	/*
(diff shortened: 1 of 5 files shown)


tests/test_multi_async.py
```diff
@@ -178,3 +178,45 @@ def test_012_ifup_node4():
...
+
+    ssn = ""
+    eq_(node2.get_synchronous_standby_names(), ssn)
+    eq_(node2.get_synchronous_standby_names_local(), ssn)
+
+def test_015_stop_primary():
+    node2.fail()
+
+    print()
+    assert node3.wait_until_state(target_state="report_lsn")
```
JelteF, line 204: I don't think this should happen. It should stay a `secondary` (or should become `catchingup` maybe).
This causes some weird behavior as a side effect. When node2 comes up now, without node1 coming up as well, suddenly node2 becomes `primary` instead of staying `wait_primary`.

JelteF: This also seems to mess up some other things. When node1 comes up after what I described above, then it will stay in demoted forever.

DimCitus: Okay I'm going to see about adding another series of tests with node1 restarted first then. I still thing REPORT_LSN is a good assignment for the non-candidate standby when there's no other node available, because that's the state that we have for a standby without a any upstream node at the moment.

src/monitor/group_state_machine.c
```diff
@@ -1028,13 +1082,55 @@ ProceedGroupStateForMSFailover(AutoFailoverNode *activeNode,
...
+	 * The currently selected node might not be marked healthy at this time
+	 * because in REPORT_LSN we shut Postgres down. We still should proceed
+	 * with the previously selected node in that case.
+	 *
+	 * Also pay attention that Postgres might be running again, as reported by
+	 * the pg_autoctl node-active protocol, while our health checks have not
+	 * picked-up on that fact yet. The IsHealthy() check requires both checks
+	 * to be all-green, finding the candidate is okay with only one of those
+	 * being good: we really need to avoid having two candidates at the same
+	 * time.
```
JelteF, line 1095: I believe the changes that this comment addresses have been (re)moved again. I don't see anything related to `isRunning` and the `REPORT_LSN` stuff is addressed below in a second comment already.

tests/test_basic_operation.py
```diff
@@ -75,31 +75,48 @@ def test_005_read_from_secondary():
...
     print("Disabling maintenance on node1")
     node1.disable_maintenance()
     assert node1.wait_until_pg_is_running()
     assert node1.wait_until_state(target_state="secondary")
     assert node2.wait_until_state(target_state="primary")
 
-    eq_(node2.get_synchronous_standby_names_local(), '*')
+    ssn = '*'
+    eq_(node2.get_synchronous_standby_names_local(), ssn)
+    eq_(node2.get_synchronous_standby_names(), ssn)
```
JelteF, line 117: It seems we're doing this type of check a lot. Maybe add it as a helper method so we can do:
```
node2.check_synchronous_standby_names("*")
```

Can also be done in a separate PR.

tests/test_multi_async.py
```diff
@@ -178,3 +178,112 @@ def test_012_ifup_node4():
...
+
+def test_014_003_restart_node1():
+    node1.run()
+
+    # node1 used to be primary, now demoted, and meanwhile node2 was primary
+    # so we can't decide to promote node1 now, it's stuck demoted
+    assert node1.wait_until_state(target_state="demoted")
+    assert node3.wait_until_state(target_state="report_lsn")
```
JelteF, line 230: I think this needs a bit of a `sleep` and check state again to confirm that `node1` doesn't move away from demoted.

tests/test_multi_ifdown.py
```diff
@@ -240,7 +243,7 @@ def test_015_finalize_failover_after_most_advanced_secondary_gets_back():
 
     # and, node2 should finally become the primary without losing any data
     print()
-    assert node2.wait_until_state(target_state="primary")
```
JelteF, line 243: Why is this change expected?

tests/test_multi_async.py
```diff
@@ -178,3 +178,112 @@ def test_012_ifup_node4():
...
+    node1.fail()
+
+    print()
+    assert node3.wait_until_state(target_state="report_lsn")
+
+def test_015_003_restart_node1():
+    node1.run()
+
+    # node1 used to be primary, now demoted, and meanwhile node2 was primary
+    # so we can't decide to promote node1 now, it's stuck demoted
```
JelteF, line 277: This comment is copy pasted from test 014 and is incorrect.

JelteF (approved):
Looks good overal, left some small comments.

tests/test_basic_operation.py
DimCitus: Yeah separate PR would be good. And in the 1.4.2 milestone too. See https://github.com/citusdata/pg_auto_failover/issues/527

tests/test_multi_ifdown.py
DimCitus: We just do the same test later, waiting for the primary node first have an higher chance of reaching a timeout compared to waiting for the secondary nodes first.

src/monitor/group_state_machine.c
DimCitus: Good point, yeah, this part of the code received several changes following some QA sessions and testing. Fixed now.

tests/test_multi_async.py
DimCitus: Added now, thanks.

tests/test_multi_async.py
DimCitus: Removed, thanks.

JelteF (approved):

src/monitor/group_state_machine.c
```diff
@@ -1031,10 +1085,45 @@ ProceedGroupStateForMSFailover(AutoFailoverNode *activeNode,
...
+			message, BUFSIZE,
+			"Active " NODE_FORMAT
+			" found failover candidate " NODE_FORMAT
+			" being promoted, it reported state \"%s\"",
+			NODE_FORMAT_ARGS(activeNode),
+			NODE_FORMAT_ARGS(nodeBeingPromoted),
+			ReplicationStateGetName(nodeBeingPromoted->reportedState));
+
+		/*
+		 *
```
JelteF, line 1110: There was an extra newline in this comment.

```suggestion
```

DimCitus: Ooops, sorry about that. I guess that's one way to check the stability of the test suite.

JelteF (approved):

PR #523, opened 2020-11-26
The three standby, one async was missing one arrow.

DimCitus:


JelteF:
Wasn't this on purpose? The application would never use C as fallback because it has `candidate priority = 0`.

DimCitus:
> Wasn't this on purpose? The application would never use C as fallback because it has `candidate priority = 0`.

Ah yeah, I see that you're right now. When using the image for my presentation, the lack of symmetry appeared to me as a bug, and I proceeded to quickly fix it. My focus was on recording a talk rather than what the schema represents, and this is why this PR is misguided, I guess. Thanks for catching it!

PR #533, opened 2020-12-01
Fix FSM when a non-candidate node is back online.

### LINKED ISSUE #528: wait_primary node is made primary incorrectly when candidate-priority=0 comes online again
1.
```
make cluster -j20 TMUX_LAYOUT=tiled NODES=3
```
2.
```
pg_autoctl set formation number-sync-standbys 0 && pg_autoctl set node candidate-priority --name node3 0 && pg_autoctl set node replication-quorum --name node3 false
```
3. kill node3
4. wait for node3 assigned catching up
5. kill node1
6. wait for node2 wait_primary
7. start node3
8. node2 now becomes primary and doesn't leave it anymore.

DimCitus:
When a node with candidate priority zero is back online and we implement the
catchingup to secondary transition, refrain from assigning PRIMARY to the
leader node: we don't have a candidate for failover and so should remain
wait_primary.

Fixes #528

### DIFF
--- a/src/monitor/group_state_machine.c
+++ b/src/monitor/group_state_machine.c
@@ -359,29 +359,51 @@ ProceedGroupState(AutoFailoverNode *activeNode)
 		IsHealthy(activeNode) &&
 		WalDifferenceWithin(activeNode, primaryNode, EnableSyncXlogThreshold))
 	{
-		ReplicationState primaryGoalState =
-			IsCurrentState(primaryNode, REPLICATION_STATE_PRIMARY)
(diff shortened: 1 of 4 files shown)


JelteF (approved):
LGTM, can we add a test for this?

DimCitus:
> LGTM, can we add a test for this?

Sure, test added, based on your repro in issue #528. All with a new way to test for assigned state of a failed node, because it seems that we need that now.

JelteF (approved):

PR #534, opened 2020-12-01
Fix assigning catchingup to unhealthy secondaries.

### LINKED ISSUE #529: candidate-priority=0 node doesn't get assigned catchingup when unhealthy
1.
```
```
2.
```
pg_autoctl set formation number-sync-standbys 0 && pg_autoctl set node candidate-priority --name node3 0 && pg_autoctl set node replication-quorum --name node3 false
```
3. kill node2
4. wait for node2 to be assigned catchingup (and/or node1 to be wait_primary)
5. kill node3
6. node3 never gets "catchingup", even though it is unhealthy

DimCitus:
We can now be in wait_primary and still lose a secondary node, when that
node has a candidate priority set to zero.

Fixes #529


JelteF (approved):
LGTM, can we add a test?

DimCitus:
> LGTM, can we add a test?

Test added now.

PR #535, opened 2020-12-01
Review consequences of pg_autoctl set formation number-sync-standbys 0.

### LINKED ISSUE #530: changing number-sync-standbys to 0 doesn't make primary transition to wait_primary
1.
```
```
3. kill node3
4. wait for node3 assigned catchingup
5. kill node2
6. wait for node2 assigned catchingup
7. 
```
pg_autoctl set formation number-sync-standbys 0
```
8. node1 is expected to become wait_primary now, but it doesn't

### LINKED ISSUE #538: number-sync-standbys=0 with 2 standbys doesn't move primary to wait_primary
1.
```
make cluster -j20 TMUX_LAYOUT=tiled NODES=3 NODES_SYNC_SB=0
```
2. Kill node2 and node3 (either at the same time, or kill one and wait until it's assigned catchingup and then kill the other)
3. Wait, but node1 does not become wait_primary. Instead it continues showing these errors:
```
16:46:01 19407 ERROR Failed to fetch current replication properties from standby node: no standby connected in pg_stat_replication.
16:46:01 19407 WARN  HINT: check pg_autoctl and Postgres logs on standby nodes
16:46:01 19407 WARN  Failed to update the keeper's state from the local PostgreSQL instance.
16:46:56 19407 INFO  pg_autoctl managed to ensure current state "primary": PostgreSQL is running
```

DimCitus:
In particular setting the number-sync-standbys to zero means that when we
don't have a single healthy secondary that is also a candidate for
promotion, we are allowed to reach WAIT_PRIMARY.

As a consequence, we also want to allow changing the number-sync-standbys
setting when in WAIT_PRIMARY so as to allow blocking writes when it's all
going south, in order to limit potential data loss.

Fixes #530 
Fixed #538


JelteF (changes requested):
This indeed seems to fix #530, but I get some weird behaviour when I merge your other two PRs together with this. During cluster startup I now get these errors in the logs:
```
10:58:56 29562 INFO  Fetched current list of 1 other nodes from the monitor to update HBA rules, including 1 changes.
10:58:56 29562 INFO  Ensuring HBA rules for node 2 "node2" (localhost:5502)
10:58:56 29562 INFO  Monitor assigned new state "wait_primary"
10:58:56 29562 INFO  Creating replication slot "pgautofailover_standby_2"
10:58:56 29562 INFO  FSM transition from "single" to "wait_primary": A new secondary was added
10:58:56 29562 INFO  Transition complete: current state is now "wait_primary"
10:58:56 29562 INFO  New state for node 3 "node3" (localhost:5503): wait_standby ➜ wait_standby
10:58:56 29562 INFO  New state for node 3 "node3" (localhost:5503): wait_standby ➜ catchingup
10:58:56 29562 INFO  New state for this node (node 1, "node1") (localhost:5501): wait_primary ➜ primary
10:58:56 29562 INFO  Fetched current list of 2 other nodes from the monitor to update HBA rules, including 1 changes.
10:58:56 29562 INFO  Ensuring HBA rules for node 3 "node3" (localhost:5503)
10:58:56 29562 INFO  Monitor assigned new state "primary"
10:58:56 29562 INFO  Creating replication slot "pgautofailover_standby_3"
10:58:56 29562 INFO  FSM transition from "wait_primary" to "primary": A healthy secondary appeared
10:58:56 29562 INFO  Setting synchronous_standby_names to 'ANY 1 (pgautofailover_standby_2, pgautofailover_standby_3)'
10:58:56 29562 WARN  Failed to set the standby Target LSN because we don't have a quorum candidate yet
10:58:56 29562 ERROR Failed to transition from state "wait_primary" to state "primary", see above.
10:58:56 29562 ERROR Failed to transition to state "primary", retrying...
10:58:57 29562 INFO  Monitor assigned new state "primary"
10:58:57 29562 INFO  FSM transition from "wait_primary" to "primary": A healthy secondary appeared
10:58:57 29562 INFO  Setting synchronous_standby_names to 'ANY 1 (pgautofailover_standby_2, pgautofailover_standby_3)'
10:58:57 29562 WARN  Failed to set the standby Target LSN because we don't have a quorum candidate yet
10:58:57 29562 ERROR Failed to transition from state "wait_primary" to state "primary", see above.
10:58:57 29562 ERROR Failed to transition to state "primary", retrying...
10:58:58 29562 INFO  Monitor assigned new state "primary"
10:58:58 29562 INFO  FSM transition from "wait_primary" to "primary": A healthy secondary appeared
10:58:58 29562 INFO  Setting synchronous_standby_names to 'ANY 1 (pgautofailover_standby_2, pgautofailover_standby_3)'
10:58:58 29562 WARN  Failed to set the standby Target LSN because we don't have a quorum candidate yet
10:58:58 29562 ERROR Failed to transition from state "wait_primary" to state "primary", see above.
10:58:58 29562 ERROR Failed to transition to state "primary", retrying...
10:58:59 29562 INFO  Monitor assigned new state "primary"
10:58:59 29562 INFO  FSM transition from "wait_primary" to "primary": A healthy secondary appeared
10:58:59 29562 INFO  Setting synchronous_standby_names to 'ANY 1 (pgautofailover_standby_2, pgautofailover_standby_3)'
10:59:00 29562 WARN  Failed to set the standby Target LSN because we don't have a quorum candidate yet
10:59:00 29562 ERROR Failed to transition from state "wait_primary" to state "primary", see above.
10:59:00 29562 ERROR Failed to transition to state "primary", retrying...
10:59:01 29562 INFO  Monitor assigned new state "primary"
10:59:01 29562 INFO  FSM transition from "wait_primary" to "primary": A healthy secondary appeared
10:59:01 29562 INFO  Setting synchronous_standby_names to 'ANY 1 (pgautofailover_standby_2, pgautofailover_standby_3)'
10:59:01 29562 WARN  Failed to set the standby Target LSN because we don't have a quorum candidate yet
10:59:01 29562 ERROR Failed to transition from state "wait_primary" to state "primary", see above.
10:59:01 29562 ERROR Failed to transition to state "primary", retrying...
10:59:02 29562 INFO  Monitor assigned new state "primary"
10:59:02 29562 INFO  FSM transition from "wait_primary" to "primary": A healthy secondary appeared
10:59:02 29562 INFO  Setting synchronous_standby_names to 'ANY 1 (pgautofailover_standby_2, pgautofailover_standby_3)'
10:59:02 29562 WARN  Failed to set the standby Target LSN because we don't have a quorum candidate yet
10:59:02 29562 ERROR Failed to transition from state "wait_primary" to state "primary", see above.
10:59:02 29562 ERROR Failed to transition to state "primary", retrying...
10:59:03 29562 INFO  Monitor assigned new state "primary"
10:59:03 29562 INFO  FSM transition from "wait_primary" to "primary": A healthy secondary appeared
10:59:03 29562 INFO  Setting synchronous_standby_names to 'ANY 1 (pgautofailover_standby_2, pgautofailover_standby_3)'
10:59:03 29562 WARN  Failed to set the standby Target LSN because we don't have a quorum candidate yet
10:59:03 29562 ERROR Failed to transition from state "wait_primary" to state "primary", see above.
10:59:03 29562 ERROR Failed to transition to state "primary", retrying...
```

JelteF (changes requested):

JelteF:
Found an issue:
1. 
```
make cluster -j20 TMUX_LAYOUT=tiled NODES=3 NODES_SYNC_SB=0
```
2. wait until cluster is stable
3. kill both secondaries
4. wait until primary is wait_primary
5. connect to primary and insert something
6. 
```
pg_autoctl set formation number-sync-standbys 1
```
7. see that node1 becomes primary
8. kill node1
9. start node 2 and 3
10. They start an election and one of them becomes primary. This primary doesn't have data that was inserted at step 5, i.e. committed data got lost.

DimCitus:
> 3. wait until primary is wait_primary
> 4. connect to primary and insert something

So here the INSERT is committed while the node is wait_primary, with number-sync-standbys set to zero. By user choice, we make no guarantees about that transaction durability in the event of a failover.

> `pg_autoctl set formation number-sync-standbys 1`

That's happening too late, we can't guarantee transactions successfully committed in the past.

> 4. They start an election and one of them becomes primary. This primary doesn't have data that was inserted at step 5, i.e. committed data got lost.

Yeah, that's what can happen with number-sync-standbys set to zero, allowing asynchronous replication.

I do not believe this is something we can solve in pg_auto_failover, either in this PR or otherwise.

JelteF:
Like discussed in chat, step 7 is where the problem is, instead of fully becoming primary it should be:
goal state becomes be primary, but actual state should be wait_primary (and stay there until a secondary comes up and catches up)

DimCitus:
> Like discussed in chat, step 7 is where the problem is, instead of fully becoming primary it should be:
> goal state becomes be primary, but actual state should be wait_primary (and stay there until a secondary comes up and catches up)

So I fixed this a little differently in the current PR:

  - when `pg_autoctl set formation number-sync-standbys 1` is done in step 6 then node1, currently WAIT_PRIMARY, is assigned APPLY_SETTINGS as expected
  - then when it reaches the APPLY_SETTINGS state, the next step is computed to now be PRIMARY, because number sync standby is 1 which forbids WAIT_PRIMARY, even though the two known standby nodes are currently not available: the goal of the user setting is now to block writes and that's what we do
  - because there is not standby available, the goal state PRIMARY can't be reached, so node1 remains at APPLY_SETTINGS/PRIMARY until something changes.

So that means that step 7. does not happen anymore: node1 does not become primary.

Now when node1 is killed in step 8, no node is running, so nothing happens.

Then in step 9 nodes 2 and 3 are started again. Because the known primary node `IsUnhealthy` at this moment, both nodes get assigned and reach REPORT_LSN. One of the nodes is elected as the new primary, and needs to fetch missing WAL from step 5; which is not possible until node1 is back online. The system remains in the state with node1 down, node2 and node3 at report_lsn.

When node1 is back online, the failover goes on and in my case node2 was elected as the new primary, doing a fast_forward from node1 to fetch the missing WAL bits. 

It is possible to argue that the failover that happens here is not necessary. One way that I can think of to improve the current situation would be to only assign REPORT_LSN to trigger a failover when the known WAL lag compared to the primary node last reported LSN exceeds the `pgautofailover.promote_wal_log_threshold` GUC on the monitor.

In that case we would proceed with a full failover and lose data in your test case scenario, unless `pgautofailover.promote_wal_log_threshold` has been set to zero or some very small value, and then no failover would happen, the system would just wait until the primary is back online, so that we avoid data loss entirely.

JelteF (approved):

tests/test_multi_ifdown.py
```diff
@@ -102,11 +103,10 @@ def test_005_set_candidate_priorities():
     assert node2.get_replication_quorum()
 
     # also let's see synchronous_standby_names here
-    print("Monitor: %s" % node1.get_synchronous_standby_names())
-    print(
-        "Node 1:  %s"
-        % node1.run_sql_query("show synchronous_standby_names")[0][0]
-    )
+    # remember to sort by candidate priority then name
+    ssn = "ANY 1 (pgautofailover_standby_3, pgautofailover_standby_2)"
```
JelteF, line 107: Maybe we should remove this special case ordering. Postgres doesn't do anything special. (for another PR)

DimCitus: It's nice for the unit tests, so there's that... otherwise it's not useful, but also I think it's neutral, as Postgres will just not care about the ordering when using `ANY` anyway...

tests/test_multi_standbys.py
```diff
@@ -243,13 +248,118 @@ def test_012_fail_primary():
...
+    node3.run()
+    node2.run()
+
+    assert node3.wait_until_state(target_state="secondary")
+    assert node2.wait_until_state(target_state="secondary")
+    assert node1.wait_until_state(target_state="primary")
+
+    ssn = "ANY 1 (pgautofailover_standby_2, pgautofailover_standby_3)"
+    eq_(node1.get_synchronous_standby_names(), ssn)
+    eq_(node1.get_synchronous_standby_names_local(), ssn)
```
JelteF, line 359: I'm missing a test where we change increase `number-sync-standbys` to 1 to block the writes on the wait_primary.

DimCitus:
After careful analysis with Jelte in a meeting we have concluded that we don't want to include the primary's LSN in the computation for the most advanced (standby) node on the system, and rather rely as before on the fact that the primary was in the state PRIMARY, which implies it had at least 1 secondary node that it was waiting for to acknowledge the commits.

tests/test_multi_standbys.py
DimCitus: Fair. That said when doing that the wait_primary node is going to be apply_settings/primary and failing to reach primary for a while, so the `pg_autoctl set formation number-sync-standbys 1` command is going to time-out after 60s.

DimCitus: Test added now, using `monitor.run_sql_query()` instead of the command line, to avoid the timeout and more failure modes to take into account.

PR #543, opened 2020-12-03
Prepare for release 1.4.1

DimCitus:
Edit version number in defaults.h and edit CHANGELOG for the new version.


hanefi (approved):
LGTM
All the items in #541 seems to be here.

I did not check if all the vendored-in libraries are up-to-date however I believe you checked them already.

hanefi:
> I did not check if all the vendored-in libraries are up-to-date however I believe you checked them already.

Clarification: I only checked that the vendored-in libraries in this repo are touched more recently than the original sources. Assuming that 1.4.0 was fine, this one is fine as well.

PR #544, opened 2020-12-03
Implement pg_autoctl do azure commands.

DimCitus:
The following commands allow to easily deploy a testing or QA environment
for pg_auto_failover using Azure VMs, which looks more like what typical
users and customers are using than our default `make cluster` approach where
all the nodes are running on localhost.

In particular we can test system integration and packages based
provisioning, so that's better for end-to-end QA.

Here's the list of new commands provided:

    pg_autoctl do azure
    + provision  provision azure resources for a pg_auto_failover demo
    + tmux       Run a tmux session with an Azure setup for QA/testing
    + show       show azure resources for a pg_auto_failover demo
      deploy     Deploy a pg_autoctl VMs, given by name
      create     Create an azure QA environment
      drop       Drop an azure QA environment: resource group, network, VMs
      ls         List resources in a given azure region
      ssh        Runs ssh -l ha-admin <public ip address> for a given VM name
      sync       Rsync pg_auto_failover sources on all the target region VMs
  
    pg_autoctl do azure provision
      region  Provision an azure region: resource group, network, VMs
      nodes   Provision our pre-created VM with pg_autoctl Postgres nodes
  
    pg_autoctl do azure tmux
      session  Create or attach a tmux session for the created Azure VMs
      kill     Kill an existing tmux session for Azure VMs
  
    pg_autoctl do azure show
      ips    Show public and private IP addresses for selected VMs
      state  Connect to the monitor node to show the current state

And there's the new top-level commands:

    make azcluster
    make azdrop

Contrary to `make cluster` when exiting the tmux session the resources are not dropped, and it's possible to connect again to the same tmux session with either `make azcluster` or `pg_autoctl do azure tmux session`.

### DIFF
--- a/src/bin/pg_autoctl/azure.c
+++ b/src/bin/pg_autoctl/azure.c
@@ -1173,7 +1173,9 @@ azure_prepare_target_versions(KeyVal *env)
  * azure_prepare_debian_command prepares the debian command to install our
  * target pg_auto_failover package on the Azure VMs.
  *
- *   sudo apt-get install -q -y postgresql-13-auto-failover-1.4=1.4.1-1
+ *   sudo apt-get install -q -y                 \
+ *      postgresql-13-auto-failover-1.4=1.4.1-1 \
+ *      pg-auto-failover-cli-1.4=1.4.0-1
  *
  * We are using environment variables to fill in the actual version numbers,
  * and we hard-code some defaults in case the environment has not be provided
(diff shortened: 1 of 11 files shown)


src/bin/pg_autoctl/azure.c
```diff
@@ -0,0 +1,2521 @@
...
+		{
+			log_level(errorLogLevel, "%s", errorLines[lineNumber]);
+		}
+	}
+}
+
+
+/*
+ * run_az_command runs a command line using the azure CLI command, and when
+ * azureScript is true instead of running the command it only shows the command
```
JelteF, line 104:
```suggestion
 * dryRun is true instead of running the command it only shows the command
```

JelteF (changes requested):

JelteF (approved):
I skimmed the code in this PR and it looked reasonable. Since this is essentially just dev helpers I think it's fine to merge this without me doing a very thorough review. I ran `make azcluster` and `make azdrop`. They worked, but there's some pg_hba lines that we probably want to add. I think it's fine to do that in a later PR though.

src/bin/pg_autoctl/azure.c
DimCitus: Good catch, done.

PR #546, opened 2020-12-07
Actually run black in CI

JelteF:
I realised after pressing merge that #216 didn't actually add `black` as
part of CI. This addresses that, so that we know formatting is always
done.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,8 +36,11 @@ before_install:
   - sudo apt-get install python3-pip
   - sudo pip3 install --upgrade pip
   - pip3 install --user pipenv
+  - pip3 install --user black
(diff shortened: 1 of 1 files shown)


DimCitus (approved):

DimCitus:
Well it seems we need to fix that before merging:

```
Oh no! 💥 💔 💥
22 files would be reformatted, 21 files would be left unchanged.
```

DimCitus:
> Well it seems we need to fix that before merging (seems like the command is not running in the right current working directory):

I think we need `make check-indent` and `make check-banned` targets, and maybe a `make lint` target that just depends on the previous two, and use that in Travis.

PR #550, opened 2020-12-11
Fix bytes mangling opportunity.

DimCitus:
The read() call might have left garbage in the buffer after the last byte,
so enforce end-of-string at the boundary returned.

### DIFF
--- a/src/bin/lib/subcommands.c/runprogram.h
+++ b/src/bin/lib/subcommands.c/runprogram.h
@@ -596,6 +596,9 @@ read_into_buf(Program *prog, int filedes, PQExpBuffer buffer, bool error)
 
 	if (bytes > 0)
 	{
+		/* terminate the buffer after the length we read bytes */
+		temp_buffer[bytes] = '\0';
+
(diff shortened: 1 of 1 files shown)


src/bin/lib/subcommands.c/runprogram.h
```diff
@@ -596,6 +596,9 @@ read_into_buf(Program *prog, int filedes, PQExpBuffer buffer, bool error)
 
 	if (bytes > 0)
 	{
+		/* terminate the buffer after the length we read bytes */
+		temp_buffer[bytes] = '\0';
```
marcocitus, line 600: We should increase the buffer size by 1 and set the last bit to 0. Main reason is that we assume this to be a 0-terminated string below.

DimCitus: Good point about the +1. The last bit to zero is done already as far as I can see.

marcocitus: true, the +1 should be sufficient, but maybe leave this line here to make it easier on the reader.

marcocitus (approved):

PR #553, opened 2020-12-14
Test/run make installcheck in docker and travis

### LINKED ISSUE #464: Running postgres regression tests

DimCitus:
At the moment the `make -C src/monitor intallcheck` target fails. We often forget to update our test suite for the monitor. This PR includes this step in the travis tests.

Fixes #464

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,11 +1,8 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the PostgreSQL License.
 
-FSM = docs/fsm.png
-CONTAINER_NAME = pg_auto_failover
 TEST_CONTAINER_NAME = pg_auto_failover_test
 DOCKER_RUN_OPTS = --privileged  -ti --rm
-PDF = ./docs/_build/latex/pg_auto_failover.pdf
 
 NOSETESTS = $(shell which nosetests3 || which nosetests)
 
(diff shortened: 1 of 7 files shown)


JelteF (approved):
Tried this out after rebasing on top of #556. The test failed when I edited the `.out` file. So it works and code LGTM.

PR #561, opened 2020-12-23
Implement pg_autoctl create postgres --pg-hba-lan option.

DimCitus:
With this option pg_autoctl edits the pg_hba.conf for Postgres to grant
connection privileges on the detected LAN for the --dbname database and for
the --username user.

The LAN detection is done the same way as with the monitor.

The idea is to simplify steps for running demos using docker-compose, where a single command can be used to start processes in each contained, and sync between running containers and extra actions isn't easily implemented.

### DIFF
--- a/src/bin/pg_autoctl/cli_do_tmux.c
+++ b/src/bin/pg_autoctl/cli_do_tmux.c
@@ -646,7 +646,9 @@ tmux_pg_autoctl_create_postgres(PQExpBuffer script,
 
 	tmux_add_send_keys_command(script,
 							   "%s create postgres %s "
-							   "--monitor %s --name %s "
+							   "--monitor %s "
+							   "--name %s "
+							   "--dbname demo --pg-hba-lan "
 							   "--replication-quorum %s "
 							   "--candidate-priority %d "
 							   "--run",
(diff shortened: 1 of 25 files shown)


JelteF (approved):

src/bin/pg_autoctl/pgsetup.c
```diff
@@ -1873,3 +1887,63 @@ pg_setup_standby_slot_supported(PostgresSetup *pgSetup, int logLevel)
...
+		}
+
+		case HBA_EDIT_MINIMAL:
+		{
+			return "minimal";
+		}
+
+		case HBA_EDIT_APP:
+		{
+			return "app";
```
JelteF, line 1940: maybe call it `"lan"` instead of `"app"` and use `HBA_EDIT_LAN` instead of `HBA_EDIT_APP`

DimCitus: Yeah the naming there had me a little confused, as you can see. I am not fond on the name of the option `--pg-hba-lan` though it might work well enough, and might only get used in our docker-compose tests anyway.

src/bin/pg_autoctl/cli_do_tmux.c
```diff
@@ -472,7 +472,9 @@ tmux_pg_autoctl_create_postgres(PQExpBuffer script,
 
 	tmux_add_send_keys_command(script,
 							   "%s create postgres %s "
-							   "--monitor %s --name %s "
+							   "--monitor %s "
+							   "--name %s "
+							   "--dbname demo --pg-hba-lan "
```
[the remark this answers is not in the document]
DimCitus: Fixed, thanks.

src/bin/pg_autoctl/pgsetup.c
DimCitus: Renamed that way now.

PR #572, opened 2021-01-18
Fix the return code when stop the node that is started by pg_autoctl

gfphoenix78:
When the node is not started by pg_autoctl, the pid file doesn't
exist, so pg_autoctl raises an error(calls log_fatal(...)). But
log_fatal() is not a no-return function, it finally returns 0 to
its parent process.


DimCitus (approved):

DimCitus:
Thanks for your contribution @gfphoenix78 !

PR #573, opened 2021-01-18
Infrastructure to handle dynamic services

gkokolatos:
This PR is a result of the discussion found in #539 and should be considered
as part one in a series or future PRs.

* What is a dynamic service?

Until now, any failure in a service running under supervision can potentially
lead to a full node failure. That behaviour was solely controlled by the
restart policy of the service. Additionally services were statically defined
or configured and initialized during startup. Once a service was started, was
not possible to stop it individually.

A dynamic service is one that can be started or stopped at will, at any point
during the lifetime of the process group. It is run under supervision as any
other service, yet any failures will not affect any other services of the node,
regardless of its restart policy.

* Why is it useful?

It is useful if non-core services are required. Such a service could be a
connection pooler. Connection poolers can be added or removed at any node and
at any point in the lifetime of the node.

* Implementation details.

The careful reader will notice that the distinguishable characteristic of a
dynamic service does not lay with the service itself but with the supervisor of
the node. Meaning that it is the supervisor's responsibility to start, stop and
restart a service. As such, the current implementation has left the service
struct intact and has expanded the supervisor accordingly.

The supervisor interface is expanded with three new functions,
supervisor_add_dynamic() and supervisor_remove_dynamic() in order to add/start
and to remove/stop a service from supervision respectively, and
supervisor_service_exists() in order to find a service run under supervision.
All services are uniquely identifiable by name.

The caller of the supervisor can decide during startup, if it will require the
handling of any dynamic services. If it does, then it will have to provide a
dynamicHandler() function during startup. The supervisor will call it, when
signaled to do so. To accommodate for the above, the function supervisor_start()
has been expanded accordingly.

The structure defining the supervisor has been expanded with two anonymous
structs, dynamicServices and dynamicRecentlyRemoved. These structs are not meant
to be accessed directly. What is a bit more interesting is why there were two
needed. It is due to the supervisor_loop design. There exists a disconnection
between a service stopping and an action taken for it. During this, it is
required to remember which service was stopped and its previously running pid,
for the supervisor to take the appropriate action. In short, if an exited
service is found in the dynamicRecentlyRemoved array, then the supervisor can
continue as normal.

### DIFF
--- a/src/bin/pg_autoctl/cli_common.c
+++ b/src/bin/pg_autoctl/cli_common.c
@@ -1625,6 +1625,37 @@ cli_pg_autoctl_reload(const char *pidfile)
 
 
 /*
+ * cli_pg_autoctl_handle_dynamic signals the pg_autoctl to handle any dynamic
+ * services based on its configuration
+ */
+bool
+cli_pg_autoctl_handle_dynamic(const char *pidfile)
+{
+	pid_t pid;
+
+	if (read_pidfile(pidfile, &pid))
+	{
+		if (pid <= 0)
+		{
+			log_error("Failed to handle dynamic services pg_autoctl: "
+					  "pid file \"%s\" contains negative-or-zero pid %d",
+					  pidfile, pid);
+			return false;
+		}
+
+		if (kill(pid, SIGUSR1) != 0)
+		{
+			log_error("Failed to send SIGUSR1 to the pg_autoctl's pid %d: %m",
+					  pid);
+			return false;
+		}
+	}
+
+	return true;
+}
+
+
+/*
  * cli_node_metadata_getopts parses the command line options for the
  * pg_autoctl set node metadata command.
  */
--- a/src/bin/pg_autoctl/dynamic_services_config.c
+++ b/src/bin/pg_autoctl/dynamic_services_config.c
@@ -1,0 +1,398 @@
+/*
+ * src/bin/pg_autoctl/dynamic_services_config.h
+ *     Dynamic services configuration data structure and function definitions
+ *
+ * XXX: Add copywrite as required
+ */
+
+#include "postgres_fe.h"
+
+#include "config.h"
+#include "defaults.h"
+#include "dynamic_services_config.h"
+#include "env_utils.h"
+#include "log.h"
+#include "monitor.h"
+#include "parson.h"
+#include "service_monitor.h"
+#include "supervisor.h"
+
+/* Helper definitions */
+#define SERVICES_CONFIG_FILENAME "services.cfg"
+
+/*
+ * Helper structs for json parsing
+ */
+typedef struct DynamicConfigEntry
+{
+	bool enabled;
+	char role[NAMEDATALEN];
+	char name[NAMEDATALEN];
+} DynamicConfigEntry;
+
+typedef struct DynamicServicesConfig
+{
+	int entryCount;
+	DynamicConfigEntry entries[MAX_SERVICES];
+} DynamicServicesConfig;
+
+/*
+ * Holds a read only array of services that can be considered as dynamic.
+ * For a matching role, the user can use the template to fill in the service
+ * definition for a fullfletched running service if required.
+ */
+ServiceArray const serviceTemplates = {
+	.array = {
+#if 0
+
+		/* Example of a service template */
+		.role = PGBOUNCER_ROLE,
+		.name = "placeholder",
+		.policy = RP_PERMANENT,
+		.startFunction = service_pgbouncer_start,
+		.context = NULL, /* Be explicit */
+#endif
+	},
+	.serviceCount = 0,
+};
+
+static bool dynamic_services_template_from_role(const char *role,
+												Service *const service);
+static bool dynamic_services_get_config_filename(char *servicesFilename);
+static void dynamic_services_edit_json_array(JSON_Array *jsonArray,
+											 const ServiceArray *services,
+											 bool enabledValue);
+
+/*
+ * dynamic_services_role_exists checks wether the role exists in the template
+ * services definitions. The user may pass in a service to be filled in with
+ * the template if needed.
+ *
+ * Returns true if the role is defined, false otherwise.
+ */
+static bool
+dynamic_services_template_from_role(const char *role,
+									Service *const service)
+{
+	for (int serviceIndex = 0;
+		 serviceIndex < serviceTemplates.serviceCount;
+		 serviceIndex++)
+	{
+		const Service *const template = &(serviceTemplates.array[serviceIndex]);
+		if (strcmp(role, template->role) == 0)
+		{
+			if (service != NULL)
+			{
+				*service = *template;
+			}
+			return true;
+		}
+	}
+
+	return false;
+}
+
+
+/*
+ * dynamic_services_get_config_filename constructs the complete, absolute path,
+ * filename used for dynamic services configuration. It is heavily relying on
+ * environmental values because one main user of the file is the supervisor who
+ * does not have any knowledge of paths. Any other users should have set the
+ * environment accordingly before calling dynamic_services_* functions.
+ *
+ * It is the caller's responsibility that servicesFilename is large enough.
+ *
+ * Returns true on success. Otherwise the contents of servicesFilename are not
+ * to be trusted.
+ */
+static bool
+dynamic_services_get_config_filename(char *servicesFilename)
+{
+	char home[MAXPGPATH];
+	char pgdata[MAXPGPATH];
+
+	if (!get_env_pgdata(pgdata))
+	{
+		log_error("BUG: PGDATA env not set");
+		return false;
+	}
+
+	/* build_xdg_path call exit if it fails to find HOME */
+	if (!get_env_copy("HOME", home, MAXPGPATH))
+	{
+		log_error("BUG: HOME env not set");
+		return false;
+	}
+
+	if (!build_xdg_path(servicesFilename,
+						XDG_CONFIG,
+						pgdata,
+						SERVICES_CONFIG_FILENAME))
+	{
+		/* It has already logged why */
+		return false;
+	}
+
+	return true;
+}
+
+
+/*
+ * dynamic_services_edit_json_array edits a valid jsonArray to contain the
+ * service entries passed. The services array passed is not meant to be
+ * exchaustive and any existing entries in the jsonArray not present in the
+ * provided services argument, will not be altered. This makes it safe for
+ * callers of this function to be clients that have limited scope.
+ *
+ * Parameters
+ *		jsonArray: an editable jsonArray to hold the entries, can be empty.
+ *		services:  a read only ServiceArray which holds the services to be
+ *				   edited, can be NULL
+ *		enabledValue: a boolean which sets the 'enabled' value in configuration
+ */
+static void
+dynamic_services_edit_json_array(JSON_Array *jsonArray,
+								 const ServiceArray *services,
+								 bool enabledValue)
+{
+	int enabled = enabledValue ? 1 : 0; /* json is a bit silly */
+	int arrayCount = json_array_get_count(jsonArray);
+
+	/* Nothing to be done */
+	if (services == NULL)
+	{
+		return;
+	}
+
+	for (int serviceIndex = 0;
+		 serviceIndex < services->serviceCount;
+		 serviceIndex++)
+	{
+		const Service *service = &(services->array[serviceIndex]);
+		JSON_Value *jsonEntryValue;
+		JSON_Object *jsonEntryObject;
+		bool inlineEdit = false;
+
+		if (!dynamic_services_template_from_role(service->role, NULL))
+		{
+			log_info("Service role %s is not defined as dynamic",
+					 service->role);
+			continue;
+		}
+
+		/* Change inline if existing */
+		for (int i = 0; i < arrayCount; i++)
+		{
+			const char *role;
+			const char *name;
+
+			jsonEntryObject = json_array_get_object(jsonArray, i);
+			role = json_object_get_string(jsonEntryObject, "role");
+			name = json_object_get_string(jsonEntryObject, "name");
+			if (role == NULL || strcmp(role, service->role) != 0 ||
+				name == NULL || strcmp(name, service->name) != 0)
+			{
+				continue;
+			}
+
+			inlineEdit = true;
+			if (json_object_get_boolean(jsonEntryObject, "enabled") != enabled)
+			{
+				json_object_remove(jsonEntryObject, "enabled");
+				json_object_set_boolean(jsonEntryObject, "enabled", enabled);
+			}
+			break;
+		}
+
+		if (inlineEdit)
+		{
+			continue;
+		}
+
+		/* create a new entry */
+		jsonEntryValue = json_value_init_object();
+		jsonEntryObject = json_value_get_object(jsonEntryValue);
+		json_object_set_string(jsonEntryObject, "role", service->role);
+		json_object_set_string(jsonEntryObject, "name", service->name);
+		json_object_set_boolean(jsonEntryObject, "enabled", enabled);
+		json_array_append_value(jsonArray, jsonEntryValue);
+
+		/* should not free jsonEntryValue here */
+	}
+}
+
+
+/*
+ * dynamic_services_read_config
+ */
+bool
+dynamic_services_read_config(ServiceArray *enabledServices)
+{
+	JSON_Value *jsonValue;
+	const JSON_Array *jsonArray;
+	const JSON_Object *jsonObject;
+	char servicesFilename[MAXPGPATH];
+	DynamicServicesConfig dynConf = { 0 };
+
+	if (!dynamic_services_get_config_filename(servicesFilename))
+	{
+		/* it has already logged why */
+		return false;
+	}
+
+	jsonValue = json_parse_file_with_comments(servicesFilename);
+	if (jsonValue == NULL)
+	{
+		if (!file_exists(servicesFilename))
+		{
+			log_error("Failed to parse json format in file %s", servicesFilename);
+		}
+		return false;
+	}
+
+	jsonObject = json_value_get_object(jsonValue);
+	jsonArray = json_object_get_array(jsonObject, "services");
+	if (jsonArray == NULL)
+	{
+		log_error("Corrupted services configuration file %s", servicesFilename);
+		json_value_free(jsonValue);
+		return false;
+	}
+
+	/*
+	 * Do not get tempted to limit array count to MAX_SERVICES because that
+	 * limit applies only to the enabled services, there can as many disabled
+	 * services as desired.
+	 */
+	for (int i = 0; i < json_array_get_count(jsonArray); i++)
+	{
+		DynamicConfigEntry *entry = &(dynConf.entries[dynConf.entryCount]);
+		JSON_Object *jsonObject = json_array_get_object(jsonArray, i);
+		const char *role;
+		const char *name;
+		size_t buflen;
+
+		/* Skip not strictly enabled services */
+		if (json_object_get_boolean(jsonObject, "enabled") != 1)
+		{
+			continue;
+		}
+
+		role = json_object_get_string(jsonObject, "role");
+		buflen = json_object_get_string_len(jsonObject, "role");
+		if (buflen == 0 || buflen >= sizeof(entry->role))
+		{
+			continue;
+		}
+
+		name = json_object_get_string(jsonObject, "name");
+		buflen = json_object_get_string_len(jsonObject, "name");
+		if (buflen == 0 || buflen >= sizeof(entry->name))
+		{
+			continue;
+		}
+
+		strlcpy(entry->role, role, sizeof(entry->role));
+		strlcpy(entry->name, name, sizeof(entry->name));
+
+		/* not strictly needed but be a good citizen */
+		entry->enabled = true;
+
+		if (dynConf.entryCount == (MAX_SERVICES - 1))
+		{
+			break;
+		}
+		dynConf.entryCount++;
+	}
+	json_value_free(jsonValue);
+
+	/*
+	 * Now create a service struct for each of the enabled services based on
+	 * configured roles.
+	 */
+	for (int i = 0; i < dynConf.entryCount; i++)
+	{
+		const DynamicConfigEntry *entry = &(dynConf.entries[i]);
+		Service templateService = { 0 };
+		Service *enabledService =
+			&(enabledServices->array[enabledServices->serviceCount]);
+
+		if (!dynamic_services_template_from_role(entry->role, &templateService))
+		{
+			log_debug("Skipping entry for role %s", entry->role);
+			continue;
+		}
+
+		strlcpy(enabledService->role, entry->role, sizeof(enabledService->role));
+		strlcpy(enabledService->name, entry->name, sizeof(enabledService->name));
+		enabledService->policy = templateService.policy;
+		enabledService->startFunction = templateService.startFunction;
+		enabledService->context = NULL; /* be explicit */
+		enabledServices->serviceCount++;
+	}
+
+	return true;
+}
+
+
+/*
+ * dynamic_services_write_config constructs entries from the provided
+ * ServiceArrays and includes them to the configuration file.
+ *
+ * If the file does not exist, then it creates it.
+ * If the file does exist, then it modifies it.
+ *
+ * Either of the ServiceArrays can be NULL.
+ * It will go over each of the Services and will try to find the corresponding
+ * entry in the configuration file. If it finds it, then it will modify it
+ * inline if needed. If it does not find it will add it.
+ *
+ * If an entry exists in both ServiceArrays, then only the disabled one will be
+ * added.
+ *
+ * The function is written with the explicit assumption that the caller, usually
+ * a client, does not have a wholistic knowledge of all the services running and
+ * wants to modify only the service(s) the client is responsible for.
+ */
+bool
+dynamic_services_write_config(const ServiceArray *const enabledServices,
+							  const ServiceArray *const disabledServices)
+{
+	JSON_Array *jsonArray;
+	JSON_Object *jsonObject;
+	JSON_Value *jsonValue;
+	char servicesFilename[MAXPGPATH];
+
+	if (enabledServices == NULL && disabledServices == NULL)
+	{
+		return false;
+	}
+
+	if (!dynamic_services_get_config_filename(servicesFilename))
+	{
+		/* it has already logged why */
+		return false;
+	}
+
+	jsonValue = json_parse_file_with_comments(servicesFilename);
+	if (jsonValue == NULL)
+	{
+		jsonValue = json_value_init_object();
+	}
+
+	jsonObject = json_value_get_object(jsonValue);
+	jsonArray = json_object_get_array(jsonObject, "services");
+	if (jsonArray == NULL)
+	{
+		json_object_set_value(jsonObject, "services", json_value_init_array());
+		jsonArray = json_object_get_array(jsonObject, "services");
+	}
+
+	dynamic_services_edit_json_array(jsonArray, enabledServices, true);
+	dynamic_services_edit_json_array(jsonArray, disabledServices, false);
+
+	json_serialize_to_file(jsonValue, servicesFilename);
+	json_value_free(jsonValue);
+
+	return true;
+}
--- a/src/bin/pg_autoctl/keeper.c
+++ b/src/bin/pg_autoctl/keeper.c
@@ -2205,45 +2205,6 @@ keeper_reload_configuration(Keeper *keeper, bool firstLoop, bool doInit)
 
 
 /*
- * reread_services reads the possibly updated services section from the
- * configuration file and integrates accepted new values into the current setup.
- */
-bool
-keeper_reread_services(Keeper *keeper)
-{
-	KeeperConfig *config = &(keeper->config);
-	KeeperConfig newConfig = { 0 };
-
-	if (!file_exists(config->pathnames.config))
-	{
-		log_error("Missing configuration file %s", config->pathnames.config);
-		return false;
-	}
-
-	/*
-	 * Set the same configuration and state file as the current config.
-	 */
-	strlcpy(newConfig.pathnames.config, config->pathnames.config, MAXPGPATH);
-	strlcpy(newConfig.pathnames.state, config->pathnames.state, MAXPGPATH);
-
-	if (!keeper_config_read_file(&newConfig,
-								 true, /* missingPgdataIsOk */
-								 true, /* pgIsNotRunningIsOk */
-								 true /* monitorDisabledIsOk */))
-	{
-		log_error("Failed to read new config");
-		return false;
-	}
-
-#if 0
-	/* update the configuration values */
-	strlcpy(config->pgbouncer, newConfig.pgbouncer, MAXPGPATH);
-#endif
-	return true;
-}
-
-
-/*
  * keeper_call_reload_hooks loops over the KeeperReloadHooks
  * reloadFunctionArray and calls each hook in turn.
  */
--- a/src/bin/pg_autoctl/keeper.c
+++ b/src/bin/pg_autoctl/keeper.c
@@ -2205,6 +2205,42 @@ keeper_reload_configuration(Keeper *keeper, bool firstLoop, bool doInit)
 
 
 /*
+ * reread_services reads the possibly updated services section from the
+ * configuration file and integrates accepted new values into the current setup.
+ */
+bool
+keeper_reread_services(Keeper *keeper)
+{
+	KeeperConfig *config = &(keeper->config);
+	KeeperConfig newConfig = { 0 };
+
+	if (!file_exists(config->pathnames.config))
+	{
+		log_error("Missing configuration file %s", config->pathnames.config);
+		return false;
+	}
+
+	/*
+	 * Set the same configuration and state file as the current config.
+	 */
+	strlcpy(newConfig.pathnames.config, config->pathnames.config, MAXPGPATH);
+	strlcpy(newConfig.pathnames.state, config->pathnames.state, MAXPGPATH);
+
+	if (!keeper_config_read_file(&newConfig,
+								 true, /* missingPgdataIsOk */
+								 true, /* pgIsNotRunningIsOk */
+								 true /* monitorDisabledIsOk */))
+	{
+		log_error("Failed to read new config");
+		return false;
+	}
+
+	/* update the configuration values (None at present) */
+	return true;
+}
+
+
+/*
  * keeper_call_reload_hooks loops over the KeeperReloadHooks
  * reloadFunctionArray and calls each hook in turn.
  */
--- a/src/bin/pg_autoctl/keeper.c
+++ b/src/bin/pg_autoctl/keeper.c
@@ -2235,7 +2235,10 @@ keeper_reread_services(Keeper *keeper)
 		return false;
 	}
 
-	/* update the configuration values (None at present) */
+#if 0
+	/* update the configuration values */
+	strlcpy(config->pgbouncer, newConfig.pgbouncer, MAXPGPATH);
+#endif
 	return true;
 }
 
--- a/src/bin/pg_autoctl/service_keeper_init.c
+++ b/src/bin/pg_autoctl/service_keeper_init.c
@@ -36,15 +36,16 @@
  * service_keeper_handle_dynamic enables or disables dynamic services based on
  * its configuration.
  *
- * Currenty no such services are present in the configuration
- *
- * Returns the number of services, positive when added, negative when removed.
+ * Records the number of services, positive when added, negative when removed,
+ * in its last argument, diffCount.
  */
-static int
-service_keeper_handle_dynamic(Supervisor *supervisor, void *arg)
+static void
+service_keeper_handle_dynamic(Supervisor *supervisor, void *arg, int *diffCount)
 {
 	Keeper *keeper = (Keeper *) arg;
 	Service service = { 0 };
+
+	*diffCount = 0;
 
 	/*
 	 * Do not bother to figure out if the value has changed, just read the
@@ -55,7 +56,6 @@ service_keeper_handle_dynamic(Supervisor *supervisor, void *arg)
 	/*
 	 * Nothing else to be done for now since no services have been configured
 	 */
-	return 0;
 }
 
 
--- a/src/bin/pg_autoctl/service_keeper_init.c
+++ b/src/bin/pg_autoctl/service_keeper_init.c
@@ -68,34 +68,34 @@ bool
 service_keeper_init(Keeper *keeper)
 {
 	const char *pidfile = keeper->config.pathnames.pid;
-
-	Service subprocesses[] = {
-		{
-			SERVICE_NAME_POSTGRES,
-			RP_PERMANENT,
-			-1,
-			&service_postgres_ctl_start,
+	ServiceArray services = {
+		.array = {
+			{
+				SERVICE_NAME_POSTGRES,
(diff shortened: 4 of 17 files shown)


src/bin/pg_autoctl/supervisor.c
```diff
@@ -56,22 +72,277 @@ static bool supervisor_restart_service(Supervisor *supervisor,
...
+
+#define foreach_supervised_service(it, supervisor, service) \
+	for ((service) = supervisor_it_first((it), (supervisor)); \
+		 (service) != NULL; \
+		 (service) = supervisor_it_next(it))
+
+
+/*
+ * supervisor_add_dynamic starts the service and if successfull, adds it to the
+ * dynamicService array.
```
DimCitus, line 144: What about `supervisor_dynamic_service_enable` as a function name? with the same prefix, and more clarity that the service is going to be started rather than just added to our array.

src/bin/pg_autoctl/supervisor.c
```diff
@@ -56,22 +72,277 @@ static bool supervisor_restart_service(Supervisor *supervisor,
...
+				  service->name);
+		return false;
+	}
+
+	/* 0-based */
+	if (serviceIndex >= (MAXDYNSERVICES - 1))
+	{
+		log_error("Reached maximum permitted number of dynamic services, "
+				  "service %s is not added",
+				  service->name);
```
DimCitus, line 178: Our error messages begin with "Failed to" which often makes it easier for the user to grasp the context of the error and its impact on the operations. Something like "Failed to start dynamic service \"%s\": pg_autoctl supports up to %d dynamic services and all the slots are used already".

src/bin/pg_autoctl/supervisor.c
```diff
@@ -56,22 +72,277 @@ static bool supervisor_restart_service(Supervisor *supervisor,
...
+		return false;
+	}
+
+	/*
+	 * IGNORE-BANNED fits here because we want to take ownership of the service.
+	 * We copy to an address that we own from a non overlapping address.
+	 */
+	dynamicService = &supervisor->dynamicServices.services[serviceIndex];
+	memcpy(dynamicService, service, sizeof(*service)); /* IGNORE-BANNED */
+	supervisor->dynamicServices.serviceCount++;
```
DimCitus, line 230: Is there a way to just copy the structure instance directly, knowing that there is no pointer in the structure, it's all direct memory usage?

src/bin/pg_autoctl/supervisor.c
```diff
@@ -56,22 +72,277 @@ static bool supervisor_restart_service(Supervisor *supervisor,
...
+	(void) unblock_signals(&sig_mask_orig);
+
+	return true;
+}
+
+
+/*
+ * supervisor_remove_dynamic stops a running dynamic service with serviceName.
+ * Once the service is stopped, then it is removed from the dynamicService array
+ * and supervisor_loop can largely forget about it.
```
DimCitus, line 241: Naming: what about `supervisor_dynamic_service_disable` here?

src/bin/pg_autoctl/supervisor.c
```diff
@@ -56,22 +72,277 @@ static bool supervisor_restart_service(Supervisor *supervisor,
...
+ * array, false otherwise.
+ */
+bool
+supervisor_remove_dynamic(Supervisor *supervisor, const char *serviceName)
+{
+	Service *dynamicService = NULL;
+	int serviceCount = supervisor->dynamicServices.serviceCount;
+	int serviceIndex;
+
+	/* Find it in the dynamic array */
```
DimCitus, line 257: Could we use the `supervisor_service_exists` function if it had a better API for this call-site? And then I suppose it would be renamed `supervisor_dynamic_service_find` or something.

src/bin/pg_autoctl/supervisor.c
```diff
@@ -56,22 +72,277 @@ static bool supervisor_restart_service(Supervisor *supervisor,
...
+	supervisor->dynamicServices.services[serviceIndex] =
+		supervisor->dynamicServices.services[serviceCount - 1];
+	supervisor->dynamicServices.serviceCount--;
+
+	/*
+	 * Finally add the dynamic service to the recently removed array.
+	 *
+	 * If we reached the end of this array, start again at the beginning. This
+	 * is a loosely held array and we do not fuss too much.
+	 */
```
DimCitus, line 303: Oh, then I finally get what this array is useful for. What about renaming it to `pendingShutdown` or something that expresses to the reader that we keep track of services we signaled for termination so that we can match them later when `waitpid()` confirms their termination?

src/bin/pg_autoctl/supervisor.c
```diff
@@ -56,22 +72,277 @@ static bool supervisor_restart_service(Supervisor *supervisor,
...
+	}
+
+	/*
+	 * IGNORE-BANNED fits here because we want to take ownership of the service.
+	 * We copy to an address that we own from a non overlapping address.
+	 */
+	memcpy(&supervisor->dynamicRecentlyRemoved.services[    /* IGNORE-BANNED */
+			   supervisor->dynamicRecentlyRemoved.serviceCount],
+		   dynamicService,
+		   sizeof(*dynamicService));
```
DimCitus, line 316: I there a better way to just copy the immediate structure instance there? We just use the C assignment in many places in the code for that, when we know that our structures are constructed without pointers to other places in memory.

src/bin/pg_autoctl/supervisor.c
```diff
@@ -193,6 +473,37 @@ supervisor_loop(Supervisor *supervisor)
 			(void) supervisor_reload_services(supervisor);
 		}
 
+		if (asked_to_handle_dynamic)
+		{
+			int changed = 0;
+			log_info("pg_autoctl received a signal to, "
+					 "handle dynamic services");
+
+			changed = supervisor_handle_dynamic(supervisor);
```
DimCitus, line 482: As mentioned already, let's improve this API here.

src/bin/pg_autoctl/supervisor.c
```diff
@@ -153,6 +424,14 @@ supervisor_start(Service services[], int serviceCount, const char *pidfile)
...
 	}
 
+	/*
+	 * Let the dynamic handler decide about dynamic services
+	 */
+	if (dynamicHandler)
+	{
+		dynamicHandler(&supervisor, dynamicHandlerArg);
+	}
+
```
DimCitus, line 434: I still don't understand why we need an indirection here. It needs to be either better commented, or just gotten rid of, if that's possible to implement directly in the main loop.

src/bin/pg_autoctl/supervisor.c
```diff
@@ -213,7 +524,7 @@ supervisor_loop(Supervisor *supervisor)
 				if (errno == ECHILD)
 				{
 					/* no more childrens */
-					if (asked_to_stop || asked_to_stop_fast || asked_to_quit)
+					if (supervisor->shutdownSequenceInProgress)
```
DimCitus, line 527: Why is your PR needing to change this part?

src/bin/pg_autoctl/supervisor.c
```diff
@@ -256,6 +567,15 @@ supervisor_loop(Supervisor *supervisor)
 			{
 				Service *dead = NULL;
 
+				/* if this is a dynamic service which just exited, do nothing */
+				if (supervisor_find_recently_removed_dynamic(supervisor,
```
DimCitus, line 571: As mentioned already, I think we need a better name for the structure slot and the function here.

src/bin/pg_autoctl/supervisor.c
```diff
@@ -282,21 +604,52 @@ supervisor_loop(Supervisor *supervisor)
...
+	for (serviceIndex = 0; serviceIndex < serviceCount; serviceIndex++)
+	{
+		Service *service = &recentlyRemoved[serviceIndex];
+		if (service->pid == pid)
+		{
+			*result = service;
+
+			/* Now that the service is found, remove it from the array */
+			recentlyRemoved[serviceIndex] = recentlyRemoved[serviceCount - 1];
+			supervisor->dynamicRecentlyRemoved.serviceCount--;
```
DimCitus, line 629: It's surprising to have a function name with `find` that actually removes an entry from the array. Let's find something else.

src/bin/pg_autoctl/supervisor.c
```diff
@@ -282,21 +604,52 @@ supervisor_loop(Supervisor *supervisor)
...
 static bool
 supervisor_find_service(Supervisor *supervisor, pid_t pid, Service **result)
 {
-	int serviceCount = supervisor->serviceCount;
-	int serviceIndex = 0;
+	Supervisor_it sit;
+	Service *service;
 
-	for (serviceIndex = 0; serviceIndex < serviceCount; serviceIndex++)
+	foreach_supervised_service(&sit, supervisor, service)
```
DimCitus, line 648: As said before I preferred the previous coding. This macro does not look like an improvement to me, and it blurs the memory aspect of things.

src/bin/pg_autoctl/supervisor.c
```diff
@@ -612,9 +959,140 @@ supervisor_shutdown_sequence(Supervisor *supervisor)
...
+	 */
+	if (supervisor->shutdownSequenceInProgress)
+	{
+		log_trace("supervisor_handle_failed_restart: shutdownSequenceInProgress");
+		return;
+	}
+
+	/*
+	 * Loop through the dynamic service array and if it found, then mark it
+	 */
```
DimCitus, line 1022: I am pretty sure we should have (or already have) an internal API to take care of that operation.

src/bin/pg_autoctl/supervisor.c
```diff
@@ -656,7 +1134,7 @@ supervisor_restart_service(Supervisor *supervisor, Service *service, int status)
 	 */
 	if (service->policy == RP_TEMPORARY)
 	{
-		return true;
+		return false;
```
DimCitus, line 1137: Should that be part of this PR? It looks like a bug fix that is worthwhile anyway?

src/bin/pg_autoctl/supervisor.h
```diff
@@ -108,19 +119,49 @@ typedef struct Supervisor
...
-bool supervisor_start(Service services[], int serviceCount, const char *pidfile);
+	/*
+	 * Anonymous struct holding active dynamic services.
+	 * Do not access directly.
+	 */
+	struct
+	{
+		Service services[MAXDYNSERVICES];
+		int serviceCount;
+	} dynamicServices;
```
DimCitus, line 142: I'm not fond of anonymous data structures. Is there a drawback to exposing that as a first-level structure that is contained in the Supervisor structure?

src/bin/pg_autoctl/supervisor.h
```diff
@@ -108,19 +119,49 @@ typedef struct Supervisor
...
+
+	/*
+	 * Anonymous struct holding recently removed dynamic services.
+	 * Do not access directly.
+	 */
+	struct
+	{
+		Service services[MAXDYNSERVICES];
+		int serviceCount;
+	} dynamicRecentlyRemoved;
```
DimCitus, line 152: Same comment as before, let's expose the currently internal and anonymous structure. And rename it, too.

DimCitus (changes requested):
Here's a first level of review. I think we need to focus on many smaller details (naming, etc), and maybe a couple of higher level concepts:

  - do we need this dynamic handler function pointer? I can't understand how that's needed and what it does exactly...
  -  do we need to implement a restart failure policy for services, so that dynamic and static services both get a chance at specifying if pg_autoctl should get down with them? that would allow a pgbouncer specific discussion later, without technical constraint, just change an enum value and be done.

src/bin/pg_autoctl/cli_common.c
```diff
@@ -1624,6 +1624,37 @@ cli_pg_autoctl_reload(const char *pidfile)
 }
 
 
+/*
+ * cli_pg_autoctl_handle_dynamic signals the pg_autoctl to handle any dynamic
+ * services based on its configuration
+ */
+bool
+cli_pg_autoctl_handle_dynamic(const char *pidfile)
```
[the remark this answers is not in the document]
gkokolatos: Correct. You do not see it here, as this is a cut from the larger PR. You can see the permalink here
https://github.com/citusdata/pg_auto_failover/blob/bb5298b8617305cde5c5551e2327e0650d6f4c00/src/bin/pg_autoctl/cli_enable_disable.c#L1498

src/bin/pg_autoctl/keeper.c
```diff
@@ -2204,6 +2204,42 @@ keeper_reload_configuration(Keeper *keeper, bool firstLoop, bool doInit)
...
+								 true, /* missingPgdataIsOk */
+								 true, /* pgIsNotRunningIsOk */
+								 true /* monitorDisabledIsOk */))
+	{
+		log_error("Failed to read new config");
+		return false;
+	}
+
+	/* update the configuration values (None at present) */
+	return true;
```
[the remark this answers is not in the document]
gkokolatos: Correct, at the moment there is no result yet. It is just the scaffolding. You can see how it will behave with the addition of dynamic services here:
https://github.com/citusdata/pg_auto_failover/blob/bb5298b8617305cde5c5551e2327e0650d6f4c00/src/bin/pg_autoctl/keeper.c#L2217

src/bin/pg_autoctl/service_keeper_init.c
```diff
@@ -32,6 +32,32 @@
...
+/*
+ * service_keeper_handle_dynamic enables or disables dynamic services based on
+ * its configuration.
+ *
+ * Currenty no such services are present in the configuration
+ *
+ * Returns the number of services, positive when added, negative when removed.
+ */
+static int
+service_keeper_handle_dynamic(Supervisor *supervisor, void *arg)
```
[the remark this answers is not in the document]
gkokolatos: Not really, because there can be more than one services added. The intension as per the discussion in the PR which this is cut from, is the services to be defined in the config file. As such the config file might have had 2 new dynamic services enabled and 3 old ones disabled. What is of interest to the supervisor is the net change.

src/bin/pg_autoctl/supervisor.c
```diff
@@ -34,12 +34,28 @@
...
+	int index;
+} Supervisor_it;
+
 static bool supervisor_init(Supervisor *supervisor);
 static bool supervisor_loop(Supervisor *supervisor);
 
 static bool supervisor_find_service(Supervisor *supervisor, pid_t pid,
 									Service **result);
 
+static bool supervisor_find_recently_removed_dynamic(Supervisor *supervisor,
```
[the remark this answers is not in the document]
gkokolatos: Sure, will do

src/bin/pg_autoctl/supervisor.c
```diff
@@ -34,12 +34,28 @@
 #include "signals.h"
 #include "string_utils.h"
 
+/*
+ * A static struct to loop seamlessly through the active services in a
+ * supervisor struct. Those can be either 'static' or 'dynamic'.
+ */
```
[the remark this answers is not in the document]
gkokolatos: Terribly sorry but I am failing to see the point here. This is just a struct.
Are you objecting to having an iterator or how the iterator is implemented?

src/bin/pg_autoctl/supervisor.c
gkokolatos: Great! Will amend!

src/bin/pg_autoctl/supervisor.c
gkokolatos: Please if you have a specific function or call in mind, I will happily add it. I guess one can bend strcpy and friends accordingly but in essence it is quite the same. Those bytes from that address should be copied to this address.

Anything you want is fine, just point me to it please :)

src/bin/pg_autoctl/supervisor.c
gkokolatos: Will do.

src/bin/pg_autoctl/supervisor.c
gkokolatos: Those two things are solving a very different problem, or at least in my mind they do :)

supervisor_service_exists() is intended to find any active service under supervision. That is what is exposed to our callers. I do not want the callers of the supervisor to have more access to it.

Privately in supervisor.c we can add as many helper functions as desired.
What do you think?

src/bin/pg_autoctl/supervisor.c
gkokolatos: Rights, it is not very obvious at the beginning. I will rename

src/bin/pg_autoctl/service_keeper_init.c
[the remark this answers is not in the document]
DimCitus: Okay then I would still insist on returning a bool. Then either have `int *diffCount` in the arguments or maybe expand fully with two arrays `addedServices[], removedServices[]` allocated by the caller, and that we fill-in with copies of the services discovered there.

src/bin/pg_autoctl/supervisor.c
gkokolatos: Same as above, please point me to the code you suggest and I will happily change

src/bin/pg_autoctl/supervisor.c
gkokolatos: Same as above. Since this is cut from the bigger PR, the actual usage is not visible.
See here https://github.com/citusdata/pg_auto_failover/pull/539 the complete code with the usage.

src/bin/pg_autoctl/supervisor.c
gkokolatos: Absolutely. Will use your suggestions.

src/bin/pg_autoctl/supervisor.c
gkokolatos: Agreed

src/bin/pg_autoctl/supervisor.c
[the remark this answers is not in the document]
DimCitus: What I mean is that you then have introduced the following code when using the macro. Where we used to return a pointer to an existing data structure, we now return a copy of it, done in two steps, one with a static allocation on the stack in the iterator internal function and second step with a call to `memcpy`. Why is that a preferred way of operating here? Who is going to be responsible for freeing that new allocated memory?

``` diff
-			*result = &(supervisor->services[serviceIndex]);
+ 			*result = service;
```

My feeling is that we should continue as before, where this memory handling is obvious, because that's where the bugs in C are coming from first. Having to write another same loop again is rarely the source of bugs in production.

I yeah I am objecting to having an iterator abstraction altogether here.

src/bin/pg_autoctl/supervisor.c
gkokolatos: It's being a while since I coded that bit. It may well have been a bug fix that snuck in. Let me revisit the code and come back :)

src/bin/pg_autoctl/supervisor.c
DimCitus: When working with struct instances (and more so when the structure internal slots are all “direct”, we are not dealing with pointers to other parts of the memory here, it is self-contained), then just use the `=` assignment operator.

src/bin/pg_autoctl/supervisor.h
gkokolatos: I will definitely not hold a strong opinion for this.

The intention is to make as obvious as possible the 'not yours to touch' aspect, just before going to the whole `void *` private method which seems to not be fitting the rest of the code.

Will follow any paradigm you prefer as maintainer of this code :)

src/bin/pg_autoctl/supervisor.c
DimCitus: Oh I meant just about the loop over the services within the `supervisor_remove_dynamic` function itself. Not the whole thing, just that part that finds the service in the array.

src/bin/pg_autoctl/supervisor.c
DimCitus: Have a look at https://github.com/citusdata/pg_auto_failover/blob/master/src/bin/pg_autoctl/keeper_config.c#L304 and many other such places in the code for examples of how we like to copy data from a structure instance to another in this project ;-)

src/bin/pg_autoctl/supervisor.c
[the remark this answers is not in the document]
gkokolatos: I deeply apologise. I really do not mean to be contradictory. I am simply a bit at a loss here.

Where did you see this memcpy? Is it on this diff? The iterator does not do any memcpy() nor the macro.

These are the iterator functions:

```
+static inline Service *
+supervisor_it_get(Supervisor_it *sit)
+{
+       if (sit->index < sit->count)
+       {
+               return &sit->services[sit->index];
+       }
+
+       return NULL;
+}
+
+
+static inline Service *
+supervisor_it_first(Supervisor_it *sit, Supervisor *supervisor)
+{
+       *sit = (Supervisor_it) {
+               .supervisor = supervisor,
+               .services = supervisor->services,
+               .count = supervisor->serviceCount,
+               .index = 0,
+       };
+
+       return supervisor_it_get(sit);
+}
+
+
+static inline Service *
+supervisor_it_next(Supervisor_it *sit)
+{
+       sit->index++;
+
+       /* if we are done with the static services, move to the dynamic */
+       if (sit->index >= sit->count &&
+               sit->services == sit->supervisor->services)
+       {
+               sit->services = sit->supervisor->dynamicServices.services;
+               sit->count = sit->supervisor->dynamicServices.serviceCount;
+               sit->index = 0;
+       }
+
+       return supervisor_it_get(sit);
+}
``` 

And this is the macro:
```
+#define foreach_supervised_service(it, supervisor, service) \
+       for ((service) = supervisor_it_first((it), (supervisor)); \
+                (service) != NULL; \
+                (service) = supervisor_it_next(it))
```

I will happily remove the macro, no problem. I simply fail to see the memcpy here.

Again terribly sorry for this and thank you very much for your time.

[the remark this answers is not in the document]
DimCitus: Ah yes, indeed, the `memcpy` are used in parts of the code that have nothing to do with the iterators. I still find the iterator very confusing. It's a lot of code to replace 3 loops that are very easy to read, so I keep failing to see the point. It's one more thing to keep in mind and juggle with, it doesn't make anything simpler, just 3 call sites only a little easier to write. Let's remove the iterator idea, even though that's not having the problems I though it had, because I think it's just too much for the list of changes in this PR.

src/bin/pg_autoctl/supervisor.h
DimCitus: My favourite paradigm is to user-focus design and ease of maintenance. I believe that our code should be as boring as possible. My experience is that anything interesting you write in C is going to come and bite you later...

src/bin/pg_autoctl/supervisor.c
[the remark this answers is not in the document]
gkokolatos: Now I understand. Thank you for your patience and time.

I will prepare something to show you how it would like without
a) the macro
b) the macro and the iterator
and let you choose what you prefer.

gkokolatos:
@DimCitus Thank you for your comments and review so far.

As you can see, there have been a few commits that should address some of the minor and easy to fix comments. If you do not mind, let us get those straight before moving on. If you do mind, please let me know and will do a second pass.

src/bin/pg_autoctl/supervisor.c
```diff
@@ -675,12 +1162,6 @@ supervisor_restart_service(Supervisor *supervisor, Service *service, int status)
 	}
 	else
 	{
-		/* exit with a non-zero exit code, and process with shutdown sequence */
-		supervisor->cleanExit = false;
-		supervisor->shutdownSequenceInProgress = true;
-
-		(void) supervisor_stop_other_services(supervisor, service->pid);
-
```
DimCitus, line 683: This change looks highly suspicious in this PR.

src/bin/pg_autoctl/supervisor.c
```diff
@@ -699,12 +1180,6 @@ supervisor_restart_service(Supervisor *supervisor, Service *service, int status)
 	 */
 	if (service->policy == RP_TRANSIENT && returnCode == EXIT_CODE_QUIT)
 	{
-		/* exit with a happy exit code, and process with shutdown sequence */
-		supervisor->cleanExit = true;
-		supervisor->shutdownSequenceInProgress = true;
-
-		(void) supervisor_stop_other_services(supervisor, service->pid);
-
```
DimCitus, line 707: This change looks highly suspicious in this PR.

src/bin/pg_autoctl/supervisor.c
```diff
@@ -721,19 +1196,16 @@ supervisor_restart_service(Supervisor *supervisor, Service *service, int status)
 	{
 		log_fatal("Failed to restart service %s", service->name);
 
-		/* exit with a non-zero exit code, and process with shutdown sequence */
-		supervisor->cleanExit = false;
-		supervisor->shutdownSequenceInProgress = true;
-
-		(void) supervisor_stop_other_services(supervisor, service->pid);
-
```
DimCitus, line 729: This change looks highly suspicious in this PR.

DimCitus (changes requested):
I left a second round of detailed review. Thanks for your efforts, I think we're getting there!

src/bin/pg_autoctl/supervisor.h
```diff
@@ -108,19 +119,43 @@ typedef struct Supervisor
...
+	/*
+	 * dynamicServices is a nested struct used internally by the supervisor in
+	 * order to hold information about the currently running dynamic services
+	 * under it's control and the ones that are pending termination.
+	 */
+	struct dynamicServices
+	{
+		Service services[MAXDYNSERVICES];
+		int serviceCount;
+	} dynamicServicesEnabled, dynamicServicesDisabled;
```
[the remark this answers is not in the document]
DimCitus: See https://github.com/citusdata/pg_auto_failover/blob/master/src/bin/pg_autoctl/pgsql.h#L148 for an example of that idea (Array struct with count/array), or https://github.com/citusdata/pg_auto_failover/blob/master/src/bin/pg_autoctl/nodestate_utils.h#L63.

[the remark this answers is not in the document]
gkokolatos: What ever you see fit. typedef it is :)

src/bin/pg_autoctl/supervisor.c
```diff
@@ -56,22 +74,278 @@ static bool supervisor_restart_service(Supervisor *supervisor,
...
+ * Also since the service runs under supervision, there must be space for it to
+ * be added to the supervisor's dynamic services array. If there is not, then
+ * the service is not started and the function returns false.
+ * The service will not be added if a signal other than "dynamic service" has
+ * been received.
+ *
+ * Returns true if successfully starts the service.
+ */
+bool
+supervisor_dynamic_service_enable(Supervisor *supervisor, Service *service)
```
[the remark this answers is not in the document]
gkokolatos: no probs

src/bin/pg_autoctl/supervisor.c
```diff
@@ -56,22 +74,278 @@ static bool supervisor_restart_service(Supervisor *supervisor,
...
+ *
+ * However, it will get notified during waitpid, so we add this service to the
+ * recentlyRemoved array and we will hold it there until it is requested via
+ * supervisor_dynamic_remove_terminated_service() which subsequently removes it.
+ *
+ * Returns true if the service is stopped and removed from the dynamicService
+ * array, false otherwise.
+ */
+bool
+supervisor_dynamic_service_disable(Supervisor *supervisor,
```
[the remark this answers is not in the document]
gkokolatos: no probs

src/bin/pg_autoctl/supervisor.c
```diff
@@ -56,22 +74,278 @@ static bool supervisor_restart_service(Supervisor *supervisor,
...
+		return false;
+	}
+
+	/*
+	 * Since we are going to signal the service, we have to make certain that we
+	 * will not error out in the supervisor_loop on the waitpid() response. So
+	 * after killing the service and removing it from the dynamicService array,
+	 * we add it to the recentlyRemoved array to find it again later.
+	 */
+	if (!kill(dynamicService->pid, SIGTERM))
```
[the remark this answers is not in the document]
gkokolatos: :) no probs

src/bin/pg_autoctl/supervisor.c
```diff
@@ -56,22 +74,278 @@ static bool supervisor_restart_service(Supervisor *supervisor,
...
+
 /*
  * supervisor_start starts given services as sub-processes and then supervise
  * them.
  */
 bool
-supervisor_start(Service services[], int serviceCount, const char *pidfile)
+supervisor_start(Service services[], int serviceCount, const char *pidfile,
+				 void (*dynamicHandler)(Supervisor *, void *, int *),
+				 void *dynamicHandlerArg)
```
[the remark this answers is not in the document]
gkokolatos: I would disagree with you. I am sorry about it.

I tried the hardcoded way initially and is by far much more invasive and forces one to jump through hoops in many parts of the code (all the way down to config) in order to avoid simply adding a function pointer. I do not see it.

Having said that, this is your project and you are the maintainer. I am simply the code monkey here. My goal is to get this in, so I will do what you say, but will kindly ask for someone to actually try it out and have the discussion there. If possible of course.

src/bin/pg_autoctl/supervisor.c
```diff
@@ -256,6 +575,15 @@ supervisor_loop(Supervisor *supervisor)
 			{
 				Service *dead = NULL;
 
+				/* if this is a dynamic service which just exited, do nothing */
+				if (supervisor_dynamic_remove_terminated_service(supervisor,
```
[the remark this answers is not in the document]
gkokolatos: Sure, but it should not get confused with the exposed function to the users of the supervisor, no?

src/bin/pg_autoctl/supervisor.c
```diff
@@ -612,9 +969,139 @@ supervisor_shutdown_sequence(Supervisor *supervisor)
...
+
+
+/*
+ * supervisor_handle_failed_restart decides what action to take when a service
+ * has failed to restart. Currently one of the three can happen:
+ *	* continue running the rest of the services
+ *	* exit the whole program with a happy code
+ *	* exit the whole program with failed code
+ *
+ * The action is depended on whether the service is held in the dynamic array.
```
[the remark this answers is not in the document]
gkokolatos: The restart policy should be the same regardless. For some of the use cases I know off, the service will be required to be PERSISTANT, otherwise the feature is not useful.

The disconnect is not about the restart policy, it is about the everything failing if one service fails.

Users not being able to connect to one pgbouncer does not mean that there are not others to connect to. Rights?

src/bin/pg_autoctl/supervisor.c
gkokolatos: It codes the behaviour. The function is responsible for restarting based on the policy. What happens after is not part of this function.

src/bin/pg_autoctl/keeper.c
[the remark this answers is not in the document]
gkokolatos: I can add the actual code ifdef'ed if you don't mind. That will make it much much easier

src/bin/pg_autoctl/supervisor.c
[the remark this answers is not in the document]
gkokolatos: Sure, but let's leave this last. Once everything else is agreed upon I can give you the two options to choose from.

I am really not married to the idea. No worries

src/bin/pg_autoctl/supervisor.c
```diff
@@ -56,22 +74,278 @@ static bool supervisor_restart_service(Supervisor *supervisor,
...
+	Service *dynamicService;
+	sigset_t sig_mask;
+	sigset_t sig_mask_orig;
+	int serviceIndex = supervisor->dynamicServicesEnabled.serviceCount;
+	bool started;
+
+	if (supervisor_service_exists(supervisor, service->name))
+	{
+		log_error("Service %s already exists under supervision",
+				  service->name);
```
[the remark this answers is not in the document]
gkokolatos: Sure.

What would you like the behaviour to be if it is not know to the OS?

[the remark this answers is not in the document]
DimCitus: In other parts of the code, when we see a stale PID file, we remove the file and then start the service again. Here it's a little more confusing because it's a stale entry in-memory in our array of services. We have to think about race conditions with the disable approach and all in that case I suppose.

src/bin/pg_autoctl/supervisor.c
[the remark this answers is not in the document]
DimCitus: I'm surprised. In my mind I was seeing the same code block just in a different place really. The thing you tried, was it with an extra dynamic service array as input to the `supervisor_start` function, and the idea that the function (or the main loop later) would loop through the array and call `supervisor_enable_dynamic_service` with the service definition as input?

src/bin/pg_autoctl/supervisor.c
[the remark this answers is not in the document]
DimCitus: Our inspiration for the supervisor implementation is the [Erlang OTP supervisor module](https://erlang.org/doc/man/supervisor.html) and I think what you're trying to implement is the supervisor strategy (one_for_one, one_for_all, rest_for_one, etc). Though I am not sure this is exactly what you have in mind.

I would like us to come up with a strategy that we can easily change later, either because the next dynamic service after pgbouncer is going to require another way to do it, or because in the discussion we realise we need to change our mind on how pgbouncer failures to run are to be handled.

src/bin/pg_autoctl/supervisor.c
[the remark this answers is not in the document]
gkokolatos: Correct. You might want to think about the cases where you want to enable a service and add it an already running process group several days after the process group has started running. Same goes for disabling services.

You can do all that with at least one array more and some more code in other parts of the code. Or with keeping that information in the Service itself. An idea initially presented to you and shot down if you remember.

What am I missing?

src/bin/pg_autoctl/supervisor.c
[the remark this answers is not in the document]
gkokolatos: Yeah, I have thought of those race conditions and cases. A stale PID in this case should indicate a coding error imho.

Otherwise the loop should have removed a failed process. Would you agree with that?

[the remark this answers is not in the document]
DimCitus: Okay now I have to do my homework and follow it through on my side. Getting back to you later ;-)

src/bin/pg_autoctl/supervisor.c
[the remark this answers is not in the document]
DimCitus: So I think this discussion is hard to have on text only medium, and I am open for a live chat if you're available. I wanted to help drive this PR into being a quick first commit from you, but apparently my optimism gets in the way of a proper review with enough time spent thinking... sorry about that. 

The design I had in mind, and you're right, that does not work, is that the main supervisor loop uses the two arrays of dynamic services (enabled, disabled) and always process them in the same way. When a signal is received (SIGHUP or SIGUSR1) that the setup might have changed, the main loop would review those arrays and start or stop dynamic services.

This doesn't work because at the moment in the PR the service definition is lazily shared with the supervisor entity. That said, it is a static definition: the only part that is dynamic is the enabled/disabled information, and that's user-driven.

We also have the problem that the SIGUSR1 signal is currently hard-coded to have the supervisor process itself read the keeper configuration file again and update its own keeper instance. I think that's why you have a modularity violation, that you are solving with a function pointer and opaque `void *` pointer for a state. This also brings the question, what if we enable pgbouncer on the monitor? Then the supervisor also needs to maintain a Monitor struct instance in-memory. Etc.

The central question here is: how the keeper node-active process communicates with the supervisor that a known dynamic service has been enabled/disabled? 

That's the whole reason why we have this [service_postgres_ctl.c](https://github.com/citusdata/pg_auto_failover/blob/master/src/bin/pg_autoctl/service_postgres_ctl.c) intermediate process, and the communication is done through a specific state file implemented at https://github.com/citusdata/pg_auto_failover/blob/master/src/bin/pg_autoctl/state.h#L158 with the `ExpectedPostgresStatus` struct. The client process (node-active or listener) can then edit the on-disk state file with a new enum value, and the Postgres controller process that loops over reading that file every 100ms then implements the change.

I am now thinking (out loud) that we need the same thing for each dynamic service we want to run: a controller process (always running) that monitors the process expected state file (enabled/disabled in the pgbouncer case, and we could add the pause/suspend/resume bits in there) and implements the change.

With that, the keeper config file is owned by a single process and code path, the keeper node-active service. If we want to implement pgbouncer in the monitor, we can handle the monitor setup in the monitor code-path and process and have the monitor signal to the pgbouncer controller process that the expected state of the pgbouncer process has changed, but just writing new content to a (possibly new) state file.

It means that the supervisor doesn't have to handle the enabled/disabled state if we go with this design. It does not need to handle dynamic services at all, actually. It's all static controller processes that are always running and might be asked to start a service at some point.

I'm sorry that I am only reaching that conclusion now, and that we needed so much work to get there. What are your thoughts?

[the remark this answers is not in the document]
gkokolatos: Very eloquent description. Kudos for that.

I would disagree on a couple of points but I agree on most. Let us take the discussion elsewhere and return with our conclusions here.

And yes, it takes a while to reach a conclusion. No need for apologies :)

gkokolatos:
@DimCitus thank you for your reviewing efforts.

Please find some updates based on comments and off list discussions.
If there are any existing comments that can be resolved, I would much appreciate if they got marked as such, in order to focus my efforts on the remaining, or new, issues.

src/bin/pg_autoctl/supervisor.c
```diff
@@ -845,6 +1378,24 @@ supervisor_update_pidfile(Supervisor *supervisor)
...
 
+static bool
+supervisor_service_exists(Supervisor *supervisor, const char *serviceName)
+{
+	Supervisor_it sit;
+	Service *service;
+
+	foreach_supervised_service(&sit, supervisor, service)
+	{
+		if (!strncmp(service->name, serviceName, strlen(serviceName)))
```
DimCitus, line 1389: Please use `strncmp(...) == 0` or our macro `streq` if it's available in that module. The man page sayth that the return value is zero when the string is found. It sayth nothing about considering an integer as a boolean, and I don't trust C too much with that.

DimCitus (changes requested):
Lots of many things, I think we're close now.

On a larger plan, is it possible to limit the changes in `supervisor_restart_service` and `supervisor_handle_failed_restart`? I am not sure I understand how we need that already?

src/bin/pg_autoctl/dynamic_services_config.c
```diff
@@ -0,0 +1,398 @@
...
+ * does not have any knowledge of paths. Any other users should have set the
+ * environment accordingly before calling dynamic_services_* functions.
+ *
+ * It is the caller's responsibility that servicesFilename is large enough.
+ *
+ * Returns true on success. Otherwise the contents of servicesFilename are not
+ * to be trusted.
+ */
+static bool
+dynamic_services_get_config_filename(char *servicesFilename)
```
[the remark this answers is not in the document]
gkokolatos: This is a static. There is absolutely no reason for anyone else to know about it besides the users of this file. I can move it if you really want though.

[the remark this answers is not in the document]
DimCitus: All the other file paths that are needed belong to a `ConfigFilePaths` structure. They are all computed in a similar way, based on XDG components, XDG environments variables, and PGDATA. My reaction here is to say that doing the exact same processing as found in `config.c` in another module, and bypassing the use of our `ConfigFilePaths` structure, is surprising when reading the code.

Can we make it less surprising, and more in line with our current modules organisation?

PR #574, opened 2021-01-19
Teach git to ignore vim working files

gkokolatos:
For us vim/vi users.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,6 +20,7 @@ objfiles.txt
 *.gcov.out
 lcov.info
 coverage/
+*.swp
 *.vcproj
 *.vcxproj
 win32ver.rc


DimCitus:
Hi @gkokolatos ; as per git best practices personal tooling choices setup go in `$GIT_DIR/info/exclude`, see https://git-scm.com/docs/gitignore. We have had other PRs that would target the project level `gitignored` and I also rejected them on the same grounds. We have contributors using all sorts of editors and IDE (vi, Emacs, CLion, vscode, etc): we don't have a pre-requisite about this part of the tooling, and so it's up to each contributor to handle that part of their setup.

gkokolatos:
My bad! Saw the .vcxproj and friends in the file and thought you were handling there. 

Thanks for the review!

DimCitus:
I think this project is used for internal code auditing and security purposes, though I might be wrong. Will remove if that turns out to be just a specific editor setup, thanks for noticing!

JelteF:
I the same `*.swp` line in my global gitignore: https://stackoverflow.com/questions/7335420/global-git-ignore

PR #576, opened 2021-02-02
Feature/zero config pg autoctl

### LINKED ISSUE #437: How create formation before keeper node creation?
How create formation before keeper node creation?
In node host, not on monitor host!
May You add `--monitor` option to `pg_autoctl create formation` command?

DimCitus:
Implement support for `--monitor` option and PG_AUTOCTL_MONITOR environment variable in many commands that don't need to have a local node already created. Those commands are:

    pg_autoctl create formation
    pg_autoctl drop formation

    pg_autoctl get|set formation settings
    pg_autoctl get|set formation number-sync-standbys
    pg_autoctl get|set node replication-quorum
    pg_autoctl get|set node candidate-priority

    pg_autoctl show uri
    pg_autoctl show events
    pg_autoctl show state
    pg_autoctl show settings
    pg_autoctl show standby-names

    pg_autoctl perform failover
    pg_autoctl perform switchover
    pg_autoctl perform promotion

The `make cluster` is also changed to use the PG_AUTOCTL_MONITOR environment variable rather than the monitor's PGDATA in the interactive terminal/session.

Fixes #437.

### DIFF
--- a/src/bin/pg_autoctl/cli_do_tmux.c
+++ b/src/bin/pg_autoctl/cli_do_tmux.c
@@ -467,9 +467,11 @@ tmux_add_xdg_environment(PQExpBuffer script)
  * tmux_setenv adds setenv commands to the tmux script.
  */
 void
-tmux_setenv(PQExpBuffer script, const char *sessionName, const char *root)
+tmux_setenv(PQExpBuffer script,
+			const char *sessionName, const char *root, int firstPort)
 {
 	char PATH[BUFSIZE] = { 0 };
+	char monitor_pguri[MAXCONNINFO] = { 0 };
 
 	if (!get_env_copy("PATH", PATH, sizeof(PATH)))
 	{
@@ -477,7 +479,17 @@ tmux_setenv(PQExpBuffer script, const char *sessionName, const char *root)
 		exit(EXIT_CODE_INTERNAL_ERROR);
 	}
 
-	tmux_add_command(script, "set-environment -t %s PATH \"%s\"", sessionName, PATH);
+	tmux_add_command(script,
+					 "set-environment -t %s PATH \"%s\"", sessionName, PATH);
+
+	sformat(monitor_pguri, sizeof(monitor_pguri),
+			"postgres://autoctl_node@localhost:%d/pg_auto_failover?sslmode=prefer",
+			firstPort);
+
+	tmux_add_command(script,
+					 "set-environment -t %s PG_AUTOCTL_MONITOR \"%s\"",
+					 sessionName,
+					 monitor_pguri);
 
 	for (int i = 0; xdg[i][0] != NULL; i++)
 	{
@@ -635,7 +647,7 @@ tmux_pg_autoctl_create_postgres(PQExpBuffer script,
 	tmux_add_send_keys_command(script, "export PGPORT=%d", pgport);
 
 	sformat(monitor, sizeof(monitor),
-			"$(%s show uri --pgdata %s/monitor --monitor)",
+			"$(%s show uri --pgdata %s/monitor --formation monitor)",
 			pg_autoctl_argv0,
 			root);
 
@@ -680,7 +692,7 @@ prepare_tmux_script(TmuxOptions *options, PQExpBuffer script)
 	tmux_add_command(script, "set-option -g default-shell /bin/bash");
 
 	(void) tmux_add_new_session(script, root, pgport);
-	(void) tmux_setenv(script, sessionName, root);
+	(void) tmux_setenv(script, sessionName, root, options->firstPort);
 
 	/* start a monitor */
 	(void) tmux_add_xdg_environment(script);
@@ -715,7 +727,6 @@ prepare_tmux_script(TmuxOptions *options, PQExpBuffer script)
 										node->name,
 										node->replicationQuorum,
 										node->candidatePriority);
-		tmux_add_send_keys_command(script, "pg_autoctl run");
 
 		strlcpy(previousName, node->name, sizeof(previousName));
 	}
@@ -740,7 +751,6 @@ prepare_tmux_script(TmuxOptions *options, PQExpBuffer script)
 	tmux_add_command(script, "split-window -v");
 	tmux_add_command(script, "select-layout even-vertical");
 	(void) tmux_add_xdg_environment(script);
-	tmux_add_send_keys_command(script, "export PGDATA=\"%s/monitor\"", root);
 
 	if (options->numSync != -1)
 	{
--- a/src/bin/pg_autoctl/cli_formation.c
+++ b/src/bin/pg_autoctl/cli_formation.c
@@ -361,15 +361,15 @@ keeper_cli_formation_create_getopts(int argc, char **argv)
 	/* when we have a monitor URI we don't need PGDATA */
 	if (IS_EMPTY_STRING_BUFFER(options.monitor_pguri))
 	{
-		cli_common_get_set_pgdata_or_exit(&(options.pgSetup));
-	}
-	else
-	{
 		if (!IS_EMPTY_STRING_BUFFER(options.pgSetup.pgdata))
 		{
 			log_warn("Given --monitor URI, the --pgdata option is ignored");
 			log_info("Connecting to monitor at \"%s\"", options.monitor_pguri);
 		}
+	}
+	else
+	{
+		cli_common_get_set_pgdata_or_exit(&(options.pgSetup));
 	}
 
 	if (IS_EMPTY_STRING_BUFFER(options.formation) ||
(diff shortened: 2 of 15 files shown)


JelteF (changes requested):
I played around with this a bit. Some issues:
1. The `--monitor` argument doesn't actually seem to work (setting `PG_AUTOCTL_MONITOR` does work):
```
pg_autoctl show state --monitor 'postgres://autoctl_node@localhost:5500/pg_auto_failover?sslmode=prefer'
```

2. When using `--pgdata` you still see this error in the output:
```
12:02:39 14800 ERROR Failed to get value for environment variable 'PG_AUTOCTL_MONITOR', which is unset
```
3. When not specifying either `--pgdata` or `--monitor` the final fatal log should include something about `--monitor` and `PG_AUTOCTL_MONITOR` too:
```
11:55:49 4723 ERROR Failed to get value for environment variable 'PG_AUTOCTL_MONITOR', which is unset
11:55:49 4723 ERROR Failed to get value for environment variable 'PGDATA', which is unset
11:55:49 4723 FATAL Failed to set PGDATA either from the environment or from --pgdata
```

src/bin/pg_autoctl/cli_do_tmux.c
```diff
@@ -635,7 +647,7 @@ tmux_pg_autoctl_create_postgres(PQExpBuffer script,
 	tmux_add_send_keys_command(script, "export PGPORT=%d", pgport);
 
 	sformat(monitor, sizeof(monitor),
-			"$(%s show uri --pgdata %s/monitor --monitor)",
+			"$(%s show uri --pgdata %s/monitor --formation monitor)",
```
JelteF, line 650: I think if we change the args like this we would need a minor release instead of a patch. Since it could be a breaking change for some automation.

JelteF: Also, maybe we should disallow using `monitor` as an actual formation name.

src/bin/pg_autoctl/cli_formation.c
```diff
@@ -350,10 +404,21 @@ keeper_cli_formation_create(int argc, char **argv)
 	FormationConfig config = formationOptions;
 	Monitor monitor = { 0 };
 
-	if (!monitor_init_from_pgsetup(&monitor, &config.pgSetup))
```
JelteF, line 353: Use `cli_monitor_init_from_option_or_config` instead?

src/bin/pg_autoctl/cli_formation.c
```diff
@@ -388,10 +453,21 @@ keeper_cli_formation_drop(int argc, char **argv)
 		exit(EXIT_CODE_BAD_ARGS);
 	}
 
-	if (!monitor_init_from_pgsetup(&monitor, &config.pgSetup))
```
JelteF, line 391: Use `cli_monitor_init_from_option_or_config` instead?

src/bin/pg_autoctl/cli_perform.c
```diff
@@ -190,13 +204,25 @@ cli_perform_failover_getopts(int argc, char **argv)
...
-	if (!keeper_config_set_pathnames_from_pgdata(&(options.pathnames),
-												 options.pgSetup.pgdata))
+	/* when we have a monitor URI we don't need PGDATA */
+	if (cli_use_monitor_option(&options))
 	{
-		/* errors have already been logged */
-		exit(EXIT_CODE_BAD_ARGS);
+		if (!IS_EMPTY_STRING_BUFFER(options.pgSetup.pgdata))
+		{
+			log_warn("Given --monitor URI, the --pgdata option is ignored");
```
JelteF, line 212: Is this really what we want? Other options are hard error or make `--pgdata` win.

Also does the same hold for setting `PG_AUTOCTL_MONITOR` and then using `--pgdata`?

src/bin/pg_autoctl/cli_show.c
```diff
@@ -963,41 +1019,82 @@ cli_show_uri(int argc, char **argv)
...
+	{
+		if (!monitor_init(&monitor, kconfig.monitor_pguri))
+		{
+			/* errors have already been logged */
+			exit(EXIT_CODE_BAD_ARGS);
+		}
+
+		/* cook the sslmode=prefer bits */
+		ssl.active = true;
+		ssl.sslMode = SSL_MODE_PREFER;
```
JelteF, line 1074: Any chance we could get this from the `monitor_pguri`? i.e. if monitor uri has `sslmode=verify-full&sslrootcert=somecert.crt` we would use those options.

src/bin/pg_autoctl/cli_do_tmux.c
DimCitus: Agreed about the minor release, it's a breaking change, although a very minor one. In another PR coming soon we reserve the monitor name for a formation where we register the monitor node itself...

src/bin/pg_autoctl/cli_formation.c
DimCitus: Can't, we have a `FormationConfig config` structure instance, not a `KeeperConfig config` one...

src/bin/pg_autoctl/cli_formation.c
DimCitus: Same as before.

src/bin/pg_autoctl/cli_show.c
DimCitus: Done now.

src/bin/pg_autoctl/cli_perform.c
DimCitus: The other commands also prefer using --monitor over --pgdata, I think we're good here?

JelteF (changes requested):
Some more issues:
1. 
```
PG_AUTOCTL_MONITOR='postgres://autoctl_node@localhost:5500/pg_auto_failover?sslmode=require' pg_autoctl create formation --formation 'test' --kind pgsql
13:48:11 24106 ERROR BUG: keeper_config_set_pathnames_from_pgdata: empty pgdata
```

2.
```
pg_autoctl create formation --formation 'test' --kind pgsql
13:51:00 30894 ERROR BUG: keeper_config_set_pathnames_from_pgdata: empty pgdata
```
3.
```
pg_autoctl create formation --formation 'test' --kind pgsql --monitor 'postgres://autoctl_node@localhost:5500/pg_auto_failover?sslmode=require'
13:51:37 32484 ERROR Failed to get value for environment variable 'PGDATA', which is unset
13:51:37 32484 FATAL Failed to set PGDATA either from the environment or from --pgdata
```
4. drop formation doesn't say it can use `--monitor` in the error message if you supply neither --monitor or --pgdata.
5.
```
PG_AUTOCTL_MONITOR='postgres://autoctl_node@localhost:5500/pg_auto_failover?sslmode=require' pg_autoctl drop formation --formation 'test'
13:53:27 3635 ERROR Failed to get value for environment variable 'PGDATA', which is unset
13:53:27 3635 FATAL Failed to set PGDATA either from the environment or from --pgdata
```
6. `show standby-names --group 5` does not error when there's no group 5 (might be the case on master too)

src/bin/pg_autoctl/parsing.c
```diff
@@ -685,6 +685,74 @@ buildPostgresURIfromPieces(URIParams *uriParams, char *pguri)
...
+			strlcpy(ssl->crlFile, val, sizeof(ssl->crlFile));
+		}
+		else if (streq(key, "sslcert"))
+		{
+			strlcpy(ssl->serverCert, val, sizeof(ssl->serverCert));
+		}
+		else if (streq(key, "sslkey"))
+		{
+			strlcpy(ssl->serverKey, val, sizeof(ssl->serverKey));
+		}
```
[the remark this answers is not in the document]
DimCitus: I see what you mean. But in case we would reuse the function `parse_pguri_ssl_settings` ini other parts of the code I think it makes sense to still correctly parse those connection string parameters. After all those parameters are listed in https://www.postgresql.org/docs/current/libpq-connect.html.

DimCitus:
> Some more issues:

I have changed the way the `--monitor` and `--pgdata` and `PG_AUTOCTL_MONITOR` and `PGDATA` are understood in many places except for the formation code path, that I had done first. Sorry about that. I have now updated the code to match our current expectations.

> 1. `show standby-names --group 5` does not error when there's no group 5 (might be the case on master too)

Yeah, so the server-side function was returning NULL, which we didn't test against in `parseSingleValueResult`. So I have added an explicit `PQgetisnull` check in there, and also changed the server-side function to report an error in the case where we find zero nodes.

src/bin/pg_autoctl/parsing.c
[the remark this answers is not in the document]
JelteF: Ah you are totally right. My mistake, I forgot these are used for TLS user certs.

JelteF (approved):
Works well now

PR #581, opened 2021-02-04
Plug few isolated memleaks that showed up during profiling

gkokolatos:
What it says on the tin + what it seems to be a missing return in `pg_setup_get_local_connection_string`


src/bin/pg_autoctl/monitor.c
```diff
@@ -416,6 +416,10 @@ monitor_print_nodes_as_json(Monitor *monitor, char *formation, int groupId)
 		log_error("Failed to get the nodes from the monitor while running "
 				  "\"%s\" with formation %s and group %d",
 				  sql, formation, groupId);
+		if (context.strVal)
+		{
+			free(context.strVal);
```
DimCitus, line 421: I wonder if we should change the definition of SingleValueResultContext (and other similar structs) so that the structure contains a pre-allocated string buffer rather than just a pointer to dynamic memory. Though I must admit fixing what we have now is great, so let's merge that one first and think more later.

src/bin/pg_autoctl/pgsetup.c
```diff
@@ -477,16 +479,19 @@ get_pgpid(PostgresSetup *pgSetup, bool pgIsNotRunningIsOk)
 		if (kill(standalonePid, 0) == 0)
 		{
 			pgSetup->pidFile.pid = pid;
+			free(contents);
```
DimCitus, line 482: I think we should unconditionally free the contents after a successful parse, and stop worrying in all the return cases thereafter.

DimCitus (changes requested):
Thanks for the clean-up job Georgios, that's highly appreciated! I think we should simply the free management in `get_pgpid` (free once when we know we don't use the contents directly anymore), and I wonder if we should pre-allocate the `strVal` in `SingleValueResultContext` though I would understand if we want to make that another PR.

src/bin/pg_autoctl/monitor.c
gkokolatos: :+1:

src/bin/pg_autoctl/pgsetup.c
gkokolatos: Sure, will do

gkokolatos: Let me know if this resolved with the latest push

DimCitus (approved):

src/bin/pg_autoctl/pgsetup.c
DimCitus: LGTM!

PR #583, opened 2021-02-05
Plug minimal memleak in keeper's configuration (Last in a series)

gkokolatos:
Admittedly this leak could have been plugged with a couple of well placed calls
to the now obsolete keeper_config_destroy() function. However, given the
lifecycle and globality of the struct the variable is found in, ownership is
blurred in some cases.

Instead the leaking variable is now living in the stack along with the rest. A
new variable for its length is added instead of MAXCONNINFO which seems a bit
too excessive. Alignment concerns aside, some relief from the stack stress might
be desirable.

Finally, the necessity for keeper_config_destroy is now removed. Thus the
function itself is removed for the benefit of the reader of the codebase.


DimCitus (approved):
Thanks for that clean-up!

Issue #313, opened 2020-06-26
Segmentation fault (core dumped)


NishchayDBA:
Hi Team,

I am getting a Segmentation fault (core dumped) message every time while creating secondary node. I am using the below command to create secondary node.
I am able to create a monitor and primary server. 

pg_autoctl create postgres --nodename edb2 --auth trust --ssl-self-signed --pgdata /opt/PG12 --pgport 5432 --pgctl /usr/pgsql-12/bin/pg_ctl --monitor postgres://autoctl_node@edb3:5432/pg_auto_failover?sslmode=prefer

OS: CentOS 7 64bit
PGSQL- 12

Please refer to the attachment for more information. 
![image](https://user-images.githubusercontent.com/67452345/85821118-7c8a8e80-b7aa-11ea-82c8-0d8d2ea32381.png)

I also found that after creating a monitor and primary server encoding is SQL_ASCII , Collate & Ctype is C but if i do manually initdb it's showing UTF8.

<img width="507" alt="Encoding" src="https://user-images.githubusercontent.com/67452345/85821466-4e597e80-b7ab-11ea-9944-7fb763eee0c8.PNG">
<img width="953" alt="Segmentation fault" src="https://user-images.githubusercontent.com/67452345/85821467-4f8aab80-b7ab-11ea-92cd-a9ebb9b94b54.PNG">

DimCitus:
Hi @NishchayDBA ; thanks for opening this issue. We have two problems here:

  1. segfault. We need to fix that. I don't have enough information in the screenshot you have shared. Please run the command with the `-vvv` switch (for _very very verbose_) and save the output to a file that you can then attach to this issue
  2. initdb environment ; can you check that your choice of locales are exported environment variables? when that's the case, I believe `pg_autoctl` would just inherit the values and then `pg_autoctl initdb` should also see the same values. It seems not to be the case in your environment, so we need to dig deeper. Which shell are you using? Can you paste the result (as text rather than images) of commands such as `env` and `locale`?

NishchayDBA:
Hi Dimitri,

Thank you for your response. 
I have executed pg_autoctl create command with -vvv mode. Please find the output in the attachment.

For, ENCODING issue I will export the UTF8 in my environment variables and then try to create monitor and check I am getting correct encoding or not. 


[Segmentation fault (core dumped).txt]
(https://github.com/citusdata/pg_auto_failover/files/4835950/Segmentation.fault.core.dumped.txt)

[...]

PR #598, opened 2021-02-19
Vendor-in rmtree

godwhoa:
As per the suggestion from here:
https://github.com/citusdata/pg_auto_failover/issues/313#issuecomment-781976113
I have tested the change locally, it does fix the seg fault from #313.
I'm not familiar with C build tooling so I just mimicked what was done for snprintf in the Makefiles etc.

### DIFF
--- a/src/bin/lib/pg/rmtree.c
+++ b/src/bin/lib/pg/rmtree.c
@@ -1,0 +1,111 @@
+/*-------------------------------------------------------------------------
+ *
+ * rmtree.c
+ *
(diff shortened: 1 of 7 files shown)


DimCitus:
Thanks for your contribution @godwhoa ; a full review and merge should happen in the next couple weeks (vacations are also happening).

godwhoa:
Closing this in favor of #612.
After discussing with Dimitri on IRC, simpler fix for it was to call `pg_logging_init` for PG versions >= 12.

PR #609, opened 2021-03-03
Fix/skip hba propagation

DimCitus:
I believe the most important part of those changes is there:

``` diff
modified   src/bin/pg_autoctl/keeper_config.c
@@ -195,6 +195,7 @@
 		OPTION_POSTGRESQL_PROXY_PORT(config), \
 		OPTION_POSTGRESQL_LISTEN_ADDRESSES(config), \
 		OPTION_POSTGRESQL_AUTH_METHOD(config), \
+		OPTION_POSTGRESQL_HBA_LEVEL(config), \
 		OPTION_SSL_ACTIVE(config), \
 		OPTION_SSL_MODE(config), \
 		OPTION_SSL_CA_FILE(config), \
```

To easily get to test the code:

```
$ make CLUSTER_OPTS="--skip-pg-hba" cluster
```


src/bin/pg_autoctl/keeper_config.c
```diff
@@ -195,6 +195,7 @@
 		OPTION_POSTGRESQL_PROXY_PORT(config), \
 		OPTION_POSTGRESQL_LISTEN_ADDRESSES(config), \
 		OPTION_POSTGRESQL_AUTH_METHOD(config), \
+		OPTION_POSTGRESQL_HBA_LEVEL(config), \
```
JelteF, line 198: If I understand this change correctly it means we didn't write hba-level to the ini file before. How will this change interact with nodes created with a previous version of pg_auto_ctl that were created with `--skip-pg-hba`? What will users need to do to fix these nodes?

JelteF (commented):
Looks good overall. Can we add a regression test for this?

src/bin/pg_autoctl/cli_common.c
```diff
@@ -193,14 +193,23 @@ cli_common_keeper_getopts(int argc, char **argv,
...
 				log_trace("--auth %s", LocalOptionConfig.pgSetup.authMethod);
+
+				if (LocalOptionConfig.pgSetup.hbaLevel == HBA_EDIT_UNKNOWN)
+				{
+					strlcpy(LocalOptionConfig.pgSetup.hbaLevelStr,
+							pgsetup_hba_level_to_string(HBA_EDIT_MINIMAL),
+							sizeof(LocalOptionConfig.pgSetup.hbaLevelStr));
+
+					LocalOptionConfig.pgSetup.hbaLevel = HBA_EDIT_MINIMAL;
+				}
```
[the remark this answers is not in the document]
DimCitus: This is necessary because we have to play the option compatibility game, where you can have either `--auth` or `--skip-pg-hba`, and if you didn't use `--skip-pg-hba` then you can also have `--pg-hba-lan`.

src/bin/pg_autoctl/cli_do_root.c
```diff
@@ -286,9 +294,14 @@ CommandLine do_tmux_wait =
...
-				 "  --root          path where to create a cluster\n"
-				 "  --first-pgport  first Postgres port to use (5500)\n"
-				 "  --nodes         number of Postgres nodes to create (2)",
+				 "  --root            path where to create a cluster\n"
+				 "  --first-pgport    first Postgres port to use (5500)\n"
+				 "  --nodes           number of Postgres nodes to create (2)\n"
+				 "  --async-nodes     number of async nodes within nodes (0)\n"
+				 "  --node-priorities list of nodes priorities (50)\n"
+				 "  --sync-standbys   number-sync-standbys to set (0 or 1)\n"
+				 "  --skip-hba        use --skip-hba when creating nodes\n"
```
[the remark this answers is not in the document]
DimCitus: Fixed now.

src/bin/pg_autoctl/keeper_config.c
DimCitus: When the configuration file was not updated, we would fill-in the gap with using `minimal` at read-time. To fix it is possible to use `pg_autoctl set config postgresql.hba_level skip`. I don't see how to automate that, so I believe we should address the situation with a special release note at the next release?

JelteF: Okay, makes sense. I already thought we couldn't automatically fix this for existing users. Let's make a clear note of this in the next release then.

DimCitus:
> Looks good overall. Can we add a regression test for this?

I have now added tests, is that the kind of coverage you had in mind?

JelteF (approved):

PR #610, opened 2021-03-03
Fix a buffer overlap instruction that macOs libc fails to process.

DimCitus:


JelteF (approved):
From what I understand this change fixes a strlcpy in standby_init_replication_source
 where src and dst are the same (or at least overlapping). If that's indeed correct I think it's good to merge.

DimCitus:
Indeed the problem is that we had a `strlcpy` call using same src and dst address. I could have made it obvious, sorry...

PR #612, opened 2021-03-03
Add pg_logging_init for PG version 12 and above

godwhoa:
I have tested the change locally on pg versions 10, 11, 12 and 13, `rmtree` no longer segfaults for PG versions >= 12.
How I tested it:
1) Get monitor and primary up and running.
2) When setting up the secondary we set PGDATA to a mount point so if we `rmtree` it, we'll get a busy device error. 
`rmtree` logs that error using `pg_log_warn` which would segfault if we don't call `pg_logging_init` before hand.

None of the testing is automated, it was done with a finicky docker-compose setup and manually executing commands on the secondary node's container.
If we want to add a test case for this, I don't know how to do it in the existing testing framework.

### DIFF
--- a/src/bin/pg_autoctl/main.c
+++ b/src/bin/pg_autoctl/main.c
@@ -53,7 +53,7 @@ main(int argc, char **argv)
 
 	/*
 	 * Since PG 12, we need to call pg_logging_init before any calls to pg_log_*
-	 * otherwise, we get a segfault. Although we doesn't use pg_log_* directly,
+	 * otherwise, we get a segfault. Although we don't use pg_log_* directly,
 	 * functions from the common library such as rmtree do use them.
 	 * Logging change introduced in PG 12: https://git.postgresql.org/cgit/postgresql.git/commit/?id=cc8d41511721d25d557fc02a46c053c0a602fed0
 	 */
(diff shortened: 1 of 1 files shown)


src/bin/pg_autoctl/main.c
```diff
@@ -48,6 +51,16 @@ main(int argc, char **argv)
 	/* set our logging infrastructure */
 	(void) set_logger();
 
+	/*
+	 * Since PG 12, we need to call pg_logging_init before any calls to pg_log_*
+	 * otherwise, we get a segfault. Although we doesn't use pg_log_* directly,
+	 * functions from the common library such as rmtree do use them.
+	 * Logging change introduced in PG 12: https://git.io/JqvcN
```
DimCitus, line 58: Is the short URL scheme going to resist the test of time? I think we should use either https://git.postgresql.org/gitweb/?p=postgresql.git;a=commitdiff;h=cc8d41511721d25d557fc02a46c053c0a602fed0 or https://git.postgresql.org/cgit/postgresql.git/commit/?id=cc8d41511721d25d557fc02a46c053c0a602fed0.

DimCitus (changes requested):
Thanks a lot for the work, I think we're almost ready to merge, I'm just worried about the URL used in the comment to link to the Postgres commit.

DimCitus:
Fixes #313.

src/bin/pg_autoctl/main.c
godwhoa: git.io is maintained by Github itself so thought it might be fine, but who knows, valid concern.
Changed it to https://git.postgresql.org/cgit/postgresql.git/commit/?id=cc8d41511721d25d557fc02a46c053c0a602fed0

DimCitus (approved):

DimCitus:
Thanks for your first contribution @godwhoa and all the work with the previous attempts at fixing this bug. Much appreciated!

PR #616, opened 2021-03-08
Implement basis for Citus Secondaries support (cluster_name).

DimCitus:


onderkalaci (approved):
These are already reviewed on the Citus-ha repo.

Issue #608, opened 2021-03-03
drop node not really drop 


laurent6:
Hi, 
I've another issues. 
I have 3 nodes: 
node0: the monitor
node1 <=>node_16
node2 <=> node_17 

I want auto change state node. At the beginning  it'st work. node_16 promote to pimary and node_17 set secondary. I simulated breakdown node_16 but système not promote node_17 to primary. 
The first question Why not promot auto node_17 to primary when the first down ? 
After I drop node. The system change and the node_17 is single. 
Another issus when I restart node_16 ( delete and recreate it )

What can I forget or what worng in my config ? 

1] 
``` postgres@t-os-postgres-dev2:~$ pg_autoctl show state
   Name |  Node |           Host:Port |       LSN | Reachable |       Current State |      Assigned State
--------+-------+---------------------+-----------+-----------+---------------------+--------------------
node_16 |    16 | 192.168.10.166:5000 | 0/3000148 |       yes |             primary |             primary
node_17 |    17 |   192.168.10.5:5000 | 0/3000148 |       yes |           secondary |           secondary
```

2] breackdown node_16 

3] show state 
```postgres@t-os-postgres-dev2:~$ pg_autoctl show state
   Name |  Node |           Host:Port |       LSN | Reachable |       Current State |      Assigned State
--------+-------+---------------------+-----------+-----------+---------------------+--------------------
node_16 |    16 | 192.168.10.166:5000 | 0/3000148 |        no |             primary |      demote_timeout
node_17 |    17 |   192.168.10.5:5000 | 0/3000310 |       yes |    stop_replication |    stop_replication
....


postgres@t-os-postgres-dev2:~$ pg_autoctl show state
   Name |  Node |           Host:Port |       LSN | Reachable |       Current State |      Assigned State
--------+-------+---------------------+-----------+-----------+---------------------+--------------------
node_16 |    16 | 192.168.10.166:5000 | 0/3000148 |        no |             primary |             demoted
node_17 |    17 |   192.168.10.5:5000 | 0/3000310 |       yes |        wait_primary |        wait_primary

``` 
4] drop state
``` postgres@t-os-postgres-dev2:~$ pg_autoctl drop node  --hostname 192.168.10.166 --pgport 5000 node_16 ``` 

5] show state 
``` postgres@t-os-postgres-dev2:~$ pg_autoctl show state
   Name |  Node |         Host:Port |       LSN | Reachable |       Current State |      Assigned State
[...]

PR #617, opened 2021-03-09
Review pg_autoctl show state output, and docs.

DimCitus:
Make it clear which kind of connection is given by which node, in particular
that when in the SINGLE or WAIT_PRIMARY state we are still delivering a
read-write connection to the application.

The "Reachable" column is now a "Connection" column and the node health as
seen by the monitor is encoded in the single-char *, !, or ?, depending on
the health value (1, 0, -1).

See #608.

With this PR applied we can get such output:

```
$ pg_autoctl show state
 Name |  Node |      Host:Port |       LSN |   Connection |       Current State |      Assigned State
------+-------+----------------+-----------+--------------+---------------------+--------------------
node1 |     1 | localhost:5501 | 0/40001F8 | read-write * |             primary |             primary
node2 |     2 | localhost:5502 | 0/40001F8 |  read-only * |           secondary |           secondary
node3 |     3 | localhost:5503 | 0/40001F8 |       none * |         maintenance |         maintenance
```

Another example involving a `wait_primary` node:

```
$ pg_autoctl show state
 Name |  Node |      Host:Port |       LSN |   Connection |       Current State |      Assigned State
------+-------+----------------+-----------+--------------+---------------------+--------------------
node1 |     1 | localhost:5501 | 0/3000000 | read-write * |        wait_primary |        wait_primary
node2 |     2 | localhost:5502 |       0/0 |       none ? |        wait_standby |          catchingup
```


src/bin/pg_autoctl/nodestate_utils.c
```diff
@@ -382,6 +391,92 @@ nodestateHealthToString(int health)
...
+		}
+
+		case 0:
+		{
+			return '!';
+		}
+
+		case 1:
+		{
+			return '*';
```
JelteF, line 414:
```suggestion
			return '✓';
```

src/bin/pg_autoctl/nodestate_utils.c
```diff
@@ -382,6 +391,92 @@ nodestateHealthToString(int health)
...
+	switch (health)
+	{
+		case -1:
+		{
+			return '?';
+		}
+
+		case 0:
+		{
+			return '!';
```
JelteF, line 409:
```suggestion
			return '✗';
```

JelteF (approved):
I think these symbols are a bit clearer (especially the checkmark vs the *)

src/bin/pg_autoctl/nodestate_utils.c
JelteF: Maybe this being a space as also totally fine, i.e. no need to have a character shown if everything is good.

DimCitus: I don't know how to maintain table alignment when using wider unicode chars, and unfortunately both ✓ and ✗ are wider than space.

src/bin/pg_autoctl/nodestate_utils.c
DimCitus: Same unicode remark with wider-than-space characters and table alignment.

PR #619, opened 2021-03-11
Fix a pipe issue and return code

gfphoenix78:
Between closing the pipe files and waiting the program to exit,
the program may write to the closed pipe files, which causes
the program got the signal SIGPIPE and exit abnormally.
A safer way is to close these pipe files after the program does
exit.

Another issue is to extract the return code. `WEXITSTATUS(status)`
should be only employed if WIFEXITED returned true. Similarly,
if `WIFSIGNALED(status)` is true, set the 128+signo as the return code.


src/bin/lib/subcommands.c/runprogram.h
```diff
@@ -581,7 +585,20 @@ waitprogram(Program *prog, pid_t childPid)
...
 
-	prog->returnCode = WEXITSTATUS(status);
+	if (WIFEXITED(status))
+	{
+		prog->returnCode = WEXITSTATUS(status);
+	}
+	else if (WIFSIGNALED(status))
+	{
+		int signo = WTERMSIG(status);
+		prog->returnCode = signo ? signo : -1;
```
DimCitus, line 595: Why would we return -1 instead of 0, and why the signal number directly rather than what I believe the default for Unix is: 128 + Signo? I can be convinced that your current choice is the best one, with proper amount of comments and references. Thanks!

DimCitus (changes requested):

src/bin/lib/subcommands.c/runprogram.h
gfphoenix78: Thank you for your suggestion. You're right.

gfphoenix78:
@DimCitus I've updated the code according to your suggestion. Could you take some time to review it?

DimCitus (approved):

DimCitus:
Hi @gfphoenix78 ; I'm just back from some days off and merged your PR. Thanks for your contribution!

PR #582, opened 2021-02-04
Plug connection leaks found during profiling

gkokolatos:
It seems that pgsql_execute_with_params() during its lifetime has been
inconsistently altered. The latest version notes in the comments that the
connection is not persistant to facilitate error handling. However that was not
entirely true and several parts of the code assumed it to not be true. Others
assumed to be true and failed to release the connection once used.

For the sake of clarity, the function will now explicitly close the connection
that has used, regardless of wether it is a new or existing connection. That
simplifies most of the code and plugs the connection leaks.

It also unconvers an inconsistency on the connections used for notification. The
code mixed the connection it was using to listen to events from the monitor and
with others. A new PGconn member has been added in the monitor struct to
distinguish between the two distinct cases.

PR #628, opened 2021-03-16
Do not leak psycopg2 connections during testing which can lead to fla…

gkokolatos:
…kyness

According to psycopg2 documentation connections are not closed on context exit.
This is in contrast with connection cursors that actually do so.

It is not an inherent problem in itself, however, calls to get_state() in a
loop, such as from wait_until_*, can overflow the connection limits and cause
the test suite to fail. This is particularly noticable when the system is slow
to respond, for example if it is run under valgrind.

This might address other objerved flakyness in the tests due to connection
exhaustion with messages as:
"psycopg2.OperationalError: FATAL:  sorry, too many clients already"

Run into this repeatedly while profiling in order to address #582 
Might be related with the comment in #512#issuecomment-767205595

### DIFF
--- a/tests/pgautofailover_utils.py
+++ b/tests/pgautofailover_utils.py
@@ -359,14 +359,21 @@ class PGNode:
         Runs the given sql query with the given arguments in this postgres node
         and returns the results. Returns None if there are no results to fetch.
         """
-        with psycopg2.connect(self.connection_string()) as conn:
-            cur = conn.cursor()
-            cur.execute(query, args)
-            try:
-                result = cur.fetchall()
-                return result
-            except psycopg2.ProgrammingError:
-                return None
+        result = None
+        conn = psycopg2.connect(self.connection_string())
+
+        with conn:
+            with conn.cursor() as cur:
+                cur.execute(query, args)
+                try:
+                    result = cur.fetchall()
+                except psycopg2.ProgrammingError:
+                    pass
+        # leaving contexts closes the cursor, however
+        # leaving contexts doesn't close the connection
+        conn.close()
+
+        return result
 
     def pg_config_get(self, settings):
         """


tests/pgautofailover_utils.py
```diff
@@ -359,14 +359,21 @@ def run_sql_query(self, query, *args):
...
-            cur.execute(query, args)
-            try:
-                result = cur.fetchall()
-                return result
-            except psycopg2.ProgrammingError:
-                return None
+        result = None
+        conn = psycopg2.connect(self.connection_string())
+
+        with conn:
```
DimCitus, line 365: Do we still need to have a context handler on the connection object if the context handler doesn't close the connection anyway? What it is used for then?

DimCitus (approved):

gkokolatos:
On Wednesday, March 17, 2021 10:56 AM, Dimitri Fontaine ***@***.***> wrote:

> @DimCitus approved this pull request.
>
> ---------------------------------------------------------------
>
> In [tests/pgautofailover_utils.py](https://github.com/citusdata/pg_auto_failover/pull/628#discussion_r595867546):
>
>> @@ -359,14 +359,21 @@ def run_sql_query(self, query, *args):
>          Runs the given sql query with the given arguments in this postgres node
>          and returns the results. Returns None if there are no results to fetch.
>          """
> -        with psycopg2.connect(self.connection_string()) as conn:
> -            cur = conn.cursor()
> -            cur.execute(query, args)
> -            try:
> -                result = cur.fetchall()
> -                return result
> -            except psycopg2.ProgrammingError:
> -                return None
> +        result = None
> +        conn = psycopg2.connect(self.connection_string())
> +
> +        with conn:
>
> Do we still need to have a context handler on the connection object if the context handler doesn't close the connection anyway? What it is used for then?

I understand your question. I am not going to claim that I am python nor a psycopg2 expert. However the documentation
of psycopg2 ([https://www.psycopg.org](https://www.psycopg.org/docs/connection.html)) gives this format as best practice so obliged. In either case it does not hurt, does it?

DimCitus:
Ah yeah that's per psycopg2 docs. Perfect then, many thanks again for your contributions!

gkokolatos:
On Wednesday, March 17, 2021 11:52 AM, Dimitri Fontaine ***@***.***> wrote:

> Ah yeah that's per psycopg2 docs. Perfect then, many thanks again for your contributions!

Thank you for the project!

PR #632, opened 2021-03-18
Add individual pages for the pg_autoctl commands.

### LINKED ISSUE #595: get formation-settings and number-sync-standbys do not error on non existing formation
using `--formation doesnotexist` does not throw an error.

### LINKED ISSUE #599: pg_autoctl show state should default to --local when monitor is disabled
When the monitor is disabled and you run `pg_autoctl show state` you get this error:
```
11:11:32 30357 ERROR Monitor is disabled in the configuration
```

I think it makes sense to default to `--local` in that case. Or at least the error should include a hint that you should use this option instead.

### LINKED ISSUE #600: show state --local ignores --group
When using `make cluster` and running the following command:
```
pg_autoctl show state --pgdata node2 --group 3 --local
 Name |  Node |      Host:Port |       LSN | Reachable |       Current State |      Assigned State
------+-------+----------------+-----------+-----------+---------------------+--------------------
node2 |     2 | localhost:5502 | 0/3000640 |   unknown |           secondary |           secondary
```

### LINKED ISSUE #614: Question reload and restart parameters
Hello,

1 - For keeper and monitor is it possible to document wich parameters requires reload and wich restart (for pg_autoctl --config).
2 - Monitor extensions parameters all require a postgres reload. Do they require a reload/restart  of the monitor service as well ?
3 - Does pg_autoctl set,  trigg a reload (particulary when pg_autoctl service is launched as a systemd unit ) ?

Thank you very much.

### LINKED ISSUE #639: confusing log messages when using show state --local --monitor url
When running the following command it says it's connecting to the monitor and using it over pgdata, but it still actually fetches the local data of `node2`:
```
pg_autoctl show state --pgdata node2 --local --monitor "postgres://autoctl_node@localhost:5500/pg_auto_failover?sslmode=prefer"
12:23:31 11420 WARN  Given --monitor URI, the --pgdata option is ignored
12:23:31 11420 INFO  Connecting to monitor at "postgres://autoctl_node@localhost:5500/pg_auto_failover?sslmode=prefer"
 Name |  Node |      Host:Port | LSN |   Connection |       Current State |      Assigned State
------+-------+----------------+-----+--------------+---------------------+--------------------
node2 |     2 | localhost:5502 |     |       none ? |        wait_standby |          catchingup
```

I think these `--monitor` related logs should not be shown when `--local` is used. Or we should disallow specifying both of them.

DimCitus:
This PR adds a dedicated manual page for each of the pg_autoctl commands, and fixes some of the option help strings, and some other option parsing infelicities.

Fixes #595.
Fixes #599.
Fixes #600.
Fixes #614.
Fixes #639.


src/bin/pg_autoctl/cli_show.c
```diff
@@ -418,8 +423,43 @@ cli_show_state(int argc, char **argv)
...
+		{
+			log_info("Monitor is disabled, showing --local state");
+			(void) cli_show_local_state();
+			exit(EXIT_CODE_QUIT);
+		}
+	}
+
 	(void) cli_monitor_init_from_option_or_config(&monitor, &config);
 
+	if (optionGroupId != -1 && config.groupId != optionGroupId)
```
JelteF, line 456: Why are you disallowing querying other groups here?

JelteF: I guess this should instead be in `cli_show_local_state`.

JelteF (approved):
I skimmed the generated docs and they look much better than before! The code changes look okay too except for the comment I left.

src/bin/pg_autoctl/cli_show.c
DimCitus: Good point, missed that entirely. Fixed now, thanks!

PR #637, opened 2021-03-23
Fix/pgsql connection leaks

DimCitus:
Based on the work from Georgios at https://github.com/citusdata/pg_auto_failover/pull/582 ; with some of the fixes I asked for and the `pgsql_begin` (commit, rollback) api that tracks opened transactions to avoid automatic connection closing in those cases.


DimCitus:
Fixed in #582 instead. Closing now.

Issue #653, opened 2021-03-29
pg_rewind sometimes fails with fatal: target server must be shut down cleanly


DimCitus:
It appears that `pg_rewind` only works when Postgres could shut down cleanly, which is not always the case. See https://github.com/postgres/postgres/blob/master/src/bin/pg_rewind/pg_rewind.c#L703 for more details. It might be possible for pg_autoctl to restart Postgres in crash-recovery mode first, and then attempt pg_rewind, then necessary.

PR #656, opened 2021-03-29
Feature crash recovery before pg rewind

DimCitus:
Implement crash recovery (target immediate, timeline current).

When pg_rewind would fail because the target PGDATA was not shutdown
properly, have Postgres do a round of crash recovery to clean-up the data
directory and only then resort to pg_rewind.

Now pg_rewind can fail for other reasons:

    INFO  pg_rewind: connected to server
    INFO  pg_rewind: servers diverged at WAL location 0/2C000060 on timeline 21
    INFO  pg_rewind: fatal: could not find previous WAL record at 0/2C000060: invalid record length at 0/2C000060: wanted 24, got 0

See #653

### DIFF
--- a/src/bin/pg_autoctl/cli_service.c
+++ b/src/bin/pg_autoctl/cli_service.c
@@ -19,6 +19,7 @@
 #include "cli_common.h"
 #include "commandline.h"
 #include "defaults.h"
+#include "env_utils.h"
 #include "fsm.h"
 #include "keeper_config.h"
 #include "keeper.h"
@@ -285,6 +286,7 @@ cli_getopt_pgdata_and_mode(int argc, char **argv)
 		{ "pgdata", required_argument, NULL, 'D' },
 		{ "fast", no_argument, NULL, 'f' },
 		{ "immediate", no_argument, NULL, 'i' },
+		{ "sigkill", no_argument, NULL, '9' },
 		{ "version", no_argument, NULL, 'V' },
 		{ "verbose", no_argument, NULL, 'v' },
 		{ "quiet", no_argument, NULL, 'q' },
@@ -327,6 +329,26 @@ cli_getopt_pgdata_and_mode(int argc, char **argv)
 					exit(EXIT_CODE_BAD_ARGS);
 				}
 				stop_signal = SIGQUIT;
+				break;
+			}
+
+			case '9':
+			{
+				/* change the signal to send from SIGTERM to SIGKILL */
+				if (!env_exists(PG_AUTOCTL_DEBUG))
+				{
+					log_fatal("Option --sigkill is only available in debug "
+							  "environments");
+					exit(EXIT_CODE_BAD_ARGS);
+					break;
+				}
+
+				if (stop_signal != SIGTERM)
+				{
+					log_fatal("Please use either --fast or --immediate, not both");
+					exit(EXIT_CODE_BAD_ARGS);
+				}
+				stop_signal = SIGKILL;
 				break;
 			}
 
@@ -407,11 +429,29 @@ cli_service_stop(int argc, char **argv)
 
 	if (read_pidfile(keeper.config.pathnames.pid, &pid))
 	{
-		if (kill(pid, stop_signal) != 0)
+		/*
+		 * Send the signal to the top-level process only, except when using
+		 * --sigkill and then `kill -9`. The intend there is to trigger a crash
+		 * of Postgres and pg_autoctl and see how we recover from it. Target
+		 * the whole process group then.
+		 */
+		if (stop_signal == SIGKILL)
 		{
-			log_error("Failed to send %s to pg_autoctl pid %d: %m",
-					  strsignal(stop_signal), pid);
-			exit(EXIT_CODE_INTERNAL_ERROR);
+			if (killpg(pid, stop_signal) != 0)
+			{
+				log_error("Failed to send %s to pg_autoctl pid %d: %m",
+						  strsignal(stop_signal), pid);
+				exit(EXIT_CODE_INTERNAL_ERROR);
+			}
+		}
+		else
+		{
+			if (kill(pid, stop_signal) != 0)
+			{
+				log_error("Failed to send %s to pg_autoctl pid %d: %m",
+						  strsignal(stop_signal), pid);
+				exit(EXIT_CODE_INTERNAL_ERROR);
+			}
 		}
 	}
 	else
--- a/src/bin/pg_autoctl/cli_service.c
+++ b/src/bin/pg_autoctl/cli_service.c
@@ -345,7 +345,8 @@ cli_getopt_pgdata_and_mode(int argc, char **argv)
 
 				if (stop_signal != SIGTERM)
 				{
-					log_fatal("Please use either --fast or --immediate, not both");
+					log_fatal("Please use only one of either "
+							  " --sigkill, --fast or --immediate");
 					exit(EXIT_CODE_BAD_ARGS);
 				}
 				stop_signal = SIGKILL;
--- a/src/bin/pg_autoctl/parsing.c
+++ b/src/bin/pg_autoctl/parsing.c
@@ -21,6 +21,9 @@
 #include "parsing.h"
 #include "string_utils.h"
 
+static bool parse_controldata_field_dbstate(const char *controlDataString,
+											DBState *state);
+
 static bool parse_controldata_field_uint32(const char *controlDataString,
 										   const char *fieldName,
 										   uint32_t *dest);
@@ -221,7 +224,9 @@ bool
 parse_controldata(PostgresControlData *pgControlData,
 				  const char *control_data_string)
 {
-	if (!parse_controldata_field_uint32(control_data_string,
+	if (!parse_controldata_field_dbstate(control_data_string,
+										 &(pgControlData->state)) ||
+		!parse_controldata_field_uint32(control_data_string,
 										"pg_control version number",
 										&(pgControlData->pg_control_version)) ||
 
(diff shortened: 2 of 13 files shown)


DimCitus:
> Now pg_rewind can fail for other reasons:

Can't reproduce anymore after a merge with current master. I suppose it could be about the fix for the replication slots on the standby nodes per #676.

DimCitus:
The way I have been testing this is:

  1. `make NODES=3 cluster`
  2. `pg_autoctl stop --sigkill --pgdata node1`
  3. wait until a new primary has been promoted (30s for noticing primary is unavailable, 30s in stop replication)
  4. `pg_autoctl run` in the node1 shell

src/bin/pg_autoctl/primary_standby.c
```diff
@@ -1036,6 +1043,209 @@ primary_rewind_to_standby(LocalPostgresServer *postgres)
 }
 
 
+/*
+ * postgres_maybedo_crash_recovery implements a round of Postgres crash
```
JelteF, line 1047:
```suggestion
 * postgres_maybe_do_crash_recovery implements a round of Postgres crash
```

src/bin/pg_autoctl/cli_service.c
```diff
@@ -330,6 +332,26 @@ cli_getopt_pgdata_and_mode(int argc, char **argv)
...
+				{
+					log_fatal("Option --sigkill is only available in debug "
+							  "environments");
+					exit(EXIT_CODE_BAD_ARGS);
+					break;
+				}
+
+				if (stop_signal != SIGTERM)
+				{
+					log_fatal("Please use either --fast or --immediate, not both");
```
JelteF, line 348:
```suggestion
					log_fatal("Please use either --sigkill, --fast or --immediate, not multiple of them");
```

src/bin/pg_autoctl/parsing.c
```diff
@@ -220,7 +223,9 @@ bool
 parse_controldata(PostgresControlData *pgControlData,
 				  const char *control_data_string)
 {
-	if (!parse_controldata_field_uint32(control_data_string,
+	if (!parse_controldata_field_dbstate(control_data_string,
+										 &(pgControlData->state)) ||
```
JelteF, line 227: Extra newline to be consistent with the rest of this function
```suggestion
										 &(pgControlData->state)) ||
										 
```

src/bin/pg_autoctl/parsing.c
```diff
@@ -243,6 +248,66 @@ parse_controldata(PostgresControlData *pgControlData,
...
+	sformat(regex, BUFSIZE, "Database cluster state: *(.*)$");
+
+	char *match = regexp_first_match(controlDataString, regex);
+
+	if (match == NULL)
+	{
+		return false;
+	}
+
+	if (streq(match, "starting up"))
```
JelteF, line 271: Does this still work if you use a non English localization for PG? i.e. if you use French, does it still store the english strings in there? I mainly worry about this because of your other localization related stuff.

src/bin/pg_autoctl/pgctl.c
```diff
@@ -1455,11 +1469,16 @@ pg_ctl_postgres(const char *pg_ctl, const char *pgdata, int pgport,
 	args[argsIndex++] = "-p";
 	args[argsIndex++] = (char *) intToString(pgport).strValue;
 
-	if (!IS_EMPTY_STRING_BUFFER(listen_addresses))
```
JelteF, line 1458: don't we still need this check too?

src/bin/pg_autoctl/primary_standby.c
```diff
@@ -1036,6 +1043,209 @@ primary_rewind_to_standby(LocalPostgresServer *postgres)
...
+		return false;
+	}
+
+	/*
+	 * We know that Postgres is not running thanks to pg_ctl_status, and we
+	 * just grabbed the output from pg_controldata. We can now implement the
+	 * same pre-condition checks as in Postgres pg_rewind.c.
+	 */
+	if (pgSetup->control.state != DB_SHUTDOWNED &&
+		pgSetup->control.state != DB_SHUTDOWNED_IN_RECOVERY)
```
JelteF, line 1096: If we invert this check, then we can use an early `return true`.

src/bin/pg_autoctl/primary_standby.c
```diff
@@ -1036,6 +1043,209 @@ primary_rewind_to_standby(LocalPostgresServer *postgres)
...
+				log_fatal("BUG: returned from service_keeper_runprogram()");
+				exit(EXIT_CODE_INTERNAL_ERROR);
+			}
+
+			default:
+			{
+				/* wait until postgres crash recovery is done */
+				for (int attempts = 0;; attempts++)
+				{
+					int timeout = 30;
```
JelteF, line 1172: These 30 seconds seem rather arbitrary, for a busy database this is probably not enough to do the recovery. I think it would be better if we would wait as long as the postgres process that we spawned is still running (and possibly making progress in the recovery).

JelteF (changes requested):

src/bin/pg_autoctl/pgsetup.c
```diff
@@ -1959,3 +1961,48 @@ pgsetup_hba_level_to_string(HBAEditLevel hbaLevel)
...
+ * dbstateToString returns a string from a pgControlFile state enum.
+ */
+const char *
+dbstateToString(DBState state)
+{
+	switch (state)
+	{
+		case DB_STARTUP:
+		{
+			return _("starting up");
```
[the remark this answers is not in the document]
DimCitus: Oops, that's right. Copy/Paste was a little too fast on that one. Fixed by removing the get text “decorations”. We don't need localisation or at least this PR isn't the place to introduce it...

src/bin/pg_autoctl/primary_standby.c
DimCitus: Fixed now, thanks!

src/bin/pg_autoctl/cli_service.c
DimCitus: I would rather keep `--sigkill` undocumented as I can't imagine a use case in production around it.

src/bin/pg_autoctl/parsing.c
DimCitus: Fixed.

src/bin/pg_autoctl/pgctl.c
DimCitus: I don't think we do, no. The idea is to drive the `-h` argument from the extra new `listen` boolean argument, rather than second guessing the intentions of the caller from the listen_addresses string buffer value. We could protect against a caller mistakenly passing in listen to true with an empty string buffer for listen_addresses though. That would be a bug. I have added such a protection now.

src/bin/pg_autoctl/primary_standby.c
DimCitus: It is meant to mimic Postgres pg_rewind test that reads:

``` C
	if (!no_ensure_shutdown &&
		ControlFile_target.state != DB_SHUTDOWNED &&
		ControlFile_target.state != DB_SHUTDOWNED_IN_RECOVERY)
```

Note that the `--no-ensure-shutdown` is new in Postgres 13, and I just noticed that now.

src/bin/pg_autoctl/primary_standby.c
DimCitus: We have an infinite loop around a 30s timeout, so I believe we're ok here.

src/bin/pg_autoctl/cli_service.c
JelteF: That makes sense, but this is the error message when already using `--sigkill`. So teh user already knows about it.

src/bin/pg_autoctl/parsing.c
DimCitus: Yes it does, we force LANGUAGE=C when running pg_controldata.

src/bin/pg_autoctl/cli_service.c
DimCitus: Ah, yes, indeed, good point. Fixed now.

JelteF:
can you fix the merge conflict? then I can try this out locally. The code seems good.

JelteF (approved):
I think this is okay to merge, i.e. it works fine. But there's one thing you might want to look at, either before merging or in a follow up PR. During the crash recovery it shows some ERROR about not being able to connect. This seems to be fixed resolved right after:

```
14:43:40 28684 INFO  Rewinding PostgreSQL to follow new primary node 1 "node1" (localhost:5501)
14:43:40 28684 INFO  Postgres needs to enter crash recovery before pg_rewind.
14:43:40 28684 INFO  Creating the standby signal file at "/home/jelte/work/pg_auto_failover/tmux/node3/standby.signal", and replication setup at "/home/jelte/work/pg_auto_failover/tmux/node3/postgresql-auto-failover-standby.conf"
14:43:40 28684 INFO  Contents of "/home/jelte/work/pg_auto_failover/tmux/node3/postgresql-auto-failover-standby.conf" have changed, overwriting
14:43:40 28717 INFO   /home/jelte/.pgenv/pgsql-12.4/bin/postgres -D /home/jelte/work/pg_auto_failover/tmux/node3 -p 5503 -h ''
14:43:42 28684 INFO  Postgres is now serving PGDATA "/home/jelte/work/pg_auto_failover/tmux/node3" on port 5503 with pid 28717
14:43:42 28684 ERROR Connection to database failed: could not connect to server: Connection refused
14:43:42 28684 ERROR    Is the server running on host "localhost" (127.0.0.1) and accepting
14:43:42 28684 ERROR    TCP/IP connections on port 5503?
14:43:42 28684 ERROR Failed to connect to local database at "port=5503 dbname=demo host=localhost", see above for details
14:43:42 28684 ERROR Failed to get Postgres metadata, continuing
14:43:42 28684 INFO   /home/jelte/.pgenv/pgsql-12.4/bin/pg_rewind --target-pgdata /home/jelte/work/pg_auto_failover/tmux/node3 --source-server 'application_name=pgautofailover_standby_3 host=localhost port=5501 user=pgautofailover_replicator dbname=postgres sslmode=require' --progress
14:43:42 28684 INFO  pg_rewind: connected to server
```

PR #658, opened 2021-03-30
Fix/identify system before pg basebackup

DimCitus:
That way we have a better handling of the connection failures.

### DIFF
--- a/src/bin/lib/subcommands.c/runprogram.h
+++ b/src/bin/lib/subcommands.c/runprogram.h
@@ -628,7 +628,24 @@ snprintf_program_command_line(Program *prog, char *buffer, int size)
 
 	for (index = 0; prog->args[index] != NULL; index++)
 	{
-		int n = snprintf(currentPtr, remainingBytes, " %s", prog->args[index]);
+		int n;
+
+		/* replace an empty char buffer with '' */
+		if (prog->args[index][0] == '\0')
+		{
+			n = snprintf(currentPtr, remainingBytes, " ''");
+		}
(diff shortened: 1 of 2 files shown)


src/bin/lib/subcommands.c/runprogram.h
```diff
@@ -628,7 +628,24 @@ snprintf_program_command_line(Program *prog, char *buffer, int size)
...
-		int n = snprintf(currentPtr, remainingBytes, " %s", prog->args[index]);
+		int n;
+
+		/* replace an empty char buffer with '' */
+		if (prog->args[index][0] == '\0')
+		{
+			n = snprintf(currentPtr, remainingBytes, " ''");
+		}
+		/* single-quote are needed when argument contains special chars */
+		else if (strchr(prog->args[index], ' ') != NULL ||
```
JelteF, line 639: This way of escaping seems a scary from a security point of view. I believe this function is only used to print the command to stdout. If that's the case then it's fine, but if this is used to actually execute the command then this should probably be done in some other way.

JelteF (approved):

src/bin/lib/subcommands.c/runprogram.h
DimCitus: Yeah this is only for logging purposes. The idea behind this change is mostly so that it's easy to see connection strings in the form of 'host=1.2.3.4 user=foo ...' as a single argument in calls to pg_basebackup or pg_rewind, and then so that chances that you can just copy & paste the log line verbatim are better.

Issue #654, opened 2021-03-29
Allow dropping nodes in all cases, create a primary from a non-candidate


DimCitus:
At the moment we see in https://github.com/citusdata/pg_auto_failover/blob/master/src/monitor/node_active_protocol.c#L1044 that we refuse to remove the primary node in the monitor when we don't have a candidate for promotion. It would be better to allow for that node removal, and this requires the ability to create a primary node from existing secondary node(s).

Also, when adding a primary from secondary nodes is possible, this would allow pg_auto_failover setups with only non-candidate nodes. 

In passing, we might want to improve the REPORT_LSN state to allow for the secondary node to be running, probably with fetching the current LSN and marking it as the target LSN in recovery mode while in the REPORT_LSN state.

DimCitus:
Fixed in #661.

PR #661, opened 2021-04-07
Allow manual promotion of nodes with candidate priority zero.

DimCitus:
When promoting a node that is not setup as a candidate for failover, we
increment its candidate priority, perform the failover, and then reset its
candidate priority, as usual.

Allowing this mode of operation allows operating a pg_auto_failover cluster
in a fully manual way. When all the nodes have candidate priority set to
zero, it is still possible to manual trigger a promotion and target any
node.

This is a pre-requisite to being able to drop a node in all cases, including
when it's the primary node and there is no candidate for failover. Our
design is complete when it's always possible to run `pg_autoctl drop node`.

See #654


DimCitus:
At the moment the following scenario requires some more fixes, client side only:

  1. make NODES=3 NODES_PRIOS=0 cluster
  2. pg_autoctl perform failover
  3. pg_autoctl set node candidate-priority --name node3 1

In that case, we get 3 nodes all with candidate priority 0. In step 2, thanks to the PR, the failover is triggered, leaving two nodes in REPORT_LSN state and the old primary in DRAINING state. New in this PR, it is then possible to tweak a node's candidate priority, immediately triggering the end of the failover process.

The limitation we still need to fix (and I plan to include that later in this PR) is that the command in step 3 (set node candidate priority) doesn't recognise when the setting has been set. In this PR we don't always go through APPLY_SETTINGS anymore. This intermediary step is only needed when we have a stable primary node in PRIMARY or WAIT_PRIMARY...

DimCitus:
> The limitation we still need to fix (and I plan to include that later in this PR) is that the command in step 3 (set node candidate priority) doesn't recognise when the setting has been set. In this PR we don't always go through APPLY_SETTINGS anymore. This intermediary step is only needed when we have a stable primary node in PRIMARY or WAIT_PRIMARY...

Fixed now.

src/monitor/node_active_protocol.c
```diff
@@ -1352,7 +1409,8 @@ perform_promotion(PG_FUNCTION_ARGS)
 	 * would still be able to implement a failover given another secondary node
 	 * being around.
 	 */
-	if (!IsCurrentState(currentNode, REPLICATION_STATE_SECONDARY))
+	if (!IsCurrentState(currentNode, REPLICATION_STATE_SECONDARY) &&
+		!IsCurrentState(currentNode, REPLICATION_STATE_REPORT_LSN))
```
JelteF, line 1413: Comment above should be updated

src/monitor/node_active_protocol.c
```diff
@@ -1873,38 +1924,43 @@ set_node_candidate_priority(PG_FUNCTION_ARGS)
...
+		/*
+		 * If we allow setting changes during APPLY_SETTINGS we open the door
+		 * for race conditions where we can't be sure that the latest changes
+		 * have been applied.
+		 *
+		 * If we don't currently have a primary node anyway, we can just
+		 * proceed with the change.
+		 */
+		if (primaryNode &&
+			!IsCurrentState(primaryNode, REPLICATION_STATE_APPLY_SETTINGS))
```
JelteF, line 1936: What happens now if we change the settings in WAIT_PRIMARY? Doesn't this have the same issue as doing it during APPLY_SETTINGS?

src/monitor/node_active_protocol.c
```diff
@@ -2015,38 +2071,43 @@ set_node_replication_quorum(PG_FUNCTION_ARGS)
...
-		if (primaryNode == NULL)
+		/*
+		 * If we allow setting changes during APPLY_SETTINGS we open the door
+		 * for race conditions where we can't be sure that the latest changes
+		 * have been applied.
+		 *
+		 * If we don't currently have a primary node anyway, we can just
+		 * proceed with the change.
+		 */
+		if (primaryNode &&
```
JelteF, line 2082: Maybe move this check to a function, since it's used when setting both replication_quorum and candidate_priority

src/monitor/node_metadata.c
```diff
@@ -1521,6 +1521,7 @@ CanInitiateFailover(ReplicationState state)
 {
 	return state == REPLICATION_STATE_SINGLE ||
 		   state == REPLICATION_STATE_PRIMARY ||
+		   state == REPLICATION_STATE_WAIT_PRIMARY ||
```
JelteF, line 1524: Probably not needed anymore after the discussed wait-primary changes.

src/monitor/node_metadata.h
```diff
@@ -90,6 +90,7 @@ typedef enum SyncState
  * candidate.
  */
 #define MAX_USER_DEFINED_CANDIDATE_PRIORITY 100
+#define CANDIDATE_PRIORITY_INCREMENT (MAX_USER_DEFINED_CANDIDATE_PRIORITY + 1)
```
JelteF, line 93: What was this `+ 1` needed for?

src/bin/pg_autoctl/cli_do_monitor.c
```diff
@@ -263,6 +273,76 @@ cli_do_monitor_get_other_nodes(int argc, char **argv)
 }
 
 
+/*
+ * cli_do_monitor_get_failover_count contacts the pg_auto_failover monitor and
+ * retrieves the current count of failover candidate nodes.
+ */
+static void
+cli_do_monitor_get_failover_count(int argc, char **argv)
```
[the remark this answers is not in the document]
DimCitus: Good point, done.

src/bin/pg_autoctl/cli_do_monitor.c
```diff
@@ -63,6 +64,14 @@ static CommandLine monitor_get_other_nodes_command =
 				 cli_getopt_pgdata,
 				 cli_do_monitor_get_other_nodes);
 
+static CommandLine monitor_get_candidate_count_command =
+	make_command("candidate-count",
```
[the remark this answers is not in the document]
DimCitus: None of the `pg_autoctl do` commands got the memo about `--monitor`, they all need `--pgdata`... while we could modernise the do commands, I feel like that would be material for another PR.

src/bin/pg_autoctl/cli_perform.c
```diff
@@ -252,11 +252,27 @@ cli_perform_failover(int argc, char **argv)
...
+										   config.groupId,
+										   &failoverCandidateCount))
+	{
+		/* errors have already been logged */
+		exit(EXIT_CODE_MONITOR);
+	}
+
+	if (failoverCandidateCount == 0)
+	{
+		targetState = WAIT_PRIMARY_STATE;
```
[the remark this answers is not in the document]
DimCitus: Right, done.

src/monitor/node_active_protocol.c
DimCitus: Done now.

src/monitor/node_active_protocol.c
DimCitus: Not really, not anymore. Nowadays the APPLY_SETTINGS to either WAIT_PRIMARY or PRIMARY is computed in the `ProceedGroupStateForPrimaryNode` function in the Group FSM, in such a way that the state that was observed before entering APPLY_SETTINGS is not important anymore. Things might change either during the operation, or because of the new settings (e.g. number sync standbys).

The only case we don't know how to handle properly is concurrent calls to change the replication settings, and that mostly is a client side problem, where we want to return from the `pg_autoctl` to the shell (or caller) only when the setting is known to have been applied.

src/monitor/node_active_protocol.c
DimCitus: I'm not sure, I feel like it's fair game to have a simple two step conditions copy/pasted in here.

src/monitor/node_metadata.c
DimCitus: That right, fixed.

src/monitor/node_metadata.h
DimCitus: It used to be that we would only allow failover when candidate priority was in the range 1..100, and now we allow it when it's in the range 0..100. We want to get from 0 to 101, and then from 101 back to 0...

src/bin/pg_autoctl/cli_perform.c
```diff
@@ -251,8 +251,9 @@ cli_perform_failover(int argc, char **argv)
 	KeeperConfig config = keeperOptions;
 	Monitor monitor = { 0 };
 
-	char *channels[] = { "state", NULL };
+	NodeState targetState = PRIMARY_STATE;
```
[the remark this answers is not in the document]
DimCitus: Good point, done that way now.

src/bin/pg_autoctl/fsm_transition.c
```diff
@@ -1374,3 +1392,42 @@ fsm_follow_new_primary(Keeper *keeper)
...
+	/* get the primary node to follow */
+	if (!keeper_get_most_advanced_standby(keeper, &upstreamNode))
+	{
+		log_error("Failed to initialise from the most advanced standby node, "
+				  "see above for details");
+		return false;
+	}
+
+	if (!standby_init_replication_source(postgres,
+										 &upstreamNode,
```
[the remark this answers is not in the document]
DimCitus: That's a good catch, done that way now.

src/bin/pg_autoctl/monitor.c
```diff
@@ -2591,6 +2591,54 @@ printFormationURI(void *ctx, PGresult *result)
...
+								  char *formation, int groupId,
+								  int *failoverCandidateCount)
+{
+	SingleValueResultContext context = { { 0 }, PGSQL_RESULT_INT, false };
+	PGSQL *pgsql = &monitor->pgsql;
+	char *sql =
+		"select count(node.candidatepriority) "
+		"       filter(where node.candidatepriority > 0) "
+		"       as failover_candidate_count "
+		"  from pgautofailover.get_nodes('default', 0) as gn "
```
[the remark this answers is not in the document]
DimCitus: And group should be the groupId. Somehow the argument array was prepared, but positional parameters not used and hardcoded. Fixed now, thanks, good catch indeed!

PR #666, opened 2021-04-08
Use PGDATA owner for the systemd service file.

DimCitus:
The current user might be root or another interactive system user so that
it's possible to use sudo to write the systemd unit files. Typically, the
"postgres" user is not granted sudo privileges, and that's a good thing.

With this changes, if pg_autoctl show systemd is used from the root account
with a --pgdata created by the "postgres" user, we expect to obtain the
proper systemd unit file.

### DIFF
--- a/src/bin/pg_autoctl/systemd_config.c
+++ b/src/bin/pg_autoctl/systemd_config.c
@@ -10,6 +10,7 @@
 #include <pwd.h>
 #include <stdbool.h>
 #include <sys/types.h>
+#include <sys/stat.h>
 #include <unistd.h>
 
 #include "postgres_fe.h"
@@ -83,7 +84,6 @@
 void
 systemd_config_init(SystemdServiceConfig *config, const char *pgdata)
 {
-	char *user = pg_setup_get_username(&(config->pgSetup));
 	IniOption systemdOptions[] = SET_INI_OPTIONS_ARRAY(config);
 
 	/* time to setup config->pathnames.systemd */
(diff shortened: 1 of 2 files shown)


src/bin/pg_autoctl/systemd_config.c
```diff
@@ -108,8 +124,8 @@ systemd_config_init(SystemdServiceConfig *config, const char *pgdata)
 	sformat(config->EnvironmentPGDATA, BUFSIZE,
 			"'PGDATA=%s'", config->pgSetup.pgdata);
 
-	/* adjust the user to the current system user */
-	strlcpy(config->User, user, NAMEDATALEN);
+	/* adjust the user to the owner of PGDATA */
+	strlcpy(config->User, pw->pw_name, NAMEDATALEN);
```
JelteF, line 128: Maybe also set `config->Group`?

JelteF (approved):

src/bin/pg_autoctl/systemd_config.c
DimCitus: Mmm, there's no “Group” in `SystemdServiceConfig` right?

JelteF: I mean it's our own type right? So we can add it. I meant it more like, maybe we should set both `User` and `Group` in the final systemd config.

DimCitus: Oh yeah, I didn't know you could set the group for systemd and then missed your comment entirely, sorry about that. About the group, given Unix design where a user can belong to many groups and a directory only to one single group, and given how Postgres is packaged for debian and some other OS, I'm not sure it would be a good idea for us to do that, or at least I have to dive into more details.

A common security hardening guideline consists of using your own Postgres system super-user that's not named “postgres” on the system, and belongs to the “postgres” group. What would be the Systemd Group to chose there? What would be the PGDATA group set to when we call initdb in such a setup? I just don't know yet... if you think we should spend time on that aspect, I will and then open a new PR with my findings...

PR #671, opened 2021-04-19
Fix memory leak allocated by createPQExpBuffer()

gfphoenix78:
Fix memory leak allocated by createPQExpBuffer()


DimCitus (approved):
Thanks for your contribution!

PR #672, opened 2021-04-20
Add rules for the monitor to pg_hba.conf on the secondary

gfphoenix78:
When we enable pg_auto_failover for an existing postgres
cluster, the pg_hba.conf on standby hasn't the following 2 rules
compared with that of the primary.
```
host(ssl) all "pgautofailover_monitor" _host_ trust # Auto-generated by pg_auto_failover
host(ssl) "postgres" _user_ _host_ trust # Auto-generated by pg_auto_failover
```

As a result, the health check connection for the secondary
continuously generates FATAL message:
```
FATAL:  no pg_hba.conf entry for host "10.123.119.3", \
user "pgautofailover_monitor", database "postgres", SSL off
```

This patch adds the access rule for the monitor on secondary,
so the health check connection will gracefully exit.


DimCitus (changes requested):
Hi, thanks for working on pg_auto_failover!

I have a general design question about your PR, and then an implementation level remark.

  1. we limit editing HBA entries to situation when we know we have a new node in the list of other nodes in the same group, i.e. we implement cache invalidation rather than “brute force editing” at each loop. Your PR does edit the HBA at each loop to check for the monitor, which is supposed to be static. I suppose it would be better to ensure the monitor's HBA entries in `fsm_init_standby` instead.
  2. When it comes to editing the HBA file with the monitor, I believe we could create a node array with just the monitor in there and use `pghba_ensure_host_rules_exist` then...

gfphoenix78:
Sounds reasonable, I'll take some time to update the PR.

Issue #594, opened 2021-02-18
Unrecoverable state after restart due to missing replication slots


AnshumanRohella:
### Architecture

* 3 node setup with 1 primary and 2 secondary.
* One of the secondary nodes has a candidate-priority 0.
*  Reference:
Node1 (primary) Node2 (Secondary,priority=50) Node3 (Secondary, priority=0)

### Details
* When all the data nodes are restarted together Node3 may have the latest LSN position. However, it cannot be selected for promotion, so,  Node2 is put in "fast_forward" state to stream the latest LSN from Node3 ( latest LSN but 0 candidate-priority)
* However, Node2 keeps waiting for the latest LSN because the replication_slot is not created on Node3. 

#### State of the formation after restarting the nodes a couple of times. 
```
Every 0.2s: pg_autoctl show state
 Name |  Node |      Host:Port |       LSN | Reachable |       Current State |      Assigned State                        
------+-------+----------------+-----------+-----------+---------------------+--------------------
node1 |     1 | localhost:9001 | 0/A020000 |        no |            draining |            draining                            
node2 |     2 | localhost:9002 | 0/A01FCE8 |       yes |          report_lsn |        fast_forward
node3 |     3 | localhost:9003 | 0/A01FD58 |       yes |          report_lsn |          report_lsn
```
#### State of Node2 ( promotion candidate ). It is waiting for the latest LSN as expected.

```
07:13:46 24333 INFO  Postgres recovery is at LSN 0/A01FCE8, waiting for LSN 0/A01FD58                                         
07:13:47 24333 INFO  Postgres recovery is at LSN 0/A01FCE8, waiting for LSN 0/A01FD58                                         
07:13:48 24333 INFO  Postgres recovery is at LSN 0/A01FCE8, waiting for LSN 0/A01FD58                                         
07:13:49 24333 INFO  Postgres recovery is at LSN 0/A01FCE8, waiting for LSN 0/A01FD58                                    
07:13:50 24333 INFO  Postgres recovery is at LSN 0/A01FCE8, waiting for LSN 0/A01FD58                                         
07:13:51 24333 INFO  Postgres recovery is at LSN 0/A01FCE8, waiting for LSN 0/A01FD58                                         
07:13:52 24333 INFO  Postgres recovery is at LSN 0/A01FCE8, waiting for LSN 0/A01FD58  
```

[...]

PR #676, opened 2021-04-22
Fix parsing pg version string for replication slots support on standby.

DimCitus:
The previous coding was wrong on several grounds:

First, pg_control_version only changes when there is a catalog
incompatibility in a given Postgres major version. For instance Postgres
12.6 still uses pg_control_verison 1201.

Second, the pg_setup_standby_slot_supported() function logic was missing
some cases and would return false even though replication slots are
supported on the standby nodes.

See #594

### DIFF
--- a/src/bin/pg_autoctl/pgsetup.c
+++ b/src/bin/pg_autoctl/pgsetup.c
@@ -24,8 +24,9 @@
 
 #include "defaults.h"
 #include "env_utils.h"
-#include "pgctl.h"
 #include "log.h"
+#include "parsing.h"
+#include "pgctl.h"
 #include "signals.h"
 #include "string_utils.h"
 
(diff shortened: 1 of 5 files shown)


src/bin/pg_autoctl/pgsetup.c
```diff
@@ -1886,16 +1905,22 @@ pg_setup_standby_slot_supported(PostgresSetup *pgSetup, int logLevel)
...
+
 	/* Starting with Postgres 13, all versions are known to have the bug fix */
-	if (pgSetup->control.pg_control_version >= 1300)
+	if (pg_version >= 1300)
 	{
 		return true;
 	}
 
 	/* should not happen */
 	log_debug("BUG in pg_setup_standby_slot_supported(%d): "
```
JelteF, line 1921: Maybe make this `log_warn`, so that people will report this bug quicker?
```suggestion
	log_warn("BUG in pg_setup_standby_slot_supported(%d): "
```

JelteF (approved):

src/bin/pg_autoctl/pgsetup.c
DimCitus: All the other BUG messages are log_debug, so that's what I'm using here. Also we use that function at every loop in the keeper, so that's going to be a lot of warning messages, and the users can't do anything about it (other than all report the same bug). Maybe we could add a warning at startup?

JelteF: Yeah logging this every loop seems even worse. If it's easy to do at startup (1-3 line changes) I'd say let's do it. Otherwise I don't think it's important enough.

DimCitus: On the other hand we already have `pg_autoctl config check` to implement an explicit check and warning about that case. When we suspect something is wrong with replication slots maintenance on the standby nodes we can always ask to run the check command.

```
$ pg_autoctl config check --pgdata node1
11:28:39 23550 INFO  Postgres setup for PGDATA "/Users/dim/dev/MS/pg_auto_failover/tmux/node1" is ok, running with PID 5501 and port 23446
11:28:39 23550 INFO  Connection to local Postgres ok, using "port=5501 dbname=demo host=/tmp"
11:28:39 23550 INFO  Postgres configuration settings required for pg_auto_failover are ok
11:28:39 23550 INFO  Postgres version "12.6" allows using replication slots on the standby nodes
...
```

PR #680, opened 2021-04-26
When receiving SIGTERM, enable maintenance.

### LINKED ISSUE #655: Improve pg_autoctl shutdown sequence
Both systemd and Kubernetes use a shutdown sequence that sends a SIGTERM signal then later a stronger one. See https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-termination for instance, where the default is to provide pods with a 30s grade period.

At the moment pg_autoctl is not smart about the grace period. It could be made to switch the current node to MAINTENANCE mode and then have a way to automatically re-join the cluster at the next start-up.

DimCitus:
That way when the node shuts down, the group FSM has done the necessary
steps already. If the node was a primary, another node has been elected to
take over already.

In most cases (systemd, docker, kubenetes) the way we know that a local node
has been asked to get down is the SIGTERM signal, after all.

Fixes #655.


JelteF:
There should be a test added that checks for this new behaviour.

JelteF:
When trying this out locally with `pg_autoctl stop --pgdata node1` (the primary). I got this error in the logs:
```
12:26:10 18285 INFO  pg_autoctl received signal SIGTERM, terminating
12:26:13 18288 INFO  Monitor assigned new state "maintenance"
12:26:13 18288 INFO  FSM transition from "prepare_maintenance" to "maintenance": Setting up Postgres in standby mode for maintenance operations
12:26:13 18288 INFO  Creating the standby signal file at "/home/jelte/work/pg_auto_failover/tmux/node1/standby.signal", and replication setup at "/home/jelte/work/pg_auto_failover/tmux/node1/postgresql-auto-failover-standby.conf"
12:26:13 18288 INFO  Transition complete: current state is now "maintenance"
12:26:13 18288 INFO  Shutdown sequence complete: reached state "maintenance"
12:26:13 18285 INFO  Service leader with pid 18288 has terminated, now stopping other services
12:26:13 18285 ERROR Failed to send signal SIGTERM to service postgres with pid 18287
12:26:13 18285 INFO  Stop pg_autoctl
```

DimCitus:
> When trying this out locally with `pg_autoctl stop --pgdata node1` (the primary). I got this error in the logs:

I can't reproduce at the moment. I suppose that's because the Postgres service was already shut down by the time the supervisor wanted to signal it... a strange race condition, maybe we should integrate some of the logic in the main `waitpid` loop, but I'm not sure there is a good solution there (race condition between `waitpid` and `kill` seems kind of impossible to avoid in general, we could `kill -0 <pid>` to check if the pid still exists before sending the actual signal but that's still open to a race condition of course).

src/bin/pg_autoctl/service_keeper.c
```diff
@@ -321,8 +334,40 @@ keeper_node_active_loop(Keeper *keeper, pid_t start_pid)
...
 		}
 
-		if (asked_to_stop || asked_to_stop_fast || asked_to_quit)
+		/*
+		 * Upon receiving SIGTERM, start maintenance on the monitor, possibly
+		 * triggering a failover when the primary node is being shut down.
+		 */
+		if (asked_to_stop)
 		{
+			if (keeperState->current_role != MAINTENANCE_STATE &&
```
JelteF, line 343: If state is `PREPARE_MAINTENANCE_STATE`, then the user clearly wants maintenance to be the state on startup. So we should handle that state in some special way, probably we want to wait until the transition is completed, but don't want to indicate maintainance should be disabled after boot.

src/bin/pg_autoctl/service_keeper.c
```diff
@@ -511,6 +558,31 @@ keeper_node_active_loop(Keeper *keeper, pid_t start_pid)
...
+		 * we record this action in the Keeper State file. At startup, we then
+		 * disable maintenance again. We might have to retry disabling
+		 * maintenance should it fail.
+		 */
+		if (shutdownInMaintenance && !disabledMaintenanceAtStartup)
+		{
+			int nodeId = keeper->state.current_node_id;
+			bool mayRetry = false;
+
+			log_info("pg_autoctl node was put in maintenance as part of "
```
JelteF, line 572: When after shutting down pg_autoctl, you run `pg_autoctl disable maintenance` manually, then we get in an infinite loop here:
```
12:59:22 22037 INFO  pg_autoctl node was put in maintenance as part of shutting down, now disabling maintenance
12:59:22 22037 ERROR Monitor ERROR:  cannot stop maintenance when current state for node 1 "node1" (localhost:5501) is not "maintenance"
12:59:22 22037 ERROR Monitor DETAIL:  Current reported state is "maintenance" and assigned state is "catchingup"
12:59:22 22037 ERROR Failed to stop_maintenance of node 1 from the monitor
```

src/bin/pg_autoctl/service_keeper.c
```diff
@@ -321,8 +334,40 @@ keeper_node_active_loop(Keeper *keeper, pid_t start_pid)
...
+			if (keeperState->current_role != MAINTENANCE_STATE &&
+				(!shutdownSequenceInProgress ||
+				 !couldStartMaintenanceAtShutdown))
+			{
+				int nodeId = keeper->state.current_node_id;
+				bool mayRetry = false;
+
+				log_info("Enabling maintenance on this node for shutdown");
+
+				if (monitor_start_maintenance(monitor, nodeId, &mayRetry))
```
JelteF, line 352: This is flooding the logs really fast if the monitor is down.
```
13:05:27 22037 WARN  Lost connection.
13:05:27 22037 INFO  Enabling maintenance on this node for shutdown
13:05:27 22037 ERROR Connection to database failed: could not connect to server: Connection refused
13:05:27 22037 ERROR    Is the server running on host "localhost" (127.0.0.1) and accepting
13:05:27 22037 ERROR    TCP/IP connections on port 5500?
13:05:27 22037 ERROR Failed to connect to monitor database at "postgres://autoctl_node@localhost:5500/pg_auto_failover?sslmode=prefer", see above for details
13:05:27 22037 ERROR Failed to start_maintenance of node 1 from the monitor
```

JelteF (changes requested):

src/bin/pg_autoctl/state.c
```diff
@@ -223,6 +223,46 @@ keeper_state_create_file(const char *filename)
...
+ServiceStateToString(KeeperServiceState serviceState)
+{
+	switch (serviceState)
+	{
+		case SERVICE_STARTUP:
+		{
+			return "starting up";
+		}
+
+		case SERVICE_SHUTDOWNED:
```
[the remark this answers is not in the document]
DimCitus: I took inspiration from Postgres code at https://github.com/postgres/postgres/blob/master/src/include/catalog/pg_control.h#L85 where we have:

``` C
typedef enum DBState
{
	DB_STARTUP = 0,
	DB_SHUTDOWNED,
	DB_SHUTDOWNED_IN_RECOVERY,
	DB_SHUTDOWNING,
	DB_IN_CRASH_RECOVERY,
	DB_IN_ARCHIVE_RECOVERY,
	DB_IN_PRODUCTION
}
```

src/bin/pg_autoctl/service_keeper.c
```diff
@@ -321,8 +334,40 @@ keeper_node_active_loop(Keeper *keeper, pid_t start_pid)
...
+				}
+				else
+				{
+					/* warning: we're going to try again */
+					log_warn("Failed to start maintenance "
+							 "upon receiving SIGTERM");
+				}
+			}
+
+			keeperState->service_state = SERVICE_SHUTDOWNING;
```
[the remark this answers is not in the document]
DimCitus: That's a good point, I have done the changes required for supporting that use case in full now.

src/bin/pg_autoctl/service_keeper.c
DimCitus: Fixed too, as part of the previous comment with SIGINT during a SIGTERM induced shutdown sequence.

src/bin/pg_autoctl/service_keeper.c
DimCitus: That's the new code that handles this part now. Thanks to fixing the case when we're in prepare_maintenance at startup and want to finish the FSM transition to maintenance and immediately disable maintenance, I believe this comment is fixed too. At least I'm not sure how to reproduce.

``` C
		/*
		 * When we receive SIGTERM, we enable maintenance on the monitor, and
		 * we record this action in the Keeper State file. At startup, we then
		 * disable maintenance again. We might have to retry disabling
		 * maintenance should it fail.
		 *
		 * The shutdown sequence might also have been interrupted in
		 * wait_maintenance state, in which case we want to first finish the
		 * FSM transition to maintenance, and only later disable maintenance.
		 */
		if (shutdownInMaintenance &&
			!disabledMaintenanceAtStartup &&
			keeperState->assigned_role == MAINTENANCE_STATE &&
			keeperState->current_role == keeperState->assigned_role &&
			!needStateChange)
		{
			int nodeId = keeper->state.current_node_id;
			bool mayRetry = false;

			log_info("pg_autoctl node was put in maintenance as part of "
					 "shutting down, now disabling maintenance");
```

src/bin/pg_autoctl/service_keeper.c
```diff
@@ -395,6 +372,48 @@ keeper_node_active_loop(Keeper *keeper, pid_t start_pid)
...
+				{
+					couldStartMaintenanceAtShutdown = true;
+				}
+				else
+				{
+					/* warning: we're going to try again */
+					log_warn("Failed to start maintenance "
+							 "upon receiving SIGTERM");
+				}
+			}
```
JelteF, line 411: Do we need another `else` here that sets `keeperState->service_state = SERVICE_SHUTDOWNING`?

src/bin/pg_autoctl/service_keeper.c
```diff
@@ -350,11 +372,55 @@ keeper_node_active_loop(Keeper *keeper, pid_t start_pid)
...
```
JelteF, line 411: Do we need an else here that sets `keeperState->service_state = SERVICE_SHUTDOWNING`?

src/bin/pg_autoctl/service_keeper.c
JelteF: if (!disabledMaintenannceAtStartup && (
        keeperState->assigned_role != MAINTENANCE_STATE ||
        keeperState->assigned_role != PREPARE_MAINTENANCE_STATE ||
        keeperState->assigned_role != WAIT_MAINTENANCE_STATE ||
)) { 
    disabledMaintenannceAtStartup = true
}

JelteF: change !disabledMaintenanceAtStartup to shouldDisableMaintenanceAtStartup

src/bin/pg_autoctl/service_keeper.c
JelteF: Maybe related:
When shutting down a node that is already in maintenance mode, SIGTERM is not shutting it down quickly:
```
16:17:26 31705 INFO  pg_autoctl received signal SIGTERM, terminating
16:17:26 31705 INFO  Waiting for subprocesses to terminate.
16:17:31 31705 INFO  pg_autoctl services are still running, signaling them with SIGTERM.
16:17:31 31710 INFO  Postgres controller service received signal SIGTERM, terminating
16:17:31 31705 INFO  pg_autoctl received signal SIGTERM, terminating
16:17:36 31705 INFO  pg_autoctl services are still running, signaling them with SIGINT.
16:17:36 31705 INFO  Service leader with pid 31711 has terminated, now stopping other services
16:17:36 31705 INFO  Stop pg_autoctl
```

src/bin/pg_autoctl/service_keeper.c
DimCitus: Done now, to cover for the following test case / scenario:

  1. pg_autoctl stop
  2. pg_autoctl disable maintenance
  3. pg_autoctl run
  4. pg_autoctl enable maintenance

Step 4 would do a secondary->maintenance->secondary without the fix outlined here.

src/bin/pg_autoctl/service_keeper.c
JelteF: Even after the current fixes this issue is still present. If you enable maintenance on a node manually and then run `pg_autoctl stop` (or send SIGTERM). The process does not actually stop:
```
12:58:26 14138 INFO  FSM transition from "wait_maintenance" to "maintenance": Suspending standby for manual maintenance.
12:58:26 14138 INFO  Transition complete: current state is now "maintenance"
12:58:26 14138 INFO  New state for node 1 "node1" (localhost:5501): primary ➜ primary
12:58:35 14135 INFO  pg_autoctl received signal SIGTERM, terminating
12:58:35 14135 INFO  Waiting for subprocesses to terminate.
12:58:40 14135 INFO  pg_autoctl services are still running, signaling them with SIGTERM.
12:58:40 14137 INFO  Postgres controller service received signal SIGTERM, terminating
12:58:40 14135 INFO  pg_autoctl received signal SIGTERM, terminating
12:58:45 14135 INFO  pg_autoctl services are still running, signaling them with SIGINT.
12:58:45 14135 INFO  Service leader with pid 14138 has terminated, now stopping other services
12:58:45 14135 INFO  Stop pg_autoctl
```

On a related note, do we maybe need to increase the amount of time after which we send a SIGINT? I can imagine going to maintenance will take longer than 10 seconds.

DimCitus: Fixed now. About the 10 seconds, I'm not too sure about what would be a proper value... any input? should we reuse `timeout.listen_notifications_timeout` or introduce a new one there?

DimCitus: It turns out that sharing the timeout values from the keeper config isn't easy, and we can't change the value on reload anyway, so I have hard-coded a 30s timeout for now.

PR #681, opened 2021-04-27
Fix/debian cluster for the monitor

DimCitus:


JelteF (approved):

tests/test_debian_clusters.py
```diff
@@ -37,13 +69,39 @@ def test_001_custom_single():
...
     monitor.print_state()
 
-    node1.destroy()
 
-    # node is stuck in init state, destroy won't remove it from monitor
-    # force remove from monitor
-    monitor.run_sql_query(
-        "select pgautofailover.remove_node('%s', %s)"
-        % (str(node1.vnode.address), str(node1.port))
+def test_002_chmod_debian_data_directory():
```
JelteF, line 94: Why is this needed? No actual test is done after.

DimCitus: It's needed so that we can destroy the nodes, otherwise we have a permission denied issue on removing PGDATA. So the tests would fail in the teardown method.

PR #692, opened 2021-04-29
Also spawn valgrind for pg_autoctl subprocesses

JelteF:
This makes sure memory leaks in subprocesses are also detected
correctly.

### DIFF
--- a/src/bin/pg_autoctl/cli_do_tmux.c
+++ b/src/bin/pg_autoctl/cli_do_tmux.c
@@ -855,7 +855,7 @@ prepare_tmux_script(TmuxOptions *options, PQExpBuffer script)
  * tmux_start_server starts a tmux session with the given script.
  */
 bool
-tmux_start_server(const char *scriptName)
+tmux_start_server(const char *scriptName, const char *binpath)
 {
 	char *args[8];
 	int argsIndex = 0;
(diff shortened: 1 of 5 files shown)


src/bin/pg_autoctl/cli_do_tmux.c
```diff
@@ -804,8 +808,10 @@ prepare_tmux_script(TmuxOptions *options, PQExpBuffer script)
 
 		tmux_add_send_keys_command(script,
 								   "PG_AUTOCTL_DEBUG=1 "
+								   "PG_AUTOCTL_DEBUG_BIN_PATH=%s "
```
DimCitus, line 811: I'm not sure to follow that one. Can't we just use the same `pg_autoctl` path as in `pg_autoctl_argv0`, which should be `src/tools/pg_autoctl.valgrind` at this point right?

JelteF: I removed these specific lines, because it's already set for the whole tmux session now. 

This change in general is needed, because `pg_autoctl_argv0` is the argument that valgrind gets `src/tools/../src/bin/pg_autoctl/pg_autoctl`, not valgrind binary itself. I expect this is the case because valgrind modifies argv, before starting the process.

DimCitus (approved):

src/bin/pg_autoctl/cli_do_tmux.c
DimCitus: okay, thanks for adding those details, it was quite a mystery to me otherwise!

JelteF: added a comment to the code explaining it too.

Issue #679, opened 2021-04-26
memory leak in process pg_autoctl: node active


RekGRpth:
memory leak in process pg_autoctl: node active
over 2 Gb RES by one day!

RekGRpth:
pg_autoctl from git in docker alpine

JelteF:
Which commit did you use from git? There have been a few memory leak fixes in the past couple of weeks, but they have not been released yet.

RekGRpth:
> Which commit did you use from git?

last

RekGRpth:
> There have been a few memory leak fixes in the past couple of weeks

some times ago everything is been ok

RekGRpth:
since 25 April process pg_autoctl: node active eat over 9 Gb!

JelteF:
Thanks for the report, we are working through a list of valgrind issues in the linked PR.

RekGRpth:
I don't use passwords in uri

JelteF:
The leak was in checking if there was a password in the URL, it doesn't matter if there is or is not. So you would still be affected by this leak.

RekGRpth:
about hour the size of process pg_autoctl: node active is about 7 Mb

RekGRpth:
again memory leak in process pg_autoctl: node active
over 415 Mb RES by 5 day!

DimCitus:
> again memory leak in process pg_autoctl: node active
> over 415 Mb RES by 5 day!

Thanks for re-opening the issue, I just did another round of valgrind checks and we fixed a couple of leaks that way. The current approach would be to run those valgrind checks at release preparation time. Please keep testing and re-opening either this PR or another one when you witness memory leaks though, that's appreciated!

PR #693, opened 2021-04-29
Fix a memory leak in BuildNodesArrayValues.

DimCitus:
See #679

### DIFF
--- a/src/bin/pg_autoctl/pgsql.c
+++ b/src/bin/pg_autoctl/pgsql.c
@@ -1491,12 +1491,21 @@ pgsql_drop_replication_slot(PGSQL *pgsql, const char *slotName)
  * We return how many parameters we filled in paramTypes and paramValues from
  * the nodeArray.
  */
+typedef struct nodesArraysValuesParams
+{
+	int count;
+	Oid types[NODE_ARRAY_MAX_COUNT * 2];
+	char *values[NODE_ARRAY_MAX_COUNT * 2];
+	char data[NODE_ARRAY_MAX_COUNT * 2][PG_LSN_MAXLENGTH];
+} nodesArraysValuesParams;
+
+
 static int
 BuildNodesArrayValues(NodeAddressArray *nodeArray,
-					  Oid *paramTypes, char **paramValues,
+					  nodesArraysValuesParams *sqlParams,
 					  char *values, int size)
 {
-	char buffer[BUFSIZE];
+	char buffer[BUFSIZE] = { 0 };
 	int nodeIndex = 0;
 	int paramIndex = 0;
 	int valuesIndex = 0;
@@ -1514,11 +1523,18 @@ BuildNodesArrayValues(NodeAddressArray *nodeArray,
 		int idParamIndex = paramIndex;
 		int lsnParamIndex = paramIndex + 1;
 
-		paramTypes[idParamIndex] = INT4OID;
-		paramValues[idParamIndex] = strdup(nodeIdString.strValue);
+		sqlParams->types[idParamIndex] = INT4OID;
 
-		paramTypes[lsnParamIndex] = LSNOID;
-		paramValues[lsnParamIndex] = node->lsn;
+		strlcpy(sqlParams->data[idParamIndex],
+				nodeIdString.strValue,
+				PG_LSN_MAXLENGTH);
+
+		sqlParams->values[idParamIndex] = sqlParams->data[idParamIndex];
+
+		sqlParams->types[lsnParamIndex] = LSNOID;
+		strlcpy(sqlParams->data[lsnParamIndex], node->lsn, PG_LSN_MAXLENGTH);
+
+		sqlParams->values[lsnParamIndex] = sqlParams->data[lsnParamIndex];
 
 		valuesIndex += sformat(buffer + valuesIndex, BUFSIZE - valuesIndex,
 							   "%s($%d, $%d%s)",
@@ -1542,6 +1558,8 @@ BuildNodesArrayValues(NodeAddressArray *nodeArray,
 		paramIndex += 2;
 	}
 
+	sqlParams->count = paramIndex;
+
 	/* when we didn't find any node to process, return our empty set */
 	if (paramIndex == 0)
 	{
@@ -1563,6 +1581,7 @@ BuildNodesArrayValues(NodeAddressArray *nodeArray,
 			return false;
 		}
 	}
+
 	return paramIndex;
 }
 
@@ -1615,13 +1634,10 @@ pgsql_replication_slot_create_and_drop(PGSQL *pgsql, NodeAddressArray *nodeArray
 		"SELECT 'drop', slot_name, NULL::pg_lsn FROM dropped";
 	/* *INDENT-ON* */
 
-	Oid paramTypes[NODE_ARRAY_MAX_COUNT * 2] = { 0 };
-	const char *paramValues[NODE_ARRAY_MAX_COUNT * 2] = { 0 };
+	nodesArraysValuesParams sqlParams = { 0 };
 	ReplicationSlotMaintainContext context = { 0 };
 
-	int paramCount = BuildNodesArrayValues(nodeArray,
-										   paramTypes, (char **) paramValues,
-										   values, BUFSIZE);
+	(void) BuildNodesArrayValues(nodeArray, &sqlParams, values, BUFSIZE);
 
 	/* add the computed ($1,$2), ... string to the query "template" */
 	int bytes = sformat(sql, 2 * BUFSIZE, sqlTemplate, values);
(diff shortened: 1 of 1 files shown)


src/bin/pg_autoctl/pgsql.c
```diff
@@ -1563,6 +1581,7 @@ BuildNodesArrayValues(NodeAddressArray *nodeArray,
 			return false;
 		}
 	}
+
 	return paramIndex;
```
JelteF, line 1585: do we really still need to return `paramIndex`? Also returning both false and paramIndex seems confusing. Let's change the code to actually handle this error.

JelteF (changes requested):

src/bin/pg_autoctl/pgsql.c
```diff
@@ -1491,12 +1491,21 @@ pgsql_drop_replication_slot(PGSQL *pgsql, const char *slotName)
  * We return how many parameters we filled in paramTypes and paramValues from
  * the nodeArray.
  */
+typedef struct nodesArraysValuesParams
+{
+	int count;
+	Oid types[NODE_ARRAY_MAX_COUNT * 2];
+	char *values[NODE_ARRAY_MAX_COUNT * 2];
+	char data[NODE_ARRAY_MAX_COUNT * 2][PG_LSN_MAXLENGTH];
```
JelteF, line 1499: maybe split this array in two fields one for LSNs and one for nodeids. I think that makes the code a bit easier to understand.

src/bin/pg_autoctl/pgsql.c
```diff
@@ -1542,6 +1558,8 @@ BuildNodesArrayValues(NodeAddressArray *nodeArray,
 		paramIndex += 2;
 	}
 
+	sqlParams->count = paramIndex;
```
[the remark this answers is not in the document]
DimCitus: The `sqlParams->count` should be the same thing as what we return in the function, which is `paramIndex`. The details are not as obvious as they should be... we are building a SQL string of constant VALUES ($1, $2), ($3, $4) and then an array of the values of those parameters, and the count here would be 4.

src/bin/pg_autoctl/pgsql.c
DimCitus: The job of this routine is to prepare arguments for `pgsql_execute_with_params`, using the libpq interface there. The thing is, this routine wants `(const char **)` and we can't use a `char [][]` to the same effect, because of pointer arithmetic and stuff in the C language. That's why we have a `char data[][]` for storage and a `char *values[]` for pointers...

JelteF: I meant something like this:

```suggestion
	char nodeIds[NODE_ARRAY_MAX_COUNT][INT_STRING_MAX_LENGTH];
	char lsns[NODE_ARRAY_MAX_COUNT][PG_LSN_MAXLENGTH];
```

src/bin/pg_autoctl/pgsql.c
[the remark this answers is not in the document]
JelteF: ah yes makes sense I missed the `+= 2;` at the end of the loop, which handles the problem I was thinking about.

src/bin/pg_autoctl/monitor.c
```diff
@@ -3611,6 +3611,9 @@ monitor_wait_for_state_change(Monitor *monitor,
 
 	*stateHasChanged = context.stateHasChanged;
 
+	/* disconnect from monitor */
+	pgsql_finish(&monitor->notificationClient);
```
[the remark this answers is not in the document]
DimCitus: Removed that commit, let's get back to that later with valgrind results.

src/bin/pg_autoctl/pgsql.c
```diff
@@ -1514,11 +1523,18 @@ BuildNodesArrayValues(NodeAddressArray *nodeArray,
 		int idParamIndex = paramIndex;
 		int lsnParamIndex = paramIndex + 1;
 
-		paramTypes[idParamIndex] = INT4OID;
-		paramValues[idParamIndex] = strdup(nodeIdString.strValue);
+		sqlParams->types[idParamIndex] = INT4OID;
+
+		strlcpy(sqlParams->data[idParamIndex],
+				nodeIdString.strValue,
+				PG_LSN_MAXLENGTH);
```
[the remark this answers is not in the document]
DimCitus: Fixed now with proper length.

src/bin/pg_autoctl/pgsql.c
DimCitus: Fixed, thanks for noticing.

src/bin/pg_autoctl/pgsql.c
DimCitus: Yeah well that could work now that we store the pointers to the data separately. Let me have a look then, I agree it would simplify quite some things.

DimCitus: Done now, thanks for the idea.

src/bin/pg_autoctl/pgsql.c
```diff
@@ -1509,16 +1527,22 @@ BuildNodesArrayValues(NodeAddressArray *nodeArray,
...
-		IntString nodeIdString = intToString(node->nodeId);
+		char *nodeIdString = intToString(node->nodeId).strValue;
 
 		int idParamIndex = paramIndex;
 		int lsnParamIndex = paramIndex + 1;
 
-		paramTypes[idParamIndex] = INT4OID;
-		paramValues[idParamIndex] = strdup(nodeIdString.strValue);
+		sqlParams->types[idParamIndex] = INT4OID;
+		strlcpy(sqlParams->nodeIds[idParamIndex], nodeIdString, NODEID_MAX_LENGTH);
```
JelteF, line 1536:
```suggestion
		strlcpy(sqlParams->nodeIds[nodeIndex], nodeIdString, NODEID_MAX_LENGTH);
```

src/bin/pg_autoctl/pgsql.c
```diff
@@ -1509,16 +1527,22 @@ BuildNodesArrayValues(NodeAddressArray *nodeArray,
...
+
+		/* store the (char *) pointer to the data in values */
+		sqlParams->values[idParamIndex] = sqlParams->nodeIds[idParamIndex];
 
-		paramTypes[lsnParamIndex] = LSNOID;
-		paramValues[lsnParamIndex] = node->lsn;
+		sqlParams->types[lsnParamIndex] = LSNOID;
+		strlcpy(sqlParams->lsns[lsnParamIndex], node->lsn, PG_LSN_MAXLENGTH);
```
JelteF, line 1542:
```suggestion
		strlcpy(sqlParams->lsns[nodeIndex], node->lsn, PG_LSN_MAXLENGTH);
```

src/bin/pg_autoctl/pgsql.c
```diff
@@ -1509,16 +1527,22 @@ BuildNodesArrayValues(NodeAddressArray *nodeArray,
...
+
+		/* store the (char *) pointer to the data in values */
+		sqlParams->values[lsnParamIndex] = sqlParams->lsns[lsnParamIndex];
```
JelteF, line 1545:
```suggestion
		sqlParams->values[lsnParamIndex] = sqlParams->lsns[nodeIndex];
```

src/bin/pg_autoctl/pgsql.c
```diff
@@ -1509,16 +1527,22 @@ BuildNodesArrayValues(NodeAddressArray *nodeArray,
...
 		int idParamIndex = paramIndex;
 		int lsnParamIndex = paramIndex + 1;
 
-		paramTypes[idParamIndex] = INT4OID;
-		paramValues[idParamIndex] = strdup(nodeIdString.strValue);
+		sqlParams->types[idParamIndex] = INT4OID;
+		strlcpy(sqlParams->nodeIds[idParamIndex], nodeIdString, NODEID_MAX_LENGTH);
+
+		/* store the (char *) pointer to the data in values */
+		sqlParams->values[idParamIndex] = sqlParams->nodeIds[idParamIndex];
```
JelteF, line 1539:
```suggestion
		sqlParams->values[idParamIndex] = sqlParams->nodeIds[nodeIndex];
```

src/bin/pg_autoctl/pgsql.c
DimCitus: Oh. wow. Fixed now.

src/bin/pg_autoctl/pgsql.c

src/bin/pg_autoctl/pgsql.c

src/bin/pg_autoctl/pgsql.c

JelteF (approved):

Issue #683, opened 2021-04-28
Unrecoverable state after failed switchover


Navid2zp:
Hello,

I have a 3 node setup, 1 primary, and 2 secondaries.

The primary node's server went down and another node got promoted to primary just fine. I tried a switchover after all servers became available again so the primary can be changed back to the server that was down before (more resources there). switchover failed and now neither of the nodes are primary and I don't know how to recover.

**Postgresql version:**

```
$ psql --version
psql (PostgreSQL) 13.2 (Ubuntu 13.2-1.pgdg20.04+1)
```

**pg_auto_failover version:**

```
$ pg_autoctl --version
pg_autoctl version 1.5.1
pg_autoctl extension version 1.5
compiled with PostgreSQL 13.2 (Ubuntu 13.2-1.pgdg20.04+1) on x86_64-pc-linux-gnu, compiled by gcc (Ubuntu 9.3.0-17ubuntu1~20.04) 9.3.0, 64-bit
compatible with Postgres 10, 11, 12, and 13
```

**OS:**

```
lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 20.04.2 LTS
Release:        20.04
Codename:       focal
```

### switchover:

```
$ pg_autoctl perform switchover --pgdata /var/lib/postgresql/monitor --formation worker-3

19:15:08 2882298 INFO  Listening monitor notifications about state changes in formation "worker-3" and group 0
19:15:08 2882298 INFO  Following table displays times when notifications are received
    Time |    Name |  Node |          Host:Port |       Current State |      Assigned State
---------+---------+-------+--------------------+---------------------+--------------------
19:15:08 | node_16 |    19 | 157.*.*.*:5436 |             primary |            draining
19:15:08 | node_16 |    19 | 157.*.*.*:5436 |            draining |            draining
19:15:08 | node_14 |    17 | 116.*.*.*:5436 |           secondary |          report_lsn
19:15:08 | node_15 |    18 |  78.*.*.*:5436 |           secondary |          report_lsn
19:15:09 | node_14 |    17 | 116.*.*.*:5436 |          report_lsn |          report_lsn
19:16:09 2882298 ERROR Failed to receive monitor's notifications
19:16:09 2882298 ERROR Failed to wait until a new primary has been notified
```

### Current state:

```
$ pg_autoctl show state --pgdata /var/lib/postgresql/monitor --formation worker-3

[...]

PR #695, opened 2021-05-04
Fix/consider timeline when reaching secondary

DimCitus:
Check that we are on the same timeline as our upstream node.

When a node reports state SECONDARY it can be the target of a failover.
Because we want a single straight line history in streaming replication, we
need to ensure that standby nodes are on the same timeline as their upstream
node before transitioning to SECONDARY.

See #683

### DIFF
--- a/src/bin/pg_autoctl/fsm_transition.c
+++ b/src/bin/pg_autoctl/fsm_transition.c
@@ -1033,6 +1033,15 @@ fsm_rewind_or_init(Keeper *keeper)
 bool
 fsm_maintain_replication_slots(Keeper *keeper)
 {
+	LocalPostgresServer *postgres = &(keeper->postgres);
+
+	/* first. check that we're on the same timeline as the new primary */
+	if (!standby_check_timeline_with_upstream(postgres))
+	{
+		/* errors have already been logged */
+		return false;
+	}
+
 	return keeper_maintain_replication_slots(keeper);
 }
 
@@ -1390,7 +1399,14 @@ fsm_follow_new_primary(Keeper *keeper)
 	}
 
 	/* now, in case we have an init state file around, remove it */
-	return unlink_file(config->pathnames.init);
+	if (!unlink_file(config->pathnames.init))
+	{
+		/* errors have already been logged */
+		return false;
+	}
+
+	/* finally, check that we're on the same timeline as the new primary */
+	return standby_check_timeline_with_upstream(postgres);
 }
 
 
--- a/src/bin/pg_autoctl/fsm_transition.c
+++ b/src/bin/pg_autoctl/fsm_transition.c
@@ -1024,14 +1024,14 @@ fsm_rewind_or_init(Keeper *keeper)
 
 
 /*
- * fsm_maintain_replication_slots is used when going from CATCHINGUP to
- * SECONDARY, to create missing replication slots. We want to maintain a
- * replication slot for each of the other nodes in the system, so that we make
- * sure we have the WAL bytes around when a standby nodes has to follow a new
- * primary, after failover.
+ * fsm_prepare_for_secondary is used when going from CATCHINGUP to SECONDARY,
+ * to create missing replication slots. We want to maintain a replication slot
+ * for each of the other nodes in the system, so that we make sure we have the
+ * WAL bytes around when a standby nodes has to follow a new primary, after
+ * failover.
  */
 bool
-fsm_maintain_replication_slots(Keeper *keeper)
+fsm_prepare_for_secondary(Keeper *keeper)
 {
 	LocalPostgresServer *postgres = &(keeper->postgres);
 
@@ -1405,8 +1405,18 @@ fsm_follow_new_primary(Keeper *keeper)
 		return false;
 	}
 
-	/* finally, check that we're on the same timeline as the new primary */
-	return standby_check_timeline_with_upstream(postgres);
+	/*
+	 * Finally, check that we're on the same timeline as the new primary when
+	 * assigned secondary as a goal state. This transition function is also
+	 * used when going from secondary to catchingup, as the primary might have
+	 * changed also in that situation.
+	 */
+	if (keeper->state.assigned_role == SECONDARY_STATE)
+	{
+		return standby_check_timeline_with_upstream(postgres);
+	}
+
+	return true;
 }
 
 
--- a/src/bin/pg_autoctl/pgsql.c
+++ b/src/bin/pg_autoctl/pgsql.c
@@ -2878,53 +2878,56 @@ pgsql_identify_system(PGSQL *pgsql, IdentifySystem *system)
 	}
 
 	/* while at it, we also run the TIMELINE_HISTORY command */
-	TimelineHistoryResult hContext = { 0 };
-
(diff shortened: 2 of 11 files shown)


src/bin/pg_autoctl/fsm_transition.c
```diff
@@ -1033,6 +1033,15 @@ fsm_rewind_or_init(Keeper *keeper)
 bool
 fsm_maintain_replication_slots(Keeper *keeper)
```
JelteF, line 1034: maybe rename this function, since it now does more.

src/bin/pg_autoctl/fsm_transition.c
```diff
@@ -1390,7 +1399,14 @@ fsm_follow_new_primary(Keeper *keeper)
...
 	/* now, in case we have an init state file around, remove it */
-	return unlink_file(config->pathnames.init);
+	if (!unlink_file(config->pathnames.init))
+	{
+		/* errors have already been logged */
+		return false;
+	}
+
+	/* finally, check that we're on the same timeline as the new primary */
+	return standby_check_timeline_with_upstream(postgres);
```
JelteF, line 1409: Do we really need this here? This function is used for the transition from `secondary` to `catchingup`. I think it's enough to only do it in the transition from `catchingup` back to `secondary`.

src/bin/pg_autoctl/pgsql.h
```diff
@@ -170,6 +170,50 @@ typedef struct NodeAddressArray
...
+ * TimeLineID data type.
+ *
+ * Zero is used indicate an invalid pointer. Bootstrap skips the first possible
+ * WAL segment, initializing the first WAL page at WAL segment size, so no XLOG
+ * record can begin at zero.
+ */
+#define InvalidXLogRecPtr 0
+#define XLogRecPtrIsInvalid(r) ((r) == InvalidXLogRecPtr)
+
+#define PG_AUTOCTL_MAX_TIMELINES 1024
```
JelteF, line 186: It seems dangerous to put a limit on this, since the amount of timelines could reasonably be very high. It might be better to use a `malloc` instead.

src/bin/pg_autoctl/primary_standby.c
```diff
@@ -1456,3 +1456,91 @@ standby_cleanup_as_primary(LocalPostgresServer *postgres)
...
+ * upstream node setup in its replicationSource.
+ */
+bool
+standby_check_timeline_with_upstream(LocalPostgresServer *postgres)
+{
+	ReplicationSource *replicationSource = &(postgres->replicationSource);
+	NodeAddress *primaryNode = &(replicationSource->primaryNode);
+
+	/* pg_control_checkpoint() returns the local timeline, update it now */
+	if (!pgsql_checkpoint(&(postgres->sqlClient)))
```
JelteF, line 1473: Creating a checkpoint can be quite heavy. If we only do this to get the current timeline id, maybe instead we should use some other function, such as `pg_walfile_name` or `pg_control_data`: https://www.postgresql.org/message-id/3FB63E54-403A-440E-88CB-CFA6FEF3E0F7%40icloud.com

src/bin/pg_autoctl/primary_standby.c
```diff
@@ -1456,3 +1456,91 @@ standby_cleanup_as_primary(LocalPostgresServer *postgres)
...
+				 primaryNode->nodeId,
+				 primaryNode->name,
+				 primaryNode->host,
+				 primaryNode->port,
+				 upstreamTimeline,
+				 localTimeline);
+
+		return false;
+	}
+	else if (upstreamTimeline == localTimeline)
```
JelteF, line 1535: Can we also check that the moment when the timeline split is the same?

JelteF (changes requested):

src/bin/pg_autoctl/fsm_transition.c
DimCitus: New name is `fsm_prepare_for_secondary`.

src/bin/pg_autoctl/fsm_transition.c
DimCitus: Good point. Because we use the same transition function in both places, and there's a lot of common code, I just added a check that way:

``` c
	/*
	 * Finally, check that we're on the same timeline as the new primary when
	 * assigned secondary as a goal state. This transition function is also
	 * used when going from secondary to catchingup, as the primary might have
	 * changed also in that situation.
	 */
	if (keeper->state.assigned_role == SECONDARY_STATE)
	{
		return standby_check_timeline_with_upstream(postgres);
	}
```

src/bin/pg_autoctl/pgsql.h
DimCitus: I would like to avoid `malloc` and then having to deal with who's responsible for calling `free` as much as possible, here we don't have to think about the lifetime of the “cache”. It's trivial and quite cheap to enlarge that array even tenfold, if you think we need to do that?

At the moment the array (given 64 bits alignment) is going to be 128kB of memory...

src/bin/pg_autoctl/primary_standby.c
DimCitus: We can grab `pg_stat_wal_receiver.received_tli`, I believe that having received it locally is what matters in our case. If we have to restart (crash recovery or normal startup), we will be able to catch-up with the new tli, and more importantly in case of a promotion, again, we can replay the WAL files and switch to the new TLI then.

Note that using `pg_walfile_name` requires knowing which WAL file is the most recently received, and should give us the same information as `pg_stat_wal_receiver` I believe, and `pg_control_data` needs a checkpoint too and shows the same value as the `pg_control_checkpoint` view I was using before.

src/bin/pg_autoctl/primary_standby.c
DimCitus: Do we need to? I don't think so. If yes, then we need to also parse the timeline history file that's local to the standby to grab that piece of information.

src/bin/pg_autoctl/pgsql.c
```diff
@@ -2505,7 +2506,11 @@ pgsql_get_postgres_metadata(PGSQL *pgsql,
 		" then coalesce(pg_last_wal_receive_lsn(), pg_last_wal_replay_lsn())"
```
JelteF, line 2506: Why do we use `pg_last_wal_receive_lsn` here? It seems `pg_last_wal_replay_lsn` is the one that reflects what we want the most.

JelteF (approved):
I tried this out and I think it's good to merge, since it's an improvement over what we have now. I'm not entirely sure if `received_tli` is the right timeline to use though. I guess it could be that it did receive the timeline, but hasn't actually applied it because of some error. 

However, it seems that there's currently no way to get the applied timeline: https://www.postgresql.org/message-id/flat/20190723180518.635ac554%40firost

src/bin/pg_autoctl/pgsql.c
DimCitus: If memory serves, that's because that's the LSN that we can reach when restarting Postgres before promotion or for REPORT_LSN. It's safe on-disk even if it's not applied yet. Applying might have delays because of read-only traffic.

PR #697, opened 2021-05-05
Fix pg_autoctl show settings --json output.

### LINKED ISSUE #694: formation settings 'formations' property should be an array or a non-generic object
The ‘pg_autoctl get formation settings --json‘ output looks like this:

```` 
{
   ...
    "formation": {
        "value": "1",
        "context": "formation",
        "node_id": null,
        "setting": "number_sync_standbys",
        "group_id": null,
        "nodename": "default"
    }
}
````

The ‘formation’ object seems to represent a generic setting value, i.e. a kind of key/value pair. But the ‘formation’ property only contains one of them, which doesn’t seem quite right. I.e. I would expect either of these:

````
"formation": {
  "number_sync_standbys": 1,
  ...
}
````

or

```` 
"formation": [
    {
        "value": "1",
        "context": "formation",
        "node_id": null,
        "setting": "number_sync_standbys",
        "group_id": null,
        "nodename": "default"
    }
]
````
 
Did you intend for the ‘formation’ property to contain an array?

### PR DESCRIPTION
**Pull request by DimCitus**
The formation context parts should also be an array, even when we know it
contains a single element for now.

Fixes #694

### DIFF
--- a/src/bin/pg_autoctl/monitor.c
+++ b/src/bin/pg_autoctl/monitor.c
@@ -2794,7 +2794,7 @@ monitor_print_formation_settings_as_json(Monitor *monitor, char *formation)
 		" ), "
 		" f(json) as "
 		" ( "
-		"   select row_to_json(settings) "
+		"   select jsonb_agg(row_to_json(settings)) "
 		"     from settings "
 		"    where context = 'formation' "
 		" ), "

### CODE REVIEW

**Comment by DimCitus** [comment]
With this fix we have:

```
pg_autoctl show settings --json | jq '.formation'
[
  {
    "value": "1",
    "context": "formation",
    "node_id": null,
    "setting": "number_sync_standbys",
    "group_id": null,
    "nodename": "default"
  }
]
```

And that makes sense, and it's the same format as the other parameters, using JSON objects in an EAV approach.

JelteF (approved):
Makes sense. Needs to be called out as a breaking change in the changelog though.

Issue #704, opened 2021-05-07
Node left into draining state in multi node set up


gkokolatos:
Hi team,

we are noticing a behaviour which is consistently reproducible since the introduction of multi node set up. The checks were performed against postgres versions 11 and 12. We are wondering if this is the expected behaviour.

Assume a three node setup with data nodes named node[1-3], and one monitor node.
Assume that node1 is primary and nodes node2 and node3 are secondary. The system is healthy.
Bring the primary down and wait until a new primary is elected, e.g. node2. Bring down that new primary node2.
Wait until the monitor has registered the state and reacted. Now node3 should be in wait_primary state.
Bring back up node2. Now node2 is stuck into draining state. What is interesting is that node2's db is down
and the system does not seem to react to it. It stays permanently in that state.

Let us show the states of the cluster (pg_autoctl show state against the monitor node) after stopping node1.

```
  Name |  Node |      Host:Port |       LSN |   Connection |       Current State |      Assigned State
-------+-------+----------------+-----------+--------------+---------------------+--------------------
node_1 |     1 | 10.0.2.15:5001 | 0/5000108 | read-write ! |             primary |             demoted
node_2 |     2 | 10.0.2.15:5002 | 0/5000410 |   read-write |             primary |             primary
node_3 |     3 | 10.0.2.15:5003 | 0/5000410 |    read-only |           secondary |           secondary
```

This is all good and expected.

Let us show the states of the cluster after stopping node2

```
  Name |  Node |      Host:Port |       LSN |   Connection |       Current State |      Assigned State
-------+-------+----------------+-----------+--------------+---------------------+--------------------
node_1 |     1 | 10.0.2.15:5001 | 0/5000108 | read-write ! |             primary |             demoted
node_2 |     2 | 10.0.2.15:5002 | 0/5000410 | read-write ! |             primary |            draining
node_3 |     3 | 10.0.2.15:5003 | 0/5000638 |   read-write |        wait_primary |        wait_primary
```

This state does not change and node_2 remains in draining Assigned state.
The registered events at this point (pg_autoctl show events against the monitor node) show

```
                    Event Time |   Node |       Current State |      Assigned State | Comment
-------------------------------+--------+---------------------+---------------------+-----------
 2021-05-07 14:20:20.097981+00 |    0/3 |           secondary | report_lsn | Setting goal state of node 3 "node_3" (10.0.2.15:5003) to report_lsn to find the failover candidate
[...]

PR #706, opened 2021-05-14
Complete transition of a second (or greater) failed primary

gkokolatos:
Assume that there exists a healthy three datanode cluster having datanodes named
node1, node2, node3. Assume that node1 is the primary and it fails. The cluster
will correctly demote node1 and promote node2 to primary. When node2 fails, the
cluster would fail to reach a stable state due to node2 getting stuck in
draining state.

The culrpit is GetPrimaryOrDemotedNodeInGroup which contains logic to try to
find a demoted primary in case that a healty primary is not existing. To do
that, it traverses an ordered by name list of nodes in the group. When it finds
the first node that is either a demoted or draining or taken out for
maintainance primary, it chooses it and stops looking further. In the example
above, note that the ordered by name list starts with node1, which would satisfy
the criterion and node2 will never be selected.

To fix that, continue looping through the nodes. Assuming that there can be only
one primary active at any time, hence only one primary that is currently going
through demotion process, then choose that one.

Add a specific test case for it.

Fixes issue #704

### DIFF
--- a/src/monitor/node_metadata.c
+++ b/src/monitor/node_metadata.c
@@ -448,8 +448,8 @@ GetPrimaryOrDemotedNodeInGroup(char *formationId, int32 groupId)
 		AutoFailoverNode *currentNode = (AutoFailoverNode *) lfirst(nodeCell);
 
 		if (StateBelongsToPrimary(currentNode->reportedState) &&
-			(IsBeingDemotedPrimary(primaryNode) == false ||
-			 IsDemotedPrimary(currentNode) == false))
+			(!IsBeingDemotedPrimary(primaryNode) ||
+			 !IsDemotedPrimary(currentNode)))
 		{
 			primaryNode = currentNode;
 		}
--- a/src/monitor/node_metadata.c
+++ b/src/monitor/node_metadata.c
@@ -438,15 +438,20 @@ GetPrimaryOrDemotedNodeInGroup(char *formationId, int32 groupId)
 		return primaryNode;
 	}
 
-	/* maybe we have a primary that is draining or has been demoted? */
+	/*
+	 * Maybe we have a primary that is draining or has been demoted?
+	 * In case there are more than one of those, choose the one that is
+	 * currently being demoted.
+	 */
 	foreach(nodeCell, groupNodeList)
 	{
 		AutoFailoverNode *currentNode = (AutoFailoverNode *) lfirst(nodeCell);
 
-		if (StateBelongsToPrimary(currentNode->reportedState))
+		if (StateBelongsToPrimary(currentNode->reportedState) &&
+			(IsBeingDemotedPrimary(primaryNode) == false ||
+			IsDemotedPrimary(currentNode) == false))
 		{
 			primaryNode = currentNode;
-			break;
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


tests/test_multinode_alternate_primary_failures.py
```diff
@@ -0,0 +1,149 @@
...
+def test_005_001_fail_primary_again():
+    # verify that node3 is primary and stop it
+    assert node3.get_state().assigned == "primary"
+    node3.fail()
+
+    assert node3.wait_until_assigned_state(target_state="demoted")
+    assert node1.wait_until_assigned_state(target_state="primary")
+
+def test_005_002_fail_primary_again():
+    # verify that node3 is primary and stop it
```
DimCitus, line 125: copy/paste'o, I think that should read “verify that node1 is primary and stop it”

src/monitor/node_metadata.c
```diff
@@ -438,15 +438,20 @@ GetPrimaryOrDemotedNodeInGroup(char *formationId, int32 groupId)
...
+	 * currently being demoted.
+	 */
 	foreach(nodeCell, groupNodeList)
 	{
 		AutoFailoverNode *currentNode = (AutoFailoverNode *) lfirst(nodeCell);
 
-		if (StateBelongsToPrimary(currentNode->reportedState))
+		if (StateBelongsToPrimary(currentNode->reportedState) &&
+			(IsBeingDemotedPrimary(primaryNode) == false ||
+			IsDemotedPrimary(currentNode) == false))
```
DimCitus, line 452: I think we would generally use a slightly different style:

``` c
		if (StateBelongsToPrimary(currentNode->reportedState) &&
  			(!IsBeingDemotedPrimary(primaryNode) ||
  			!IsDemotedPrimary(currentNode)))
```

src/monitor/node_metadata.c
```diff
@@ -438,15 +438,20 @@ GetPrimaryOrDemotedNodeInGroup(char *formationId, int32 groupId)
...
 		{
 			primaryNode = currentNode;
-			break;
```
DimCitus, line 449: Do we need to remove the break here? My understanding is that a short cut in this loop wouldn't have an impact, given the new tests, we find only one node and may return either the first or the last one then, it should be the same node, right?

DimCitus (approved):
Thanks for this bug fix! I like how limited your code changes are.

Please review https://travis-ci.com/github/citusdata/pg_auto_failover/jobs/505381364 and run a `make indent` pass to make Travis happy, and then I'll be able to merge your PR. I have left also some comments, those are not stoppers but you might still want to consider them.

tests/test_multinode_alternate_primary_failures.py
gkokolatos: You are correct, will fix.

src/monitor/node_metadata.c
gkokolatos: Heh, I was debating as to which was preferable :) Will fix.

src/monitor/node_metadata.c
gkokolatos: I have not tested it with the break to be honest. My understanding is the same as yours. I chose to keep the full loop under a 'belts and suspenders' argument, helped by the fact that the loops are generally very short anyways.

gkokolatos:
Will push a couple of commits shortly with the ident run (always forget that one) and comment derived touches.

gkokolatos:
> Will push a couple of commits shortly with the ident run (always forget that one) and comment derived touches.

Done

DimCitus:
Also, I think we want your new test to be part of the automated test suite on Travis, and for that we need to manually edit the Makefile to add the file in the TESTS_MULTI list. (Beware that I clicked on “update branch” on Travis so your local version will have diverged and you might want to rebase / force push when you do more changes to it).

gkokolatos:
On Friday, 14 May 2021 13:10, Dimitri Fontaine ***@***.***> wrote:

> Also, I think we want your new test to be part of the automated test suite on Travis, and for that we need to manually edit the Makefile to add the file in the TESTS_MULTI list. (Beware that I clicked on “update branch” on Travis so your local version will have diverged and you might want to rebase / force push when you do more changes to it).

Sure. Will do!

gkokolatos:
> ‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐
> On Friday, 14 May 2021 13:10, Dimitri Fontaine ***@***.***> wrote: Also, I think we want your new test to be part of the automated test suite on Travis, and for that we need to manually edit the Makefile to add the file in the TESTS_MULTI list. (Beware that I clicked on “update branch” on Travis so your local version will have diverged and you might want to rebase / force push when you do more changes to it).
> Sure. Will do!

Done and travis now seems happy.

> […](#)

PR #707, opened 2021-05-17
Fix how many nodes need to report their LSN to perform a failover.

### LINKED ISSUE #682: Node is promoted to wait_primary incorrectly
```
```
1. wait until stable
2. Ctrl+C primary
3. Ctrl+C one secondary
4. Wait 1 minute, the other secondary is now wait_primary. 

This is incorrect, because with `number-sync-standbys = 1` we need to have 2 out of 3 nodes in the election to avoid losing commited data.

DimCitus:
With this patch we make sure that all the nodes that participage in the
quorum have reported their LSN before we select a candidate for failover.
Otherwise, we might lose committed data.

Fixes #682

### DIFF
--- a/src/monitor/group_state_machine.c
+++ b/src/monitor/group_state_machine.c
@@ -41,10 +41,12 @@
  */
 typedef struct CandidateList
 {
+	int numberSyncStandbys;
 	List *candidateNodesGroupList;
 	List *mostAdvancedNodesGroupList;
 	XLogRecPtr mostAdvancedReportedLSN;
 	int candidateCount;
+	int quorumCandidateCount;
 	int missingNodesCount;
 } CandidateList;
 
@@ -1221,6 +1223,11 @@ ProceedGroupStateForMSFailover(AutoFailoverNode *activeNode,
 	 * different candidateNodesGroupList in which every node has reported their
 	 * LSN position, allowing progress to be made.
 	 */
+	char *formationId = activeNode->formationId;
+	AutoFailoverFormation *formation = GetFormation(formationId);
+
+	candidateList.numberSyncStandbys = formation->number_sync_standbys;
+
 	BuildCandidateList(nodesGroupList, &candidateList);
 
 	/*
@@ -1263,8 +1270,6 @@ ProceedGroupStateForMSFailover(AutoFailoverNode *activeNode,
 	 * WAIT_PRIMARY state with all the writes blocked for lack of standby
 	 * nodes.
 	 */
-	char *formationId = activeNode->formationId;
-	AutoFailoverFormation *formation = GetFormation(formationId);
 	int minCandidates = formation->number_sync_standbys + 1;
 
 	/* no candidates is a hard pass */
@@ -1274,7 +1279,7 @@ ProceedGroupStateForMSFailover(AutoFailoverNode *activeNode,
 	}
 
 	/* not enough candidates to promote and then accept writes, pass */
-	else if (candidateList.candidateCount < minCandidates)
+	else if (candidateList.quorumCandidateCount < minCandidates)
 	{
 		char message[BUFSIZE] = { 0 };
 
@@ -1285,7 +1290,7 @@ ProceedGroupStateForMSFailover(AutoFailoverNode *activeNode,
 			"number_sync_standbys=%d in formation \"%s\", "
 			"activeNode is " NODE_FORMAT
 			" and reported state \"%s\"",
-			candidateList.candidateCount,
+			candidateList.quorumCandidateCount,
 			minCandidates,
 			formation->number_sync_standbys,
 			formation->formationId,
@@ -1459,10 +1464,19 @@ BuildCandidateList(List *nodesGroupList, CandidateList *candidateList)
 			continue;
 		}
 
-		/* grab healthy standby nodes which have reached REPORT_LSN */
+		/*
+		 * Grab healthy standby nodes which have reached REPORT_LSN.
+		 */
 		if (IsCurrentState(node, REPLICATION_STATE_REPORT_LSN))
 		{
 			candidateNodesGroupList = lappend(candidateNodesGroupList, node);
+
+			/* when number_sync_standbys is zero, quorum isn't discriminant */
+			if (node->replicationQuorum ||
+				candidateList->numberSyncStandbys == 0)
+			{
+				++(candidateList->quorumCandidateCount);
+			}
 
 			continue;
 		}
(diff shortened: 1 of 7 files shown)


src/monitor/group_state_machine.c
```diff
@@ -1377,6 +1377,25 @@ BuildCandidateList(List *nodesGroupList, CandidateList *candidateList)
...
+			 * Otherwise, we didn't get its LSN and this node might be the most
+			 * advanced LSN. Picking it now might lead to loosing commited data
+			 * that was reported to the client connection.
+			 *
+			 * Only the nodes that participate in the quorum are required to
+			 * report their LSN, because only those nodes are waited by
+			 * Postgres to report a commit to the client connection.
+			 */
+			if (node->replicationQuorum &&
+				node->reportedState != REPLICATION_STATE_REPORT_LSN)
```
JelteF, line 1394: What happens when the node is down and is in REPORT_LSN, but also had the highest LSN?

JelteF (changes requested):
Looks like a reasonable change to fix this issue, but definitly needs a test.

src/monitor/group_state_machine.c
DimCitus: First, it might not be the only node with the highest LSN, so the code path continues to see about that later in `SelectFailoverCandidateNode`. Second, if it's the highest LSN, we can still select another candidate and go through the fast forward state. Well at some point if that node is the only one with the highest LSN it needs to be back online to ensure success of the failover, I think.

DimCitus:
> Looks like a reasonable change to fix this issue, but definitly needs a test.

The new test from #706 seems to cover the situation here, and also had to be adjusted to the new situation. I think we're good in terms of test coverage.

JelteF (approved):

tests/test_multi_alternate_primary_failures.py
```diff
@@ -96,15 +96,14 @@ def test_003_002_stop_primary():
     assert node2.get_state().assigned == "primary"
     node2.fail()
 
-    # wait for node3 to become the new write node
-    assert node2.wait_until_assigned_state(target_state="demoted")
-    assert node3.wait_until_state(target_state="wait_primary")
+    # node3 can't be promoted when it's the only one reporting its LSN
+    assert node2.wait_until_assigned_state(target_state="draining")
+    assert node3.wait_until_state(target_state="report_lsn")
```
JelteF, line 101: I think it might be nice if we could wait a bit here, to ensure that `node3` doesn't become `wait_primary`.

JelteF: Something like:
```suggestion
    assert node3.wait_until_state(target_state="report_lsn")
    node3.sleep(10)
    assert node3.wait_until_state(target_state="report_lsn")
```

DimCitus:
I have now revised this PR to implement some quorum validation when selecting a new node during a failover. Your comment that we might failover to a node that's alone and then have to block writes anyway, compared to having a read-only service, that's what I considered when working on this PR. Also note that previous to those changes, the failover mechanism when given a single node in REPORT_LSN would drive it to WAIT_PRIMARY, where writes are accepted and not replicated, contrary to the number_sync_standby setting when that's non-zero.

``` c
	/*
	 * So all the expected candidates did report their LSN, no node is missing.
	 * Let's see about selecting a candidate for failover now, when we do have
	 * candidates.
	 *
	 * To start the selection process, we require at least number_sync_standbys
	 * nodes to have reported their LSN and be currently healthy, otherwise we
	 * won't be able to maintain our guarantees: we would end-up with a node in
	 * WAIT_PRIMARY state with all the writes blocked for lack of standby
	 * nodes.
	 */
	char *formationId = activeNode->formationId;
	AutoFailoverFormation *formation = GetFormation(formationId);
	int minCandidates = formation->number_sync_standbys + 1;

	/* no candidates is a hard pass */
	if (candidateList.candidateCount == 0)
	{
		return false;
	}

	/* not enough candidates to promote and then accept writes, pass */
	else if (candidateList.candidateCount < minCandidates)
	{
		char message[BUFSIZE] = { 0 };

		LogAndNotifyMessage(
			message, BUFSIZE,
			"Failover still in progress with %d candidates having reported "
			"their LSN: %d nodes are required in the quorum to satisfy "
			"number_sync_standbys = %d in formation \"%s\", "
			"activeNode is " NODE_FORMAT
			" and reported state \"%s\"",
			candidateList.candidateCount,
			minCandidates,
			formation->number_sync_standbys,
			formation->formationId,
			NODE_FORMAT_ARGS(activeNode),
			ReplicationStateGetName(activeNode->reportedState));

		return false;
	}

	/* enough candidates to promote and then accept writes, let's do it! */
	else
	{
```

tests/test_multi_async.py
```diff
@@ -256,31 +260,39 @@ def test_014_003_restart_node1():
...
+#   demoted      wait_primary secondary
+#                <down>
+#   demoted      demoted      report_lsn
+#                <up>
+#   demoted      wait_primary secondary
+#   <up>
+#   secondary    primary      secondary
+#
+#   ~~~
+#
```
JelteF, line 295: I think this new order should be instead of the one below

src/monitor/group_state_machine.c
```diff
@@ -1256,8 +1256,47 @@ ProceedGroupStateForMSFailover(AutoFailoverNode *activeNode,
...
+	int minCandidates = formation->number_sync_standbys + 1;
+
+	/* no candidates is a hard pass */
+	if (candidateList.candidateCount == 0)
+	{
+		return false;
+	}
+
+	/* not enough candidates to promote and then accept writes, pass */
+	else if (candidateList.candidateCount < minCandidates)
```
JelteF, line 1277: Should only use the count of the sync candidates here.

JelteF (changes requested):
(see other comment)

tests/test_multi_async.py
DimCitus: Thanks, that was a leftover that should have been cleaned-up. Done now.

src/monitor/group_state_machine.c
DimCitus: That's a good point. Done now.

src/monitor/group_state_machine.c
```diff
@@ -1256,8 +1263,45 @@ ProceedGroupStateForMSFailover(AutoFailoverNode *activeNode,
...
+	}
+
+	/* not enough candidates to promote and then accept writes, pass */
+	else if (candidateList.quorumCandidateCount < minCandidates)
+	{
+		char message[BUFSIZE] = { 0 };
+
+		LogAndNotifyMessage(
+			message, BUFSIZE,
+			"Failover still in progress with %d candidates having reported "
```
JelteF, line 1288: Let's make this error message a bit more explicit to avoid confusion:

```suggestion
			"Failover still in progress with %d candidates that participate in the quorum having reported "
```

DimCitus: Thanks, added.

JelteF (approved):

PR #711, opened 2021-05-26
Minor tweaks during the registration of healthcheck workers

gkokolatos:
It is possible for a dynamic background worker to fail registration. It is
unlikely, yet possible. If that happens during the monitor's launch of
healthcheck workers, the function StartHealthCheckWorker() will emmit an info
level log message saying to acknowledge the fact. Then the caller of the
function will try again to register that worker until it succeeds or the process
receives a signal to terminate. If the registration is successfull, it will note
the fact and then proceed to start the background worker. Then it will never try
again to register nor restart the worker.

The above works great in most scenarios. However there exists a corner case
where it is not optimal. Assume that a healthcheck worker for a database
receives a SIGTERM. Then the system will never try to restart it. As a
consequence the system will seize to update the health information for the nodes
of that database.

On similar lines, if a health check worker fails to register after a monitor
restart, then the system will use whatever entries where last registered in its
state and database entries. The logs will contain only a info level log line
informing of the failure to register the worker and the system will appear
otherwise healthy.

The current patch takes non intrucive steps in addressing some noticed above.

* It raises the log level of the failure to register a dynamic background worker
  from info to warn.
* It adds a hint to said message in accordance to postgres documentation and
  usage in core, e.g. logical replication launcer
* Improves clarity by renamiing StartHealthCheckWorker to
  RegisterHealthCheckWorker since the function is responsible for registering
  said worker and we do not, nor can we know that the worker has actually
  started.
* Updates the process name of the worker to include that it is a healthcheck
  worker of a named database
* Adds a check that the process has actually started after registration and only
  then proceeds to mark it as active
* Adds a comment in the relevant part to explain the above
* Adds a checking mechanism in the loop and detects if an already registered and
  started background worker is still running. If not it emmits a warn level
  message and tries to register it again.

The above do not address all the problems mentioned, notably updating the health
information of the nodes after N amount of failures to register/start. However
it is believed that this is a sufficient first step to a `right direction.

### DIFF
--- a/src/monitor/health_check_worker.c
+++ b/src/monitor/health_check_worker.c
@@ -105,7 +105,6 @@ typedef struct HealthCheckHelperDatabase
 	/* hash key: database to run on */
 	Oid dboid;
 	pid_t workerPid;
-	bool isActive;
 	BackgroundWorkerHandle *handle;
 } HealthCheckHelperDatabase;
 
@@ -262,7 +261,7 @@ HealthCheckWorkerLauncherMain(Datum arg)
 			HealthCheckHelperDatabase *dbData = hash_search(HealthCheckWorkerDBHash,
 															(void *) &entry->dboid,
 															HASH_ENTER, &isFound);
-			if (isFound && dbData->isActive)
+			if (isFound)
 			{
 				handle = dbData->handle;
 
@@ -275,19 +274,27 @@ HealthCheckWorkerLauncherMain(Datum arg)
 				 * actually running. Note that it is not possible to get
 				 * BGWH_NOT_YET_STARTED at this point, because this is not first
 				 * time we try to register the worker due to the isFound value
-				 * above, and we will WaitForBackgroundWorkerStartup after
-				 * registering before setting isActive to true. Thus we can only
-				 * get BGWH_STARTED or BGWH_STOPPED.
+				 * above. The HealthCheckWorkerDBHash only maintains verified
+				 * started entries. Thus we can only get BGWH_STARTED or
+				 * BGWH_STOPPED.
 				 *
 				 * NOTE: For PostgreSQL versions prior to 12, it is possible to
-				 * get erroneous results in GetBackgroundWorkerPid due to the
-				 * slot generation number being out of date. This is especially
-				 * true after signalling the Postmaster to reload the
-				 * configuration.
+				 * get erroneous results in GetBackgroundWorkerPid. In that
+				 * function, the handle we are providing is checked against a
+				 * localy held and maintained array. In affected versions, there
+				 * exists a window where our stored handle and that array to
+				 * have fallen out of sync. Such a window is more probable when
+				 * signalling the Postmaster e.g. to reload the configuration.
+				 *
+				 * When the entries are out of sync handle will be returned with
+				 * status BGWH_STOPPED and it will never be upgraded to
+				 * BGWH_STARTED, whereas in reality the bgworker is running and
+				 * most probably is healthy. From that moment onwards, we will
+				 * never be able to correctly assess the worker's status via the
+				 * provided API.
 				 */
 				if (GetBackgroundWorkerPid(handle, &pid) != BGWH_STARTED)
 				{
-					dbData->isActive = false;
 					#if (PG_VERSION_NUM >= 120000)
 					ereport(WARNING,
 							(errmsg(
@@ -302,6 +309,18 @@ HealthCheckWorkerLauncherMain(Datum arg)
 								 entry->dbname)));
 					#endif
 
+					/*
+					 * Now we know that either the worker has stopped or we
+					 * encountered the issue described in the NOTE above. We use
+					 * StopHealthCheckWorker to remove the entry from the
+					 * HealthCheckWorkerDBHash. That will force a retry in the
+					 * next scan of the databaselist. Furthermore, if the
+					 * status from GetBackgroundWorkerPid was not the correct
+					 * one, then StopHealthCheckWorker will also make certain
+					 * that the worker that we can no longer monitor correctly,
+					 * will be stopped. That will leave HealthCheckWorkerDBHash
+					 * in a consistent state.
+					 */
 					StopHealthCheckWorker(entry->dboid);
 				}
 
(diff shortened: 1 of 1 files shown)


DimCitus (approved):
Thanks for your contribution @gkokolatos ; looks good to me!

gkokolatos:
On Wednesday, 26 May 2021 17:50, Dimitri Fontaine ***@***.***> wrote:

> @DimCitus approved this pull request.
>
> Thanks for your contribution @gkokolatos ; looks good to me!

Thank you!

DimCitus:
Seems like a `make indent` pass is necessary, too...

gkokolatos:
On Wed, May 26, 2021 at 18:56, Dimitri Fontaine ***@***.***> wrote:

> Seems like a make indent pass is necessary, too...

This time I had actually done that :) Seems that I used snprintf which is banned.

Will amend first thing tomorrow morning.

gkokolatos:
@DimCitus Longest string fix in history :P 

While working on the string fix, decided to do some exhaustive tests in all supported versions locally and run into an upstream issue in postgres versions prior to 12.

I redesigned a bit so a second review pass is required.

src/monitor/health_check_worker.c
```diff
@@ -262,21 +264,59 @@ HealthCheckWorkerLauncherMain(Datum arg)
...
+								 entry->dbname)));
+					#else
+					ereport(WARNING,
+							(errmsg(
+								 "stopping worker for pg_auto_failover "
+								 "health checks in \"%s\"",
+								 entry->dbname)));
+					#endif
+
+					StopHealthCheckWorker(entry->dboid);
```
DimCitus, line 305: Can you add a comment about we stop the worker here? I can guess why, but I'm not sure, and would appreciate being able to read the reasoning here.

DimCitus (changes requested):
Thanks for the extra work on the PG<12 support. I think we need a couple more comments to understand the race condition bits and why we call `StopHealthCheckWorker` in the start code path...

gkokolatos:
On Thursday, 27 May 2021 14:57, Dimitri Fontaine ***@***.***> wrote:

> @DimCitus requested changes on this pull request.
>
> Thanks for the extra work on the PG<12 support. I think we need a couple more comments to understand the race condition bits and why we call StopHealthCheckWorker in the start code path...

Thank you. I will do it.

> ---------------------------------------------------------------
>
> In [src/monitor/health_check_worker.c](https://github.com/citusdata/pg_auto_failover/pull/711#discussion_r640596078):
>
>> +					dbData->isActive = false;
> +					#if (PG_VERSION_NUM >= 120000)
> +					ereport(WARNING,
> +							(errmsg(
> +								"found stopped worker for pg_auto_failover "
> +								"health checks in \"%s\"",
> +								entry->dbname)));
> +					#else
> +					ereport(WARNING,
> +							(errmsg(
> +								"stopping worker for pg_auto_failover "
> +								"health checks in \"%s\"",
> +								entry->dbname)));
> +					#endif
> +
> +					StopHealthCheckWorker(entry->dboid);
>
> Can you add a comment about we stop the worker here? I can guess why, but I'm not sure, and would appreciate being able to read the reasoning here.

Absolutely!

> ---------------------------------------------------------------
>
> In [src/monitor/health_check_worker.c](https://github.com/citusdata/pg_auto_failover/pull/711#discussion_r640596429):
>
>> +					dbData->handle = handle;
> +					ereport(LOG,
> +							(errmsg(
> +								"started worker for pg_auto_failover "
> +								"health checks in \"%s\"",
> +								entry->dbname)));
> +				}
> +				else
> +				{
> +					ereport(WARNING,
> +							(errmsg(
> +								"failed to start worker for pg_auto_failover "
> +								"health checks in \"%s\"",
> +								entry->dbname)));
> +
> +					StopHealthCheckWorker(entry->dboid);
>
> Here too I think we want a short comment about why we Stop the Worker that we failed to start...

Sure.

gkokolatos:
@DimCitus, any inputs on the latest failures? The log line regarding dropping the node and the python trace do not seem to match.

[3103.1](https://travis-ci.com/github/citusdata/pg_auto_failover/jobs/508630947)

gkokolatos:
On Thursday, 27 May 2021 15:07, Dimitri Fontaine ***@***.***> wrote:

>> [@DimCitus](https://github.com/DimCitus), any inputs on the latest failures? The log line regarding dropping the node and the python trace do not seem to match.
>>
>> [3103.1](https://travis-ci.com/github/citusdata/pg_auto_failover/jobs/508630947)
>
> I can tell you we have some instability in the test suite at the moment, that I can't reproduce locally. Usually it's not just Travis acting up, but those test failures are hard to fix. I sometimes just click restart on some of the jobs. When I have a whole series of 4 (the same test suite on 4 different versions of Postgres) failing at the same time mark or abouts, though, that's when I spend extra efforts on debugging the situation...

Yeah, I am not ready to drop this as instability, at least not at the moment. It is only a series of 3, with v12 passing.

Any hints as to where to start looking?

gkokolatos:
> @DimCitus, any inputs on the latest failures? The log line regarding dropping the node and the python trace do not seem to match.

Now I get that mismatch. I run the tests locally and they have been successful. Even on a successful run I get output as


```
test_multi_async.test_017_003_restart_nodes ... state of /tmp/multi_async/node3 is 'secondary', done waiting
state of /tmp/multi_async/node2 is 'catchingup', waiting for 'secondary' ...
state of /tmp/multi_async/node2 is 'secondary', done waiting
state of /tmp/multi_async/node1 is 'wait_primary', waiting for 'primary' ...
state of /tmp/multi_async/node1 is 'primary', done waiting
ok
pg_autoctl drop node --destroy failed
/home/georgios/postgres/build/bin/pg_autoctl drop node --destroy --pgdata /tmp/multi_async/node4

14:08:05 445933 FATAL Expected configuration file does not exists: "/home/georgios/.config/pg_autoctl/tmp/multi_async/node4/pg_autoctl.cfg"
14:08:05 445933 WARN  HINT: Check your PGDATA setting: "/tmp/multi_async/node4"


Ran 29 tests in 327.357s

OK
```

That is because there exists a specific case to destroy node4

```
def test_013_drop_node4():
    node4.destroy()
```

and at the end of the suite a call to cluster.destroy() is made, which will call node4.destroy() again. Pretty much any test case after  test_013_drop_node4 will contain that FATAL output regardless of the state of the suite.

Issue #717, opened 2021-06-07
Assertion failure in postgres related to background worker code


JelteF:
https://github.com/citusdata/pg_auto_failover/pull/711 introduces an assertion failure in postgres code. This in turn causes the monitor to go in a crash loop, when PG is built with assertions enabled. This happens on PG11, PG12 and PG13. The below backtrace is from PG13:

```
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
#1  0x00007fc5377dd921 in __GI_abort () at abort.c:79
#2  0x000055e32b5aa44b in ExceptionalCondition (conditionName=conditionName@entry=0x55e32b7039d0 "handle->slot < max_worker_processes", errorType=errorType@entry=0x55e32b602c08 "FailedAssertion", fileName=fileName@entry=0x55e32b7039f4 "bgworker.c", lineNumber=lineNumber@entry=1094) at assert.c:67
#3  0x000055e32b3d4ffa in GetBackgroundWorkerPid (handle=handle@entry=0x55e32c2828f0, pidp=pidp@entry=0x7ffd1c921d04) at bgworker.c:1094
#4  0x00007fc53500375e in HealthCheckWorkerLauncherMain (arg=<optimized out>) at health_check_worker.c:296
#5  0x000055e32b3d4a09 in StartBackgroundWorker () at bgworker.c:879
#6  0x000055e32b3e387a in do_start_bgworker (rw=rw@entry=0x55e32c207cf0) at postmaster.c:5870
#7  0x000055e32b3e39f1 in maybe_start_bgworkers () at postmaster.c:6095
#8  0x000055e32b3e3ba4 in reaper (postgres_signal_arg=<optimized out>) at postmaster.c:3040
#9  <signal handler called>
#10 0x00007fc5378b3dd7 in __GI___select (nfds=nfds@entry=12, readfds=readfds@entry=0x7ffd1c922c80, writefds=writefds@entry=0x0, exceptfds=exceptfds@entry=0x0, timeout=timeout@entry=0x7ffd1c922bf0) at ../sysdeps/unix/sysv/linux/select.c:41
#11 0x000055e32b3e48e8 in ServerLoop () at postmaster.c:1703
#12 0x000055e32b3e5d6f in PostmasterMain (argc=7, argv=<optimized out>) at postmaster.c:1412
#13 0x000055e32b32975d in main (argc=7, argv=0x55e32c1e1540) at main.c:210
[...]

PR #719, opened 2021-06-07
Keep the context around for as long as you need it

gkokolatos:
A previous commit [1], rather embarrassingly, failed to notice the reset of used
memory contexts. Additionally it was missdiagnosed for PostgreSQL versions 10
and 11. A simple run with assertions enabled was enough to expose the real bug.

This commit ammends that error and in passing restructures a bit the code in
HealthCheckWorkerLauncherMain to be hopefully cleaner.

Now the launcherContext is still reset in every database scan, but objects that
need to persist, do reside in the parent context.

[1] commit 785fd533bb

Addresses issue #717

### DIFF
--- a/src/monitor/health_check_worker.c
+++ b/src/monitor/health_check_worker.c
@@ -218,6 +218,8 @@ InitializeHealthCheckWorker(void)
 void
 HealthCheckWorkerLauncherMain(Datum arg)
 {
+	MemoryContext originalContext;
+
 	/* Establish signal handlers before unblocking signals. */
 	pqsignal(SIGHUP, pg_auto_failover_monitor_sighup);
 	pqsignal(SIGINT, SIG_IGN);
@@ -240,13 +242,17 @@ HealthCheckWorkerLauncherMain(Datum arg)
 														  ALLOCSET_DEFAULT_INITSIZE,
 														  ALLOCSET_DEFAULT_MAXSIZE);
 
-	MemoryContextSwitchTo(launcherContext);
 
 	while (!got_sigterm)
 	{
+		List *databaseList;
 		ListCell *databaseListCell;
 
-		List *databaseList = BuildDatabaseList();
+		originalContext = MemoryContextSwitchTo(launcherContext);
+
+		databaseList = BuildDatabaseList();
+
+		MemoryContextSwitchTo(originalContext);
 
 		foreach(databaseListCell, databaseList)
 		{
@@ -257,6 +263,7 @@ HealthCheckWorkerLauncherMain(Datum arg)
 			bool isFound = false;
 
 			LWLockAcquire(&HealthCheckHelperControl->lock, LW_EXCLUSIVE);
+
 
 			HealthCheckHelperDatabase *dbData = hash_search(HealthCheckWorkerDBHash,
 															(void *) &entry->dboid,
@@ -277,49 +284,26 @@ HealthCheckWorkerLauncherMain(Datum arg)
 				 * above. The HealthCheckWorkerDBHash only maintains verified
 				 * started entries. Thus we can only get BGWH_STARTED or
 				 * BGWH_STOPPED.
-				 *
-				 * NOTE: For PostgreSQL versions prior to 12, it is possible to
-				 * get erroneous results in GetBackgroundWorkerPid. In that
-				 * function, the handle we are providing is checked against a
-				 * localy held and maintained array. In affected versions, there
-				 * exists a window where our stored handle and that array to
-				 * have fallen out of sync. Such a window is more probable when
-				 * signalling the Postmaster e.g. to reload the configuration.
-				 *
-				 * When the entries are out of sync handle will be returned with
-				 * status BGWH_STOPPED and it will never be upgraded to
-				 * BGWH_STARTED, whereas in reality the bgworker is running and
-				 * most probably is healthy. From that moment onwards, we will
-				 * never be able to correctly assess the worker's status via the
-				 * provided API.
 				 */
 				if (GetBackgroundWorkerPid(handle, &pid) != BGWH_STARTED)
 				{
-					#if (PG_VERSION_NUM >= 120000)
 					ereport(WARNING,
 							(errmsg(
 								 "found stopped worker for pg_auto_failover "
 								 "health checks in \"%s\"",
 								 entry->dbname)));
-					#else
-					ereport(WARNING,
-							(errmsg(
-								 "stopping worker for pg_auto_failover "
-								 "health checks in \"%s\"",
-								 entry->dbname)));
-					#endif
 
 					/*
-					 * Now we know that either the worker has stopped or we
-					 * encountered the issue described in the NOTE above. We use
+					 * Now we know that the worker has stopped. We use
 					 * StopHealthCheckWorker to remove the entry from the
 					 * HealthCheckWorkerDBHash. That will force a retry in the
-					 * next scan of the databaselist. Furthermore, if the
-					 * status from GetBackgroundWorkerPid was not the correct
-					 * one, then StopHealthCheckWorker will also make certain
-					 * that the worker that we can no longer monitor correctly,
-					 * will be stopped. That will leave HealthCheckWorkerDBHash
-					 * in a consistent state.
+					 * next scan of the databaselist.
+					 *
+					 * Furthermore, if the status from GetBackgroundWorkerPid
+					 * was not the correct one, then StopHealthCheckWorker will
+					 * also make certain that the rogue worker will be stopped.
+					 * That will leave HealthCheckWorkerDBHash in a consistent
+					 * state.
 					 */
 					StopHealthCheckWorker(entry->dboid);
 				}
@@ -336,18 +320,16 @@ HealthCheckWorkerLauncherMain(Datum arg)
 				 * pid.
 				 */
 				dbData->workerPid = 0;
-			}
 
-			LWLockRelease(&HealthCheckHelperControl->lock);
-
-			if (handle)
-			{
 				/*
 				 * We need to release the lock for the worker to be able to
 				 * complete its startup procedure: the per-database worker
 				 * takes the control lock in SHARED mode to edit its own PID in
 				 * its own entry in HealthCheckWorkerDBHash.
-				 *
+				 */
+				LWLockRelease(&HealthCheckHelperControl->lock);
+
+				/*
 				 * WaitForBackgroundWorkerStartup will wait for worker to start;
 				 * thus, BGWH_NOT_YET_STARTED is never returned. However, if the
 				 * postmaster has died, it will give up and return
@@ -355,13 +337,6 @@ HealthCheckWorkerLauncherMain(Datum arg)
 				 * signaled to stop and we will exit further down. For good
 				 * measure though, do verify the process did actually start
 				 * before marking it as Active.
-				 *
-				 * NOTE: Since WaitForBackgroundWorkerStartup is internally
-				 * using GetBackgroundWorkerPid, it suffers from the same
-				 * symptoms as in the comment above for PostgreSQL versions
-				 * prior to 12. Since we do not want to mark it as Active
-				 * without really knowning, it is removed from the list and will
-				 * be readded on the next round.
 				 */
 				if (WaitForBackgroundWorkerStartup(handle, &pid) == BGWH_STARTED)
 				{
@@ -371,32 +346,28 @@ HealthCheckWorkerLauncherMain(Datum arg)
 								 "started worker for pg_auto_failover "
 								 "health checks in \"%s\"",
 								 entry->dbname)));
-				}
-				else
-				{
-					ereport(WARNING,
-							(errmsg(
-								 "failed to start worker for pg_auto_failover "
-								 "health checks in \"%s\"",
-								 entry->dbname)));
-
-					/*
-					 * Similarly to the comment above, we either failed to start
-					 * the worker, or we encountered the issue described in the
-					 * NOTE. We use StopHealthCheckWorker to remove the entry
-					 * from the HealthCheckWorkerDBHash so that it will be
-					 * retried in the next databaselist scan. The call to kill()
-					 * the failed worker in StopHealthCheckWorker() will take
-					 * place only if the worker has actually started, in which
-					 * case it means that we encountered the issue described in
-					 * NOTE and we can no longer properly assess the state of
-					 * that worker. After the call to StopHealthCheckWorker, we
-					 * know that HealthCheckWorkerDBHash will contain only
-					 * started and able to monitor entries.
-					 */
-					StopHealthCheckWorker(entry->dboid);
+					continue;
 				}
 			}
+
+			LWLockRelease(&HealthCheckHelperControl->lock);
+
+			/*
+			 * Similarly to the comment above, we either failed to start
+			 * the worker, or we failed to register it.
+			 *
+			 * NOTE. We use StopHealthCheckWorker to remove the entry
+			 * from the HealthCheckWorkerDBHash so that it will be
+			 * retried in the next databaselist scan. The call to kill()
+			 * the failed worker in StopHealthCheckWorker() will take
+			 * place only if a handle was registered.
+			 */
+			ereport(WARNING,
+					(errmsg("failed to %s worker for pg_auto_failover "
+							"health checks in \"%s\"",
+							handle ? "start" : "register",
+							entry->dbname)));
+			StopHealthCheckWorker(entry->dboid);
 		}
 
 		MemoryContextReset(launcherContext);
@@ -409,6 +380,9 @@ HealthCheckWorkerLauncherMain(Datum arg)
 			ProcessConfigFile(PGC_SIGHUP);
 		}
 	}
+
+	MemoryContextReset(launcherContext);
+	MemoryContextSwitchTo(originalContext);
 }
 
 
--- a/src/monitor/health_check_worker.c
+++ b/src/monitor/health_check_worker.c
@@ -218,7 +218,7 @@ InitializeHealthCheckWorker(void)
 void
 HealthCheckWorkerLauncherMain(Datum arg)
 {
-	MemoryContext originalContext;
+	MemoryContext originalContext = CurrentMemoryContext;
 
 	/* Establish signal handlers before unblocking signals. */
 	pqsignal(SIGHUP, pg_auto_failover_monitor_sighup);


JelteF:
This is definitly a lot better than before. I tried this locally and it allows me to run pg_auto_failover again when PG is compiled with assertions enabled. 

I'm not entirely sure this fixes everything though. Travis still randomly fails sometimes with "FATAL: the database is in recovery mode", which indicates that there has been a crash before:
```
test_multi_async.test_014_001_fail_node1 ... state of /tmp/multi_async/node2 is 'secondary', waiting for 'stop_replication' ...
ERROR
ERROR: test_multi_async.test_014_001_fail_node1
  File "/opt/python/3.7.6/lib/python3.7/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/travis/build/citusdata/pg_auto_failover/tests/test_multi_async.py", line 241, in test_014_001_fail_node1
    assert node2.wait_until_state(target_state="stop_replication", timeout=120)
  File "/home/travis/build/citusdata/pg_auto_failover/tests/pgautofailover_utils.py", line 1153, in wait_until_state
    current_state, assigned_state = self.get_state()
  File "/home/travis/build/citusdata/pg_auto_failover/tests/pgautofailover_utils.py", line 1244, in get_state
    self.group,
  File "/home/travis/build/citusdata/pg_auto_failover/tests/pgautofailover_utils.py", line 368, in run_sql_query
    conn = psycopg2.connect(self.connection_string())
  File "/opt/python/3.7.6/lib/python3.7/site-packages/psycopg2/__init__.py", line 130, in connect
    conn = _connect(dsn, connection_factory=connection_factory, **kwasync)
psycopg2.OperationalError: FATAL:  the database system is in recovery mode
````

**Comment by gkokolatos** [comment]
On Tuesday, 8 June 2021 10:34, Jelte Fennema ***@***.***> wrote:

> This is definitly a lot better than before. I tried this locally and it allows me to run pg_auto_failover again when PG is compiled with assertions enabled.

Thanks for that!


>
> I'm not entirely sure this fixes everything though. Travis still randomly fails sometimes with "FATAL: the database is in recovery mode", which indicates that there has been a crash before:


I started to look at that myself. I have proven in an environment that is irrelevant to this commit.

Basically, if a primary node (node1 In this case) fails, then other nodes get promoted and change state. During this process,
a node that was secondary (node2) will be stopped, some configuration on it will change, and then get started again. This is
controlled by pg_autoctl. If the timing is such, while it is waiting for a state change, it will issue a connect command and it
might fail either with connection refused or the database is in recovery. Both of those are expected parts of the transition.

I am betting that some test instability exists around wait_*_state calls.

**Comment by gkokolatos** [comment]
On Tuesday, June 8th, 2021 at 10:45, Dimitri Fontaine ***@***.***> wrote:

> > This is controlled by pg_autoctl. If the timing is such, while it is waiting for a state change, it will issue a connect command and it might fail either with connection refused or the database is in recovery. Both of those are expected parts of the transition. I am betting that some test instability exists around wait_*_state calls.
>
> That is surprising to me. In the test failures I had a look at, including the one here shared by Jelte, the query that fails to connect runs `self.monitor.run_sql_query(...)`: so the monitor Postgres instance has crashed. That's not expected, and this bug here looks like a smoking gun...

Fair enough. Let me do some more digging then.

**Comment by gkokolatos** [comment]

On Tuesday, June 8th, 2021 at 10:48, ***@***.***> wrote:

> ‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐
>
> On Tuesday, June 8th, 2021 at 10:45, Dimitri Fontaine ***@***.*** wrote:
>
> > > This is controlled by pg_autoctl. If the timing is such, while it is waiting for a state change, it will issue a connect command and it might fail either with connection refused or the database is in recovery. Both of those are expected parts of the transition. I am betting that some test instability exists around wait_*_state calls.
> >
> > That is surprising to me. In the test failures I had a look at, including the one here shared by Jelte, the query that fails to connect runs `self.monitor.run_sql_query(...)`: so the monitor Postgres instance has crashed. That's not expected, and this bug here looks like a smoking gun...
>
> Fair enough. Let me do some more digging then.

Debugging this was indeed fun. I mean it. Now I have to explain my findings which is not as straight forward.


I will try to do it here and I accept that I might fail to communicate well what I have found. Please correct me
where you think I am wrong.

The setup was as follows:

To start, I tried to run the test case in a loop until it fails. I am using the latest pg13 version.
Also I wanted to have the logs and data files available in the case that a failure happened. I modified
utils so that the teardown step will not remove them by removing the drop node:

--- a/tests/pgautofailover_utils.py
+++ b/tests/pgautofailover_utils.py
@@ -1109,10 +1109,9 @@ class DataNode(PGNode):
         try:
             destroy = PGAutoCtl(self)
             destroy.execute(
-                "pg_autoctl drop node --destroy",
+                "pg_autoctl drop node",
                 "drop",
                 "node",
-                "--destroy",
                 timeout=3,
             )
         except Exception as e:
@@ -1697,7 +1696,7 @@ class MonitorNode(PGNode):
         try:
             destroy = PGAutoCtl(self)
             destroy.execute(
-                "pg_autoctl destroy monitor", "drop", "monitor", "--destroy"
+                "pg_autoctl destroy monitor", "drop", "monitor"
             )
         except Exception as e:
             print(str(e))


Then I wrote a small bash function to remove the state/config files that pg_autoctl
is using along with the datafiles and named it remove_pgauf_state.

Finally I run the offending test in a loop like so:
$ for i in {1..100}; do remove_pgauf_state; TEST=test_multi_async make test || break; done;

True enough after a few successful runs, the test failed at the same point.


ERROR: test_multi_async.test_014_001_fail_node1
  File "/usr/lib/python3/dist-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/georgios/pg_auto_failover/tests/test_multi_async.py", line 241, in test_014_001_fail_node1
    assert node2.wait_until_state(target_state="stop_replication", timeout=120)
  File "/home/georgios/pg_auto_failover/tests/pgautofailover_utils.py", line 1152, in wait_until_state
    current_state, assigned_state = self.get_state()
  File "/home/georgios/pg_auto_failover/tests/pgautofailover_utils.py", line 1236, in get_state
    results = self.monitor.run_sql_query(
  File "/home/georgios/pg_auto_failover/tests/pgautofailover_utils.py", line 368, in run_sql_query
    conn = psycopg2.connect(self.connection_string())
  File "/usr/lib/python3/dist-packages/psycopg2/__init__.py", line 126, in connect
    conn = _connect(dsn, connection_factory=connection_factory, **kwasync)
psycopg2.OperationalError: FATAL:  the database system is in recovery mode



Great, now there is a reproduction, core files and log files.

Starting with the core, we see this:


#0  0x00007f3230e24cb8 in TupleToAutoFailoverNode (tupleDescriptor=0x55733a632e10, heapTuple=0x7f7f7f7f7f7f7f7f) at node_metadata.c:107
#1  0x00007f3230e2ddad in SetNodeHealthState (nodeId=4, nodeName=0x55733a5e1a20 "node_4", nodeHost=0x55733a5e1a40 "172.27.1.6", nodePort=5432, previousHealthState=1, healthState=0) at health_check_metadata.c:223
#2  0x00007f3230e2aab1 in ManageHealthCheck (healthCheck=0x55733a5e1c40, currentTime=...) at health_check_worker.c:850
#3  0x00007f3230e2a528 in DoHealthChecks (healthCheckList=0x55733a5e1c98) at health_check_worker.c:686
#4  0x00007f3230e2a25c in HealthCheckWorkerMain (arg=16385) at health_check_worker.c:577

and easy enough heapTuple is a nope: 0x7f7f7f7f7f7f7f7f

heapTuple comes from SPI_tuptable->vals[0]
which was set by SPI_execute


The code is found in SetNodeHealthState() and is the following:


    if (HaMonitorHasBeenLoaded())
    {
        initStringInfo(&query);
        appendStringInfo(&query,
                         "UPDATE " AUTO_FAILOVER_NODE_TABLE
                         "   SET health = %d, healthchecktime = now() "
                         " WHERE nodeid = %d "
                         "   AND nodehost = %s AND nodeport = %d "
                         " RETURNING node.*",
                         healthState,
                         nodeId,
                         quote_literal_cstr(nodeHost),
                         nodePort);

        pgstat_report_activity(STATE_RUNNING, query.data);

        spiStatus = SPI_execute(query.data, false, 0);
        Assert(spiStatus == SPI_OK_UPDATE_RETURNING);

        if (healthState != previousHealthState)
        {
            HeapTuple heapTuple = SPI_tuptable->vals[0];
            AutoFailoverNode *pgAutoFailoverNode =
                TupleToAutoFailoverNode(SPI_tuptable->tupdesc, heapTuple);


This codepath assumes that the UPDATE is successful by checking (asserting but is the same)
spiStatus. However, what it does not do, is to check the number of returning rows. In this
run SPI_processed is actually 0.

(gdb) p SPI_processed
$12 = 0

That actually means that the quals of the update did not match any of the rows.

True enough, the logs have captured that node_4 has actually being removed:

 [267769] STATEMENT:  SELECT pgautofailover.remove_node($1, $2)
 [267769] LOG:  Setting goal state of node 1 "node_1" (172.27.1.3:5432) to apply_settings after removing standby node 4 "node_4" (172.27.1.6:5432) from formation default.
 [267769] STATEMENT:  SELECT pgautofailover.remove_node($1, $2)
<snip>
 [265002] LOG:  background worker "pg_auto_failover monitor healthcheck worker pg_auto_failover" (PID 265013) was terminated by signal 11: Segmentation fault
 [265002] DETAIL:  Failed process was running: UPDATE pgautofailover.node   SET health = 0, healthchecktime = now()  WHERE nodeid = 4    AND nodehost = '172.27.1.6' AND nodeport = 5432  RETURNING node.*



Given the above, I understand that there are two processes involved and one of them removes an
entry from the table
    const char *deleteQuery =
        "DELETE FROM " AUTO_FAILOVER_NODE_TABLE
        " WHERE nodeid = $1";
and another one executes an UPDATE expecting definitely one result on the same table. There does
seem that there exists a window that the this will not hold true.


What have I missed?

**Comment by DimCitus** [comment]
> This codepath assumes that the UPDATE is successful by checking (asserting but is the same) spiStatus. However, what it does not do, is to check the number of returning rows. In this run SPI_processed is actually 0. (gdb) p SPI_processed $12 = 0 That actually means that the quals of the update did not match any of the rows. 
>True enough, the logs have captured that node_4 has actually being removed: [267769] STATEMENT: SELECT pgautofailover.remove_node($1, $2) [267769] 

> Given the above, I understand that there are two processes involved and one of them removes an entry from the table const char *deleteQuery = "DELETE FROM " AUTO_FAILOVER_NODE_TABLE " WHERE nodeid = $1"; and another one executes an UPDATE expecting definitely one result on the same table. There does seem that there exists a window that the this will not hold true. What have I missed?

Thanks for the debugging sessions @gkokolatos ; great detective work! As you're saying, with a repro that leaves hints behind that's a better use of our time, to thanks for taking to time to make that happen. I think your analysis is spot-on, and even if other problems might cause this kind of bug, this one has to be fixed. I'm going to cook a patch/PR for this one and then link it to this issue here. Thanks again!

**Comment by gkokolatos** [comment]
On Tuesday, June 8th, 2021 at 17:01, Dimitri Fontaine ***@***.***> wrote:

> > This codepath assumes that the UPDATE is successful by checking (asserting but is the same) spiStatus. However, what it does not do, is to check the number of returning rows. In this run SPI_processed is actually 0. (gdb) p SPI_processed $12 = 0 That actually means that the quals of the update did not match any of the rows.
> >
> > True enough, the logs have captured that node_4 has actually being removed: [267769] STATEMENT: SELECT pgautofailover.remove_node($1, $2) [267769]
>
> > Given the above, I understand that there are two processes involved and one of them removes an entry from the table const char *deleteQuery = "DELETE FROM " AUTO_FAILOVER_NODE_TABLE " WHERE nodeid = $1"; and another one executes an UPDATE expecting definitely one result on the same table. There does seem that there exists a window that the this will not hold true. What have I missed?
>
> Thanks for the debugging sessions @gkokolatos ; great detective work! As you're saying, with a repro that leaves hints behind that's a better use of our time, to thanks for taking to time to make that happen. I think your analysis is spot-on, and even if other problems might cause this kind of bug, this one has to be fixed. I'm going to cook a patch/PR for this one and then link it to this issue here. Thanks again!

Thank you Dimitri. It is a pleasure to help.

I am looking forward to your PR. I have cooked something small to move things forward but better to wait for your solution first.

While at it, I have found the cause of the second failure in travis. I was able to reproduce at least once ever 10 runs.

The output in travis and indeed a failed run is:

4007test_multi_alternate_primary_failures.test_002_001_add_two_standbys ... state of /tmp/multi_alternate_primary_failures/node2 is 'secondary', done waiting
4008ERROR
4009
4010======================================================================
4011ERROR: test_multi_alternate_primary_failures.test_002_001_add_two_standbys
4012----------------------------------------------------------------------
4013Traceback (most recent call last):
4014  File "/opt/python/3.7.6/lib/python3.7/site-packages/nose/case.py", line 198, in runTest
4015    self.test(*self.arg)
4016  File "/home/travis/build/citusdata/pg_auto_failover/tests/test_multi_alternate_primary_failures.py", line 56, in test_002_001_add_two_standbys
4017    assert node2.has_needed_replication_slots()
4018  File "/home/travis/build/citusdata/pg_auto_failover/tests/pgautofailover_utils.py", line 1561, in has_needed_replication_slots
4019    if self.pgmajor() == 10:
4020  File "/home/travis/build/citusdata/pg_auto_failover/tests/pgautofailover_utils.py", line 607, in pgmajor
4021    self.pgversion()
4022  File "/home/travis/build/citusdata/pg_auto_failover/tests/pgautofailover_utils.py", line 597, in pgversion
4023    self.run_sql_query("show server_version_num")[0][0]
4024  File "/home/travis/build/citusdata/pg_auto_failover/tests/pgautofailover_utils.py", line 368, in run_sql_query
4025    conn = psycopg2.connect(self.connection_string())
4026  File "/opt/python/3.7.6/lib/python3.7/site-packages/psycopg2/__init__.py", line 130, in connect
4027    conn = _connect(dsn, connection_factory=connection_factory, **kwasync)
4028psycopg2.OperationalError: could not connect to server: Connection refused
4029	Is the server running on host "172.27.1.4" and accepting
4030	TCP/IP connections on port 5432?

Now what happens here is a small timing problem.

node2.has_needed_replication_slots() requires to check the version, which runs a query on that node
(call self.run_sql_query("show server_version_num")[0][0] the trace above). That node has been assigned
and reached state secondary yet the primary has not reached its primary state. So when there is a window
that the query will be executed while node2 is reloading config/state due to the change in node1.

By changing the order of the calls in that test, to call has_needed_replication_slots() after node1 has
finished the transition and the cluster is stable, I was able to execute the tests successfully more than
100 times in a row. Note that this pattern (calling has_needed_replication_slots after all nodes have
reached the desired state) is used throughout the rest of the suite.

In short, I think that this failure is not relevant to the current patch. I am more than happy to provide
with a patch to fix the testing if you agree.

**Comment by DimCitus** [comment]
```
4007test_multi_alternate_primary_failures.test_002_001_add_two_standbys ... state of /tmp/multi_alternate_primary_failures/node2 is 'secondary', done waiting 
4008ERROR 
4009 
4010====================================================================== 
4011ERROR: test_multi_alternate_primary_failures.test_002_001_add_two_standbys 
4012---------------------------------------------------------------------- 
4013Traceback (most recent call last): 
4014 File "/opt/python/3.7.6/lib/python3.7/site-packages/nose/case.py", line 198, in runTest 
4015 self.test(*self.arg) 
4016 File "/home/travis/build/citusdata/pg_auto_failover/tests/test_multi_alternate_primary_failures.py", line 56, in test_002_001_add_two_standbys 
4017 assert node2.has_needed_replication_slots() 
4018 File "/home/travis/build/citusdata/pg_auto_failover/tests/pgautofailover_utils.py", line 1561, in has_needed_replication_slots 
4019 if self.pgmajor() == 10: 
4020 File "/home/travis/build/citusdata/pg_auto_failover/tests/pgautofailover_utils.py", line 607, in pgmajor 
4021 self.pgversion() 
4022 File "/home/travis/build/citusdata/pg_auto_failover/tests/pgautofailover_utils.py", line 597, in pgversion 
4023 self.run_sql_query("show server_version_num")[0][0] 
4024 File "/home/travis/build/citusdata/pg_auto_failover/tests/pgautofailover_utils.py", line 368, in run_sql_query 
4025 conn = psycopg2.connect(self.connection_string()) 
4026 File "/opt/python/3.7.6/lib/python3.7/site-packages/psycopg2/__init__.py", line 130, in connect 
4027 conn = _connect(dsn, connection_factory=connection_factory, **kwasync) 
4028psycopg2.OperationalError: could not connect to server: Connection refused 
4029 Is the server running on host "172.27.1.4" and accepting 
4030 TCP/IP connections on port 5432? 
```

> Now what happens here is a small timing problem. node2.has_needed_replication_slots() requires to check the version, which runs a query on that node (call self.run_sql_query("show server_version_num")[0][0] the trace above). That node has been assigned and reached state secondary yet the primary has not reached its primary state. So when there is a window that the query will be executed while node2 is reloading config/state due to the change in node1. By changing the order of the calls in that test, to call has_needed_replication_slots() after node1 has finished the transition and the cluster is stable, I was able to execute the tests successfully more than 100 times in a row. Note that this pattern (calling has_needed_replication_slots after all nodes have reached the desired state) is used throughout the rest of the suite. In short, I think that this failure is not relevant to the current patch. I am more than happy to provide with a patch to fix the testing if you agree.

I see. For the sake of reducing Travis duties and wait time, I suppose it'd be quite nice to also have this test suite fix part of this very PR, if you'd agree to it?

DimCitus (approved):
LGTM

DimCitus:
> I see. For the sake of reducing Travis duties and wait time, I suppose it'd be quite nice to also have this test suite fix part of this very PR, if you'd agree to it?

Well. Let me take that back. One PR per task is cleaner. I'll merge this one after an update, and please have a look at refactoring/fixing the test suite the way you said, in another PR?

gkokolatos:
Sent from ProtonMail for iOS

On Tue, Jun 8, 2021 at 17:55, Dimitri Fontaine ***@***.***> wrote:

>> I see. For the sake of reducing Travis duties and wait time, I suppose it'd be quite nice to also have this test suite fix part of this very PR, if you'd agree to it?
>
> Well. Let me take that back. One PR per task is cleaner. I'll merge this one after an update, and please have a look at refactoring/fixing the test suite the way you said, in another PR?

It should absolutely be a different PR.
I didn’t phrase it well.
I will submit one soon(tm)

PR #720, opened 2021-06-08
Fix a race condition issue around health check updates.

DimCitus:
When a node is being dropped, the DELETE and the health check UPDATE might
work concurrently, and the monitor's code that drives the UPDATE failed to
check that a row was processed by the SPI query.

See #719 where @gkokolatos has done a great detective work!


gkokolatos:
:+1: from me! I have tested similar and it works.

Also thanks for the shout out :)

DimCitus:
> 👍 from me! I have tested similar and it works.

Unfortunately (not sure why or how to fix that) GitHub doesn't allow me to add you as a reviewer, nor to assign you on some issues/PRs, depending on who creates them.

JelteF (approved):

PR #728, opened 2021-06-14
Add build parameters on travis file about postgres 14

gurkanindibay:
postgres14


DimCitus:
Thanks @gurkanindibay ! we had a couple of issues in the code itself, and I have fixed them in #716 where I also copied your work here, and then done some Travis improvements for reducing the run-time.

PR #733, opened 2021-06-17
Use sysctl(3) on BSD

eradman:
sysctlbyname(3) is a convenience wrapper for sysctl(3) that is implemented on FreeBSD and MacOS, but not on OpenBSD.

Include `sys/param.h` to ensure that BSD is defined.


src/bin/pg_autoctl/system_utils.c
```diff
@@ -73,26 +74,33 @@ get_system_info_linux(SystemInfo *sysInfo)
...
 /*
- * FreeBSD, OpenBSD, and darwin use the sysctlbyname(3) API.
+ * FreeBSD, OpenBSD, and darwin use the sysctl(3) API.
  */
 #if defined(__APPLE__) || defined(BSD)
 static bool
 get_system_info_bsd(SystemInfo *sysInfo)
 {
 	unsigned int ncpu = 0;      /* the API requires an integer here */
+	int mib[2];
```
DimCitus, line 84: What about using the following style, that would look more like what we use in the rest of the codebase?

``` c
    int ncpuMIB[2] = { CTL_HW, HW_NCPU };
```

src/bin/pg_autoctl/system_utils.c
```diff
@@ -73,26 +74,33 @@ get_system_info_linux(SystemInfo *sysInfo)
...
 		log_error("Failed to probe number of CPUs: %m");
 		return false;
 	}
 
 	sysInfo->ncpu = (unsigned short) ncpu;
 
-	if (sysctlbyname("hw.memsize", &(sysInfo->totalram), &memSize, NULL, 0) != 0)
+	mib[0] = CTL_HW;
+	mib[1] = HW_PHYSMEM;
+
```
DimCitus, line 102: What about using the following style, that would look more like what we use in the rest of the codebase?

``` c
    int ramMIB[2] = { CTL_HW, HW_PHYSMEM };
```

DimCitus (commented):
I think we can improve the style to look more like what we do elsewhere in the code. Also, can you list which OSes you have tested already with the changes?

src/bin/pg_autoctl/system_utils.c
eradman: Good suggestion; will do

DimCitus:
Hi @eradman ; I just had a try of your branch locally on my laptop, and it fails to detect the amount of RAM properly when using your current code. I suppose the old API needs to use another MIB?

``` diff
- 12:23:02 43377 DEBUG pgtuning.c:85: Detected 12 CPUs and 16 GB total RAM on this server
+ 12:22:26 42248 DEBUG pgtuning.c:85: Detected 12 CPUs and 2048 MB total RAM on this server
```

That's the diff when running the following command once in the master's branch, then once in your PR branch:

```
$ PG_AUTOCTL_DEBUG=1 ./src/bin/pg_autoctl/pg_autoctl do pgsetup tune --pgdata /tmp -vv
```

eradman (commented):
This change was not nearly as easy as I thought since sysctl API has diverged between various BSD-like platforms

Other options:

* Patch this in OpenBSD ports
* Create a compatibility function sysctlbyname(3) for OpenBSD

DimCitus:
I just tried it with a macOS and confirm this now works, and I think we're good with the code the way you've submitted it. I trust you have validated OpenBSD and FreeBSD variants, and intend to merge after Travis confirmed a test run.

DimCitus (approved):

DimCitus:
Thanks for your contribution @eradman !

PR #737, opened 2021-06-23
Call pg_rewind when creating a secondary from an existing PGDATA.

DimCitus:
The existing PGDATA might have come from a former primary that's not in the
new timeline and where pg_rewind is needed to go before the fork LSN. Also,
it could be a standby node that has been used for PITR.

We should decide if we want to add such a process to `fsm_follow_new_primary`...


JelteF:
FYI Tests are failing

PR #740, opened 2021-06-24
Introduce hooks to implement refreshing the otherNodes array.

DimCitus:
This way we allow extensions to pg_auto_failover to implement specific
processing when the otherNodes array changes, when that's needed.

### DIFF
--- a/src/bin/pg_autoctl/keeper.c
+++ b/src/bin/pg_autoctl/keeper.c
@@ -2039,9 +2039,7 @@ keeper_refresh_other_nodes(Keeper *keeper, bool forceCacheInvalidation)
 	Monitor *monitor = &(keeper->monitor);
 	KeeperConfig *config = &(keeper->config);
 
-	NodeAddressArray *otherNodesArray = &(keeper->otherNodes);
 	NodeAddressArray newNodesArray = { 0 };
-	NodeAddressArray diffNodesArray = { 0 };
 
 	int64_t nodeId = keeper->state.current_node_id;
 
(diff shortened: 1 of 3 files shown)


src/bin/pg_autoctl/keeper.c
```diff
@@ -2064,30 +2062,83 @@ keeper_refresh_other_nodes(Keeper *keeper, bool forceCacheInvalidation)
...
+						  NodeAddressArray *newNodesArray,
+						  bool forceCacheInvalidation)
+{
+	bool success = true;
+
+	for (int index = 0; KeeperRefreshHooks[index]; index++)
+	{
+		KeeperNodesArrayRefreshFunction hookFun = KeeperRefreshHooks[index];
+
+		bool ret = (*hookFun)(keeper, newNodesArray, forceCacheInvalidation);
```
onderkalaci, line 2096: I think my only concern here is that what if the first hook succeeds and has some side-effects, and the second hook fails.

We'd report this as a failure to the upper levels of the function but have done the side-effects.

Do you see any problems with that?

DimCitus: I've done it that way on purpose, as usual the hook should expect to be called again when it reports false. I think it's a good way to approach these hooks, because we need to decide if we should go with the new array of nodes, and we should certainly call the hooks again with the same diff/changes when they have failed previously to install the changes...

onderkalaci (approved):
I asked an internal question on teams, please feel free to merge after having a quick look on that

PR #743, opened 2021-06-28
Have drop node --force --destroy work as expected again

JelteF:
The following command would get stuck for 30 seconds when the pg_autoctl service was still running.
```
pg_autoctl drop node --pgdata node1 --force --destroy
```

### DIFF
--- a/src/bin/pg_autoctl/cli_drop_node.c
+++ b/src/bin/pg_autoctl/cli_drop_node.c
@@ -608,28 +608,42 @@ cli_drop_local_node(KeeperConfig *config, bool dropAndDestroy)
 	pid_t pid = 0;
 
 	/*
-	 * If --force isn't used then any running pg_autoctl process will detect
-	 * that it is dropped and clean itself up nicely and finally it will exit.
-	 * We give the process 30 seconds to exit by itself, if it didn't probably
-	 * something is wrong and we manually kill it. We will then complete the
-	 * transition to dropped from this process instead.
+	 * Before continuing we need to make sure that a currently running service
+	 * has stopped.
+	 *
+	 * If --force isn't used then a running pg_autoctl process will detect that
+	 * it is dropped and clean itself up nicely and finally it will exit. We
+	 * give the process 30 seconds to exit by itself.
 	 *
 	 * If --force is used, we skip the transition to "dropped". So a currently
 	 * running process won't realise it's dropped, so it will not exit by
-	 * itself. So there's no point in waiting for 30 seconds. Instead we
-	 * manually kill it using SIGQUIT right away, because we would still like
-	 * the process to be stopped in that case.
+	 * itself. So there's no point in waiting for 30 seconds.
 	 */
-	int quit_timeout = dropForce ? 0 : 30;
-
-	if (!wait_for_pid_to_exit(config->pathnames.pid, quit_timeout, &pid))
+	bool stopped;
+	if (!dropForce)
 	{
-		if (pid == 0)
+		if (!wait_for_process_to_stop(config->pathnames.pid, 30, &stopped, &pid))
 		{
 			/* errors have already been logged */
 			exit(EXIT_CODE_INTERNAL_ERROR);
 		}
+	}
+	else
+	{
+		if (!is_process_stopped(config->pathnames.pid, &stopped, &pid))
+		{
+			/* errors have already been logged */
+			exit(EXIT_CODE_INTERNAL_ERROR);
+		}
+	}
 
+	/*
+	 * If the service is not stopped yet,  we just want to process to exit
+	 * so we can take over. This can happen either because --force was used
+	 * or because 30 seconds was not enough time for the service to exit.
+	 */
+	if (!stopped)
+	{
 		/* if the service isn't terminated, signal it to quit now */
 		log_info("Sending signal %s to pg_autoctl process %d",
 				 signal_to_string(SIGQUIT),
@@ -641,7 +655,8 @@ cli_drop_local_node(KeeperConfig *config, bool dropAndDestroy)
 			exit(EXIT_CODE_INTERNAL_ERROR);
 		}
 
-		if (!wait_for_pid_to_exit(config->pathnames.pid, 30, &pid))
+		if (!wait_for_process_to_stop(config->pathnames.pid, 30, &stopped, &pid) ||
+			!stopped)
 		{
 			log_fatal("Failed to stop the pg_autoctl process with pid %d", pid);
 			exit(EXIT_CODE_INTERNAL_ERROR);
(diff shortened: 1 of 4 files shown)


src/bin/pg_autoctl/cli_drop_node.c
```diff
@@ -578,7 +578,9 @@ cli_drop_local_node(KeeperConfig *config, bool dropAndDestroy)
 	 */
 	pid_t pid = 0;
 
-	if (!wait_for_pid_to_exit(config->pathnames.pid, 30, &pid))
+	int quit_timeout = dropForce ? 0 : 30;
```
DimCitus, line 581: I think this requires some comments, I'm not sure what's your intention there, or why we need different timeouts depending on whether the user is using `--force` or not?

src/bin/pg_autoctl/pidfile.c
```diff
@@ -510,8 +510,11 @@ wait_for_pid_to_exit(const char *pidfile, int timeout, pid_t *pid)
 	{
 		if (read_pidfile(pidfile, pid))
 		{
-			log_info("An instance of pg_autoctl is running with PID %d, "
-					 "waiting for it to stop.", *pid);
+			if (timeout > 0)
+			{
+				log_info("An instance of pg_autoctl is running with PID %d, "
+						 "waiting for it to stop.", *pid);
+			}
```
DimCitus, line 517: Here too some comments about our intention with the timeout-zero special case would be nice to have.

tests/pgautofailover_utils.py
```diff
@@ -1113,6 +1113,7 @@ def destroy(self):
                 "drop",
                 "node",
                 "--destroy",
+                "--force",
```
DimCitus, line 1116: Do we really want to change the way we run all of our test suite to use `--force`?

DimCitus (changes requested):
I think this needs more comments... it's not obvious what your intention are in those code paths...

src/bin/pg_autoctl/cli_drop_node.c
```diff
@@ -578,9 +578,23 @@ cli_drop_local_node(KeeperConfig *config, bool dropAndDestroy)
...
+	 * If --force isn't used then any running pg_autoctl process will detect
+	 * that it is dropped and clean itself up nicely and finally it will exit.
+	 * We give the process 30 seconds to exit by itself, if it didn't probably
+	 * something is wrong and we manually kill it. We will then complete the
+	 * transition to dropped from this process instead.
+	 *
+	 * If --force is used, we skip the transition to "dropped". So a currently
+	 * running process won't realise it's dropped, so it will not exit by
+	 * itself. So there's no point in waiting for 30 seconds. Instead we
+	 * manually kill it using SIGQUIT right away.
```
DimCitus, line 591: Can we bypass calling `wait_for_pid_to_exit` entirely in this case, then?

src/bin/pg_autoctl/pidfile.c
```diff
@@ -510,8 +510,15 @@ wait_for_pid_to_exit(const char *pidfile, int timeout, pid_t *pid)
...
 		if (read_pidfile(pidfile, pid))
 		{
-			log_info("An instance of pg_autoctl is running with PID %d, "
-					 "waiting for it to stop.", *pid);
+			/*
+			 * If we have a timeout of 0 passed in, this log message doesn't
+			 * make sense. We are not waiting for the process to stop.
+			 */
+			if (timeout > 0)
+			{
```
DimCitus, line 518: I think it would be easier with an early exit in this case.

DimCitus (changes requested):

src/bin/pg_autoctl/cli_drop_node.c
JelteF: I think it's still nice to stop a currently running pg_autoctl process in this case. The first `wait_for_pid_to_exit` then simply becomes an easy way to check if there is a running process.

src/bin/pg_autoctl/pidfile.c
JelteF: done

src/bin/pg_autoctl/cli_drop_node.c
```diff
@@ -578,9 +578,44 @@ cli_drop_local_node(KeeperConfig *config, bool dropAndDestroy)
...
+	 * If --force isn't used then a running pg_autoctl process will detect that
+	 * it is dropped and clean itself up nicely and finally it will exit. We
+	 * give the process 30 seconds to exit by itself.
+	 *
+	 * If --force is used, we skip the transition to "dropped". So a currently
+	 * running process won't realise it's dropped, so it will not exit by
+	 * itself. So there's no point in waiting for 30 seconds.
+	 */
+	bool stopped;
+	if (!dropForce)
```
DimCitus, line 594: Nitpick: if we're going to have code paths for the both true and false values, can we do them in the easy to read order, `if (dropForce) { ... } else { ... }`?

src/bin/pg_autoctl/cli_drop_node.c
```diff
@@ -578,9 +578,44 @@ cli_drop_local_node(KeeperConfig *config, bool dropAndDestroy)
...
+	{
+		if (!wait_for_process_to_stop(config->pathnames.pid, 30, &stopped, &pid))
+		{
+			/* errors have already been logged */
+			exit(EXIT_CODE_INTERNAL_ERROR);
+		}
+	}
+	else
+	{
+		if (!is_process_stopped(config->pathnames.pid, &stopped, &pid))
```
DimCitus, line 604: The function `is_process_stopped` is used a lot and often with a negation before it. I think it would be all easier if we had the function `pidfile_process_is_running` (similar to `pg_setup_is_running` and others).

src/bin/pg_autoctl/pidfile.c
```diff
@@ -499,47 +499,59 @@ pidfile_as_json(JSON_Value *js, const char *pidfile, bool includeStatus)
 }
 
 
+bool
+is_process_stopped(const char *pidfile, bool *stopped, pid_t *pid)
```
DimCitus, line 503: Missing function level header. Also, as before, what about having `pidfile_process_is_running` instead?

DimCitus (changes requested):
Thanks for the changes. I think we only need a last pass of small changes!

src/bin/pg_autoctl/cli_drop_node.c
JelteF: Is discussed over chat, flipping this brings in other consistency problems. Leaving it like this.

src/bin/pg_autoctl/pidfile.c
JelteF: Is discussed over chat, flipping this brings in other consistency problems. Leaving it like this.

DimCitus (approved):

PR #744, opened 2021-06-29
Review and update docs with the recent PRs.

### LINKED ISSUE #667: `pg_autoctl config set replication.password` is missing from the documentation
It is mentioned in the last code block of https://pg-auto-failover.readthedocs.io/en/master/security.html#authentication-with-passwords, but is missing from the man page https://pg-auto-failover.readthedocs.io/en/master/ref/pg_autoctl_config_set.html

DimCitus:
Fixes #667 
See also #382

### DIFF
--- a/docs/ref/pg_autoctl_show_state.rst
+++ b/docs/ref/pg_autoctl_show_state.rst
@@ -79,9 +79,10 @@ The ``pg_autoctl show state`` outputs includes the following columns:
 
   - TLI: LSN
 
-	Timeline identifier and Postgres Log Sequence Number, which is the
-	current position in the WAL stream. This is a hexadecimal number. See
-	`pg_lsn`__ for more information.
+	Timeline identifier (TLI) and Postgres Log Sequence Number (LSN).
+
+	The LSN is the current position in the Postgres WAL stream. This is a
+	hexadecimal number. See `pg_lsn`__ for more information.
 
 	__ https://www.postgresql.org/docs/current/datatype-pg-lsn.html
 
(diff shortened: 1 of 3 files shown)


docs/ref/pg_autoctl_show_state.rst
```diff
@@ -77,14 +77,20 @@ The ``pg_autoctl show state`` outputs includes the following columns:
...
 
-  - LSN
+  - TLI: LSN
 
-	Postgres Log Sequence Number, which is the current position in the WAL
-	stream. This is a hexadecimal number. See `pg_lsn`__ for more
-	information.
+	Timeline identifier and Postgres Log Sequence Number, which is the
+	current position in the WAL stream. This is a hexadecimal number. See
+	`pg_lsn`__ for more information.
```
JelteF, line 84: This is a bit confusing to read, since it talks about both the TLI and the LSN. It then only goes into what the LSN means without really making clear that that's what's happening. I would suggest something like this:

> 	Postgres Timeline Identifier and Log Sequence Number.
> 
>   The current `timeline identifier`__ is incremented each time a failover happens, or
>	when doing Point In Time Recovery. A node can only reach the secondary
>	state when it is on the same timeline as its primary node.
>
>   The Log Sequence Number is the current position in the WAL stream. This is a 
>   hexadecimal number. See `pg_lsn`__ for more information.

JelteF (approved):

docs/ref/pg_autoctl_show_state.rst
DimCitus: Good point, edited that way now.

PR #745, opened 2021-06-29
Desultory refactoring of the code.

DimCitus:


src/monitor/pgautofailover--1.5--1.6.sql
```diff
@@ -554,26 +554,27 @@ CREATE FUNCTION pgautofailover.current_state
...
 
-grant execute on function pgautofailover.current_state(text)
-   to autoctl_node;
-
 comment on function pgautofailover.current_state(text)
         is 'get the current state of both nodes of a formation';
 
+grant execute on function pgautofailover.current_state(text)
+   to autoctl_node;
+
```
JelteF, line 577: again this grant is only in the upgrade, not in the `pgautofailover.sql`. I think it might be because they are pure SQL functions.

JelteF (approved):

src/monitor/pgautofailover--1.5--1.6.sql
DimCitus: In the SQL extension script, from this function on, GRANT statements are missing. It would be cleaner to go a special PR for fixing that, I believe. Merged as-is, then.

PR #747, opened 2021-06-30
Use -std=c11

eradman:
`_Static_assert` is not available in older compilers

https://gcc.gnu.org/wiki/C11Status
https://clang.llvm.org/c_status.html#c11

### DIFF
--- a/src/bin/pg_autoctl/Makefile
+++ b/src/bin/pg_autoctl/Makefile
@@ -29,7 +29,7 @@ COMMON_LIBS += -I${SRC_DIR}../lib/parson/
 
 CC = $(shell $(PG_CONFIG) --cc)
 
-DEFAULT_CFLAGS = -std=c99 -D_GNU_SOURCE -g
+DEFAULT_CFLAGS = -std=c11 -D_GNU_SOURCE -g
 DEFAULT_CFLAGS += -I $(shell $(PG_CONFIG) --includedir)
 DEFAULT_CFLAGS += -I $(shell $(PG_CONFIG) --includedir-server)
 DEFAULT_CFLAGS += -I $(shell $(PG_CONFIG) --pkgincludedir)/internal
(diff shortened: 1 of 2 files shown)


DimCitus:
Hi @eradman ; thanks for your continued contributions! Would it be possible to detect the situation at compile time and only use the Assert when the compiler has support for it? I think that'd be better than reducing the portability to requiring C11, but that's a gut feeling at this point.

The _Static_assert we are using are good enough if only the maintenance team sees them, I believe, and the CI too.

eradman:
@DimCitus, this is a good observation: compile-time assertions is not required at runtime, so we can maintain compatibility with C99. Unfortunately testing directly for `_Static_assert` doesn't seem possible since it's a keyword. We can test for macros defined in `assert.h`:

```c
#include <assert.h>

#ifdef static_assert
/* ... */
#endif
```
I'll ammend this PR to do this

eradman:
Unfortunately `assert.h` doesn't define `static_assert` when `-std=c99` is used, so there's no obvious way to conditionally test for this functionality

```c
#if defined(__STDC_VERSION__) && __STDC_VERSION__ >= 201112
#define static_assert _Static_assert
#endif
```

DimCitus:
Does this STDC_VERSION test works well in your environment, with keeping `--std=c99`? if yes, then let's edit the patch that way and I'll try locally also, and we'll be able to see what Travis makes of it.

eradman:
Here are the results of some manual tests to print the value of `__STDC_VERSION__`

```
Platform      Compiler       Default    -std=c99
------------- -------------- ---------- ---------
OpenBSD  6.9  clang 10.0.1   201112     199901
OpenBSD  6.9  gcc    4.2.1   undefined  199901
OpenBSD  6.9  gcc    8.4.0   201710     199901
MacOS   11.4  clang 12.0.5   201710     199901
Ubuntu  20.10 gcc   10.3.0   201710     199901
FreeBSD 13.0  clang 11.0.1   201710     199901
```

So using `-std` flag always changes weather the `static_assert` macro is defined, but since `_Static_assert` is a keyword it remains available if the compiler supports it.

eradman:
Closing as I am convinced there is no solution to this in C. For any platform that has a very old compiler a simple patch can remove the use of `_Static_assert`.

DimCitus:
Fair enough, thanks for diving into the topic. Hopefully getting a recent enough compiler is easy enough in all the current platforms...

PR #748, opened 2021-07-01
Have pg_autoctl drop node comment wait until the node has been removed.

DimCitus:


src/bin/pg_autoctl/cli_drop_node.c
```diff
@@ -395,7 +400,12 @@ cli_drop_node(int argc, char **argv)
 			exit(EXIT_CODE_BAD_ARGS);
 		}
 
-		(void) cli_drop_node_from_monitor(&config);
+		int64_t nodeId = -1;
+		int groupId = -1;
+
+		(void) cli_drop_node_from_monitor(&config, &nodeId, &groupId);
+
+		(void) wait_until_node_is_dropped(&config, nodeId, groupId);
```
JelteF, line 408: I think it would be good to have a log here, or at the end of `wait_until_node_is_dropped` saying something like: "Node dropped successfully"

src/bin/pg_autoctl/cli_drop_node.c
```diff
@@ -851,3 +868,49 @@ stop_postgres_and_remove_pgdata_and_config(ConfigFilePaths *pathnames,
...
+wait_until_node_is_dropped(KeeperConfig *config, int64_t nodeId, int groupId)
+{
+	bool dropped = false;
+	Monitor monitor = { 0 };
+
+	(void) cli_monitor_init_from_option_or_config(&monitor, config);
+
+
+	while (!dropped)
+	{
```
JelteF, line 887: Should have some check that calls `exit` after some timeout (probably `PG_AUTOCTL_LISTEN_NOTIFICATIONS_TIMEOUT`). Otherwise we can get stuck forever here.

JelteF (changes requested):

src/bin/pg_autoctl/cli_drop_node.c
DimCitus: Good point, added now.

src/bin/pg_autoctl/cli_drop_node.c
DimCitus: Good point also, added too.

JelteF (approved):

Issue #746, opened 2021-06-30
BUG: requested to open an already open connection in non PGSQL_CONNECTION_MULTI_STATEMENT mode


JelteF:
When using make cluster I got this error on the primary:
![image](https://user-images.githubusercontent.com/1162278/123952274-17154300-d9a6-11eb-99fa-9f6fd23241ae.png)

JelteF:
this might be related too, the monitor starts up much slower on current master than on 1.5.2:
![image](https://user-images.githubusercontent.com/1162278/124144459-48683e80-da8c-11eb-9d34-7a3aafc564dc.png)

JelteF:
A few extra logs for some more context:
![image](https://user-images.githubusercontent.com/1162278/124251840-a21d4700-db26-11eb-89ac-afdf0fff1e91.png)

DimCitus:
I believe this is all solved now thanks to your work in #751 and #752.

PR #751, opened 2021-07-02
Fix transaction begin failure handling

JelteF:
When opening a transaction using `pgsql_begin` we would set the
connection type to multi statement mode. However, if we failed to open a
connection, we would not reset it to single statement mode.

This changes it such that we only set the connection type to multi
statement mode once we have successfully opened a transaction.

See #746

### DIFF
--- a/src/bin/pg_autoctl/pgsql.c
+++ b/src/bin/pg_autoctl/pgsql.c
@@ -478,7 +478,7 @@ log_connection_error(PGconn *connection, int logLevel)
 /*
  * pgsql_open_connection opens a PostgreSQL connection, given a PGSQL client
  * instance. If a connection is already open in the client (it's not NULL),
- * then pgsql_open_connection reuses it and returns it immediately.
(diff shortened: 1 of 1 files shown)


src/bin/pg_autoctl/pgsql.c
```diff
@@ -873,17 +873,21 @@ pgAutoCtlDebugNoticeProcessor(void *arg, const char *message)
 bool
 pgsql_begin(PGSQL *pgsql)
 {
-	PGconn *connection;
-
-	pgsql->connectionStatementType = PGSQL_CONNECTION_MULTI_STATEMENT;
-	connection = pgsql_open_connection(pgsql);
-	if (connection == NULL)
+	if (!pgsql_execute(pgsql, "BEGIN"))
```
JelteF, line 876: pgsql_execute already opens a connection if needed, so no need to do so manually.

DimCitus (approved):
Looks good to me!

PR #753, opened 2021-07-05
Add the reported timeline id to the monitor events table.

DimCitus:


src/monitor/pgautofailover--1.5--1.6.sql
```diff
@@ -405,6 +454,91 @@ CREATE TRIGGER disable_secondary_check
...
+     limit count
+)
+select * from last_events order by eventtime, eventid;
+$$;
+
+comment on function pgautofailover.last_events(text,int,int)
+        is 'retrieve last COUNT events for given formation and group';
+
+grant execute on function pgautofailover.last_events(text,int,int)
+   to autoctl_node;
```
JelteF, line 539: All these grants are actualy not in `pgautofailover.sql`, I guess it doesn't need access to these functions.

JelteF (approved):
Apart from the inconsistent grants I think this is good to merge.

src/monitor/pgautofailover--1.5--1.6.sql
DimCitus: Other grants on functions are edited the same (only in the upgrade script), and I came to the theory that the upgrade script is run by the superuser and default grants would not be adequate?

DimCitus: Actually most functions have the grant execute lines, no reason to bypass this one. Added now.

PR #756, opened 2021-07-06
Set candidate priority without waiting for state changes in more cases.

DimCitus:
When dropping the last node with candidate priority greater than zero, we
might want to change the candidate priority of the last remaining node. The
client-side implementation of the command would wait for the APPLY_SETTINGS
cycle in some cases with dropped nodes, which is unfortunate: this PR fixes
that.

### DIFF
--- a/src/bin/pg_autoctl/cli_get_set_properties.c
+++ b/src/bin/pg_autoctl/cli_get_set_properties.c
@@ -680,10 +680,10 @@ set_node_candidate_priority(Keeper *keeper, int candidatePriority)
 								   config->groupId,
 								   &nodesArray))
 	{
-		log_fatal("Failed to get the list of all the nodes in formation \"%s\" "
-				  "from the monitor, see above for details",
-				  keeper->config.formation);
-		return false;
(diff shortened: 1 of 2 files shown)


src/bin/pg_autoctl/cli_get_set_properties.c
```diff
@@ -668,22 +668,27 @@ static bool
...
-		log_warn("Failed to get_nodes() on the monitor");
+	if (!monitor_get_current_state(&(keeper->monitor),
+								   config->formation,
+								   config->groupId,
+								   &nodesArray))
+	{
+		log_fatal("Failed to get the list of all the nodes in formation \"%s\" "
+				  "from the monitor, see above for details",
+				  keeper->config.formation);
+		return false;
```
JelteF, line 686: We now do a hard failure here. Is that what we want? (see old comment about ignoring the error and just not waiting in that case).

DimCitus: Good point, fixed now!

JelteF (approved):

PR #767, opened 2021-07-09
Set the pgautofailover_monitor user's password to NULL.

### LINKED ISSUE #763: Password of pgautofailover_monitor? Why it is always setup as trust in pg_hba?
We are trying to use pg_auto_failover with Docker swarm. In Docker swarm the DNS reverse lookup of the container IP does not match its hostname, e.g. `db-monitor -> 192.168.1.10 -> db-monitor.<slot-id>.<task-id>.<network>`, so pg_auto_failover issues some warnings and then uses the container IP for `pg_hba.conf`. 

However given that container IPs change regularly and that `--pg-hba-lan` does not affect replication entries in `pg_hba.conf` we decided to use `--skip-pg-hba`. But now we can not use `--auth scram-sha-256` anymore and have to include our own `postgres-custom.conf` which configures postgres to use scram-sha-256 again. Why is that? Why can't we use `--skip-pg-hba` and `--auth scram-sha-256` at the same time?

This generally works, but now on the postgres nodes we see a user `pgautofailover_monitor` which still has a MD5 password while the `pgautofailover_replicator` user has a scram password (because we need to set that password manually after postgres node creation). So technically that user should not work, however we have copied the `pg_hba.conf` generated by pg_auto_failover prior using `--skip-pg-hba` and it contains an entry `hostssl all "pgautofailover_monitor" x.x.x.x/24 trust`. So the MD5 password isn't used anyways.

We don't like the fact that we have a trust based user for a whole subnet in our `pg_hba.conf` so we would like to change it to `scram-sha-256` just like all the other entries in that file. This brings us to the following questions:

1. Why does this user have a MD5 password, even though our postgres-custom.conf file defines scram-sha-256?
2. What is the password of that user so that we can update it to scram-sha-256?
3. Shouldn't the password be configurable to begin with and pg_auto_failover should not use `trust` for that user in `pg_hba.conf`

DimCitus:
We don't actually need to have a password for this user, given that the
monitor doesn't authenticate when doing health checks. Not having a password
to manage also makes it easier to by compliant with password storage
policies of our users (e.g. md5 vs scram-sha-256).

Fixes #763.

### DIFF
--- a/src/bin/pg_autoctl/cli_do_misc.c
+++ b/src/bin/pg_autoctl/cli_do_misc.c
@@ -150,7 +150,7 @@ keeper_cli_create_monitor_user(int argc, char **argv)
 
 	if (!primary_create_user_with_hba(&postgres,
 									  PG_AUTOCTL_HEALTH_USERNAME,
-									  PG_AUTOCTL_HEALTH_PASSWORD,
+									  NULL, /* no password */
 									  monitorHostname,
 									  "trust",
 									  HBA_EDIT_MINIMAL,
--- a/src/bin/pg_autoctl/defaults.h
+++ b/src/bin/pg_autoctl/defaults.h
@@ -121,7 +121,6 @@
 
 /* pg_auto_failover monitor related constants */
 #define PG_AUTOCTL_HEALTH_USERNAME "pgautofailover_monitor"
-#define PG_AUTOCTL_HEALTH_PASSWORD "pgautofailover_monitor"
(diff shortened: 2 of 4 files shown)


JelteF:
I don't think this change is good in the current form. If you have the password set to NULL and set pg_hba.conf to only allow scram-sha-256 (or md5). Then postgres logs on node1 are being flooded with the following FATAL errors (every second a new log):
afbeelding

JelteF:
See also https://github.com/citusdata/pg_auto_failover/pull/672 which has similar errors.

PR #768, opened 2021-07-09
Refrain from using PGPASSWORD for pg_basebackup connection.

### LINKED ISSUE #766: Wrong password used by pg_autoctl during monitor connect
pg_auto_failover version 1.5.2

During first initialization of cluster (monitor + 2 postgres nodes) we discovered that secondary postgres node uses the replication password when connecting to the monitor after initial base backup has been done and postgres starts the first time.

During initial testing we used the same password for `autoctl_node` and `pgautofailover_replicator` and everything did work as expected. Then we decided to use two different passwords and pg_auto_failover was able to register the secondary postgres node with the monitor and  do a base backup from the primary node. But after creating the base backup pg_autoctl fails to connect to the monitor node to ask for other nodes. Reverting the replication password back to the monitor password resolves the issue.

Also worth mentioning that we are using `.pgpass` file and do not have any passwords in monitor URL. The `.pgpass` file has been verified to contain the correct passwords for `autoctl_node` and `pgautofailover_replicator` and connecting manually from console as postgres user works as expected. Adding the monitor password directly to the monitor URL fixes the problem as well. So the issue only occurs when using `.pgpass`.

This method does not return true after several attempts to connect to the monitor:
https://github.com/citusdata/pg_auto_failover/blob/125bb19c13ea8af1d6aacd5f2f9f7cd51b406e74/src/bin/pg_autoctl/keeper.c#L1840

It seems that the pg client instance attached to the monitor is not correctly configured.

The following log is after a successful base backup and first start of postgres (db001-m is the monitor node):
```
14:22:51 30 DEBUG pgsql.c:500 Connecting to [monitor] "postgres://autoctl_node@db001-m:5432/pg_auto_failover?sslmode=require"
14:22:51 30 WARN  pgsql.c:598 Failed to connect to "postgres://autoctl_node@db001-m:5432/pg_auto_failover?sslmode=require", retrying until the server is ready
14:22:51 30 DEBUG pgsql.c:643 PQping(postgres://autoctl_node@db001-m:5432/pg_auto_failover?sslmode=require): slept 5 ms on attempt 2
14:22:51 30 DEBUG pgsql.c:657 PQping OK after 2 attempts
14:22:52 30 DEBUG pgsql.c:464 Connection to database failed: FATAL:  password authentication failed for user "autoctl_node"
14:22:52 30 WARN  pgsql.c:708 Failed to connect after successful ping
14:22:52 30 DEBUG pgsql.c:423 Disconnecting from [monitor] "postgres://autoctl_node@db001-m:5432/pg_auto_failover?sslmode=require"
14:22:52 30 DEBUG pgsql.c:643 PQping(postgres://autoctl_node@db001-m:5432/pg_auto_failover?sslmode=require): slept 9 ms on attempt 3
14:22:52 30 DEBUG pgsql.c:657 PQping OK after 3 attempts
...
...
14:22:54 30 DEBUG pgsql.c:657 PQping OK after 11 attempts
14:22:54 30 DEBUG pgsql.c:423 Disconnecting from [monitor] "postgres://autoctl_node@db001-m:5432/pg_auto_failover?sslmode=require"
14:22:54 30 ERROR pgsql.c:623 Failed to connect to "postgres://autoctl_node@db001-m:5432/pg_auto_failover?sslmode=require" after 11 attempts in 2271 ms, pg_autoctl stops retrying now
14:22:54 30 ERROR monitor.c:498 Failed to get other nodes from the monitor while running "SELECT * FROM pgautofailover.get_other_nodes($1) ORDER BY node_id" with node id 56
14:22:54 30 ERROR keeper.c:1842 Failed to get_other_nodes() on the monitor
14:22:54 30 ERROR fsm_transition.c:872 Failed to update HBA rules after a base backup
14:22:54 30 DEBUG fsm.c:501 Transition function returned: false
14:22:54 30 ERROR fsm.c:519 Failed to transition from state "wait_standby" to state "catchingup", see above.
14:22:54 30 FATAL cli_do_service.c:572 Failed to initialize the node active service, see above for details
14:22:54 1 ERROR supervisor.c:653 pg_autoctl service node-active exited with exit status 12
14:22:54 1 DEBUG supervisor.c:790 supervisor_may_restart: service "node-active" restarted 1 times, most recently at Fri Jul  9 14:22:43 2021, 11 seconds ago
14:22:54 1 INFO  supervisor.c:738 Restarting service node-active
```

DimCitus:
Using the environment variable allows to omit the replication password from
the logs, though messes with next connection attempts. We should either call
unsetenv() once the pg_basebackup command has finished, or use the password
on the connection string.

In this tentative PR we use the password in the connection string.

Should fix #766.

### DIFF
--- a/src/bin/pg_autoctl/pgctl.c
+++ b/src/bin/pg_autoctl/pgctl.c
@@ -1268,6 +1268,7 @@ pg_basebackup(const char *pgdata,
 	int argsIndex = 0;
 
 	char command[BUFSIZE];
+	char pgpassword[BUFSIZE] = { 0 };
 
 	log_debug("mkdir -p \"%s\"", replicationSource->backupDir);
 	if (!ensure_empty_dir(replicationSource->backupDir, 0700))
@@ -1289,6 +1290,14 @@ pg_basebackup(const char *pgdata,
 
 	if (!IS_EMPTY_STRING_BUFFER(replicationSource->password))
 	{
+		if (env_exists("PGPASSWORD"))
+		{
+			if (!get_env_copy("PGPASSWORD", pgpassword, sizeof(pgpassword)))
+			{
+				/* errors have already been logged. */
+				return false;
+			}
+		}
 		setenv("PGPASSWORD", replicationSource->password, 1);
 	}
 	setenv("PGAPPNAME", replicationSource->applicationName, 1);
@@ -1356,6 +1365,19 @@ pg_basebackup(const char *pgdata,
 
 	(void) execute_subprogram(&program);
 
+	/* clean-up the environment again */
+	if (!IS_EMPTY_STRING_BUFFER(replicationSource->password))
+	{
+		if (IS_EMPTY_STRING_BUFFER(pgpassword))
+		{
+			unsetenv("PGPASSWORD");
+		}
+		else
+		{
+			setenv("PGPASSWORD", pgpassword, 1);
+		}
+	}
+
 	returnCode = program.returnCode;
 	free_program(&program);
 
@@ -1409,6 +1431,7 @@ pg_rewind(const char *pgdata,
 	int argsIndex = 0;
 
 	char command[BUFSIZE];
+	char pgpassword[BUFSIZE] = { 0 };
 
 	/* call pg_rewind*/
 	path_in_same_directory(pg_ctl, "pg_rewind", pg_rewind);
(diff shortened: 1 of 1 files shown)


JelteF (approved):
As discussed privately: Can you double check that the password is not in the logs after this change, even when using `-vvv`. If it's not there, then I think this is fine to move.

DimCitus:
Given my reading of the code I believe you've spotted a real problem with this attempt:

``` c
	args[argsIndex++] = primaryConnInfo;

        ...

	/* log the exact command line we're using */
	int commandSize = snprintf_program_command_line(&program, command, BUFSIZE);
```

Making it so that we don't include the password in the logs is going to require too much fiddling around, I suppose I should work on using `unsetenv` instead.

DimCitus:
I just pushed a better version of this fix, using `unsetenv`, or `setenv` again to the previous value of the PGPASSWORD environment variable if it happened to be already set before. Might be useful...

Issue #749, opened 2021-07-01
Docker: Upgrading pg_auto_failover monitor causes postgres nodes to terminate and cluster downtime


jnehlmeier:
Hi,

we are planning to use pg_auto_failover with Docker and the upgrade process introduced in 1.4 sounds nice but doesn't really work well with Docker. The reason is that the pg_autoctl process terminates and tries to restart as soon as it detects a version mismatch with the (updated) monitor. This causes the postgres cluster to fully terminate until the docker image of postgres nodes has also been updated and restarted.

Are there any recommendations how to update pg_auto_failover with docker without downtime? Even if pg_autoctl would get an option to not restart itself as soon as it detects a version mismatch, switchover can probably not be done since nodes can not be coordinated anymore.

I tink in an ideal world it should work similar to

1. Monitor will be upgraded from version A to B. Monitor can still speak to postgres nodes using the old protocol version A, but it also supports now the new protocol version B.
2. Update docker image of postgres secondaries so they now also speak protocol version B
3. Do a switchover (coordinating using the correct protocol version A or B, depending on the node)
4. Update docker image of old postgres primary so it speaks protocol version B.

JelteF:
Hmm, this is indeed a scenario we didn't really think about when designing the upgrade story. The flow you propose does sound nice, but there's quite a lot of edge cases with running mixed versions that are nice not to have to support (e.g. additions/changes to the FSM). 

A flow that might work for you is the following that introduces an "upgrade container" in a way like this:
1. Have nodes on version A
2. Create docker container with version A installed, but with the package for version B downloaded as a `deb`/`rpm`
3. Replace secondary containers with these upgrade containers
4. Do a switchover
5. Replace primary container with this upgrade container
6. Run this in all docker containers `exec dpkg -i pgautofailover-vB.deb`. binaries are swapped B, but running version is still A
7. Upgrade the monitor to the container with only version B (no need for the "upgrade container" here)
8. Inplace upgrade to B will happen on the other nodes now, since the binaries have been swapped
9. Upgrade the secondaries to a container with only version B
10. Do a switchover
11. Upgrade the old primary to a container with only version B

jnehlmeier:
This could work. We will try it once 1.6.x packages are available.

However this isn't really scaleable if you have multiple clusters. So I hope you will think about a more cloud native upgrade solution that does not need that much manual intervention.

I can see that changing available states in FSM is problematic so maybe a compromise would be:
[...]

PR #771, opened 2021-07-12
Only exit at upgrade when the on-disk binary is ready.

DimCitus:
See #749.

### DIFF
--- a/src/bin/pg_autoctl/ipaddr.c
+++ b/src/bin/pg_autoctl/ipaddr.c
@@ -32,6 +32,7 @@
 #include "ipaddr.h"
 #include "log.h"
 #include "pgsetup.h"
+#include "pgsql.h"
 #include "string_utils.h"
 
 static unsigned int countSetBits(unsigned int n);
@@ -42,6 +43,10 @@ static bool fetchIPAddressFromInterfaceList(char *localIpAddress, int size);
 static bool ipaddr_sockaddr_to_string(struct addrinfo *ai,
 									  char *ipaddr, size_t size);
 static bool ipaddr_getsockname(int sock, char *ipaddr, size_t size);
+static bool GetAddrInfo(const char *restrict node,
+						const char *restrict service,
+						const struct addrinfo *restrict hints,
+						struct addrinfo **restrict res);
 
 
 /*
@@ -74,14 +79,12 @@ fetchLocalIPAddress(char *localIpAddress, int size,
 	hints.ai_socktype = SOCK_STREAM; /* we only want TCP sockets */
 	hints.ai_protocol = IPPROTO_TCP; /* we only want TCP sockets */
 
-	int error = getaddrinfo(serviceName,
-							intToString(servicePort).strValue,
-							&hints,
-							&lookup);
-	if (error != 0)
+	if (!GetAddrInfo(serviceName,
+					 intToString(servicePort).strValue,
+					 &hints,
+					 &lookup))
 	{
-		log_warn("Failed to parse host name or IP address \"%s\": %s",
-				 serviceName, gai_strerror(error));
+		/* errors have already been logged */
 		return false;
 	}
 
@@ -473,11 +476,9 @@ findHostnameLocalAddress(const char *hostname, char *localIpAddress, int size)
 	struct addrinfo *dns_addr;
 	struct ifaddrs *ifaddrList, *ifaddr;
 
-	int error = getaddrinfo(hostname, NULL, 0, &dns_lookup_addr);
-	if (error != 0)
+	if (!GetAddrInfo(hostname, NULL, 0, &dns_lookup_addr))
 	{
-		log_warn("Failed to resolve DNS name \"%s\": %s",
-				 hostname, gai_strerror(error));
+		/* errors have already been logged */
 		return false;
 	}
 
@@ -628,19 +629,17 @@ findHostnameFromLocalIpAddress(char *localIpAddress, char *hostname, int size)
 	struct addrinfo *lookup, *ai;
 
 	/* parse ipv4 or ipv6 address using getaddrinfo() */
-	int ret = getaddrinfo(localIpAddress, NULL, 0, &lookup);
-	if (ret != 0)
+	if (!GetAddrInfo(localIpAddress, NULL, 0, &lookup))
 	{
-		log_warn("Failed to resolve DNS name \"%s\": %s",
-				 localIpAddress, gai_strerror(ret));
+		/* errors have already been logged */
 		return false;
 	}
 
 	/* now reverse lookup (NI_NAMEREQD) the address with getnameinfo() */
 	for (ai = lookup; ai; ai = ai->ai_next)
 	{
-		ret = getnameinfo(ai->ai_addr, ai->ai_addrlen,
-						  hbuf, sizeof(hbuf), NULL, 0, NI_NAMEREQD);
+		int ret = getnameinfo(ai->ai_addr, ai->ai_addrlen,
+							  hbuf, sizeof(hbuf), NULL, 0, NI_NAMEREQD);
 
 		if (ret != 0)
 		{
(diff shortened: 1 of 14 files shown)


src/bin/pg_autoctl/keeper.c
```diff
@@ -1194,7 +1198,24 @@ keeper_node_active(Keeper *keeper, bool doInit,
 		 * with the expected version of pg_autoctl that matches the monitor's
 		 * extension version.
 		 */
-		exit(EXIT_CODE_MONITOR);
+		KeeperVersion keeperVersion = { 0 };
```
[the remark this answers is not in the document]
DimCitus: Updated. I just removed the “with luck” part, the rest seemed okay as-is to me.

src/bin/pg_autoctl/keeper.c
```diff
@@ -1402,21 +1423,22 @@ keeper_check_monitor_extension_version(Keeper *keeper)
 	}
 
 	/* from a member of the cluster, we don't try to upgrade the extension */
-	if (strcmp(version.installedVersion, PG_AUTOCTL_EXTENSION_VERSION) != 0)
+	if (strcmp(version->installedVersion, PG_AUTOCTL_EXTENSION_VERSION) != 0)
 	{
 		log_fatal("The monitor at \"%s\" has extension \"%s\" version \"%s\", "
```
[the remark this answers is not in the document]
DimCitus: The message is now at the INFO level.

src/bin/pg_autoctl/keeper.c
```diff
@@ -1402,21 +1423,22 @@ keeper_check_monitor_extension_version(Keeper *keeper)
...
 				  "this pg_autoctl version requires version \"%s\".",
 				  keeper->config.monitor_pguri,
 				  PG_AUTOCTL_MONITOR_EXTENSION_NAME,
 				  PG_AUTOCTL_EXTENSION_VERSION,
-				  version.installedVersion);
+				  version->installedVersion);
 		log_info("Please connect to the monitor node and restart pg_autoctl.");
```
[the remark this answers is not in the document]
DimCitus: I think the message is meant for the situation where the client has been upgraded (say, to 1.6) already when the monitor is still running 1.5 ; and so it would mean that you need to restart the monitor. That said, I agree that it's confusing and not helpful, so I have removed it.

src/bin/pg_autoctl/keeper.c
```diff
@@ -1187,14 +1191,36 @@ keeper_node_active(Keeper *keeper, bool doInit,
...
+		if (strcmp(monitorVersion.installedVersion,
+				   keeperVersion.required_extension_version) == 0)
+		{
+			log_info("pg_autoctl version \"%s\" with compatibility with "
+					 "monitor extension \"%s\" has been found on-disk, "
+					 "exiting for a restart of the node-active process.",
+					 keeperVersion.pg_autoctl_version,
+					 keeperVersion.required_extension_version);
+			exit(EXIT_CODE_MONITOR);
+		}
```
JelteF, line 1223: Should we do anything here in the `else` case? e.g. log something and/or return false?

DimCitus: That's a good point indeed. I have added extra logging (but that will happen every second) and returned false in that case, so that we don't try to use a monitor protocol version that we don't know. I have yet to figure a nice way to test this code path now (if I just upgrade the binary locally in a `make cluster` environment, because it's the same binary for every node, the right thing will happen...).

JelteF: My trick for testing this locally was to change the version number everywhere to `1.7`, but change this to compare against hardcoded `"1.6"` instead of the `PG_AUTOCTL_EXTENSION_VERSION` constant:
https://github.com/citusdata/pg_auto_failover/blob/0fecd653f9ef43019e4d76e347afce378adcc481/src/bin/pg_autoctl/keeper.c#L1405

JelteF: actually no, I think it was a different line where I did this trick. I made `pg_autoctl version` always return `1.6` I think.

DimCitus: Okay I added preliminary support for running an upgrade from within a docker compose environment, where it's possible to use the following command to replace a running container with a new container freshly built from the repository:

```
$ docker compose up -d --no-deps --build monitor
```

I added a little Makefile in `tests/upgrade` along with the docker-compose file to use. Then the idea would be to install such a patch as the following, and run `make upgrade`, which replaces the monitor and we can see the new error message about the upgrade to be done on the other nodes. 

The docker compose network is sometimes acting up though, and I have had other errors, such as the following, making it pretty hard to reliably test for upgrade paths. Still, it's a promising work in progress, I believe.

```
monitor_1  | 21:29:44 25 WARN  Failed to resolve DNS name "79a3a3ebe79e": Temporary failure in name resolution
monitor_1  | 21:29:44 25 WARN  Failed to find IP address for hostname "79a3a3ebe79e", see above for details
monitor_1  | 21:29:44 25 WARN  Failed to grant connection to local network.
monitor_1  | 21:29:44 1 ERROR pg_autoctl service listener exited with exit status 12
monitor_1  | 21:29:44 1 INFO  Restarting service listener
```

``` diff
diff --git a/src/bin/pg_autoctl/defaults.h b/src/bin/pg_autoctl/defaults.h
index 8e7e2eaf..643dc96a 100644
--- a/src/bin/pg_autoctl/defaults.h
+++ b/src/bin/pg_autoctl/defaults.h
@@ -17,7 +17,7 @@
 #define PG_AUTOCTL_VERSION "1.6.1"
 
 /* version of the extension that we requite to talk to on the monitor */
-#define PG_AUTOCTL_EXTENSION_VERSION "1.6"
+#define PG_AUTOCTL_EXTENSION_VERSION "1.7"
 
 /* environment variable to use to make DEBUG facilities available */
 #define PG_AUTOCTL_DEBUG "PG_AUTOCTL_DEBUG"
diff --git a/src/monitor/Makefile b/src/monitor/Makefile
index 8e6bf321..478fb2a7 100644
--- a/src/monitor/Makefile
+++ b/src/monitor/Makefile
@@ -2,7 +2,7 @@
 # Licensed under the PostgreSQL License.
 
 EXTENSION = pgautofailover
-EXTVERSION = 1.6
+EXTVERSION = 1.7
 
 SRC_DIR := $(dir $(abspath $(lastword $(MAKEFILE_LIST))))
 
diff --git a/src/monitor/metadata.h b/src/monitor/metadata.h
index 9e50ab72..0f8b39a3 100644
--- a/src/monitor/metadata.h
+++ b/src/monitor/metadata.h
@@ -15,7 +15,7 @@
 
 #include "storage/lockdefs.h"
 
-#define AUTO_FAILOVER_EXTENSION_VERSION "1.6"
+#define AUTO_FAILOVER_EXTENSION_VERSION "1.7"
 #define AUTO_FAILOVER_EXTENSION_NAME "pgautofailover"
 #define AUTO_FAILOVER_SCHEMA_NAME "pgautofailover"
 #define AUTO_FAILOVER_FORMATION_TABLE "pgautofailover.formation"
diff --git a/src/monitor/pgautofailover.control b/src/monitor/pgautofailover.control
index a649eb76..d0504410 100644
--- a/src/monitor/pgautofailover.control
+++ b/src/monitor/pgautofailover.control
@@ -1,5 +1,5 @@
 comment = 'pg_auto_failover'
-default_version = '1.6'
+default_version = '1.7'
 module_pathname = '$libdir/pgautofailover'
 relocatable = false
 requires = 'btree_gist'
```

DimCitus: So what happened with the DNS resolution is actually that on the first run in the docker compose environment, the hostname would be set to a dynamic hash (the container hash/name). We would store that as the default `--hostname` in our pg_autoctl configuration file. And at the next start-up of course the container name/hash would change, but pg_autoctl would still try to resolve the previous container name, with the DNS now knows nothing about.

I have added a retry loop to the DNS temporary resolution failures, and more importantly fixed the docker compose file to use known hostnames thanks to the `hostname` configuration. 

I have also improved the `tests/upgrade/Makefile` to handle the patching of the source code for the 1.7 version, which we would have to maintain for other versions later of course. And added a README that makes it easy enough to navigate the tooling.

Oh, in passing, I have validated that the current PR actually works as intended. The Makefile also allows to run a full upgrade of all the nodes, using the docker compose commands.

src/bin/pg_autoctl/ipaddr.c
```diff
@@ -867,3 +863,55 @@ ipaddrGetLocalHostname(char *hostname, size_t size)
 
 	return true;
 }
+
+
+/*
+ * GetAddrInfo calls getaddrinfo and
```
JelteF, line 869: Sentence is incomplete

DimCitus: Thanks, fixed now.

JelteF (approved):

PR #772, opened 2021-07-12
Avoid re-electing primary during a switchover.

### LINKED ISSUE #770: pg_autoctl perform switchover doesn't actually fail over since 1.6.1
### To reproduce
```
pg_autoctl perform switchover
```

### What happens
Election happens and `node1` becomes primary again.

DimCitus:
When all the nodes are up and running, including the primary, a switchover
has a great chance of finding that the current primary is then the best
candidate to with the primary election, making the operation a no-op.

In this PR we tweak the current primary node's candidate priority in a way
that should make it lose the election, and reset the priority when the
election is over.

Fixes #770.

### DIFF
--- a/src/monitor/node_active_protocol.c
+++ b/src/monitor/node_active_protocol.c
@@ -1394,6 +1394,32 @@ perform_failover(PG_FUNCTION_ARGS)
 
 		SetNodeGoalState(primaryNode, REPLICATION_STATE_DRAINING, message);
 
+		/*
+		 * When a failover is performed with all the nodes up and running, the
+		 * old primary is often in the best situation to win the election. In
+		 * that case, we trick the candidate priority in a way that makes the
+		 * node lose the election.
+		 */
+		if (primaryNode)
+		{
+			primaryNode->candidatePriority -= CANDIDATE_PRIORITY_INCREMENT;
+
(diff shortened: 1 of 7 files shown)


JelteF:
Needs a test

src/monitor/node_active_protocol.c
```diff
@@ -1394,6 +1394,32 @@ perform_failover(PG_FUNCTION_ARGS)
 
 		SetNodeGoalState(primaryNode, REPLICATION_STATE_DRAINING, message);
 
+		/*
+		 * When a failover is performed with all the nodes up and running, the
+		 * old primary is often in the best situation to win the election. In
+		 * that case, we trick the candidate priority in a way that makes the
+		 * node lose the election.
```
JelteF, line 1401:
```suggestion
		 * node lose the election. We undo this change in priority once the election completes.
```

DimCitus: Done now, thanks.

DimCitus:
> Needs a test

I tweaked an existing test to have the current primary and the target primary set to the same candidate priority, making sure that we still promote the target node rather than re-elect the current primary.

JelteF (approved):

PR #779, opened 2021-08-03
Improve error messages for missing configuration files.

### LINKED ISSUE #778: Misleading error message when pg_ident.conf (or pg_hba.conf) are missing on existing cluster.
When any of 3 files do not exist
https://github.com/citusdata/pg_auto_failover/blob/ac12fcd702397719a408546d639147d652a8dc06/src/bin/pg_autoctl/debian.c#L497-L507

The error message given to the user is about `postgresql.conf` not existing
https://github.com/citusdata/pg_auto_failover/blob/ac12fcd702397719a408546d639147d652a8dc06/src/bin/pg_autoctl/debian.c#L221-L235

This is misleading as the user will see:

```
18:57:34 16816 WARN Failed to find Postgres configuration files in PGDATA, as expected: "/pgdata/<redacted>/postgresql.conf" does not exists
18:57:34 16816 WARN Failed to match PGDATA at "/pgdata/<redacted>" with a debian setup following the data_directory template '/var/lib/postgresql/%v/%c'
18:57:34 16816 FATAL Failed to setup your Postgres instance the PostgreSQL way, see above for details
```

This was actually encountered by a customer that is migrating from Citus Cloud to onprem.
In the end it turned out `pg_ident.conf` was missing, but they required help from support to figure that out.

Would be great if these files are required to actually understand which file is missing and report accordingly.

DimCitus:
Fixes #778

### DIFF
--- a/src/bin/pg_autoctl/debian.c
+++ b/src/bin/pg_autoctl/debian.c
@@ -175,15 +175,24 @@ debian_find_postgres_configuration_files(PostgresSetup *pgSetup,
 		return true;
 	}
 
-	/* is it a debian postgres-common style setup then? */
-	if (debian_init_postgres_config_files(pgSetup,
-										  &debianConfFiles,
-										  PG_CONFIG_TYPE_DEBIAN))
+	/*
+	 * Is it a debian postgresql-common style setup then?
+	 *
+	 * We only search for debian style setup when the main postgresql.conf file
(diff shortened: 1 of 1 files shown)


thanodnl (approved):
Great! The error messages are now way more actionable.

Bit strange that they print 5 times though, however, I guess that is the retrying mechanism.

PR #786, opened 2021-08-25
Fix maintenance state related transitions.

DimCitus:
We used to disallow starting maintenance on a node in some cases, but it
seems that the user should be able to decide about when they need to operate
maintenance on their own nodes. After all, we don't stop Postgres when going
to maintenance, so users may change their mind without impacting their
service. A WARNING message is now displayed in some cases that were
previously prevented.

Also, the transition from WAIT_MAINTENANCE to MAINTENANCE was failing since
we improved the Group State Machine for the primary node, which would go
from JOIN_PRIMARY to PRIMARY without waiting for the other nodes to reach
their assigned state of WAIT_MAINTENANCE.


DimCitus:
For interactive testing and QA, use the following setup:

```
$ make NODES=3 NODES_PRIOS=50,50,0 cluster
```

Then you can play around putting node2 and node3 to maintenance, and then both of them together.

src/monitor/node_active_protocol.c
```diff
@@ -1665,50 +1665,34 @@ start_maintenance(PG_FUNCTION_ARGS)
...
-		}
+				 errmsg("Starting maintenance on " NODE_FORMAT
+						" will block writes on the primary " NODE_FORMAT,
+						NODE_FORMAT_ARGS(currentNode),
+						NODE_FORMAT_ARGS(primaryNode)),
+				 errdetail("we now have %d "
+						   "healthy node(s) left in the \"secondary\" state "
+						   "and formation \"%s\" number-sync-standbys requires "
+						   "%d sync standbys",
+						   secondaryNodesCount - 1,
```
JelteF, line 1693: This `- 1` seems incorrect in some cases the final number is `-1`:
image
image

JelteF:
I got into an unrecoverable state with these steps:
```
pg_autoctl enable maintenance --pgdata node2
pg_autoctl enable maintenance --pgdata node3
pg_autoctl enable maintenance --pgdata node1 --allow-failover
```
Cluster will then be stuck in this state:
image


Disabling maintenance on node1 does not work:
image

Disabling maintenance on node2 (or node3) also does not work, because it gets in this loop when you try:
image

State then stays like this:
image

DimCitus:
> I got into an unrecoverable state with these steps:

Now fixed with the following error. We might want to avoid the first WARNING, what do you think?

```
$ pg_autoctl enable maintenance --pgdata node1 --allow-failover
12:27:29 65490 WARN  WARNING:  Starting maintenance on node 1 "node1" (localhost:5501) will block writes on the primary node 1 "node1" (localhost:5501)
12:27:29 65490 WARN  DETAIL:  we now have 0 healthy node(s) left in the "secondary" state and formation "default" number-sync-standbys requires 1 sync standbys
12:27:29 65490 ERROR Monitor ERROR:  Starting maintenance on node 1 "node1" (localhost:5501) in state "primary" is not currently possible
12:27:29 65490 ERROR Monitor DETAIL:  there is currently 0 candidate nodes available
12:27:29 65490 ERROR Failed to start_maintenance of node 1 from the monitor
12:27:29 65490 FATAL Failed to enable maintenance of node 1 on the monitor, see above for details
```

tests/test_multi_maintenance.py
```diff
@@ -173,34 +173,59 @@ def test_006_maintenance_and_failover():
...
+    assert node3.get_state().assigned == "primary"
+    node3.fail()
+
+    # check that even after 30s node3 is still not set to draining
+    node3.sleep(30)
+    assert not node3.get_state().assigned == "draining"
+    assert node3.get_state().assigned == "primary"
+
+
+def test_008a_start_primary():
```
JelteF, line 217:
```suggestion
def test_008b_start_primary():
```

src/monitor/node_active_protocol.c
```diff
@@ -1725,6 +1715,22 @@ start_maintenance(PG_FUNCTION_ARGS)
 		AutoFailoverNode *firstStandbyNode = linitial(standbyNodesGroupList);
 		char message[BUFSIZE] = { 0 };
 
+		/*
+		 * We need at least one candidate note to initiate a failover and allow
```
JelteF, line 1719:
```suggestion
		 * We need at least one candidate node to initiate a failover and allow
```

tests/test_multi_maintenance.py
```diff
@@ -203,6 +203,22 @@ def test_008b_node2_to_maintenance():
...
+    node3.run()
+    assert node3.wait_until_state(target_state="primary")
+
+
```
JelteF, line 221: Can we add a test that `enable maintenance --allow-failover` fails here?

JelteF:
Again I found a way to reach an unrecoverable state:
```
pg_autoctl enable maintenance --pgdata node3
pg_autoctl enable maintenance --pgdata node1 --allow-failover
```
Cluster will then be stuck in this state:
image

Disabling maintenance on node1 does not work:
image

Disabling maintenance on node3 also does not work, because it gets in this loop when you try:
image

State then stays like this:
image

DimCitus:
> Again I found a way to reach an unrecoverable state:
> 
> ```
> make cluster -j20 TMUX_LAYOUT=tiled NODES=3
> pg_autoctl enable maintenance --pgdata node3
> pg_autoctl enable maintenance --pgdata node1 --allow-failover
> ```

This now can be unblocked by running `pg_autoctl disable maintenance --pgdata node3` ; but still fails when trying to disable maintenance on node1. I am looking at a transition from prepare_maintenance back to primary, if that makes sense.

tests/test_multi_maintenance.py
DimCitus: Fixed, thanks.

src/monitor/node_active_protocol.c
DimCitus: Fixed, thanks.

tests/test_multi_maintenance.py
DimCitus: Added, thanks for catching it!

src/monitor/node_active_protocol.c
```diff
@@ -1867,21 +1918,60 @@ stop_maintenance(PG_FUNCTION_ARGS)
...
 									   currentNode->groupId);
 
-	if (primaryNode == NULL)
+	/*
+	 * When there is no primary, we might be in trouble, we just want to join
+	 * the possibily ongoing election.
+	 */
+	if (totalNodesCount == 1)
+	{
+		(void) ProceedGroupState(currentNode);
```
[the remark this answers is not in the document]
DimCitus: Good catch, thanks! Fixed.

src/bin/pg_autoctl/fsm.c
```diff
@@ -362,6 +362,9 @@ KeeperFSMTransition KeeperFSM[] = {
 	 */
 	{ SECONDARY_STATE, REPORT_LSN_STATE, COMMENT_SECONDARY_TO_REPORT_LSN, &fsm_report_lsn },
 	{ CATCHINGUP_STATE, REPORT_LSN_STATE, COMMENT_SECONDARY_TO_REPORT_LSN, &fsm_report_lsn },
+	{ MAINTENANCE_STATE, REPORT_LSN_STATE, COMMENT_SECONDARY_TO_REPORT_LSN, &fsm_report_lsn },
+	{ PREPARE_MAINTENANCE_STATE, REPORT_LSN_STATE, COMMENT_SECONDARY_TO_REPORT_LSN, &fsm_report_lsn },
```
JelteF, line 366: We should allow going from PREPARE_MAINTENANCE_STATE to CATCHINGUP_STATE too.

DimCitus: Good point. I'm not sure how to make it happen, but I can see the kind of “race condition” that could raise the situation. Fixed now.

JelteF (approved):

JelteF:
I think this is good to merge. I ran into some more issues with wait_maintenance and opened a PR to fix some of those: #794 
I don't want to block this PR on that though.

PR #794, opened 2021-09-02
Only use wait_maintenance to wait for wait_primary

JelteF:
Before we would use wait_maintenance together with join_primary. This has no
function anymore with the current code.


src/monitor/node_active_protocol.c
```diff
@@ -1793,28 +1793,37 @@ start_maintenance(PG_FUNCTION_ARGS)
...
-			" to wait_maintenance "
+			" to %s "
 			"after a user-initiated start_maintenance call.",
 			NODE_FORMAT_ARGS(primaryNode),
 			ReplicationStateGetName(primaryGoalState),
-			NODE_FORMAT_ARGS(currentNode));
+			NODE_FORMAT_ARGS(currentNode),
+			ReplicationStateGetName(maintenanceGoalState));
 
 		SetNodeGoalState(primaryNode, primaryGoalState, message);
```
DimCitus, line 1825: I think we should only do that when the primaryGoalState has actually changed.

DimCitus (approved):

DimCitus (approved):

PR #796, opened 2021-09-03
Get rid of the JOIN_PRIMARY state.

DimCitus:
We still need to maintain backward compatibility of course, in case there is
an upgrade where one node is in JOIN_PRIMARY and then we want to be able to
continue working with the node.

Now, any node that reports JOIN_PRIMARY is unconditionnally assigned PRIMARY
instead, and we never assign JOIN_PRIMARY anymore. We still have some code
that matches the state in the group level FSM, and we keep the full compat
on the client side for the moment.


src/monitor/group_state_machine.c
```diff
@@ -403,6 +404,9 @@ ProceedGroupState(AutoFailoverNode *activeNode)
 		/* start replication */
 		AssignGoalState(activeNode, REPLICATION_STATE_CATCHINGUP, message);
 
+		/* edit synchronous_standby_names to add the new standby now */
+		AssignGoalState(primaryNode, REPLICATION_STATE_APPLY_SETTINGS, message);
```
JelteF, line 408: This is not part of the log above

JelteF (changes requested):
During startup of the cluster I now get this error on the primary, it resoleves later by itself:
image

The command I used to init the cluster was:
```
make cluster -j20 TMUX_LAYOUT=tiled NODES=5 NODES_ASYNC=1 NODES_PRIOS=50,50,50,50,0
```

DimCitus:
> During startup of the cluster I now get this error on the primary, it resoleves later by itself:

We do the apply_settings to primary transition on the monitor while the nodes are still catching-up, in that case I suppose that the standby nodes are still doing the pg_basebackup part of their initialisation. As this happens during the wait_standby to catching-up transition, we could also assign APPLY_SETTINGS only after the node has reached CATCHINGUP. Will see about that, thanks for the comment.

src/monitor/group_state_machine.c
DimCitus: Fixed now, and moved to another place to also fix the other part of your review. Thanks!

src/monitor/group_state_machine.c
```diff
@@ -431,12 +432,17 @@ ProceedGroupState(AutoFailoverNode *activeNode)
...
+			" to secondary after it caught up and " NODE_FORMAT
+			" to apply_settings to add the secondary to synchronous_standby_names",
+			NODE_FORMAT_ARGS(activeNode),
+			NODE_FORMAT_ARGS(primaryNode));
 
 		/* node is ready for promotion */
 		AssignGoalState(activeNode, REPLICATION_STATE_SECONDARY, message);
 
+		/* edit synchronous_standby_names to add the new standby now */
+		AssignGoalState(primaryNode, REPLICATION_STATE_APPLY_SETTINGS, message);
```
JelteF, line 444: This should not be done when the current primary is in state wait_primary, because it still has to wait to be sure that WAL has propagated to the secondary. This is done in the transition from wait_primary->primary but not from wait_primary->apply_settings. Once the primary is in state apply_settings/apply_settings, we consider it a full primary and a failover is allowed.

JelteF: also the comment at the top of this if statement seems like it might be a bit outdated.

DimCitus: Should be fixed now, thanks!

src/monitor/group_state_machine.c
```diff
@@ -406,6 +406,34 @@ ProceedGroupState(AutoFailoverNode *activeNode)
...
+			" to %s to edit synchronous_standby_names.",
+			NODE_FORMAT_ARGS(activeNode),
+			NODE_FORMAT_ARGS(primaryNode),
+			ReplicationStateGetName(primaryNode->reportedState));
+
+		/* start replication */
+		AssignGoalState(activeNode, REPLICATION_STATE_CATCHINGUP, message);
+
+		/* edit synchronous_standby_names to add the new standby now */
+		AssignGoalState(primaryNode, REPLICATION_STATE_APPLY_SETTINGS, message);
```
JelteF, line 432: Is this necessary when the node going to catchingup is a non replication quorum node?

DimCitus: Playing the _Betteridge's law of headlines_ I see. Good move. I suppose we don't have to, and I'm testing that theory now, pushing soon.

JelteF (approved):

PR #797, opened 2021-09-06
Use more cores while installing

SaitTalhaNisanci:
We can also get the core number dynamically but using `10` is probably sufficient in many cases.


SaitTalhaNisanci:
@DimCitus thanks for adding the relevant labes and milestone. I guess I should do this myself for the next PR:)

DimCitus (approved):

PR #798, opened 2021-09-06
Edit maintenance nodes out of synchronous_standby_names.

DimCitus:
When a node is ongoing maintenance, the operator can use our maintenance
state to avoid our automation to continue driving the node. We're asked to
get pg_auto_failover out of the wheel.

In that situation, we want to avoid the node in maintenance to be connected
to the primary and acknowledge transaction commits: if then later we would
have to failover to a new node, we would open the hazard of selecting a node
that doesn't have all the reported commits. Because a commit could have made
it to a node in maintenance, and this node will not be taking part of the
failover process, neither as a candidate, not as a WAL source.

To ensure that commits won't get accepted by nodes in maintenance, we now
set synchronous_standby_names to 'pgautofailover_maintenance_blocks_writes',
a standby name that is otherwise never used. That ensures blocking all the
commits on the primary while every single one of the standby nodes is
ongoing maintenance.

To unblock the situation, it is possible to either pg_autoctl disable
maintenance on a standby node, or to pg_autoctl set formation
number-sync-standbys 0.

### DIFF
--- a/src/monitor/group_state_machine.c
+++ b/src/monitor/group_state_machine.c
@@ -958,6 +958,35 @@ ProceedGroupStateForPrimaryNode(AutoFailoverNode *primaryNode)
 	}
 
 	/*
+	 * when another node wants to go to maintenance:
+	 *  apply_settings -> primary
+	 *  wait_maintenance -> maintenance
+	 */
+	if (IsCurrentState(primaryNode, REPLICATION_STATE_APPLY_SETTINGS))
+	{
+		ListCell *nodeCell = NULL;
+
+		foreach(nodeCell, otherNodesGroupList)
+		{
+			AutoFailoverNode *otherNode = (AutoFailoverNode *) lfirst(nodeCell);
+
+			if (IsCurrentState(otherNode, REPLICATION_STATE_WAIT_MAINTENANCE))
+			{
+				char message[BUFSIZE] = { 0 };
+
+				LogAndNotifyMessage(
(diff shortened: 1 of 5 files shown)


src/monitor/group_state_machine.c
```diff
@@ -957,6 +957,35 @@ ProceedGroupStateForPrimaryNode(AutoFailoverNode *primaryNode)
...
+	 */
+	if (IsCurrentState(primaryNode, REPLICATION_STATE_APPLY_SETTINGS))
+	{
+		ListCell *nodeCell = NULL;
+
+		foreach(nodeCell, otherNodesGroupList)
+		{
+			AutoFailoverNode *otherNode = (AutoFailoverNode *) lfirst(nodeCell);
+
+			if (IsCurrentState(otherNode, REPLICATION_STATE_WAIT_MAINTENANCE))
```
JelteF, line 973: this is currently never reached because there's the this check that will cause a transition from wait_maintenance to maintenance whenever target state of the primary isn't wait_primary:
https://github.com/citusdata/pg_auto_failover/blob/master/src/monitor/group_state_machine.c#L506-L507

JelteF (changes requested):
I get an error with the following steps:
```bash
# Ctrl+C node1 and wait for failover to be finished
pg_autoctl enable maintenance {secondary-node}
```
The primary will then never finish applying the settings and will continuously show this in its logs:
image

PR #799, opened 2021-09-07
Per valgrind, fix some memory leaks.

### LINKED ISSUE #679: memory leak in process pg_autoctl: node active

DimCitus:
We would be using PQfreemem on a PQExpBuffer, where it's expected to be
using destroyPQExpBuffer instead.

Also we were short of a call to finish the
connection to the monitor notification client when exiting from the service
keeper process, although adding this clean-up doesn't seem to impact the
valgrind reports.

Fixes #679 (again).

### DIFF
--- a/src/bin/pg_autoctl/pgsql.c
+++ b/src/bin/pg_autoctl/pgsql.c
@@ -1717,8 +1717,8 @@ pgsql_replication_slot_maintain(PGSQL *pgsql, NodeAddressArray *nodeArray)
 	if (!BuildNodesArrayValues(nodeArray, &sqlParams, values))
 	{
 		/* errors have already been logged */
-		PQfreemem(query);
-		PQfreemem(values);
+		destroyPQExpBuffer(query);
+		destroyPQExpBuffer(values);
 
 		return false;
 	}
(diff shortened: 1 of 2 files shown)


src/bin/pg_autoctl/pgsql.c
```diff
@@ -1640,8 +1640,8 @@ pgsql_replication_slot_create_and_drop(PGSQL *pgsql, NodeAddressArray *nodeArray
 	if (!BuildNodesArrayValues(nodeArray, &sqlParams, values))
 	{
 		/* errors have already been logged */
-		PQfreemem(query);
-		PQfreemem(values);
+		destroyPQExpBuffer(query);
+		destroyPQExpBuffer(values);
```
JelteF, line 1644: This same issue seems to be present in the function `pgsql_replication_slot_maintain`

JelteF (changes requested):

src/bin/pg_autoctl/pgsql.c
DimCitus: Fixed now. Curiously didn't appear in the valgrind report.

JelteF (approved):

PR #800, opened 2021-09-07
Add tests that we keep the replication slots OIDs constants.

DimCitus:
We want the replication slots to remain active during failovers, to
guarantee that all the nodes can reconnect and use pg_rewind, including the
old primary.

### DIFF
--- a/tests/pgautofailover_utils.py
+++ b/tests/pgautofailover_utils.py
@@ -922,6 +922,7 @@ class DataNode(PGNode):
         self.listen_flag = listen_flag
         self.formation = formation
         self.monitorDisabled = None
+        self.slots = None
 
     def create(
         self,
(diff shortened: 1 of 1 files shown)


tests/pgautofailover_utils.py
```diff
@@ -1545,15 +1546,35 @@ def list_replication_slot_names(self):
         """
         Returns a list of the replication slot names on the local Postgres.
         """
+        slots = {}
+        nameList = []
+
         query = (
-            "select slot_name from pg_replication_slots "
+            "select oid, slot_name from pg_replication_slots "
```
DimCitus, line 1553: Surprisingly, I could complete some of the unit testing on my local setup, and now I realise that the `pg_replication_slots` view does not have an OID column, and replication slots are identified with their name, I don't see that they would expose an OID. Cancelling that PR.

PR #806, opened 2021-09-14
Refactor the same error used in 6-7 different places

SaitTalhaNisanci:


src/monitor/formation_metadata.c
```diff
@@ -773,3 +771,14 @@ AutoFailoverFormationGetDatum(FunctionCallInfo fcinfo, AutoFailoverFormation *fo
...
+/*
+ * ErrorOutForNotFoundPrimary gives an error for when a
+ * primary node is not found in the given formation and group.
+ */
+void ErrorOutForNotFoundPrimary(const char* formationId, int groupId) {
+	ereport(ERROR,
+		(errmsg("Couldn't find the primary node in "
+					"formation \"%s\", group %d",
+							formationId, groupId)));
+}
```
DimCitus, line 784: I am not fond of this function's name, what about `ReportPrimaryNotFoundError` or something like that?

Then, the formatting of this piece of code isn't per the project standard. Try and run `make indent`, and if that doesn't do the job, see other function definitions around in the file. Mind about the prototype (2 lines), the brackets positions (they must be on their own lines), etc.

DimCitus (changes requested):
See the comment about the function definition (name, syntax, indentation).

Now, I don't think such refactoring are very important. Which means I will accept that we merge the PR when it's ready, but also that I can't be too excited about those code changes.

SaitTalhaNisanci:
>Now, I don't think such refactoring are very important. Which means I will accept that we merge the PR when it's ready, but also that I can't be too excited about those code changes.

The main reason for this refactor was that it seems that we want to give the same error message in those 7 places, so when we want to update one of the error messages, we will need to update all of the places. I thought that it is better that we update only one place, what downside do you see with using one place? Is that about "the error is not visible?"? If you can give the reasoning, I am okay with not merging this PR as well

DimCitus:
> The main reason for this refactor was that it seems that we want to give the same error message in those 7 places, so when we want to update one of the error messages, we will need to update all of the places. I thought that it is better that we update only one place, what downside do you see with using one place? Is that about "the error is not visible?"? If you can give the reasoning, I am okay with not merging this PR as well

Well if you spend the efforts to refactor that way, then I think it's worth it, because it's now done and as you say there is something nice about having the error message centralised in one place. That said when learning the code you now have one more indirection to learn, and if we want to change/enrich the error message in one place we need to think about changing all places or just that one, etc.

The win is not obvious, and the effort is real. But now that you've done the efforts...

Issue #718, opened 2021-06-07
pg_autoctl show settings errors during an election


JelteF:
During an election you cannot run `show settings`. It would be much better if it still showed the settings that could be found, because this makes debugging of a stuck election much harder:
```
$ pg_autoctl show settings
14:18:37 11488 ERROR Monitor ERROR:  Couldn't find the primary node in formation "default", group 0
14:18:37 11488 ERROR Monitor CONTEXT:  SQL function "formation_settings" statement 1
14:18:37 11488 ERROR Failed to retrieve formation settings from the monitor
```

PR #807, opened 2021-09-14
Not require primary node to be up for synchronous_standby_names

SaitTalhaNisanci:
We don't need to require the primary to be up to see `synchronous_standby_names`. 

Also there seems to be other places where we require `primary` node to be up. It makes sense to go over them and see if it is a requirement or we can loose the condition.

Fixes https://github.com/citusdata/pg_auto_failover/issues/718

### DIFF
--- a/src/monitor/node_active_protocol.c
+++ b/src/monitor/node_active_protocol.c
@@ -2392,13 +2392,6 @@ synchronous_standby_names(PG_FUNCTION_ARGS)
 	/* when we have more than one node, fetch the primary */
 	AutoFailoverNode *primaryNode = GetPrimaryNodeInGroup(formationId, groupId);
 
-	if (primaryNode == NULL)
-	{
-		ereport(ERROR,
-				(errmsg("Couldn't find the primary node in formation \"%s\", "
-						"group %d", formationId, groupId)));
-	}
-
 	List *standbyNodesGroupList = AutoFailoverOtherNodesList(primaryNode);
 
 	/*


src/monitor/node_active_protocol.c
```diff
@@ -2389,17 +2388,11 @@ synchronous_standby_names(PG_FUNCTION_ARGS)
...
-		ereport(ERROR,
-				(errmsg("Couldn't find the primary node in formation \"%s\", "
-						"group %d", formationId, groupId)));
-	}
+	AutoFailoverNode *primaryNode =
+		GetPrimaryNodeInGroup(formationId, groupId);
 
-	List *standbyNodesGroupList = AutoFailoverOtherNodesList(primaryNode);
+	List *standbyNodesGroupList =
+		AutoFailoverOtherNodesList(primaryNode);
```
DimCitus, line 2395: This looks like the only change is the indentation. I'd rather not review indent-only changes in a bug fix PR, can we get rid of this “noise”?

src/monitor/node_active_protocol.c
```diff
@@ -2448,7 +2441,8 @@ synchronous_standby_names(PG_FUNCTION_ARGS)
 		int count = list_length(syncStandbyNodesGroupList);
 
 		if (count == 0 ||
-			IsCurrentState(primaryNode, REPLICATION_STATE_WAIT_PRIMARY))
+			(primaryNode != NULL && IsCurrentState(primaryNode,
+												   REPLICATION_STATE_WAIT_PRIMARY)))
```
DimCitus, line 2445: Given the following current definition of `IsCurrentState`, it looks to me that this change is not needed either:

``` c
bool
IsCurrentState(AutoFailoverNode *pgAutoFailoverNode, ReplicationState state)
{
	return pgAutoFailoverNode != NULL &&
		   pgAutoFailoverNode->goalState == pgAutoFailoverNode->reportedState &&
		   pgAutoFailoverNode->goalState == state;
}
```

DimCitus (changes requested):
I think your PR can be simplified even more. Please have a look at my comments, what do you think?

src/monitor/node_active_protocol.c
```diff
@@ -2344,7 +2344,6 @@ update_node_metadata(PG_FUNCTION_ARGS)
 	{
 		nodePort = PG_GETARG_INT32(3);
 	}
-
```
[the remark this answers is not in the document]
SaitTalhaNisanci: Yeah I was trying something else initially, but the solution is simpler so these were leftovers, I removed them

DimCitus (approved):

PR #809, opened 2021-09-17
Feature/watch

DimCitus:
Implement a new `pg_autoctl watch` command that uses the Ncurses API to display a auto-updating dashboard for pg_auto_failover. The dashboard refreshes every 500ms and reacts to terminal size changes. Depending on the terminal size the output contains a different number of columns.


src/bin/pg_autoctl/cli_show.c
```diff
@@ -82,6 +84,7 @@ CommandLine show_events_command =
 				 "  --formation   formation to query, defaults to 'default' \n"
 				 "  --group       group to query formation, defaults to all \n"
 				 "  --count       how many events to fetch, defaults to 10 \n"
+				 "  --watch       display an auto-updating dashboard\n"
```
JelteF, line 87: Should this still show the state too? Or only the events. Right now it displays both.

src/bin/pg_autoctl/watch_colspecs.h
```diff
@@ -0,0 +1,173 @@
...
+		0,
+		{
+			{ COLUMN_TYPE_NAME, "Name", 0 },
+			{ COLUMN_TYPE_ID, "Node", 0 },
+			{ COLUMN_TYPE_REPLICATION_QUORUM, "Quorum", 0 },
+			{ COLUMN_TYPE_CANDIDATE_PRIORITY, "Priority", 0 },
+			{ COLUMN_TYPE_HOST_PORT, "Host:Port", 0 },
+			{ COLUMN_TYPE_TLI_LSN, "TLI: LSN", 0 },
+			{ COLUMN_TYPE_CONN_REPORT_LAG, "Report-Lag", 0 },
+			{ COLUMN_TYPE_CONN_HEALTH_LAG, "Health-Lag", 0 },
```
JelteF, line 162: Maybe use Last Report and Last Health instead of Lag, because that was considered confusing by some people.
```suggestion
			{ COLUMN_TYPE_CONN_REPORT_LAG, "Report-Lag", 0 },
			{ COLUMN_TYPE_CONN_HEALTH_LAG, "Health-Lag", 0 },
```

src/bin/pg_autoctl/watch_colspecs.h
```diff
@@ -0,0 +1,173 @@
...
+			{ COLUMN_TYPE_REPORTED_STATE, "Reported State", 0 },
+			{ COLUMN_TYPE_ASSIGNED_STATE, "Assigned State", 0 },
+			{ COLUMN_TYPE_LAST, "", 0 }
+		}
+	},
+	{
+		"terse",
+		0,
+		{
+			{ COLUMN_TYPE_ID, "Id", 0 },
```
JelteF, line 83: I think Name is more important to the users than Id (same for very terse and minimal)
```suggestion
			{ COLUMN_TYPE_ID, "Name", 0 },
```

src/bin/pg_autoctl/watch.c
```diff
@@ -0,0 +1,1102 @@
...
+
+/*
+ * catch_sigwinch is registered as the SIGWINCH signal handler.
+ */
+static void
+catch_sigwinch(int sig)
+{
+	window_size_changed = 1;
+	pqsignal(sig, catch_sigwinch);
+}
```
JelteF, line 82: Probably nice to move this to signals.c, even though it's only used here.

src/bin/pg_autoctl/watch.c
```diff
@@ -0,0 +1,1102 @@
...
+
+
+/*
+ * watch_display_window takes over the terminal window and displays the state
```
JelteF, line 86:
```suggestion
 * watch_main_loop takes over the terminal window and displays the state
```

src/bin/pg_autoctl/watch.c
```diff
@@ -0,0 +1,1102 @@
...
+ * and events in there, refreshing the output often, as when using the watch(1)
+ * command, or similar to what top(1) would be doing.
+ */
+void
+watch_main_loop(WatchContext *context)
```
JelteF, line 91: Maybe change these function names to be prefixed with `cli_` like we have in most others, especially since otherwise `watch` can be interpreted as a verb. i.e. this function watches the main loop
```suggestion
cli_watch_main_loop(WatchContext *context)
```

src/bin/pg_autoctl/watch.c
```diff
@@ -0,0 +1,1102 @@
...
+		 * strokes. We are going to update our data set twice per second, and
+		 * we want to react to key strokes and other events much faster than
+		 * this, every 50ms.
+		 */
+		step = (step + 1) % 10;
+
+		if (!watch_update(context, step))
+		{
+			/* we quit when watch_update returns false */
+			break;
```
JelteF, line 114: Because of this stopping the monitor will cause the watch command to exit, without any indication as to why this happened or even a non zero exit code. I think in this case there's two options for us:

1. Exit with non zero exit code and write something to stderr about the error.
2. Change the contents of the watch window to include some message like: Updates from monitor could not be retrieved.

As a user I'd prefer option 2, but I can imagine that's harder to implement.

src/bin/pg_autoctl/watch.c
```diff
@@ -0,0 +1,1102 @@
...
+
+		/*
+		 * First, update the data that we want to display, and process key
+		 * strokes. We are going to update our data set twice per second, and
+		 * we want to react to key strokes and other events much faster than
+		 * this, every 50ms.
+		 */
+		step = (step + 1) % 10;
+
+		if (!watch_update(context, step))
```
JelteF, line 111: I think it's okay for now, but if fetching results takes a while inputs are blocked until that is done. A common way to handle this is to do the update in a thread, so that UI interaction is not blocked.

src/bin/pg_autoctl/watch.c
```diff
@@ -0,0 +1,1102 @@
...
+
+		(void) watch_render(context);
+
+		/* and then sleep for the rest of the 50 ms */
+		INSTR_TIME_SET_CURRENT(duration);
+		INSTR_TIME_SUBTRACT(duration, start);
+
+		int sleepMs = 50 - INSTR_TIME_GET_MILLISEC(duration);
+
+		pg_usleep(sleepMs * 1000);
```
JelteF, line 131: What happens when we have spent more than 50ms on this step? We should probably have a check for negative numbers somewhere.

JelteF (commented):
Suggestions:

1. Reverse the order of the logs, this will allow us to implement scrolling at some point. It also means that the newest log is close to the state window, so you don't have to look up and down.

src/bin/pg_autoctl/monitor.c
```diff
@@ -2438,6 +2480,255 @@ printLastEvents(void *ctx, PGresult *result)
...
+	int nTuples = PQntuples(result);
+
+	int errors = 0;
+
+	log_trace("getLastEvents: %d tuples", nTuples);
+
+	if (nTuples > EVENTS_ARRAY_MAX_COUNT)
+	{
+		log_error("Query returned %d rows, pg_auto_failover supports only up "
+				  "to %d standby nodes at the moment",
```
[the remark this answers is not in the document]
DimCitus: Fixed, thanks.

src/bin/pg_autoctl/monitor.h
```diff
@@ -48,6 +48,34 @@ typedef struct StateNotification
 	int nodePort;
 } StateNotification;
 
+typedef struct MonitorEvent
+{
+	int64_t eventId;
+	char eventTime[NAMEDATALEN];
```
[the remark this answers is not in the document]
DimCitus: Fixed now.

src/bin/pg_autoctl/cli_show.c
DimCitus: Yeah we have a single watch command and 3 ways to call it. I think I like that we can add `--watch` to the existing commands and have the new dashboard, even though it's not just `show state` or `show events` anymore. I would vote for keeping it that way?

src/bin/pg_autoctl/watch_colspecs.h
DimCitus: Yeah I changed that to “Last Check” and “Last Report” in the long forms, and made the health check column move just before the connection column, then we have the last report column, and only then we have the reported/assigned states. The short forms are now just “Check” and “Report” and it might be explicit enough...

src/bin/pg_autoctl/watch_colspecs.h
DimCitus: I edited that now. The reason I picked the Id over the Name is that it's very terse indeed, allowing for more space for other information. That said, I suppose users will want to see the names of the nodes rather than their id in most situations.

src/bin/pg_autoctl/watch.c
DimCitus: I don't feel like this would help clarify the code, so I didn't do that yet. Let's chat about it if you feel more strongly about it?

src/bin/pg_autoctl/watch.c
DimCitus: Fixed now, thanks.

src/bin/pg_autoctl/watch.c
DimCitus: But the source file name is “watch.c” and we have a “cli_watch.c” that as usual implements the command line parsing! That said, I agree that “watch” being a verb may cause confusion, so I did that change anyway.

src/bin/pg_autoctl/watch.c
DimCitus: Fixed now. We go back to “cooked” mode for the terminal, and that means we see the normal stderr logs of pg_autoctl, and when we can connect to the monitor again, we're back to the Ncurses display.

src/bin/pg_autoctl/watch.c
DimCitus: I am not sure I want to implement a multi-threaded thing just for this use-case, really. I think we can live with slow updates when it takes more than 500ms to grab the result of our 3 SQL queries... at least for now...

src/bin/pg_autoctl/watch.c
DimCitus: I added a protection for negative numbers, good catch.

src/bin/pg_autoctl/watch.c
```diff
@@ -0,0 +1,1102 @@
...
+									 context->groupId,
+									 EVENTS_BUFFER_COUNT,
+									 eventsArray))
+		{
+			/* errors have already been logged */
+			return false;
+		}
+
+		/* time to finish our connection */
+		pgsql_finish(pgsql);
```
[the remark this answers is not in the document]
DimCitus: IIRC any error in the processing of the SQL query will internally close the connection already?

src/bin/pg_autoctl/watch.c
```diff
@@ -0,0 +1,1102 @@
...
+					context->rows = size.ws_row;
+					context->cols = size.ws_col;
+
+					resizeterm(context->rows, context->cols);
+				}
+			}
+		}
+
+		/* left and right moves are conditionnal / relative */
+		else if (ch == KEY_LEFT || ch == 'b' || ch == 'h')
```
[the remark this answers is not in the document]
DimCitus: I didn't know how to process CTRL+key before, so this was kind of a TODO. It's done now, thanks!

src/bin/pg_autoctl/watch.c
```diff
@@ -0,0 +1,1102 @@
...
+
+
+/*
+ * print_watch_footer prints the last line of the screen, an help message.
+ */
+static int
+print_watch_footer(WatchContext *context)
+{
+	int r = context->rows - 1;
+	char *help = "Press F1 to exit";
```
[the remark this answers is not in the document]
DimCitus: Apparently F1 is common in Ncurses land... or at least in the docs from previous century. ESC is a can of worms, because of the way a qwerty keyboard is wired, ESC then a is the same thing as M-a (or Alt+a depending on how you want to write it). So at the moment we stay away from ESC and Meta / Alt.

DimCitus:
> Suggestions:
> 
> 1. Reverse the order of the logs, this will allow us to implement scrolling at some point. It also means that the newest log is close to the state window, so you don't have to look up and down.

I have done that now, with all the other changes. I am not sure I prefer it that way, but I have added the Event Id in the output to make it obvious that what's happening here. Also allows better communication (“can you look at event 98” makes it easier than spelling out the time/date and then picking one of the entries that happened in that same second).

DimCitus:
> I think this is good to merge, except for one thing:
> 
> The "Last Check" column will only show somewhere between 1 or 5 seconds for me. Even if I took down a node. It will still reset back to 1 second even if the postgres node it is checking is down. So either there's a bug here or at least the contents of this column are not very useful to show. Since they don't seem to mean anything really.

I think this column is showing the actual monitor's health check worker behaviour and that we should get exposed to that and then see about either improving it, or fixing what we display. At the moment it's not clear to me, but I think we might want to fix the health check worker now that we see how it behaves actually.

PR #811, opened 2021-09-21
Refrain from using atexit() to clean-up our logs semaphore.

### LINKED ISSUE #742: Failed to release/acquire a lock with semaphore x: Invalid argument
On all of my instances I am seeing the following errors around every single log line.

```
3549194 Failed to acquire a lock with semaphore 46: Invalid argument
--- Log line here ---
Failed to release a lock with semaphore 46: Invalid 
```

Here is an example:
```
Jun 23 15:27:03 postgres-c pg_autoctl[547]: 547 Failed to acquire a lock with semaphore 0: Invalid argument
Jun 23 15:27:03 postgres-c pg_autoctl[547]: 15:27:03 547 ERROR Connection to database failed: timeout expired
Jun 23 15:27:03 postgres-c pg_autoctl[547]: Failed to release a lock with semaphore 0: Invalid argument
Jun 23 15:27:03 postgres-c pg_autoctl[547]: 547 Failed to acquire a lock with semaphore 0: Invalid argument
Jun 23 15:27:03 postgres-c pg_autoctl[547]: 15:27:03 547 ERROR Failed to connect to monitor database at "removed", see above for details
Jun 23 15:27:03 postgres-c pg_autoctl[547]: Failed to release a lock with semaphore 0: Invalid argument
Jun 23 15:27:03 postgres-c pg_autoctl[547]: 547 Failed to acquire a lock with semaphore 0: Invalid argument
Jun 23 15:27:03 postgres-c pg_autoctl[547]: 15:27:03 547 ERROR Failed to get node state for node 3 in group 0 of formation "default" with initial state "secondary", replication state "", and current lsn "1A/1786E9D0", see previous lines for details
Jun 23 15:27:03 postgres-c pg_autoctl[547]: Failed to release a lock with semaphore 0: Invalid argument
Jun 23 15:27:03 postgres-c pg_autoctl[547]: 547 Failed to acquire a lock with semaphore 0: Invalid argument
Jun 23 15:27:03 postgres-c pg_autoctl[547]: 15:27:03 547 ERROR Failed to get the goal state from the monitor
Jun 23 15:27:03 postgres-c pg_autoctl[547]: Failed to release a lock with semaphore 0: Invalid argument
Jun 23 15:27:06 postgres-c pg_autoctl[547]: 547 Failed to acquire a lock with semaphore 0: Invalid argument
Jun 23 15:27:06 postgres-c pg_autoctl[547]: 15:27:06 547 INFO  Successfully got the goal state from the monitor
Jun 23 15:27:06 postgres-c pg_autoctl[547]: Failed to release a lock with semaphore 0: Invalid argument
Jun 23 15:27:15 postgres-c pg_autoctl[547]: 547 Failed to acquire a lock with semaphore 0: Invalid argument
Jun 23 15:27:15 postgres-c pg_autoctl[547]: 15:27:15 547 WARN  Failed to connect to "redacted", retrying until the server is ready
Jun 23 15:27:15 postgres-c pg_autoctl[547]: Failed to release a lock with semaphore 0: Invalid argument
Jun 23 15:27:16 postgres-c pg_autoctl[547]: 547 Failed to acquire a lock with semaphore 0: Invalid argument
Jun 23 15:27:16 postgres-c pg_autoctl[547]: 15:27:16 547 INFO  Successfully connected to "removed" after 2 attempts in 2303 ms.
Jun 23 15:27:16 postgres-c pg_autoctl[547]: Failed to release a lock with semaphore 0: Invalid argument
```

Is this something to worry about or only related to logging?

DimCitus:
When using the atexit() facility, we might run into situations where an
early failure to call execve() would then call the atexit() functions, and
clean-up the parent process semaphore from underneath it.

As per the manual page for atexit(3):

       When a child process is created via fork(2), it inherits copies
       of its parent's registrations.  Upon a successful call to one of
       the exec(3) functions, all registrations are removed.

So rather than register the semaphore_finish call with the atexit()
machinery, we just manually call it from the main() function after the
command that to be run is done. We also manually call it when exiting upon
receiving the SIGQUIT signal.

Should fix #742


rheaton:
Hi @DimCitus,

This does appear to fix the issue of the error coming out, when you are trying to set `pg_ctl` path, but @jchampio and I noticed that the semaphores are no longer cleaned up if the command fails. 

For example:
1. Set the monitor `pg_ctl` to a thing that doesn't exist `pg_autoctl config set postgresql.pg_ctl /usr/local/bin/pg_ctl_not_there --pgdata ~/monitor/`
2. Attempt to fix your error using the command line `pg_autoctl config set postgresql.pg_ctl /usr/local/bin/pg_ctl --pgdata ~/monitor/`
```
10:00:39 81690 ERROR Failed to run "pg_ctl --version" using program "/usr/local/bin/pg_ctl_not_there": Undefined error: 0
10:00:39 81690 FATAL Failed to read configuration file "/Users/me/.config/pg_autoctl/Users/me/monitor/pg_autoctl.cfg"
```
3. Notice that `ipcs -s` keeps adding another semaphore each time the above is called. 

The issue with this fix is that the semaphores are no longer being cleaned up. 


Additionally, we believe there is another issue:
In the case of setting `pg_ctl` for the monitor, it attempts to call `pg_ctl --version` for the path set in the config file before setting it to the new one. We expect that it would check the one that we are setting, rather than the one in the config file (which may prevent this second issue if it didn't allow one to set to a bad binary).

Shall we open an additional Github issue for the above?


Thank you,
Rachel & Jacob

DimCitus:
> This does appear to fix the issue of the error coming out, when you are trying to set `pg_ctl` path, but @jchampio and I noticed that the semaphores are no longer cleaned up if the command fails.

After reading the code again, fixing the `exit()` calls looks like a never ending task. Instead, I have re-introduced the atexit() registration, so that we clean-up the semaphore on the way out. The way I fixed it now is to export the created semaphore Id in the environment always, so that programs started with `run_program` will have the environment set, and as a result will not clean-up the semaphore, even when an early exit happens.

> For example:
> 
> 1. Set the monitor `pg_ctl` to a thing that doesn't exist `pg_autoctl config set postgresql.pg_ctl /usr/local/bin/pg_ctl_not_there --pgdata ~/monitor/`

I think we should fix that by checking that we can run `pg_ctl --version` with the given PATH here. I suppose this check should happen in `monitor_config_accept_new`, and maybe we have the same check to implement at `keeper_config_accept_new`.

> 2. Attempt to fix your error using the command line `pg_autoctl config set postgresql.pg_ctl /usr/local/bin/pg_ctl --pgdata ~/monitor/`
> 
> ```
> 10:00:39 81690 ERROR Failed to run "pg_ctl --version" using program "/usr/local/bin/pg_ctl_not_there": Undefined error: 0
> 10:00:39 81690 FATAL Failed to read configuration file "/Users/me/.config/pg_autoctl/Users/me/monitor/pg_autoctl.cfg"
> ```

I don't see those anymore.

>  1. Notice that `ipcs -s` keeps adding another semaphore each time the above is called.

This is yet unclear to me. I had a couple semaphore listed after a run, but I don't trust my environment to have been clean to start with, so that's an open question. Please have a look if you can, and I will get back to testing this in a clean docker environment or something later too.

> Additionally, we believe there is another issue:
> In the case of setting `pg_ctl` for the monitor, it attempts to call `pg_ctl --version` for the path set in the config file before setting it to the new one. We expect that it would check the one that we are setting, rather than the one in the config file (which may prevent this second issue if it didn't allow one to set to a bad binary).
> 
> Shall we open an additional Github issue for the above?

Yes please. Sounds like something we should fix independently.

DimCitus:
> > This does appear to fix the issue of the error coming out, when you are trying to set `pg_ctl` path, but @jchampio and I noticed that the semaphores are no longer cleaned up if the command fails.
> 
> After reading the code again, fixing the `exit()` calls looks like a never ending task. Instead, I have re-introduced the atexit() registration, so that we clean-up the semaphore on the way out. The way I fixed it now is to export the created semaphore Id in the environment always, so that programs started with `run_program` will have the environment set, and as a result will not clean-up the semaphore, even when an early exit happens.

So after more thinking and playing around the idea, I have now added a semaphore owner (set to `gettpid()` result). The owner of the semaphore is of course only set when we create a new semaphore, it's left to zero (0) when a pre-existing semaphore is just opened in a sub-process.

I suppose this should be it now.

rheaton:
> > 1. Attempt to fix your error using the command line `pg_autoctl config set postgresql.pg_ctl /usr/local/bin/pg_ctl --pgdata ~/monitor/`
> > 
> > ```
> > 10:00:39 81690 ERROR Failed to run "pg_ctl --version" using program "/usr/local/bin/pg_ctl_not_there": Undefined error: 0
> > 10:00:39 81690 FATAL Failed to read configuration file "/Users/me/.config/pg_autoctl/Users/me/monitor/pg_autoctl.cfg"
> > ```
> 
> I don't see those anymore.
> 
> > 1. Notice that `ipcs -s` keeps adding another semaphore each time the above is called.
> 
> This is yet unclear to me. I had a couple semaphore listed after a run, but I don't trust my environment to have been clean to start with, so that's an open question. Please have a look if you can, and I will get back to testing this in a clean docker environment or something later too.


@DimCitus Tried out these changes, they fix the issue on my system. Additionally, I don't have any semaphores hanging around with the latest update 


> > Additionally, we believe there is another issue:
> > In the case of setting `pg_ctl` for the monitor, it attempts to call `pg_ctl --version` for the path set in the config file before setting it to the new one. We expect that it would check the one that we are setting, rather than the one in the config file (which may prevent this second issue if it didn't allow one to set to a bad binary).
> > Shall we open an additional Github issue for the above?
> 
> Yes please. Sounds like something we should fix independently.

Issue created: https://github.com/citusdata/pg_auto_failover/issues/813

src/bin/lib/subcommands.c/runprogram.h
```diff
@@ -183,6 +183,15 @@ execute_subprogram(Program *prog)
 	int outpipe[2] = { 0, 0 };
 	int errpipe[2] = { 0, 0 };
 
+	/* first level sanity check */
+	if (access(prog->program, F_OK | X_OK) == -1)
+	{
+		fprintf(stderr, "Failed to find executable program at \"%s\": %s",
+				prog->program,
+				strerror(errno));
+		return;
```
jchampio, line 192: Do `prog->returnCode` and `prog->error` need to be set in these code paths?

DimCitus: Good point, thanks! Done now.

JelteF:
When I manually check `ipcs -s` on this branch there's no semaphores created at all. So I think there's a bug in some code here somewhere.

rheaton:
@DimCitus We are working on #813 off of this branch, and noticed that when we call `pg_autoctl create` for both `monitor` and `postgres` we get the following new error in the log:

`11:38:16 12191 FATAL BUG: semaphore_finish semId 131072 owner is 12186, getpid is 12191`

PR #815, opened 2021-10-04
Work around pg_replication_slot_advance xmin maintenance bug.

### LINKED ISSUE #814: Failover results in stale/stuck replication slots
_I believe this is a duplicate of #662, which was closed not too long ago. FYI @porsager_

We've run into a production instance that was complaining about `WARNING:  oldest xmin is far in the past`, a few days after a failover occurred. We discovered that an inactive replication slot on the _secondary_ was pinning an old xmin value. (Dropping and recreating the _primary's_ replication slots did not help, which is potentially why that workaround did not help #662?) Dropping the secondary's slots released the xmin.

We were able to reproduce this situation with the following steps (though we don't know for sure that this is what happened to the production cluster):
1. Set up a two-node instance. Add a table with some data. At this point, the primary should have an active replication slot with a recent xmin, and the secondary should have a blank, inactive slot.
2. Kill the primary, and wait for the demotion timeout and for the secondary to be promoted to primary.
3. Bring the demoted node back online and wait for it to rewind.
4. Connect to the new secondary and note that a stale, inactive replication slot is still holding onto that old xmin.
5. Connect to the primary and commit some transactions. Note that the primary's replication slot never advances its xmin (since the secondary's slot is still pinning it).

Performing a manual switchover also triggers this issue. But if I manually kill the primary and bring it back up _before_ the demotion timeout, I don't see this problem, even after the failover.

We've reproduced against Postgres 13.2, with pg_auto_failover versions 1.5.2 and 1.6.2.

DimCitus:
When the Postgres function pg_replication_slot_advance is called on a slot
that used to be active (using streaming replication) and is now
inactive (maintained manually) then the xmin of the slot is not maintained
anymore.

To avoid situations where Postgres won't VACUUM dead rows and otherwise
maintains itself correctly, when a former primary node is available again
and joind a pg_auto_failover as a standby, we drop the pre-existing
replication slot and create new ones (where this time xmin is NULL).

Fixes #814


JelteF:
For reference this is the bugreport to postgres about this: https://www.postgresql.org/message-id/flat/VI1PR83MB01897A49AE5C54A82F841D8999B09%40VI1PR83MB0189.EURPRD83.prod.outlook.com

PR #838, opened 2021-11-24
Quick adjusting for Postgres 15devel compatibility.

DimCitus:


thanodnl (approved):
Compiles on pg15, which the main branch doesn't 🎉 
🚢

PR #853, opened 2021-12-15
Add support for tablespaces

### LINKED ISSUE #844: No failover with a separate tablespace
Hi

We are currently testing the pg_auto_failover tool for us.
The installation went off without any problems. 
There are three nodes - 1x monitor and 2x dbs with pg_auto_failover 1.6.3 
The nodes are ipv6 only. PostreSQL version is 12 and the OS is RedHat 8.
After creating a tablespace under the path /srv/tblspc/app for example, a failover is not possible.

- from monitor:
]$ pg_autoctl perform failover --pgdata ./monitor

- log output:
pg_autoctl[1577]: 14:13:15 1608 INFO   /usr/pgsql-12/bin/pg_basebackup -w -d 'application_name=pgautofailover_standby_1 host=xxx.fqdn.tld>
pg_autoctl[1577]: 14:13:15 1608 INFO  pg_basebackup: initiating base backup, waiting for checkpoint to complete
pg_autoctl[1577]: 14:13:15 1608 INFO  pg_basebackup: checkpoint completed
pg_autoctl[1577]: 14:13:15 1608 INFO  pg_basebackup: write-ahead log start point: 1/28000028 on timeline 12
pg_autoctl[1577]: 14:13:15 1608 INFO  pg_basebackup: error: directory "/srv/tblspc/app" exists but is not empty
pg_autoctl[1577]: 14:13:15 1608 INFO  pg_basebackup: removing contents of data directory "/var/lib/pgsql/backup/node_1"
pg_autoctl[1577]: 14:13:15 1608 ERROR Failed to run pg_basebackup: exit code 1
pg_autoctl[1577]: 14:13:15 1608 ERROR Failed to become standby server, see above for details
pg_autoctl[1577]: 14:13:15 1608 ERROR Failed to transition from state "demoted" to state "catchingup", see above.
pg_autoctl[1577]: 14:13:15 1608 ERROR Failed to transition to state "catchingup", retrying...

The content must always be deleted under the path (/srv/tblspc/app/PG_xxx) where the tablespace is located. Then the failover works.

Are we doing something wrong or is that perhaps a bug?

rheaton:
This allows up to 16 tablespace-mappings to be defined when creating nodes. 

Fixes #844

Notes: 
- This is not moving the mapped directory after the pg_basebackup as it requires more work than just a simple `mv` to do it. What are the advantages or disadvantages of this technique? 
- Additionally, I'm not in love with `--tablespace-mappings=/mnt/giant=/mnt/giantbackup;/mnt/superfast=/mnt/superfastbackup` as a UX, but it appeared to involve the fewest changes.
 - At first I attempted to detect the existing tablespaces by running a query against the primary, but it's not trivial to get them as a replication type role. In production, I expect the user would want to set up their directory structure by hand and would need to know that the volume mounts need to exist. The python tests don't currently have a way to replicate a more production-like scenario with volume mounts of the same name for each node.

I'll be around a few days next week and this week to make updates from feedback/changes. 🙂

### DIFF
--- a/docs/ref/pg_autoctl_create_postgres.rst
+++ b/docs/ref/pg_autoctl_create_postgres.rst
@@ -41,6 +41,7 @@ registered too, and is known to be healthy).
      --candidate-priority    priority of the node to be promoted to become primary
      --replication-quorum    true if node participates in write quorum
      --maximum-backup-rate   maximum transfer rate of data transferred from the server during initial sync
+     --tablespace-mappings   tablespace mapping options for pg_basebackup; semicolon separated
 
 Description
 -----------
@@ -303,3 +304,14 @@ The following options are available to ``pg_autoctl create postgres``:
 --server-cert
 
   Set the Postgres ``ssl_cert_file`` to that file path.
+
+--tablespace-mappings
+
+  Sets the tablespace mapping options for ``pg_basebackup``.
+  Tablespace mappings should be semicolon separated.
+
+  This option should be used when creating nodes for a cluster with additional tablespaces.
+  To add additional an additional tablespace after the fact, see the Postgres documentation.
+
+  __ https://www.postgresql.org/docs/current/manage-ag-tablespaces.html
+  __ https://www.postgresql.org/docs/current/warm-standby.html#STANDBY-SERVER-OPERATION
--- a/src/bin/pg_autoctl/cli_common.c
+++ b/src/bin/pg_autoctl/cli_common.c
@@ -573,6 +573,20 @@ cli_common_keeper_getopts(int argc, char **argv,
 				break;
 			}
 
+			case 'T':
+			{
+				strlcpy(LocalOptionConfig.tablespaceMappingsStr, optarg,
+						PG_AUTOCTL_MAX_TABLESPACES_STR);
+				log_trace("--tablespace-mappings %s",
+						  LocalOptionConfig.tablespaceMappingsStr);
+				if (!parse_tablespace_mappings(&LocalOptionConfig))
+				{
+					/* error has already been logged */
+					errors++;
+				}
+				break;
+			}
+
 			default:
 			{
 				/* getopt_long already wrote an error message */
@@ -705,6 +719,7 @@ cli_common_keeper_getopts(int argc, char **argv,
  *      { "server-crt", required_argument, &ssl_flag, SSL_SERVER_CRT_FLAG },
  *      { "server-key", required_argument, &ssl_flag, SSL_SERVER_KEY_FLAG },
  *      { "ssl-mode", required_argument, &ssl_flag, SSL_MODE_FLAG },
+ *      { "tablespace-mappings", no_argument, NULL, 'T' },
  *		{ NULL, 0, NULL, 0 }
  *	};
  *
--- a/src/bin/pg_autoctl/cli_create_node.c
+++ b/src/bin/pg_autoctl/cli_create_node.c
@@ -92,7 +92,8 @@ CommandLine create_postgres_command =
 		KEEPER_CLI_SSL_OPTIONS
 		"  --candidate-priority    priority of the node to be promoted to become primary\n"
 		"  --replication-quorum    true if node participates in write quorum\n"
-		"  --maximum-backup-rate   maximum transfer rate of data transferred from the server during initial sync\n",
+		"  --maximum-backup-rate   maximum transfer rate of data transferred from the server during initial sync\n"
+		"  --tablespace-mappings   tablespace mapping options for pg_basebackup; semicolon separated\n",
 		cli_create_postgres_getopts,
 		cli_create_postgres);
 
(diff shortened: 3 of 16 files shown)


DimCitus:
Thanks a lot for working on this @rheaton ! So my first question is: what happens if non-default (extra) table spaces have been created on an instance and the new option `--tablespace-mappings` is not used at all? My current (albeit quick) reading of your patch makes me think we have the same problem as before in that case, is that true?

Reading through [pg_basebackup](https://github.com/postgres/postgres/blob/master/src/bin/pg_basebackup/pg_basebackup.c#L1265) source code and then [streaming replication protocol BASE_BACKUP docs](https://www.postgresql.org/docs/current/protocol-replication.html#PROTOCOL-REPLICATION-BASE-BACKUP) I'm left wondering if we can make it work our way. 

The better option that I can think of right now consists of using the `--format=tar` option to `pg_basebackup` so that we fetch the files from upstream as tar archives in our `replication.backup_directory` location. That would make sense actually, and also help understand our two phases approach. We would get a list of table spaces that way (one tar archive file per table space), and we can also inspect the `base.tar` file for the `tablespace_map` file and have our default mapping from there.

Then about providing a table space mapping. Do we actually want to do that? I mean, I like the idea that every node in the HA system should be the same, right?

In passing, another discussion internally led us to rethink the way we are handling the base backup transition in pg_autoctl. At the moment I made the choice to first download the new PGDATA copy, and then when it is known that we have a new copy, we delete the old PGDATA and replace it. That's because of my years as a DBA: I want to be able to do forensics on the old PGDATA copy for as long as possible. The drawback is that you know need double the disk space. The idea is to make this behaviour an configuration option, the default being remove PGDATA first then fetch a new copy, and a tuning would allow for delete-as-late-as-possible.

rheaton:
I like keeping everything the same, the idea with this patch is that the `tablespace-mappings` parameter is _required_ if you are using tablespaces, so we can make sure the directories are cleared. I believe if the user configures the tablespace-mappings to link to the same thing (e.g. originaldir=originaldir) we would have the desired behavior. Which makes me think we should just require the user to give us `tablespaces=/mnt/superfastdisk;/mnt/superhugedisk` and then we can make sure those locations exist and are empty. 

I worry about the tar solution for the same reason as the double-the-disk-space issue you mentioned (and one was filed today). 

I think users who have tablespaces can give us the list and we can make this work pretty simply. We are going to have to figure out a way to have separate containers with matching sets of mount dirs to automate these tests in a more realistic scenario. 

What do you think about requiring the user to specify their tablespaces like so: `tablespaces=/mnt/superfastdisk;/mnt/superhugedisk` and trying to figure out a way to test it? I'm willing to take this on, but it might not be completed before the Christmas-time holiday.

DimCitus:
My first reaction is still the same, we should find a way that pg_auto_failover just work without user intervention, even when they're using non-default table spaces. Maybe what we could do as a first step is implement _delete PGDATA first, then pg_basebackup in place_ so that it works for everyone. And then we can add the option of keeping PGDATA around for as long as possible, which is very costly in terms of disk space, and in that case we're good with using `pg_basebackup --format=tar` I suppose. After all this would be for paranoid users who know they have the necessary disk space or accept the failures when they don't.

rheaton:
I think that we'd have the same issue for tablespaces, as there isn't currently a reasonable way to get the tablespace locations prior to backing up. We won't be able to delete those locations like we do for PGDATA. 

I agree that we should do the PGDATA deletion, in addition to something for tablespaces.

DimCitus:
> I think that we'd have the same issue for tablespaces, as there isn't currently a reasonable way to get the tablespace locations prior to backing up. We won't be able to delete those locations like we do for PGDATA.

There must be a way to read the list of table spaces from the local PGDATA. A quick test locally seems to show that we can read the local `pg_tblspc` directory. It contains a list of symlinks and that's where we need to remove stuff. If a new table space has been created on the new primary, it's not a problem, because if we don't have it listed in the local PGDATA that we want to destroy then we know we won't have a conflict at pg_basebackup time.

```
$ ls -l monitor/pg_tblspc/
total 0
lrwx------  1 dim  staff  8 Dec 15 21:01 17165 -> /tmp/foo

$ psql -X -p 5500 -d postgres -c "select oid, spcname, pg_catalog.pg_tablespace_location(oid) AS location from pg_tablespace"
  oid  |  spcname   | location
-------+------------+----------
  1663 | pg_default |
  1664 | pg_global  |
 17165 | foo        | /tmp/foo
(3 rows)
```

What do you think?

rheaton:
Oh good idea! I didn't think about just looking in the existing directory. Obviously there can be cases where this won't work, but I think pgbackrest will just fail out with a decent error message. 

Where should we go next for this? Would you like me to remove the mapping bit and have it "just work" for tablespaces with in-place restore, or should we wait for the PGDATA in-place work to be complete first? 
And if so, is there a timeline for that?

DimCitus:
Awesome, I'm happy that you like the idea. I think we should make the thing “just work” with table spaces first, which means introducing the configuration knob to default to removing PGDATA first. I suppose the PR that does that could skip the implementation of configuration knob turned to paranoid/forensic mode (keep PGDATA for as long as possible, use `pg_basebackup --format=tar`) and just raise an error for now. Another PR could implement that behaviour I suppose.

In terms of timeline, we are planning our next release to happen sometime in January. If this work misses the release schedule, then it will be part of the March release.

jchampio:
My concern with ever using `pg_basebackup --format=tar` is that it not only requires you to have double the disk space available, but it also requires you to be able to fit the data from all tablespaces on a single filesystem, and then pay the cost of recopying all those bits across any mount points you have (since `mv`/`rename()` won't work). I think that's likely to be impossible for several real-world tablespace use cases.

DimCitus:
> My concern with ever using `pg_basebackup --format=tar` is that it not only requires you to have double the disk space available, but it also requires you to be able to fit the data from all tablespaces on a single filesystem, and then pay the cost of recopying all those bits across any mount points you have (since `mv`/`rename()` won't work, atomically at least). I think that's likely to be impossible for several real-world tablespace use cases.

I share the concern. I think using `-z -Z9` and dealing with `.tar.gz` then is going to be necessary, but won't alleviate the problem entirely. At the moment I'm not seeing a better option though for the forensics options, and that's also why I am saying it should then stop being the default... activate only if you know what you're doing (either not using table spaces at all, or you live within the requirements for the forensics abilities, etc).

rheaton:
Closing in favour of #870 per discussion above

PR #856, opened 2021-12-15
Correct typos in documentation and tests

eradman:


eradman:
I also considered changing "formated" to "formatted", but perhaps that should be a separate commit?

DimCitus:
> I also considered changing "formated" to "formatted", but perhaps that should be a separate commit?

I suppose it would be nice to have it in the same PR, even if in a commit on-top of the existing one?

DimCitus (approved):
Thanks for your contribution! Fixing typos is very much appreciated!

PR #862, opened 2022-01-13
Check return value of strdup calls

danielgustafsson:
Albeit unlikely on modern server systems, memory allocation can still fail and the result, if unchecked, be null pointer dereferences. This adds a check to affected callsites of strdup to ensure allocation was successful, or abort if not.

### DIFF
--- a/src/bin/pg_autoctl/debian.c
+++ b/src/bin/pg_autoctl/debian.c
@@ -282,6 +282,11 @@ buildDebianDataAndConfDirectoryNames(PostgresSetup *pgSetup,
 	char clusterDir[MAXPGPATH] = { 0 };
 	char versionDir[MAXPGPATH] = { 0 };
 
+	if (pgmajor == NULL)
+	{
+		log_error(ALLOCATION_FAILED_ERROR);
+		return false;
+	}
 
 	/* we need to work with the absolute pathname of PGDATA */
 	if (!normalize_filename(pgSetup->pgdata, pgdata, MAXPGPATH))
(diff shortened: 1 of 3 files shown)


DimCitus:
Hi @danielgustafsson ; thanks for your contribution!

Can you have a look at using [ALLOCATION_FAILED_ERROR](https://github.com/citusdata/pg_auto_failover/blob/master/src/bin/pg_autoctl/defaults.h#L175) instead of the error format you've been using here? It's already used in a couple places: https://github.com/citusdata/pg_auto_failover/search?q=ALLOCATION_FAILED_ERROR&type=code.

danielgustafsson:
> Hi @danielgustafsson ; thanks for your contribution!
> 
> Can you have a look at using [ALLOCATION_FAILED_ERROR](https://github.com/citusdata/pg_auto_failover/blob/master/src/bin/pg_autoctl/defaults.h#L175) instead of the error format you've been using here? It's already used in a couple places: https://github.com/citusdata/pg_auto_failover/search?q=ALLOCATION_FAILED_ERROR&type=code.

Oh nice, that's a much bette fit. I'll fix that.

DimCitus:
Sorry about that, I missed it before. For clarity, I think I would prefer using `value == NULL` rather than `!value` in the tests. What do you think? I like code better when it's obvious, and `== NULL` can't be more obvious, I believe...

danielgustafsson:
> Sorry about that, I missed it before. For clarity, I think I would prefer using `value == NULL` rather than `!value` in the tests. What do you think? I like code better when it's obvious, and `== NULL` can't be more obvious, I believe...

I don't have strong opinions either way, so I've updated to patch to your preference.

DimCitus (approved):

danielgustafsson:
Thanks!

PR #870, opened 2022-01-27
Add support for tablespaces

### LINKED ISSUE #844: No failover with a separate tablespace

rheaton:
There is a new folder of tests, using volumes to keep data persistent
against pod restarts and to allow for interaction within and without of
the running containers. Unfortunately, it was non-trivial to re-use some of the existing testing libraries, but the additions were minimal. There are minor changes to the test Dockerfile to add permissions to the additional data directories.

This adds a newly _destructive_ action that removes the data within a tablespace
directory. Perhaps this can be rebased off of the work that will do in-place
restore before merging, depending on the caution of the core team.

Implementation note: this permits tablespace data directories to
created at the top level of mounted volumes.

Fixes #844

### DIFF
--- a/src/bin/pg_autoctl/pgctl.c
+++ b/src/bin/pg_autoctl/pgctl.c
@@ -83,7 +83,7 @@ static bool pg_write_recovery_conf(const char *pgdata,
 								   ReplicationSource *replicationSource);
 static bool pg_write_standby_signal(const char *pgdata,
 									ReplicationSource *replicationSource);
-
+static bool ensure_empty_tablespace_dirs(const char *pgdata);
 
 /*
  * Get pg_ctl --version output in pgSetup->pg_version.
@@ -1172,6 +1172,83 @@ prepare_guc_settings_from_pgsetup(const char *configFilePath,
 }
 
 
+bool
+ensure_empty_tablespace_dirs(const char *pgdata)
+{
+	char *dirName = "pg_tblspc";
+	struct dirent *dirEntry = NULL;
+
+	char pgTblspcFullPath[MAXPGPATH] = { 0 };
+	sformat(pgTblspcFullPath, MAXPGPATH, "%s/%s", pgdata, dirName);
+
+	if (!directory_exists(pgTblspcFullPath))
+	{
+		log_debug("Postgres dir pg_tblspc does not exist at \"%s\"",
+				  pgTblspcFullPath);
+		return true;
+	}
+
+
+	/* open and scan through the Postgres tablespace directory */
+	DIR *tblspcDir = opendir(pgTblspcFullPath);
+
+	if (tblspcDir == NULL)
+	{
+		log_error("Failed to open Postgres tablespace directory \"%s\" at \"%s\"",
+				  dirName, pgdata);
+		return false;
+	}
+
+	while ((dirEntry = readdir(tblspcDir)) != NULL)
+	{
+		char tblspcDataDirPath[MAXPGPATH] = { 0 };
+
+		/* skip non-symlinks, as all tablespace dirs are symlinks */
+		if (dirEntry->d_type == DT_UNKNOWN)
+		{
+			struct stat structStat;
+			char dirEntryFullPath[MAXPGPATH] = { 0 };
+			sformat(dirEntryFullPath, MAXPGPATH, "%s/%s", pgTblspcFullPath,
+					dirEntry->d_name);
+			if (lstat(dirEntryFullPath, &structStat) != 0)
+			{
+				log_error("Failed to get file information for \"%s/%s\": %m",
+						  pgTblspcFullPath, dirEntry->d_name);
+				return false;
+			}
+			if (!S_ISLNK(structStat.st_mode))
+			{
+				log_debug("Non-symlink file found in tablespace directory: \"%s/%s\"",
+						  pgTblspcFullPath, dirEntry->d_name);
+				continue;
+			}
+		}
+		else if (dirEntry->d_type != DT_LNK)
+		{
+			log_debug("Non-symlink file found in tablespace directory: \"%s/%s\"",
+					  pgTblspcFullPath, dirEntry->d_name);
+			continue;
+		}
+
+		join_path_components(tblspcDataDirPath, pgTblspcFullPath, dirEntry->d_name);
+
+		log_debug("Removing contents of tablespace data directory \"%s\"",
+				  tblspcDataDirPath);
+
+		/* remove contents of tablespace data directory */
+		if (!rmtree(tblspcDataDirPath, false))
+		{
+			log_error(
+				"Failed to remove contents of existing tablespace data directory \"%s\": %m",
+				tblspcDataDirPath);
+			return false;
+		}
+	}
+
+	return true;
+}
+
+
 /*
  * Call pg_basebackup, using a temporary directory for the duration of the data
  * transfer.
@@ -1194,6 +1271,12 @@ pg_basebackup(const char *pgdata,
 
 	log_debug("mkdir -p \"%s\"", replicationSource->backupDir);
 	if (!ensure_empty_dir(replicationSource->backupDir, 0700))
+	{
+		/* errors have already been logged. */
+		return false;
+	}
+
+	if (!ensure_empty_tablespace_dirs(pgdata))
 	{
 		/* errors have already been logged. */
 		return false;
--- a/tests/tablespaces/Dockerfile
+++ b/tests/tablespaces/Dockerfile
@@ -1,0 +1,11 @@
+ARG PGVERSION=10
+
+FROM pg_auto_failover_test:pg${PGVERSION} as test
+
+USER root
+RUN mkdir -p /extra_volumes/extended_a && chown -R docker /extra_volumes/extended_a
+RUN mkdir -p /extra_volumes/extended_b && chown -R docker /extra_volumes/extended_b
+RUN mkdir -p /extra_volumes/extended_c && chown -R docker /extra_volumes/extended_c
+RUN mkdir -p /var/lib/postgres && chown -R docker /var/lib/postgres
+
+USER docker
--- a/tests/tablespaces/Makefile
+++ b/tests/tablespaces/Makefile
@@ -2,23 +2,23 @@ PGVERSION ?= 10
 
 build:
 	$(MAKE) -C ../../ build-test-pg$(PGVERSION)
-	docker compose build --build-arg "PGVERSION=$(PGVERSION)"
+	docker-compose build --build-arg "PGVERSION=$(PGVERSION)"
 
 teardown:
-	docker compose down --volumes --remove-orphans
+	docker-compose down --volumes --remove-orphans
 
 tail:
-	docker compose logs -f
+	docker-compose logs -f
 
 state:
-	docker compose exec monitor pg_autoctl show state
+	docker-compose exec monitor pg_autoctl show state
 
 watch:
-	docker compose exec monitor pg_autoctl watch
+	docker-compose exec monitor pg_autoctl watch
 
 test_01:
-	docker compose up -d node1
-	TEST_NUM="01" docker compose up --exit-code-from=test test
+	docker-compose up -d node1
+	TEST_NUM="01" docker-compose up --exit-code-from=test test
 
 test_02:
 	docker-compose up -d node2
@@ -26,21 +26,21 @@ test_02:
 
 test_03:
 	# Pollute the data dir so pg_rewind fails
-	docker compose exec node2 mkdir '/var/lib/postgres/data/base/1/99999999'
-	docker compose exec monitor pg_autoctl perform failover
-	TEST_NUM="03" docker compose up --exit-code-from=test test
+	docker-compose exec node2 mkdir '/var/lib/postgres/data/base/1/99999999'
+	docker-compose exec monitor pg_autoctl perform failover
+	TEST_NUM="03" docker-compose up --exit-code-from=test test
 
 test_04:
-	docker compose pause node1
-	TEST_NUM="04" docker compose up --exit-code-from=test test
+	docker-compose pause node1
+	TEST_NUM="04" docker-compose up --exit-code-from=test test
 
 test_05:
-	docker compose unpause node1
-	TEST_NUM="05" docker compose up --exit-code-from=test test
+	docker-compose unpause node1
+	TEST_NUM="05" docker-compose up --exit-code-from=test test
 
 test_06:
-	docker compose exec monitor pg_autoctl perform failover
-	TEST_NUM="06" docker compose up --exit-code-from=test test
+	docker-compose exec monitor pg_autoctl perform failover
+	TEST_NUM="06" docker-compose up --exit-code-from=test test
 
 test: build test_01 test_02 test_03 test_04 test_05 test_06
 
--- a/tests/tablespaces/Makefile
+++ b/tests/tablespaces/Makefile
@@ -1,0 +1,55 @@
+PGVERSION ?= 10
+
+build:
+	$(MAKE) -C ../../ build-test-pg$(PGVERSION)
+	docker compose build --build-arg "PGVERSION=$(PGVERSION)"
+
+teardown:
+	docker compose down --volumes --remove-orphans
+
+tail:
+	docker compose logs -f
+
+state:
+	docker compose exec monitor pg_autoctl show state
+
+watch:
+	docker compose exec monitor pg_autoctl watch
+
+test_01:
+	docker compose up -d node1
+	TEST_NUM="01" docker compose up --exit-code-from=test test
+
+test_02:
+	docker-compose up -d node2
+	TEST_NUM="02" docker-compose up --exit-code-from=test test
+
+test_03:
+	# Pollute the data dir so pg_rewind fails
+	docker compose exec node2 mkdir '/var/lib/postgres/data/base/1/99999999'
+	docker compose exec monitor pg_autoctl perform failover
+	TEST_NUM="03" docker compose up --exit-code-from=test test
+
+test_04:
+	docker compose pause node1
+	TEST_NUM="04" docker compose up --exit-code-from=test test
+
+test_05:
+	docker compose unpause node1
+	TEST_NUM="05" docker compose up --exit-code-from=test test
+
+test_06:
+	docker compose exec monitor pg_autoctl perform failover
+	TEST_NUM="06" docker compose up --exit-code-from=test test
+
+test: build test_01 test_02 test_03 test_04 test_05 test_06
+
+#  The test containers only connect to each other via the postgres protocol.
+#    For this reason, the Makefile uses  `docker-compose exec` to perform
+#    `pg_autoctl` commands, such as failover. Starting/stopping the pg_autoctl
+#    process is also done using `docker-compose` so as not to complicate the
+#    command used to run each container.
+#  See CONTRIBUTING.md for more detailed instructions.
+run-test: test teardown
+
+.PHONY: build state watch run-test test tail teardown test_01 test_02 test_03 test_04 test_05 test_06
--- a/tests/tablespaces/docker-compose.yml
+++ b/tests/tablespaces/docker-compose.yml
@@ -1,0 +1,69 @@
+version: "3.9"  # optional since v1.27.0
+services:
+  monitor:
+    build: .
+    environment:
+      PGDATA: /tmp/pgdata
+      PG_AUTOCTL_DEBUG: 1
+    command: pg_autoctl create monitor --ssl-self-signed --auth trust --run
+    expose:
+      - 5432
+    container_name: monitor
+  node1:
+    build: .
+    environment:
+      PGDATA: /var/lib/postgres/data
+      PGPORT: 5433
+      PG_AUTOCTL_MONITOR: "postgresql://autoctl_node@monitor:5432/pg_auto_failover"
+      HOSTNAME: "node1.tablespaces_default"
+    command: pg_autoctl create postgres --ssl-self-signed --auth trust --pg-hba-lan --run
+    expose:
+      - 5433
+    links:
+      - monitor
+    container_name: node1
+    volumes:
+      - node1-volume:/var/lib/postgres:rw
+      - node1-a-volume:/extra_volumes/extended_a:rw
+      - node1-b-volume:/extra_volumes/extended_b:rw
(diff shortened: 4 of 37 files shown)


Dockerfile
```diff
@@ -165,6 +165,13 @@ COPY --from=build /usr/lib/postgresql/${PGVERSION}/bin/pg_autoctl /usr/local/bin
 RUN mkdir -p /var/lib/postgres \
  && chown -R docker /var/lib/postgres
 
+RUN mkdir -p /tmp/extended_a && chown -R docker /tmp/extended_a
+RUN mkdir -p /tmp/extended_b && chown -R docker /tmp/extended_b
+RUN mkdir -p /tmp/extended_c && chown -R docker /tmp/extended_c
+RUN mkdir -p /tmp/pg/data \
+    && chown -R docker /tmp/pg \
+    && chown -R docker /tmp/pg/data
+
```
DimCitus, line 174: Is there a way to keep those directories outside of the main Dockerfile, and also, in another directory... maybe in /var/lib/postgres/tablespaces or something. About the Dockerfile, we could have a specific Dockerfile in the new test directory, and use `build: .` in the docker-compose file. The new Dockerfile could inherit (as in `FROM ...`) from the local image. I think `FROM pg_auto_failover` would do the trick.

src/bin/pg_autoctl/pgctl.c
```diff
@@ -1172,6 +1172,82 @@ prepare_guc_settings_from_pgsetup(const char *configFilePath,
...
+	while ((tblspcDirEntry = readdir(tblspcDir)) != NULL)
+	{
+		char tblspcDataDirPath[MAXPGPATH] = { 0 };
+
+		/* skip non-symlinks, as all tablespace dirs are symlinks */
+		if (tblspcDirEntry->d_type == DT_UNKNOWN)
+		{
+			struct stat structStat;
+			char pgTblspcEntryFullPath[MAXPGPATH] = { 0 };
+			sprintf(pgTblspcEntryFullPath, "%s/%s", pgTblspcFullPath, tblspcDirEntry->d_name);
```
DimCitus, line 1211: Should we use `sformat` here, to avoid an error when then using `./ci/banned.h.sh` checker?

DimCitus (changes requested):
Thanks! On a quick review, it looks good to me! Simpler than before too, if I remember correctly. Now that's a lot of testing attached to it, and I didn't review that part yet. Will do soon.

Dockerfile
rheaton: Yes, I was hoping we could clean this up a bit. Thanks for the guidance.

rheaton: I'll try this out now-- I was following what exists in the upgrade docker-compose

rheaton:
Cleaning up a few compiler warnings and I'll push up fixes today

Dockerfile
rheaton: Pushed a fix

src/bin/pg_autoctl/pgctl.c
rheaton: Pushed a fix!

tests/tablespaces/docker-compose.yml
```diff
@@ -0,0 +1,81 @@
...
+      PG_AUTOCTL_DEBUG: 1
+    command: [
+    "pg_autoctl", "create", "postgres",
+    "--pgport", "5433",
+    "--ssl-self-signed",
+    "--auth", "trust",
+    "--pg-hba-lan",
+    "--hostname", "node1.tablespaces_default",
+    "--monitor", "postgresql://autoctl_node@monitor/pg_auto_failover",
+    "--run"]
```
DimCitus, line 25: Is it possible to use a very short command and move all the options to the environment, as done in https://github.com/citusdata/pg_auto_failover/blob/93f995812dffef29695f6f0ce63c5e69f5a34a74/tests/upgrade/docker-compose.yml#L24 for instance?

rheaton: On it

rheaton: I found a small behavior difference when working on this between env variable and function args when creating/running a data node:

`--pgport=1234` if the monitor is passed in with `PG_AUTOCTL_MONITOR` or `--monitor` without a port, it assumes the default port for the monitor of 5432

`PGPORT=1234` if the monitor is passed in with `PG_AUTOCTL_MONITOR` or `--monitor` without a port, it constructs a monitor URI with _no_ port (e.g. `postgresql://autoctl_node@monitor:/pg_auto_failover` and connections fail. passing in the monitor port in the monitor string works, but this is surprising.

Is this desired for some postgres-y reason I am not familiar with, or shall I file an issue?

rheaton:
@DimCitus I've addressed the comments and did a bit more cleanup. I haven't added the new tests to the travis suite -- I am not very familiar with it so please either let me know where is best or add it for me :) 
Also, the new top-level `make test-tablespaces` works with `PGVERSION`. (I tested it with versions 10-14)

src/bin/pg_autoctl/pgctl.c
```diff
@@ -1172,6 +1172,82 @@ prepare_guc_settings_from_pgsetup(const char *configFilePath,
...
+
+	while ((tblspcDirEntry = readdir(tblspcDir)) != NULL)
+	{
+		char tblspcDataDirPath[MAXPGPATH] = { 0 };
+
+		/* skip non-symlinks, as all tablespace dirs are symlinks */
+		if (tblspcDirEntry->d_type == DT_UNKNOWN)
+		{
+			struct stat structStat;
+			char pgTblspcEntryFullPath[MAXPGPATH] = { 0 };
```
DimCitus, line 1210: This variable is used after the if block, it should be moved up to the while block/scope.

src/bin/pg_autoctl/pgctl.c
```diff
@@ -1172,6 +1172,82 @@ prepare_guc_settings_from_pgsetup(const char *configFilePath,
...
+			if (lstat(pgTblspcEntryFullPath, &structStat) != 0)
+			{
+				log_error("Failed to get file information for \"%s/%s\": %m",
+						  pgTblspcFullPath,
+						  tblspcDirEntry->d_name);
+				return false;
+			}
+			if (!S_ISLNK(structStat.st_mode))
+			{
+				continue;
```
DimCitus, line 1222: We might want to log_debug about those non-symlink entries found in pg_tblspc?

src/bin/pg_autoctl/pgctl.c
```diff
@@ -1172,6 +1172,82 @@ prepare_guc_settings_from_pgsetup(const char *configFilePath,
...
+			}
+		}
+		else if (tblspcDirEntry->d_type != DT_LNK)
+		{
+			continue;
```
DimCitus, line 1227: Same here with log_debug?

src/bin/pg_autoctl/pgctl.c
DimCitus: Actually it's not, it's just that all those long names look too much alike each other. Is it possible to find better names for those variable, maybe just dropping the the pgTblspc prefix for the readdir() related ones to start with?

tests/tablespaces/Dockerfile
```diff
@@ -0,0 +1,12 @@
+ARG PGVERSION=10
+
+FROM pg_auto_failover_test:pg${PGVERSION} as test
+
+USER root
+RUN mkdir -p /tmp/extended_a && chown -R docker /tmp/extended_a
+RUN mkdir -p /tmp/extended_b && chown -R docker /tmp/extended_b
+RUN mkdir -p /tmp/extended_c && chown -R docker /tmp/extended_c
+RUN mkdir -p /tmp/extended_c && chown -R docker /tmp/extended_c
+RUN mkdir -p /var/lib/postgres && chown -R docker /var/lib/postgres
```
DimCitus, line 10: Can we install the table spaces not in /tmp?

I am thinking we could use /var/lib/postgres/tablespaces/{a,b,c,d} ; given that PGDATA is going to be found in /var/lib/postgres/10/main for PGVERSION 10. What do you think?

(Using /tmp triggers too much overhead thinking about what kind of mount point that is, file system, sticky bits on the directory, etc, and how that impacts the test, do we want to test those elements etc).

tests/tablespaces/Makefile
```diff
@@ -0,0 +1,37 @@
...
+	# Pollute the data dir so pg_rewind fails
+	docker compose exec node2 mkdir '/var/lib/postgres/data/base/1/99999999'
+	docker compose exec monitor pg_autoctl perform failover
+	TEST_NUM="03" docker compose up --exit-code-from=test test
+	docker compose pause node1
+	TEST_NUM="04" docker compose up --exit-code-from=test test
+	docker compose unpause node1
+	TEST_NUM="05" docker compose up --exit-code-from=test test
+	docker compose exec monitor pg_autoctl perform failover
+	TEST_NUM="06" docker compose up --exit-code-from=test test
```
DimCitus, line 33: Let's refactor those bits and have each test with its own Makefile target, and then the main test just depend on them.

``` Makefile
test: build test_01 test_02 ... ;

test_01:
	docker compose up -d node1
  	TEST_NUM="01" docker compose up --exit-code-from=test test

... 
```

tests/tablespaces/tablespace_utils.py
```diff
@@ -0,0 +1,113 @@
...
+            + self.port
+            + "'"
+        )
+        result = ""
+        if results:
+            result = results[-1]
+            if result:
+                result = result[-1]
+
+        return result
```
DimCitus, line 71: I tink in other tests we'd just use `return result[0][0]` or something? (the run_sql_query() method returns a list of tuples, and here we have a single item in the list, and a single column in the tuple, so [0][0] is the magic bit).

tests/tablespaces/tablespace_utils.py
```diff
@@ -0,0 +1,113 @@
...
+                cur.execute(query, args)
+                try:
+                    result = cur.fetchall()
+                except psycopg2.ProgrammingError:
+                    pass
+        # leaving contexts closes the cursor, however
+        # leaving contexts doesn't close the connection
+        conn.close()
+
+        return result
```
DimCitus, line 39: Sounds like we could inherit from class PGNode here, or teach a few new tricks to it?

tests/tablespaces/tablespace_utils.py
```diff
@@ -0,0 +1,113 @@
...
+                + "/pg_auto_failover"
+            )
+        else:
+            return (
+                "postgresql://docker@"
+                + self.service_name
+                + ":"
+                + self.port
+                + "/postgres"
+            )
```
DimCitus, line 57: Same worries again duplication of code and maintenance compared to enhancing the existing class or inheriting from it.

tests/tablespaces/tablespace_utils.py
DimCitus, line 71: Ditto.

tests/tablespaces/tablespace_utils.py
```diff
@@ -0,0 +1,113 @@
...
+                    print(
+                        "state of %s is '%s', waiting for '%s' ..."
+                        % (self.service_name, current_state, target_state)
+                    )
+
+            if current_state == target_state:
+                return True
+
+            prev_state = current_state
+        raise Exception("never got to state")
```
DimCitus, line 97: Again, that looks very familiar. Is it from PGNode? How can we re-use existing code instead of duplicating it?

DimCitus (changes requested):
Thanks for all the work, it looks better at each review. This time I left a couple of nitpicking comments about the C code, and then had a good look at the tests.

I like that you've been using docker-compose, which I think I prefer overall compared to the specific solution we have baked for testing, but I don't know how to run docker-compose in the CI environment. It seems to me that it would be best if you could re-use the existing framework and approach for your new tests. Put them all in a single file, use the usual `def test_01()` convention in Python, etc.

Or maybe you tried using the usual framework but it won't fit with the specific needs of the table space testing? in that case we should document that in a README file for the tests, I suppose...

src/bin/pg_autoctl/pgctl.c
rheaton: We'd be logging about `.` and `..` every time, but it's debug log so could be ok?

tests/tablespaces/Makefile
rheaton: Sure, the only caveat being that the `failover` one is run twice and make dependencies will miss that, so those will have to not be make targets. (which is fine)

tests/tablespaces/tablespace_utils.py
rheaton: Yes, I can change this, this was done but an interim test case during development when a new node was being created.

tests/tablespaces/tablespace_utils.py
rheaton: I can look into it, I'd worry about surprising behavior for things that don't work for future devs, but I can take a look and maybe have "not implemented" for things that won't work?

rheaton:
I decided to use docker-compose for a couple reasons:
1) It's already in use by the upgrade tests
2) the existing test behavior won't work for tablespaces because it uses a single volume which doesn't translate to this behavior. We need matching mounts across nodes.

tests/tablespaces/Dockerfile
rheaton: I'll use something that isn't `tmp` but I want to demonstrate that they are external volume mounts (or at least imply it) I will review! :)

tests/tablespaces/tablespace_utils.py
rheaton: I've addressed all the other comments. I'm looking into a nice way to do this. The number of functions that aren't possible in this way of running (in that you need to run things from outside of the test to interact with them as they aren't run in the same container) makes it so most of the PGNode functions and DataNode functions can't be shared. 

There are subtle differences that needed to be addressed, so I haven't had any luck sharing the code yet. I tried a few approaches but they weren't great.

rheaton: I'll create a readme today for the new tests and re-use the `wait_for_state` and `run_sql_queries` with some class inheritance (i'll make new super-classes because my attempt to inherit from existing had too much cruft, IMO). :)

rheaton: Take a look at the last commit, it simplifies the tablespace_utils.py file, but add a wee bit of complexity to the main test utils. Let me know if you think it's worth it! (I expect that it is)

src/bin/pg_autoctl/pgctl.c
DimCitus: We might want to skip those too, that's a good comment, thanks!

DimCitus (commented):
Will dive more into the testing bits again, it looks already cleaner. Also looks like we can actually run docker-compose (docker service) in the Travis CI environment, which is good. We had this thinking about replacing the Network Namespace implementation of our test framework with just using docker-compose, your work might as well be a good first step in that direction, thanks!

rheaton:
@DimCitus Confirmed we can get docker-compose to work! https://app.travis-ci.com/github/citusdata/pg_auto_failover/jobs/558452771 
For some reason it doesn't like `docker compose` versus `docker-compose` but I'll take it 🥇

tests/tablespaces/Dockerfile
```diff
@@ -0,0 +1,12 @@
+ARG PGVERSION=10
+
+FROM pg_auto_failover_test:pg${PGVERSION} as test
+
+USER root
+RUN mkdir -p /extra_volumes/extended_a && chown -R docker /extra_volumes/extended_a
+RUN mkdir -p /extra_volumes/extended_b && chown -R docker /extra_volumes/extended_b
+RUN mkdir -p /extra_volumes/extended_c && chown -R docker /extra_volumes/extended_c
+RUN mkdir -p /extra_volumes/extended_c && chown -R docker /extra_volumes/extended_c
```
DimCitus, line 9: Twice `extended_c` looks like a typo where d would be missing?

tests/tablespaces/Makefile
```diff
@@ -0,0 +1,49 @@
...
+	TEST_NUM="01" docker-compose up --exit-code-from=test test
+
+test_02:
+	docker-compose up -d node2
+	TEST_NUM="02" docker-compose up --exit-code-from=test test
+
+test_03:
+	# Pollute the data dir so pg_rewind fails
+	docker-compose exec node2 mkdir '/var/lib/postgres/data/base/1/99999999'
+	docker-compose exec monitor pg_autoctl perform failover
```
DimCitus, line 30: The split between the “infra” part in the Makefile, Dockerfile and docker-compose files and the “test/code” parts in the Python code is almost self-explanatory all along, apart from this bit here. Is there a strong reason why we would control the failover from the Makefile here rather than from the Python test file as usual?

If there is one, that's okay, it should be documented via a comment in the Makefile I think. It's not that trivial after all.

tests/tablespaces/test_tablespaces_03.py
```diff
@@ -0,0 +1,16 @@
+from nose.tools import *
+from tests.tablespaces.tablespace_utils import node1
+from tests.tablespaces.tablespace_utils import node2
+
+
```
DimCitus, line 5: That's where I expected to see the test_001 launching the failover as in the other (non-tablespace) tests, if that's possible. If we keep the approach where a docker-compose command is used in the Makefile, I suppose a comment here would make sense too.

tests/tablespaces/test_tablespaces_04.py
```diff
@@ -0,0 +1,14 @@
+from nose.tools import *
+from tests.tablespaces.tablespace_utils import node2
+
+
```
DimCitus, line 4: Same as before, the pause/unpause approach needs some comments as what to expect and why from the test file. I'm not sure that this time we can stop the pg_autoctl process without stopping the whole container, so we might just have to accept the control-from-outside-the-test-file idea there.

tests/tablespaces/Makefile
```diff
@@ -0,0 +1,49 @@
...
+	TEST_NUM="03" docker-compose up --exit-code-from=test test
+
+test_04:
+	docker-compose pause node1
```
DimCitus, line 34: This requires a comment too, as we would usually stop the pg_autoctl process from the Python test file. I'm not sure if we can do that when using docker-compose, as I suppose it would also stop the container entirely. Comparing the two approaches in a comment here looks like a good idea.

tests/tablespaces/Makefile
```diff
@@ -0,0 +1,49 @@
...
+	TEST_NUM="04" docker-compose up --exit-code-from=test test
+
+test_05:
+	docker-compose unpause node1
```
DimCitus, line 38: Same as the previous entry about pause.

tests/tablespaces/test_tablespaces_05.py
```diff
@@ -0,0 +1,15 @@
+from nose.tools import *
+from tests.tablespaces.tablespace_utils import node1
+from tests.tablespaces.tablespace_utils import node2
+
+
```
DimCitus, line 5: Same as the pause comment, a comment about the Makefile running unpause is necessary here.

DimCitus (approved):
I have left comments and changes that are still needed before we merge, but it's all about cleaning and documenting/commenting what you've come up with. Thanks for all the work on that one, and also creating a good reference how to use docker-compose instead of what we have at the moment.

I still have some reserves about the approach as you will see, if we manage to improve those points we should be able to clean-up our test infra and just use docker-compose for the heavy lifting. That would be good. But a lot of work ;-)

tests/tablespaces/Dockerfile
DimCitus: Actually it's just one extra line to remove apparently.

tests/tablespaces/Makefile
rheaton: Added some comments to the test files as well as the Makefile with some reasoning. 


I've been thinking about the other tests, if we wanted to move to a docker-compose approach:
We could have more python-test control if we added ssh access to the containers (this is non-trivial, but not terribly difficult, I believe). The caveat being stopping/starting the processes would require either a re-think of how the container commands run or external `docker-compose exec` like is being done here, since Docker will happily restart the container for you and there could be potential race conditions in your test (if it restarts it before you've done whatever work you wanted to do). 

Let me know if the updates are sufficient for this chunk of work!

rheaton:
I'm glad we didn't go forward with the mappings implementation, as it is not yet supported on Windows (noticed in the postgres Slack): https://www.postgresql.org/message-id/CAB7nPqRU7tiF8zXv9c3zrV-1kbtryJFHDRE0_zfUv6ncwG3%2B-A%40mail.gmail.com

PR #871, opened 2022-01-27
Minor: Remove duplication add error handling

rheaton:
Minor fixups found during #844 

- removing duplicate key from docker-compose.yml
- Handle [non-standard](https://www.gnu.org/software/libc/manual/html_node/Directory-Entries.html) field `d_type` in log file handling

### DIFF
--- a/src/bin/pg_autoctl/pgctl.c
+++ b/src/bin/pg_autoctl/pgctl.c
@@ -1666,11 +1666,6 @@ pg_log_startup(const char *pgdata, int logLevel)
 		char pgLogFilePath[MAXPGPATH] = { 0 };
 		struct stat pgLogFileStat;
 
-		/* our logFiles are regular files, skip . and .. and others */
-		if (logFileDirEntry->d_type != DT_REG)
-		{
-			continue;
-		}
 
 		/* build the absolute file path for the logfile */
 		join_path_components(pgLogFilePath,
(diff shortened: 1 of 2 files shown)


src/bin/pg_autoctl/pgctl.c
```diff
@@ -1684,6 +1679,19 @@ pg_log_startup(const char *pgdata, int logLevel)
...
 		}
+
+		/* our logFiles are regular files, skip . and .. and others */
+		if (logFileDirEntry->d_type == DT_UNKNOWN)
+		{
+			if (!S_ISREG(pgLogFileStat.st_mode))
+			{
+				continue;
+			}
+		} else if (logFileDirEntry->d_type != DT_REG)
```
DimCitus, line 1690: Nitpicking (style): please have else if on its own line.

More importantly, can you add a comment about why it's important to first try the d_type DT_UNKNOWN and the ST_ISREG macro and then check the d_type DT_REG? I'm curious, and then I think future us will be too ;-)

DimCitus (changes requested):

src/bin/pg_autoctl/pgctl.c
rheaton: Sure thing!

rheaton: Updated

DimCitus (approved):

Issue #874, opened 2022-03-23
fail to get uri 


changchichung:
I drop the monitor node first , then create a new monitor
```
postgres@hqs150:~$ pg_autoctl drop monitor
17:24:09 95082 INFO  An instance of pg_autoctl is running with PID 95054, waiting for it to stop.
17:24:11 95082 INFO  The pg_autoctl instance with pid 95054 has now terminated.
17:24:11 95082 WARN  Preserving configuration file: "/var/lib/postgresql/.config/pg_autoctl/database/main/pg_autoctl.cfg"
17:24:11 95082 WARN  Preserving Postgres Data Directory: "/database/main"
17:24:11 95082 INFO  pg_autoctl drop node keeps your data and setup safe, you can still run Postgres or re-join a pg_auto_failover cluster later
17:24:11 95082 INFO  HINT: to completely remove your local Postgres instance and setup, consider `pg_autoctl drop node --destroy`

postgres@hqs150:~$ pg_autoctl create monitor --hostname pg_monitor --auth trust --run --no-ssl
17:24:34 95112 WARN  No encryption is used for network traffic! This allows an attacker on the network to read all replication data.
17:24:34 95112 WARN  Using --ssl-self-signed instead of --no-ssl is recommend to achieve more security with the same ease of deployment.
17:24:34 95112 WARN  See https://www.postgresql.org/docs/current/libpq-ssl.html for details on how to improve
17:24:34 95112 INFO  Using default --ssl-mode "prefer"
17:24:34 95112 FATAL pg_autoctl is already running with pid 95085
```
then try to get the uri but failed
```
pg_autoctl show uri --monitor --pgdata /database/main
17:25:20 95123 ERROR Failed to parse connection string "--pgdata": missing "=" after "--pgdata" in connection info string
 
17:25:20 95123 FATAL Failed to parse --monitor connection string, see above for details.
```
and remove the pgdata parameter
```
postgres@hqs150:~$ pg_autoctl show uri --monitor
uri: option '--monitor' requires an argument
17:32:33 95546 ERROR Failed to parse command line, see above for details.
pg_autoctl show uri: Show the postgres uri to use to connect to pg_auto_failover nodes
usage: pg_autoctl show uri  [ --pgdata --monitor --formation --json ] 

  --pgdata      path to data directory
  --monitor     show the monitor uri
  --formation   show the coordinator uri of given formation
  --json        output data in the JSON format

```

missing an argument ? but I can not find what's the correct command ??
[...]

PR #880, opened 2022-04-18
Updated README to use '--formation monitor'.

chewygranola:
… README to use '--formation monitor' #874.

### DIFF
--- a/src/bin/pg_autoctl/cli_show.c
+++ b/src/bin/pg_autoctl/cli_show.c
@@ -67,9 +67,8 @@ static void print_all_uri(SSLOptions *ssl,
 CommandLine show_uri_command =
 	make_command("uri",
 				 "Show the postgres uri to use to connect to pg_auto_failover nodes",
-				 " [ --pgdata --monitor --formation --json ] ",
+				 " [ --pgdata --formation --json ] ",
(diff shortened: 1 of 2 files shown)


src/bin/pg_autoctl/cli_show.c
```diff
@@ -922,7 +921,6 @@ cli_show_uri_getopts(int argc, char **argv)
 
 	static struct option long_options[] = {
 		{ "pgdata", required_argument, NULL, 'D' },
-		{ "monitor", required_argument, NULL, 'm' },
```
DimCitus, line 925: Do we really want to remove the `--monitor` option from the command? It should be possible still to do the following:

```
$ pg_autoctl show uri --monitor postgres://monitor:5432/pg_auto_failover --formation foo
```

I suppose it's only documentation that should be updated.

DimCitus (changes requested):
Thanks for the PR! I believe the code editing wasn't necessary, even though the current help string and docs might be misleading.

src/bin/pg_autoctl/cli_show.c
chewygranola: Alright, --monitor option has been put back.

DimCitus (approved):

PR #913, opened 2022-06-27
Override style to ensure there is a focus ring on buttons

jonels-msft:
Fixes accessibility issue 1793792


jonels-msft:
@DimCitus this change looks good locally. Please give it a look and merge when you get a chance.

DimCitus (approved):

PR #922, opened 2022-08-17
Fix monitor extension for new Postgres 15 shared memory hooks.

DimCitus:


DimCitus:
We should probably edit the GitHub Actions settings and replace Postgres 10 with Postgres 15 as part of this PR. To test manually, do the following:

```
$ make TEST=test_installcheck PGVERSION=15 run-test
```

src/monitor/pg_auto_failover.c
```diff
@@ -77,10 +83,35 @@ _PG_init(void)
 						 "configuration variable in postgresql.conf.")));
 	}
 
+#if PG_VERSION_NUM >= PG_VERSION_15
```
onderkalaci, line 86: I guess you have not defined PG_VERSION_15. For Citus code, we have it here: https://github.com/citusdata/citus/blob/e2a94953342642f28d87e5ab846f55e29d636c50/src/include/distributed/pg_version_constants.h#L16

onderkalaci (approved):
I think the C code changes look good (other than the compile issue), I'll ask @hanefi to have a quick look to the Makefile and docker changes.

DimCitus:
> I think the C code changes look good (other than the compile issue), I'll ask @hanefi to have a quick look to the Makefile and docker changes.

Thanks for the review, I fixed the PG_VERSION_15 macro... because I didn't get a warning on compile I though it might have been provided by Postgres and I would have missed it somehow. I'm back to using `150000` which is good enough here I believe.

Also about the Docker changes, I think it's good to go. Our Dockerfile organisation made it very easy to include that, I'm not worried about these parts, and they're easy to test locally too.

PR #958, opened 2022-10-21
{WIP} Implement on_primary hooks.

### LINKED ISSUE #878: Feature request post failover command
Salut Dimitri,

Aim is to provide a parameter to execute a command (whatever command DBAs set) once a failover successfully happened

Use case is:
 -  sometimes after a failover we have to switch some settings (example switching Barman passive repository to active repository) 
 -  another example is indirection (a.k.a switch-ip)  if client does not support native high availability (e.g. they can't set two hosts on connection strings, example ODBC or embedded drivers configuration ), we could use this parameter (for example to reconfigure pgbouncer)
 
Because the only component which knows that a failover completely and succefullly happened is pg_autofailover.

- This command should be decorralated, if an error happens it simply should be list in the pg_autoctl log file.
- Most HA tools know provide this add-hoc parameter
Merci

Tiago

DimCitus:
This allows users of pg_auto_failover to setup their own scripts/actions to complement a failover. The hooks are run in a process that is separate from the main FSM, so as to prevent the system from making progress.

As a result, it's not possible for the hooks to change how things are implemented in pg_auto_failover itself.

The hook system also allows running a user-defined "service", which is a long running process or a deamon that belongs to pg_autoctl process tree.

 - [x] Implement a new internal service for running user-defined hooks
 - [ ] Implement pg_autoctl enable|disable run-hooks
 - [ ] Implement pg_autoctl create listener
 - [ ] Implement support for running script/commands (man system)
 - [ ] Add unit testing support for user-defined hooks
 - [ ] Document the new hook system, including tutorial
 - [ ] Add documentation examples covering pgbouncer as a hooked system

Fixes #878.

### DIFF
--- a/src/bin/pg_autoctl/cli_do_service.c
+++ b/src/bin/pg_autoctl/cli_do_service.c
@@ -27,6 +27,7 @@
 #include "service_keeper.h"
 #include "service_monitor.h"
 #include "service_postgres_ctl.h"
+#include "service_run_hooks.h"
 #include "signals.h"
 #include "supervisor.h"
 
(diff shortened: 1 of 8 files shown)


abubakarm94:
This feature will be extremely valuable. Any thoughts on when we can expect this?

PR #992, opened 2023-06-01
fix PG_AUTOCTL_MAX_TIMELINES

dbamu:
ERROR history file "00000202.history" contains 1024 lines, pg_autoctl only supports up to 1023 lines
https://github.com/hapostgres/pg_auto_failover/issues/991


dimitri:
Proper way to fix would be to use dynamic memory.

c2main:
What's the point in reading so much just to output to client ? Aren't the latest lines of the file enough ? ( this is just used for information purpose, isn't it ?)

dimitri:
Wee actually I did fix the problem in https://github.com/hapostgres/pg_auto_failover/pull/995

PR #1017, opened 2023-11-23
Prepare for release 2.1

gkokolatos:
* Add CHANGELOG entry for the release
* Update version number throughout
* Update vendored source dependency for parson
* Update vendored Postgres sources snprintf, strerror


gkokolatos:
There is a consistent test failure. It needs a bit more work I suspect.

dimitri:
I think it's the monitor extension scripts that need updating, including the https://github.com/hapostgres/pg_auto_failover/blob/main/src/monitor/pgautofailover--2.0--dummy.sql that should be 2.1--dummy now for the tests to pass. Another way to do it would be to release version 2.0.1 where the monitor extension is still 2.0 and didn't change.

gkokolatos:
> I think it's the monitor extension scripts that need updating, including the https://github.com/hapostgres/pg_auto_failover/blob/main/src/monitor/pgautofailover--2.0--dummy.sql that should be 2.1--dummy now for the tests to pass. Another way to do it would be to release version 2.0.1 where the monitor extension is still 2.0 and didn't change.

You are, of course, correct. I force pushed an update with the related test version fixes. I did keep the version number at 2.1 but I will happily change it to 2.0.1 if that is preferred.

dimitri (changes requested):
I think we should also edit the version string of `pg_autoctl --version` so that we list Postgres 16 compat there. Maybe remove Postgres 10 from the list also? It so happens that we're still compatible, but I don't want to make promises that we will continue to ensure that now.

gkokolatos:
> I think we should also edit the version string of `pg_autoctl --version` so that we list Postgres 16 compat there. Maybe remove Postgres 10 from the list also? It so happens that we're still compatible, but I don't want to make promises that we will continue to ensure that now.

Happy to do, however in my environment (TM) I am getting:
```bash
 $ pg_autoctl --version
pg_autoctl version 2.0.26.gf9f4bc5.dirty
pg_autoctl extension version 2.1
compiled with PostgreSQL 16.1 on x86_64-pc-linux-gnu, compiled by gcc (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0, 64-bit
compatible with Postgres 11, 12, 13, 14, 15, and 16
```

What am I missing?

dimitri:
> What am I missing?

You? nothing it seems ;-) Me... just what you showed. Thanks!

dimitri (approved):

dimitri:
We usually Squash and merge in this project, and then delete the work branch altogether. Please consider merging that way from the GitHub UI before going on to tag and publish the release! cheers

gkokolatos:
> We usually Squash and merge in this project, and then delete the work branch altogether. Please consider merging that way from the GitHub UI before going on to tag and publish the release! cheers

Excellent, will do exactly that

PR #1019, opened 2023-12-06
Add 2.0 -> 2.1 sql file to fix update path

### LINKED ISSUE #1018: Upgrade to 2.1 fails: extension "pgautofailover" has no update path from version "2.0" to version "2.1"
Today on our Ubuntu-based cluster, we installed updates via apt and our cluster broke.

Logs on the monitor node said
```
Dec 06 10:19:07 monitor01 pg_autoctl[67687]: 10:19:07 67687 ERROR Error 22023 while running Postgres query: ALTER EXTENSION "pgautofailover" UPDATE TO "2.1":
Dec 06 10:19:07 monitor01 pg_autoctl[67687]: 10:19:07 67687 ERROR ERROR:  extension "pgautofailover" has no update path from version "2.0" to version "2.1"
```

A bit of research told me that postgresql extension updates require an sql file with the filename `<extension>--<old version>--<new version>.sql`.

I've looked in the repo and found them [here](https://github.com/hapostgres/pg_auto_failover/tree/main/src/monitor), but a file for 2.0 to 2.1 is indeed missing. Instead there's "2.0 to dummy" at https://github.com/hapostgres/pg_auto_failover/blob/main/src/monitor/pgautofailover--2.1--dummy.sql.

I've looked at the update sql file for 1.6 to 2.0 which didn't contain anything except comments, so I've copied it and named it appropriately:
```
$ cd /usr/share/postgresql/14/extension
/usr/share/postgresql/14/extension$ cp pgautofailover--1.6--2.0.sql pgautofailover--2.0--2.1.sql
```
Now the cluster was able to start again. I just don't know if the actual upgrade file would have had important instructions that we have not applied.

Akkowicz:
Fixes #1018 

It looks like it was missed during release process.
It looks like there are no changes required to be made in sql update files.

### DIFF
--- a/src/monitor/pgautofailover--2.0--2.1.sql
+++ b/src/monitor/pgautofailover--2.0--2.1.sql
@@ -1,0 +1,7 @@
+--
+-- extension update file from 2.0 to 2.1
+--
+-- complain if script is sourced in psql, rather than via CREATE EXTENSION
+\echo Use "CREATE EXTENSION pgautofailover" to load this file. \quit
+
+-- no changes, just the version number


dimitri:
The problem is that release 2.1 has shipped already, we can't change it. 

What needs to happen is that we include this fix in the next release of course, but the next release is going to be v2.1.1. So we need to add the release prep for that release version string, with the CHANGELOG edits, debian control edits, version string in the code, etc etc, all the jazz. And then build new packages, publish them. 

So the proper fix for that problem is a whole new release. Ping @gkokolatos in case you want to be involved?

porsager:
Are there any plans for a new release that solves this?

dimitri (approved):
LGTM

porsager:
Very useful indeed! pg_auto_failover has been chugging along for us for several years now — enjoying the stability so much! 

I unfortunately know all to well how it is not being able to find the time we'd like for our passion projects.. I would love to help out any way I could, but I'm unfortunately a bit in the same situation as you with my own project Postgres.js.. Perhaps some day🤞

dimitri:
Fixed in #1069, see also https://github.com/hapostgres/pg_auto_failover/releases/tag/v2.2

PR #1061, opened 2025-01-18
Add support for PostgreSQL 17

### LINKED ISSUE #1048: PostgreSQL 17 support
Hi,

v17 GA is due soon. The attached patch fixed build problems on my servers, but I don't know whether it is enough to support v17 or not. Can you please take a look? Thanks! DEvrim

[pg_auto_failover-pg17.patch.txt](https://github.com/user-attachments/files/17090522/pg_auto_failover-pg17.patch.txt)

hashworks:
Added build support and test matrix entry for v17. Updated `citusdata/tools`. Dropped v10 support over the board since it was inconsistent in multiple files.

I also changed Debian buster and v14 to bookworm and v16 in the README to reference the current versions.

Fixes #1048.

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -6,8 +6,8 @@ Installing pg_auto_failover
 We provide native system packages for pg_auto_failover on most popular Linux
 distributions.
 
-Use the steps below to install pg_auto_failover on PostgreSQL 16. At the
-current time pg_auto_failover is compatible with PostgreSQL 13 to 16.
+Use the steps below to install pg_auto_failover on PostgreSQL 17. At the
+current time pg_auto_failover is compatible with PostgreSQL 13 and newer.
 
 Ubuntu or Debian
 ----------------
@@ -20,13 +20,13 @@ Binary packages for debian and derivatives (ubuntu) are available from
 documentation and then::
 
   $ sudo apt-get install pg-auto-failover-cli
-  $ sudo apt-get install postgresql-16-auto-failover
+  $ sudo apt-get install postgresql-17-auto-failover
 
 __ https://wiki.postgresql.org/wiki/Apt
 
 The Postgres extension named "pgautofailover" is only necessary on the
 monitor node. To install that extension, you can install the
-``postgresql-16-auto-failover`` package when using Postgres 16. It's
+``postgresql-17-auto-failover`` package when using Postgres 17. It's
 available for other Postgres versions too.
 
 Avoiding the default Postgres service
@@ -42,12 +42,12 @@ debian package, follow those steps:
 ::
 
   $ curl https://www.postgresql.org/media/keys/ACCC4CF8.asc | apt-key add -
-  $ echo "deb http://apt.postgresql.org/pub/repos/apt buster-pgdg main" > /etc/apt/sources.list.d/pgdg.list
+  $ echo "deb http://apt.postgresql.org/pub/repos/apt bookworm-pgdg main" > /etc/apt/sources.list.d/pgdg.list
 
   # bypass initdb of a "main" cluster
   $ echo 'create_main_cluster = false' | sudo tee -a /etc/postgresql-common/createcluster.conf
   $ apt-get update
-  $ apt-get install -y --no-install-recommends postgresql-14
+  $ apt-get install -y --no-install-recommends postgresql-17
 
 That way when it's time to :ref:`pg_autoctl_create_monitor` or
 :ref:`pg_autoctl_create_postgres` there is no confusion about how to handle
@@ -91,7 +91,7 @@ Here's a sample output from the command:
    WorkingDirectory = /var/lib/postgresql
    Environment = 'PGDATA=/var/lib/postgresql/monitor'
    User = postgres
-   ExecStart = /usr/lib/postgresql/10/bin/pg_autoctl run
+   ExecStart = /usr/lib/postgresql/17/bin/pg_autoctl run
    Restart = always
    StartLimitBurst = 0
 
@@ -128,7 +128,7 @@ pg_auto_failover uses the same build dependencies:
 
 ::
 
-   $ sudo apt-get build-dep -y --no-install-recommends postgresql-14
+   $ sudo apt-get build-dep -y --no-install-recommends postgresql-17
 
 Then build pg_auto_failover from sources with the following instructions:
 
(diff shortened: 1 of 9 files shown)


hashworks:
Not tested in production, let's see if the new v17 test workflow (which you need to approve) works.

dimitri:
It seems like you need to update Citus version for a Postgres 17 compatible one. Also, you can run the CI tests locally, it's using docker images that you can build etc. It's meant to be the same unit testing on your local dev env and in the CI.

porsager:
If I understand correctly, it unfortunately doesn't look llke there is a Citus version with support for v17 yet. It seems this issue is tracking it, but no ETA so far https://github.com/citusdata/citus/issues/7605 .

porsager:
Ok Citus 13 is out and it seems there where only a few extra changes needed to make all tests pass. Unfortunately the argument handling for pg_ctl changed in v17 (failed with `pg_ctl: too many command-line arguments`), so I had to change them to pass the subcommand as the last parameter to work around that issue. I think that's a fine solution, but would probably be good to dig deeper into why their change fails with the original code in this case.

I also switched from docker-compose to `docker compose` since it seems github runners changed old images to not include `docker-compose`.

I've got a branch with the additional changes here https://github.com/porsager/pg_auto_failover/tree/chore/postgres17, and you can see a successful test run here: https://github.com/porsager/pg_auto_failover/actions/runs/13053936252

Would you mind pulling the needed changes into this PR @hashworks , or should I just make a new PR with everything @dimitri ?

If there's anything else I can do to help push a new pg_auto_failover version with v17 support let me know!

dimitri:
Thanks a lot for your time and contribution @porsager ; please open a series of PR. 

I think we need a separate one for the switch from `docker-compose` to `docker compose`, and I have seen in the code some replacements that need more thinking: we want to search for `docker` in the $PATH now I suppose.

Then another PR to upgrade Citus to v13 in our tests, and same one or another one for Postgres 17 support?

Please go ahead and open your own PRs @porsager ; I believe that's the best way to make progress and respect everyone's time on this instance.

hashworks:
Sounds good to me, I can rebase this branch when the citus v13 is in master.

porsager:
That's great :)

I've made a PR for your branch @hashworks with the pg_ctl fixes here https://github.com/hashworks/pg_auto_failover/pull/1

src/monitor/version_compat.c
```diff
@@ -1,68 +0,0 @@
...
- *
- * Copyright (c) Microsoft Corporation. All rights reserved.
- * Licensed under the PostgreSQL License.
- *
- *-------------------------------------------------------------------------
- */
-
-#include "postgres.h"
-
-#if (PG_VERSION_NUM < 110000)
```
hashworks, line 14: If you want I can keep the file and just drop this block.

src/monitor/version_compat.h
```diff
@@ -14,38 +14,11 @@
...
-#endif
-
-#if (PG_VERSION_NUM < 120000)
-
-#define table_beginscan_catalog heap_beginscan_catalog
-#define TableScanDesc HeapScanDesc
-
-#endif
-
 #if (PG_VERSION_NUM >= 120000)
```
hashworks, line 22: I guess we can drop the #if, #endif around this since we are always >= 120000?

dimitri: While it's good to unclutter the CI and stop promising support for older Postgres versions (including in the docs and reference material), I see no reason to drop the actual code. If it doesn't incur development headaches, let's just keep that around.

src/monitor/version_compat.c
dimitri: Let's keep the file wholesale, I think that's best (see other comment).

src/monitor/version_compat.h
```diff
@@ -14,38 +14,11 @@
 
 #include "postgres.h"
 
-/* we support Postgres versions 10, 11, 12, 13, 14, 15, and 16. */
-#if (PG_VERSION_NUM < 100000 || PG_VERSION_NUM >= 170000)
+/* we support Postgres versions 13, 14, 15, 16, and 17. */
+#if (PG_VERSION_NUM < 130000 || PG_VERSION_NUM >= 180000)
```
dimitri, line 18: Let's allow for building with older Postgres versions. Make it easy for people to tinker with it, unless it's a direct impediment to progress.

dimitri (changes requested):

src/monitor/version_compat.h
hashworks: In that case I'll drop the whole block.

src/monitor/version_compat.h
```diff
@@ -14,11 +14,6 @@
 
 #include "postgres.h"
 
-/* we support Postgres versions 10, 11, 12, 13, 14, 15, and 16. */
-#if (PG_VERSION_NUM < 100000 || PG_VERSION_NUM >= 170000)
-#error "Unknown or unsupported postgresql version"
-#endif
-
```
dimitri, line 21: I believe we want to keep that part, that's the only build-time check we have for Postgres compatibility. Just amend it to include Postgres 17.

dimitri (changes requested):

porsager:
Was something not working with 11 and 12? How about just getting v17 support released, and then removing the older support later if it's not a burden at the moment?

src/monitor/version_compat.h
hashworks: Dropped the removal-commit, re-added build-support for v10.

docs/install.rst
```diff
@@ -6,8 +6,8 @@ Installing pg_auto_failover
 We provide native system packages for pg_auto_failover on most popular Linux
 distributions.
 
-Use the steps below to install pg_auto_failover on PostgreSQL 16. At the
-current time pg_auto_failover is compatible with PostgreSQL 13 to 16.
+Use the steps below to install pg_auto_failover on PostgreSQL 17. At the
+current time pg_auto_failover is compatible with PostgreSQL 13 and newer.
```
dimitri, line 10: I think I prefer the spelling “PostgreSQL 13 to 17”, because soon enough PostgreSQL 18 will be out.

dimitri (changes requested):

docs/install.rst
hashworks: Changed in 50051c6

dimitri (approved):

dimitri:
Fixes #1048.

PR #1066, opened 2025-01-31
Drop i386 support

hashworks:
Debian, Ubuntu and Arch derivates have dropped i386 support since a while, so I don't see a good reason to keep it here.


df7cb:
The PostgreSQL extensions in Debian have disabled 32-bit architectures because the users-to-bugs ratio there is too low to justify spending effort on it. (The PG server packages still support 32-bit because it still works and doesn't cause problems.)

So yes, pg_a_f should do it as well.

df7cb (approved):
LGTM.

hashworks:
Can you re-run 15-monitor in case it's a race condition?

PR #1068, opened 2025-02-01
Update to Debian Bookworm, drop deprecated apt-key usage

hashworks:
As documented in https://wiki.postgresql.org/wiki/Apt#Manual_Repository_Configuration .


docs/tutorial/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM debian:bullseye-slim
+FROM debian:bookworm-slim
 
 ARG PGVERSION=14
 ARG CITUS=postgresql-14-citus-11.1
```
dimitri, line 4: Can we also update that to Postgres 17 and Citus 13? That would fix https://github.com/hapostgres/pg_auto_failover/pull/1029 I believe.

hashworks:
> > RUN pip3 install pyroute2>=0.5.17
> 
> This should be installed as a Debian package instead, it's version 0.7.2 in bookworm.

Added in 0c61f11. But I see there is a `Pipfile`... I had to look up what [that is](https://github.com/pypa/pipfile). As far as I see the project is dead, should I:
A) Simple remove `Pipfile` and `Pipfile.lock` since we expect people to use docker or
B) Convert it to `requirements.txt` ?

docs/tutorial/Dockerfile
hashworks: Added in 392f825

dimitri:
A Quick Look shows that the PIP file was used to install python packages pyroute2, nose, and psycopg2 all from pypi. I see that we have `apt-get` coverage for those 3 now in the Dockerfile (with your addition of pyroute2 there), so please see about removing the pipfiles  entirely, and maybe also `python3-pip` from the package list installed with `apt-get`.

dimitri (approved):

dimitri:
> Test failures seem to be due to the updated pyroute2 package, I'll take a look tomorrow.

My advice would then be to remove that change from this PR, and make another PR that modernizes the installation of Python dependencies for our unit testing.

hashworks:
> My advice would then be to remove that change from this PR, and make another PR that modernizes the installation of Python dependencies for our unit testing.

Moved to #1070.

dimitri (approved):

hashworks:
> ```
> ERROR: test_multi_ifdown.test_012_prepare_replication_quorums
> ----------------------------------------------------------------------
> Traceback (most recent call last):
>   File "/usr/lib/python3/dist-packages/nose/case.py", line 197, in runTest
>     self.test(*self.arg)
>   File "/usr/src/pg_auto_failover/tests/test_multi_ifdown.py", line 207, in test_012_prepare_replication_quorums
>     assert node2.set_replication_quorum("false")
>            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
...
>   File "/usr/lib/python3/dist-packages/pyroute2/nslink/nspopen.py", line 224, in proxy
>     return _handle(self.channel_in.get())
>            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/usr/lib/python3/dist-packages/pyroute2/nslink/nspopen.py", line 25, in _handle
>     raise result['data']
> UnicodeDecodeError: 'utf-8' codec can't decode byte 0x90 in position 289: invalid start byte
> ```

I wasn't able to find a fix for this, if anyone has any ideas I would appreciate it.

dimitri:
Still having the same issues when updating python3/pyroute2 (UnicodeDecodeError). Let's go ahead with the release and see about modernizing our CI later?

Issue #905, opened 2022-05-22
pg_autoctl: unrecognized option on Alpine Linux


asafza:
When running the following Dockerfile

```
FROM alpine:3.15.4

RUN apk add postgresql-dev openssl-dev krb5-dev ncurses-dev \
    build-base linux-headers \
    libxml2-dev lz4-dev libxslt-dev libedit-dev readline-dev \
    postgresql14 git

RUN git clone --depth 1 --branch v1.6.4 https://github.com/citusdata/pg_auto_failover && \
    cd pg_auto_failover && \
    make install -j8

RUN /usr/libexec/postgresql14/pg_autoctl create monitor \
    --auth trust \
    --ssl-self-signed \
    --pgdata monitor \
    --pgctl /usr/bin/pg_ctl
```

The `pg_autoctl` command results with 
```
/usr/libexec/postgresql14/pg_autoctl: unrecognized option: auth
/usr/libexec/postgresql14/pg_autoctl: unrecognized option: ssl-self-signed
/usr/libexec/postgresql14/pg_autoctl: unrecognized option: pgdata
/usr/libexec/postgresql14/pg_autoctl: unrecognized option: pgctl
pg_autoctl: pg_auto_failover control tools and service
usage: pg_autoctl [ --verbose --quiet ]


Available commands:
  pg_autoctl
  + create   Create a pg_auto_failover node, or formation
  + drop     Drop a pg_auto_failover node, or formation
  + config   Manages the pg_autoctl configuration
  + show     Show pg_auto_failover information
  + enable   Enable a feature on a formation
  + disable  Disable a feature on a formation
  + get      Get a pg_auto_failover node, or formation setting
  + set      Set a pg_auto_failover node, or formation setting
  + perform  Perform an action orchestrated by the monitor
    run      Run the pg_autoctl service (monitor or keeper)
    watch    Display a dashboard to watch monitor's events and state
    stop     signal the pg_autoctl service for it to stop
    reload   signal the pg_autoctl for it to reload its configuration
    status   Display the current status of the pg_autoctl service
    help     print help message
    version  print pg_autoctl version
```

DimCitus:
See also #223 ; where my last question remains. I can't see how to fix getopt_long parsing with POSIXLY_CORRECT in a way that works both on GNU libc, macos libc, and also Alpine libc. Or maybe I'm just missing something? help!

DimCitus:
[...]

PR #1084, opened 2025-05-15
Replace `POSIXLY_CORRECT` for `getopt_long`

elipsion:
* Removes the use of `POSIXLY_CORRECT` environment variable in favor of passing this directive in the optstring. This should solve issues for users of muslc (aka Alpine Linux).


felixbuenemann:
Related to the minimal fix in #1090 and issue #905

Issue #1113, opened 2026-02-23
Last database stuck in report_lsn


nmcc1212:
I have 2 dbs in site 1 and 1 db in site 2
I am testing a scenario where both DBs in site 1 are down/unavailable
But the DB in site 2 does not promote to primary (as I would expect) and instead is stuck in 'report_lsn' 

I have tried setting replication quorum on the site 2 DB to false and setting number-sync-standbys to 0, but still have the same result

Is this expected behaviour?

<img width="1554" height="77" alt="Image" src="https://github.com/user-attachments/assets/5ba66a83-26ec-420b-9759-e9614d7332a6" />

nmcc1212:
Possibly related to #858 - tried setting db3 (second site node) to priority 0, with replication quorum disabled and setting number-sync-standbys to 0 attempting to mimic the example in the docs https://pg-auto-failover.readthedocs.io/en/main/architecture-multi-standby.html#sample-architecture-with-three-standby-nodes-one-async but still same issue

dimitri:
This is fixed in PR #1142 (now merged). Your scenario is exactly the case the feature targets.

The root cause: both site-1 nodes had `replication_quorum = true`, so the monitor's `missingNodesCount` guard blocked promotion of the site-2 node indefinitely. Setting `number-sync-standbys = 0` did not help because that controls a *different* guard — `missingNodesCount` is checked independently.

**To unblock a stuck cluster like this:**
```bash
pg_autoctl perform failover --allow-data-loss --formation default
```
Run this against the monitor. It bypasses both guards and promotes the most advanced surviving standby. Any transactions acknowledged by the missing site-1 nodes but not yet replicated to site-2 may be lost — that risk is why this is opt-in and named explicitly.

If the missing nodes come back later they will rejoin as standbys; diverged WAL is truncated via `pg_rewind` during re-join.

See the updated [configuration reference](https://pg-auto-failover.readthedocs.io/en/main/ref/configuration.html) for `pgautofailover.guard_data_loss` and the [perform failover reference](https://pg-auto-failover.readthedocs.io/en/main/ref/pg_autoctl_perform_failover.html) for full details.

nmcc1212:
Thanks for this, is a release planned soon so we can use this?

dimitri:
[...]

Issue #858, opened 2021-12-16
Multi-node failure requires manual intervention


rheaton:
Hello, we have been testing a variety of scenarios using pg_autofailover and disaster recovery. 

If we have a 6 node setup, and we destroy 3 that includes the primary, it takes manual intervention to get out of `report_lsn` state and promote a primary from the remaining healthy nodes. 
This was surprising to us.

Is there a reason that we don't have a timeout for [this scenario](https://github.com/citusdata/pg_auto_failover/blob/master/src/monitor/group_state_machine.c#L1710)? 

Some output from our scenario, for your perusal follows.

_Postgres version: 14.1
pg_autofailover version: 1.6.3
OS: centos7_

After cutting network access to/from node_1, node_2, and node_3, we see the following:

```bash
-bash-4.2$ pg_autoctl show state
  Name |  Node |                               Host:Port |       TLI: LSN |   Connection |      Reported State |      Assigned State
-------+-------+-----------------------------------------+----------------+--------------+---------------------+--------------------
node_1 |     1 | hadr-node-a.c.data-pcf-db.internal:5432 |   1: 0/B002508 | read-write ! |             primary |            draining
node_2 |     2 | hadr-node-b.c.data-pcf-db.internal:5432 |   1: 0/B002508 |  read-only ! |           secondary |           secondary
node_3 |     3 | hadr-node-c.c.data-pcf-db.internal:5432 |   1: 0/B002508 |  read-only ! |           secondary |           secondary
node_4 |     4 | hadr-node-d.c.data-pcf-db.internal:5432 |   1: 0/B002580 |    read-only |          report_lsn |          report_lsn
node_5 |     5 | hadr-node-e.c.data-pcf-db.internal:5432 |   1: 0/B002580 |    read-only |          report_lsn |          report_lsn
node_6 |     6 | hadr-node-f.c.data-pcf-db.internal:5432 |   1: 0/B002580 |    read-only |          report_lsn |          report_lsn
```
and 

```bash
-bash-4.2$ pg_autoctl show settings
  Context |    Name |                   Setting | Value
----------+---------+---------------------------+------
formation | default |      number_sync_standbys | 1
  primary |  node_1 | synchronous_standby_names | ''
     node |  node_1 |        candidate priority | 50
     node |  node_2 |        candidate priority | 50
     node |  node_3 |        candidate priority | 50
     node |  node_4 |        candidate priority | 10
     node |  node_5 |        candidate priority | 10
[...]

PR #1114, opened 2026-02-24
monitor: fix report_lsn deadlock when quorum nodes are permanently unavailable

nmcc1212:
## Problem

When a multi-node cluster loses enough nodes to include the primary, the
remaining healthy standbys can get permanently stuck in `report_lsn` with no
automatic recovery, requiring manual `pg_autoctl drop node` intervention to
escape. This affects any topology where a replication-quorum node becomes
permanently unreachable without first reporting its LSN to the monitor.

Reported in #858 and #1113.

### Root cause

`BuildCandidateList` incremented `missingNodesCount` for *every* quorum node
that was unhealthy and not reporting — treating a permanently dead node the
same as a transiently slow one. Because `ProceedGroupStateForMSFailover`
refuses to proceed while `missingNodesCount > 0`, the failover would block
forever once any quorum node disappeared without reporting.

## Solution

Separate the two kinds of "missing" nodes with a new `deadMissingNodesCount`
field on `CandidateList`:

| Counter | Meaning |
|---|---|
| `missingNodesCount` | Alive nodes assigned `report_lsn` that haven't reported yet — we wait for these |
| `deadMissingNodesCount` | Quorum nodes that are unreachable (unhealthy + not reporting) and whose LSN we can never recover |

After all alive nodes have reported, the **pigeonhole principle** determines
whether automatic failover is safe:

- **`deadMissingNodesCount < number_sync_standbys`** → proceed automatically.
  Every acknowledged commit required `number_sync_standbys` standbys to
  confirm it. With fewer than that many dead, at least one alive candidate
  must hold the most recent commit, so promoting the most-advanced alive node
  is data-safe.

- **`deadMissingNodesCount >= number_sync_standbys`** → block and emit a clear
  log message directing the operator to run `pg_autoctl drop node` for each
  unreachable node. Data-safety cannot be automatically guaranteed.

- **`number_sync_standbys = 0`** → always proceed; no synchronous-replication
  guarantee was ever in effect.

## Changes

- `src/monitor/group_state_machine.c`
  - Add `deadMissingNodesCount` to `CandidateList`
  - `BuildCandidateList`: increment `deadMissingNodesCount` for unreachable quorum nodes
  - `ProceedGroupStateForMSFailover`: apply pigeonhole check after the
    `missingNodesCount` guard; log message when
    automatic failover cannot proceed

### DIFF
--- a/src/monitor/group_state_machine.c
+++ b/src/monitor/group_state_machine.c
@@ -48,6 +48,7 @@ typedef struct CandidateList
 	int candidateCount;
 	int quorumCandidateCount;
 	int missingNodesCount;
+	int deadMissingNodesCount;
 } CandidateList;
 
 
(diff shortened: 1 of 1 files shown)


nmcc1212:
Closing in favour of #1142

PR #1116, opened 2026-04-08
Fix missing dbname in primary_conninfo for PG 17 slot sync

RToledo07:
## Summary

`pg_autoctl` generates `primary_conninfo` for standbys without the `dbname=` parameter, because all internal callers of `prepare_primary_conninfo()` pass `NULL` as the `dbname` argument. PostgreSQL 17+ slot synchronization requires `dbname=` in `primary_conninfo` and rejects connections that lack it with:

```
LOG:  replication slot synchronization requires "primary_conninfo" to be set
```

As a result, logical replication slots created with `failover = true` cannot be synchronized to physical standbys managed by pg_auto_failover, breaking logical replication continuity across failovers on PG 17 and later.

## Fix

Add a `dbname` field to `ReplicationSource` and propagate the configured database name (`config->pgSetup.dbname`) through `standby_init_replication_source()` to all callers of `prepare_primary_conninfo()` (and the related helpers `pg_basebackup()`, `prepare_recovery_settings()` and `pgctl_identify_system()`).

`prepare_primary_conninfo()` now also guards against an empty (non-`NULL`) `dbname` with `IS_EMPTY_STRING_BUFFER` to avoid emitting a malformed `dbname=` token in the connection string.

### Files changed (8 files, +24 / -4)

| File | Change |
|---|---|
| `src/bin/pg_autoctl/pgsql.h` | Add `char dbname[NAMEDATALEN]` field to `ReplicationSource` |
| `src/bin/pg_autoctl/primary_standby.h` | Add `const char *dbname` parameter to `standby_init_replication_source()` |
| `src/bin/pg_autoctl/primary_standby.c` | Implementation: copy `dbname` into `ReplicationSource` (`NULL`-guarded) |
| `src/bin/pg_autoctl/pgctl.c` | Replace `NULL` with `replicationSource->dbname` in 3 calls (`pg_basebackup`, `prepare_recovery_settings`, `pgctl_identify_system`); add `IS_EMPTY_STRING_BUFFER` guard in `prepare_primary_conninfo` |
| `src/bin/pg_autoctl/cli_do_misc.c` | Pass `config.pgSetup.dbname` in 3 callers of `standby_init_replication_source()` |
| `src/bin/pg_autoctl/fsm_transition.c` | Pass `config->pgSetup.dbname` in 7 callers (all FSM transitions that initialize replication) |
| `src/bin/pg_autoctl/keeper.c` | Pass `config->pgSetup.dbname` in `keeper_ensure_configuration()` |
| `src/bin/pg_autoctl/keeper_pg_init.c` | Pass `config->pgSetup.dbname` in `wait_until_primary_has_created_our_replication_slot()` |

### Note on `pg_rewind` (intentionally untouched)

`pg_rewind()` in `pgctl.c` (around line 1454) is the **fourth** caller of `prepare_primary_conninfo()` but is intentionally **not** modified by this patch. It already passes a hardcoded `"postgres"` as the dbname with the existing comment `/* pg_rewind needs a database */`. This is intentional and defensive: `pg_rewind` runs during failover/recovery scenarios where the user-configured database may be unavailable or in a degraded state, while the `postgres` database is guaranteed to exist on every standard PostgreSQL installation. Switching to `replicationSource->dbname` here would risk regressions in recovery paths.

## Reproduction (without the fix)

1. Set up a 1+2 pg_auto_failover cluster on PG 17 (1 monitor, 1 primary, 2 standbys).
2. On the standbys, set `sync_replication_slots = on` and `hot_standby_feedback = on`, restart.
3. Inspect any standby's logs - you will see (repeatedly):
   ```
   LOG: replication slot synchronization requires "primary_conninfo" to be set
   ```
   even though `primary_conninfo` is in fact set, because `dbname=` is missing from it.
4. Inspect `SHOW primary_conninfo;` on a standby:
   ```
   application_name=pgautofailover_standby_2 host=... port=... user=pgautofailover_replicator sslmode=prefer
   ```
   No `dbname=` token.
5. On the primary, create a subscription that requests slot failover support:
   ```sql
   CREATE TABLE events(id bigserial PRIMARY KEY, payload text);
   CREATE PUBLICATION pub_events FOR TABLE events;
   ```
   Then on a separate subscriber instance:
   ```sql
   CREATE SUBSCRIPTION sub_events
     CONNECTION '...'
     PUBLICATION pub_events
     WITH (failover = true, create_slot = true, slot_name = 'sub_events_slot');
   ```
6. Inspect `pg_replication_slots` on the standbys: `sub_events_slot` is **not** present (slot sync worker cannot synchronize it).
7. `pg_autoctl perform failover` - the new primary does not have `sub_events_slot` - the subscriber loses its slot and must be recreated with full data resync.

## Validation (with the fix)

End-to-end test performed locally on a 3-node cluster + independent subscriber instance:

| # | Step | Expected (with fix) | Result |
|---|---|---|---|
| 1 | Inspect `primary_conninfo` on standbys | Contains `dbname=postgres` | PASS |
| 2 | Standby logs after restart | Slot sync worker reaches `started` state without the "requires primary_conninfo" error | PASS |
| 3 | Create logical subscription with `failover = true` from a separate subscriber instance | Slot `sub_events_slot` created on primary | PASS |
| 4 | Inspect `pg_replication_slots` on standbys after a few seconds | `sub_events_slot` appears with `synced=t`, `failover=t` | PASS |
| 5 | Standby logs show `"newly created replication slot ... is sync-ready now"` | Slot sync confirmed | PASS |
| 6 | INSERT data on primary | Subscriber receives the rows | PASS |
| 7 | Run `pg_autoctl perform failover` | Topology switches, new primary elected | PASS |
| 8 | Inspect `primary_conninfo` on the **new** standbys (former primary + remaining standby) | Still contains `dbname=postgres`, now pointing at the new primary | PASS |
| 9 | Inspect `sub_events_slot` on the new primary | Slot present, `active=t` | PASS |
| 10 | INSERT data on the new primary | Subscriber receives the new rows **without** restarting the subscription | PASS |
| 11 | `pg_stat_subscription` on the subscriber | Streaming live (`pid` set, `received_lsn` advancing) | PASS |

Without this patch, step 4 never succeeds and steps 9-11 fail because the slot cannot be synchronized to standbys in the first place.

## Compatibility

The patch contains no version-specific PostgreSQL APIs, no new `#include`s, no `PG_VERSION_NUM` checks. It only touches internal pg_auto_failover structs and parameter passing, so it should compile and work identically across the full range of supported PG versions (`PGVERSIONS = 13 14 15 16 17 18` from the Makefile).

The behavior change is observable only on PG 17+ where slot synchronization to standbys was introduced; on earlier versions, including `dbname=` in `primary_conninfo` is harmless.

## Backwards compatibility

- The new field `ReplicationSource.dbname` is a fixed-size buffer; the `NULL`-guarded `strlcpy` in `standby_init_replication_source()` ensures callers that pass `NULL` (none currently exist after this patch) leave the field as zero bytes.
- The `IS_EMPTY_STRING_BUFFER` guard in `prepare_primary_conninfo()` ensures that, even if a future caller forgets to set `dbname`, the resulting connection string is still well-formed (just omitting the `dbname=` token, restoring the previous behavior).


dimitri:
Thank you for the detailed report and the patch, @RToledo07 — the reproduction steps and validation table made this really easy to confirm and act on.

This fix has been folded into #1144, which lands broader PostgreSQL 17/18/19 compatibility work and includes the same `dbname` threading through `ReplicationSource`/`prepare_primary_conninfo()` that you proposed here. That's why this PR is being closed rather than merged standalone — the change itself is now on `main` as part of #1144.

Thanks again for catching this and for the thorough write-up!

PR #1111, opened 2026-02-04
Add PostgreSQL 18 support

jhf:
## Summary
- Add PostgreSQL 18 to the compatibility version string in `pg_autoctl --version`
- Enable PG18 Docker builds using Citus commit `803f0ac` from `release-14.0` branch
- Add `libnuma-dev` dependency required by PostgreSQL 18

## Test plan
- [x] Single standby tests pass (88 tests)
- [x] Multi-standby tests pass (101 tests)
- [x] `pg_autoctl --version` shows "compatible with Postgres 13, 14, 15, 16, 17 and 18"

```
$ ./src/bin/pg_autoctl/pg_autoctl --version
pg_autoctl version ...
pg_autoctl extension version 2.2
compiled with PostgreSQL 18.1
compatible with Postgres 13, 14, 15, 16, 17 and 18
```

🤖 Generated with [Claude Code](https://claude.ai/code)

PR #1120, opened 2026-05-18
Add PostgreSQL 18 support

shirkevich:
Builds on #1111 and adds one missing PG18 path: `pg_autoctl do tmux compose` still rejected `PGVERSION=18` because its local Postgres/Citus compatibility matrix stopped at 17.

Changes:
- keep the existing PG18 Docker and version-string updates from #1111
- add PG18 to the tmux-compose Citus matrix using the same Citus commit (`803f0ac`)
- allow `PGVERSION=18` in the tmux-compose range check

I could not retrieve the old Actions job logs for #1111 because GitHub returned HTTP 410 for the expired job logs, but the failed PG18 check annotations still point at the workflow commands exiting non-zero.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -33,6 +33,7 @@ RUN apt-get update \
     libedit-dev \
     libkrb5-dev \
     liblz4-dev \
+    libnuma-dev \
     libncurses6 \
     libpam-dev \
     libreadline-dev \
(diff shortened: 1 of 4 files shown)


dimitri:
Thank you @shirkevich for picking up where #1111 left off and tracking down the missing tmux-compose Citus matrix entry — that kind of follow-through on an incomplete fix is valuable.

#1144 (merged 2026-07-15) supersedes both #1111 and this PR. It provides full PostgreSQL 18 and 19 support, the correct Citus alignment (`v14.1.0` for PG18, skipped for PG19 pending an upstream release), and comprehensive CI coverage across PG 14–19 including pgaftest and pytest suites.

Closing in favour of #1144 — thank you for the contribution!

